PR #3, opened 2021-10-31
feat: add translate_package_lock rule

alexeagle:


js/private/translate_package_lock.bzl
```diff
@@ -0,0 +1,97 @@
...
+        bzl_out.extend([_NPM_IMPORT_TMPL.format(
+            package = name,
+            version = dep["version"],
+            url = dep["resolved"],
+            integrity = dep["integrity"],
+            deps = deps,
+        )])
+
+def _translate_package_lock(repository_ctx):
+    bzl_content = ["""# Generated by package_lock.bzl from {package_lock}
```
mattem, line 75: Should this have an `@generated` banner?

alexeagle: I don't understand, you mean this comment should have different syntax?

mattem: In addition to, if this is a generated file that's check in, some IDEs and code review tools etc read the `@generated` style comments and treat them differently.

mattem (approved):

PR #8, opened 2021-12-04
fix: translate_package_lock handles semver ranges

alexeagle:
We have to do a second lookup through the file to get the actual version
of the package that satisfied the dependency.

Also add the ability to patch npm packages as we fetch them,
since there's no way to use the patch-package package in this setup
where npm packages are installed individually.

These fixes needed for rules_swc


gregmagolan (approved):
🍕

PR #13, opened 2021-12-21
fix: only check for --enable_runfiles on windows

alexeagle:
By default on Mac/Linux, the flag isn't set so we get a false positive here, requiring users to put --enable_runfiles in bazelrc


gregmagolan (approved):
🌮

PR #14, opened 2022-01-10
wip: starlark only dynamic linker

thesayyn:
starlark & node_path only dynamic linker. I have investigated a bit to see if we could rely on bazel to do dynamic linking for us. so far I have tried a few approaches but none of them were successful except the one with derived NODE_PATH which seems to work atm.

The only caveat is that you end up having two distinct node_modules which look like below.

```
path/to/bin
     _bin_.runfiles/node_modules
     _bin_launcher.sh
    
path/to/makesiterule
     node_modules
```

the magic is here to tell nodejs to look up for both node_modules when trying to resolve a package. though not sure this approach will work for all of nodejs packages out there. I will confirm that it works with de-facto bundlers out there.

Even if this doesn’t work, I still have a trick up my sleeve to get a unified node_modules directory.

this is just a small work done to find out possible ways to link make-site dependencies without writing a linker that runs before every action.


example/BUILD.bazel
```diff
@@ -223,3 +224,18 @@ diff_test(
...
+
+
+rollup_bundle(
+    name = "bundle",
+    srcs = [
+        "require_acorn.js",
+    ],
+    deps = [
+        "@npm_deps//@gregmagolan/test-b",
+        "@npm_acorn-8.4.0",
```
alexeagle, line 236: is there a reason not to use `@npm_deps//acorn` here for consistency?

js/merge_deps.bzl
```diff
@@ -0,0 +1,30 @@
+def merge_deps(ctx, tool_attr, dep_attr):
```
alexeagle, line 1: I think this file belongs under private/

js/merge_deps.bzl
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Collects third-party dependecies and merges them with first-party dependencies provides them as tools for actions to consume those dependecies.
```
alexeagle, line 3: spelling: dependencies
(generally maybe your editor could underline those)

js/private/nodejs_binary.bzl
```diff
@@ -144,13 +144,17 @@ def _bash_launcher(ctx, linkable, args):
...
         node_path = ""
-    node_path = "export NODE_PATH=" + ":".join(node_paths)
+    node_path = ":".join(node_paths)
     ctx.actions.write(
         launcher,
         """#!{bash}
 {rlocation_function}
 set -o pipefail -o errexit -o nounset
-{node_path}
+if [[ -n "${{NODE_PATH-}}" ]]; then
```
alexeagle, line 153: does there need to be a corresponding fix in the windows cmd.bat version?

js/private/nodejs_package.bzl
```diff
@@ -152,6 +152,11 @@ def _nodejs_package_impl(ctx):
     return [
         DefaultInfo(files = files, runfiles = runfiles),
         LinkablePackageInfo(package_name = ctx.attr.package_name, files = [output]),
+        # TODO: figure this out why we need this
```
alexeagle, line 155: you can just comment
"Always assume that packages provide typings, so we don't need to use an action to inspect the package.json#typings field or search for .d.ts files in the package."

linker/BUILD.bazel
```diff
@@ -0,0 +1,11 @@
+load("//js:nodejs_package.bzl", "nodejs_package")
+
+nodejs_package(
+    name = "linker",
+    package_name = "@rules_js/linker",
```
alexeagle, line 5: what does `@rules_js` mean here? we don't have such a scope on npm and don't plan to publish one. do we actually resolve from this package name?

linker/index.js
```diff
@@ -0,0 +1,11 @@
+const fs = require("fs");
```
alexeagle, line 1: you should `pre-commit install` to get prettier running on the files

rollup/rollup.js
```diff
@@ -0,0 +1,7 @@
+const {link} = require("@rules_js/linker");
```
alexeagle, line 1: can we use `--require` to make the linker call happen from the shell launcher, so we don't need to modify rollup?

ts/BUILD.typescript
```diff
@@ -26,6 +29,8 @@ write_file(
     out = "tsc.js",
     content = [
         "const {dirname} = require('path')",
+        "const {link} = require('@rules_js/linker')",
```
alexeagle, line 32: same here, soon we'll want to remove this JS file and point `nodejs_binary` directly at the file in external, so we want `--require` to install the linker

ts/private/ts_project.bzl
```diff
@@ -6,12 +6,29 @@ load("@rules_nodejs//nodejs/private:ts_project.bzl", _ts_project_lib = "ts_proje
...
+    )
+
+def _ts_project_impl(ctx):
+    return _ts_project_lib.implementation(ctx, run_action = _run_tsc)
+
 ts_project = rule(
-    implementation = _ts_project_lib.implementation,
-    attrs = _ts_project_lib.attrs,
+    implementation = _ts_project_impl,
+    attrs = dicts.add(_ts_project_lib.attrs, {"link_workspace_root": attr.bool()}),
```
alexeagle, line 33: I don't think we want a feature like link_workspace_root anymore - but I see, you're using the run_action argument to ts_project's implementation which expects such an attribute.
I think maybe we still want to attach the linker to the deps attribute using an aspect, like build_bazel_rules_nodejs does with the module_mappings aspect.
Not sure though, we should get @gregmagolan in here

alexeagle (commented):
nice job making a short PR for this so we can reason about it

example/BUILD.bazel
thesayyn: not really. just wanted to make sure that it works both ways.

js/merge_deps.bzl
thesayyn: rules authors will need this helper so I am not sure where it belongs. I have the feeling that it should be considered a public API

js/merge_deps.bzl
thesayyn: oh, thanks.

js/private/nodejs_binary.bzl
thesayyn: Yes, we will need that. I'll make another pass on whole thing

linker/BUILD.bazel
thesayyn: Kind of yes. the reason I gave it the name is because rule authors will want to invoke it at a given time so it has to be reachable to the users. One might add a dependency on it like `@aspect_rules_js//linker` since it is released with the rules itself rather than on npm.  One use case that comes to my mind is a rule that has worker mode enabled thus has to invoke this to link again.

Honestly, this linker package is just a tiny function that links from `MERGED_NODE_MODULES`  to `pwd/node_modules`. Basically `ln -s $MERGED_NODE_MODULES $(pwd)/node_modules`.  I am not sure if we should have this at all.

thesayyn: I was thinking maybe we should get rid of NODE_PATH and change the laucher script to link from MERGED_NODE_MODULES which would point to `runfiles/node_modules` for statically linked modules and `bazel-out/node_modules` for dynamically linked modules (merged via merge_deps.bzl hence it has all needed dependencies)

so in both cases, we just rely on bazel to create the node_modules tree and then we symlink that directory next to CWD.

rollup/rollup.js
thesayyn: yes. I just need to change the linker a bit to call the linker function when it is being called by node directly rather than being imported.

ts/private/ts_project.bzl
thesayyn: Yeah. nodejs doesn't like it either.

example/BUILD.bazel
thesayyn: also, it is a standalone `npm_import` so it is not part of the @npm_deps repo.

alexeagle:
replaced by https://github.com/aspect-build/rules_js/pull/31

PR #19, opened 2022-02-27
perf: use merge_all to avoid deeply nested runfiles depsets

mattem:
Prefer to use `merge_all` when merging runfiles in a loop to avoid creating deeply nested depset.

See the docs for more info: https://bazel.build/rules/lib/runfiles#merge_all


alexeagle (approved):
Nice Catch

PR #21, opened 2022-03-04
[dev branch] TS gazelle extension

alexeagle:
This will be a long-lived feature branch while we get to a working MVP


sluongng:
Hi folks,

Just a drive by comment but I thought you guys might be interested in my proposal over here https://github.com/bazelbuild/bazel-gazelle/issues/1189.

Thanks for creating this 🤗

gazelle/tests/simple_import_generated/BUILD.out
```diff
@@ -0,0 +1,14 @@
...
+genrule(
+    name = "generator",
+    outs = ["genlib.js"],
+    cmd = "echo 'export const ANSWER = 42;' > $@",
+)
+
+ts_project(
+    name = "simple_import_generated",
+    srcs = ["main.ts"],
+    deps = ["//:genlib.js"],
```
alexeagle, line 13: in reality this isn't allowed: deps of ts_project must provide DeclarationInfo

alexeagle (commented):
very confusing, but in TypeScript `rootDirs` is not the plural of `rootDir` - they have totally different meanings.
`rootDir` affects what files go in the program
`rootDirs` is for the module resolver to apply an extra pathmapping to treat two different directories as logical siblings.

It's a problem where we can develop here and have wrong typescript semantics, and nothing tells us. I think we need to extend the test harness so you also ask TypeScript whether the tsconfig/files in a given test fixture have the result you expect.

alexeagle:
Now that we've decided to split TypeScript into its own ruleset, we should move this to aspect-build/rules_ts - I'll start on it.

jbedard:
Replaced with https://github.com/aspect-build/rules_ts/pull/4

PR #22, opened 2022-03-26
refactor: allow more reuse of nodejs_binary starlark lib

alexeagle:
I'm using this for a new jest_test rule which wants to return a similar launcher binary but control the args


kormide (approved):
👍

bazaglia:
@alexeagle Do you have an example of what your jest_test rule looks like? Trying to achieve the same in rules_js.

alexeagle:
@bazaglia I have some code started in a private rules_jest repo, but haven't made progress on it in a while. I guess I could make it public, or just invite you for now

bazaglia:
@alexeagle That'd be great. If you invite me I might be able to help you with a PR.

PR #23, opened 2022-03-28
Allow users to reference the repositories.bzl file

### LINKED ISSUE #10: Allow for translate_package_lock to store result in the repo
Rather than dynamically create `@some_external_repo//:repositories.bzl` for the user to then `load()` and call from their WORKSPACE, we could act like rules_go's [update-repos command](https://github.com/bazelbuild/bazel-gazelle/#update-repos).

We would write the `npm_import` calls into the WORKSPACE or some designated helper macro.

This has a couple advantages:
- user can hand-modify the result
- WORKSPACE install is simpler so if you're a ruleset you might want to make your setup harder in exchange for making users setup easier
- Workaround for https://github.com/bazelbuild/bazel/issues/14445 (not sure yet if that's a bug/missing feature in bzlmod)

alexeagle:
This way they can use write_source_files to vendor it into their repo.
https://github.com/aspect-build/rules_swc/compare/bzl shows how it can be used

Fixes #10


mattem (approved):
Personal perference I guess, but I prefer if docs don't use the language "you can do x", 

eg:
> You can immediately load from the generated repositories.bzl file in your WORKSPACE.

 vs

> Immediately load from the generated repositories.bzl file in the WORKSPACE.

Or 

> You can check in the repositories.bzl file in your repo

vs

> Check in the repositories.bzl file into the repo

I'm not strongly opinionated either way, just think it reads better for technical documentation.

PR #25, opened 2022-04-01
feat: apply patches through translate_package_lock

kormide:


WORKSPACE
```diff
@@ -53,6 +53,16 @@ npm_import(
...
+    patch_args = {
+        "@gregmagolan/test-a@0.0.1": [
+            "-p1",
+        ],
+    },
+    patches = {
+        "@gregmagolan/test-a@0.0.1": [
+            "//example:test-a.patch",
+        ],
+    },
```
kormide, line 65: Planning to also support wildcards `"*"` to patch everything (ngcc case), and the package name minus the version to patch all version. But will do that in a separate PR.

js/npm_import.bzl
```diff
@@ -1,6 +1,7 @@
 "repository rules for importing packages from npm"
 
 load("//js/private:translate_package_lock.bzl", lib = "translate_package_lock")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "patch")
```
kormide, line 4: This is what `http_archive` uses and unlike the `repository_ctx` is supports `patch_args` for git patches as well as other stuff.

js/private/translate_package_lock.bzl
```diff
@@ -105,7 +116,7 @@ _NPM_IMPORT_TMPL = """\
         integrity = "{integrity}",
         package = "{package}",
         version = "{version}",
-        deps = {deps},
+        deps = {deps},{maybe_patches}{maybe_patch_args}
```
kormide, line 119: A bit awkward, but I didn't want to output `patches = [...]` and `patch_args = [...]` if there weren't any.

alexeagle (approved):

js/private/translate_package_lock.bzl
alexeagle: yeah I was commenting about it but came to the same conclusion, sometimes the generated thing will be read

WORKSPACE
kormide: Actually I think the wildcard is only applicable for postinstall, not patches.

kormide:
I added a second commit to support patching all versions of a particular package.

gregmagolan (approved):
Beauty

js/private/translate_package_lock.bzl
```diff
@@ -140,13 +155,25 @@ def npm_repositories():
         if "requires" in dep.keys():
             for n in dep["requires"].keys():
                 deps.append("@" + _repo_name(n, [d["version"] for (p, d) in lockfile["dependencies"].items() if p == n][0]))
+
+        name_with_version = _repo_name_and_version(name, dep["version"])
+
+        patches = patches_map.get(name, [])[:]
+        patches.extend(patches_map.get(name_with_version, []))
+
+        patch_args = patch_args_map.get(name, [])
```
gregmagolan, line 164: nit: I would make this symmetrical here with
```
patch_args.extend(patch_args.get(name_with_version, []))
```
Not likely to be used but it keeps the semantics for the two dicts the same

kormide: Updated

gregmagolan (approved):
:taco:

PR #27, opened 2022-04-04
Demonstrate running ts_project with no npm install

alexeagle:
Based on refactoring in https://github.com/bazelbuild/rules_nodejs/pull/3386

Duplicates the wrapper macro code from rules_nodejs, but the rest of the starlark implementation can now be shared.

- The typescript dependency is a simple http_archive, there is no npm_install, no `@bazel` npm packages, and no linker.
- It's impossible to cause an eager load of npm packages as reported in https://github.com/bazelbuild/rules_nodejs/issues/3193
- Depends only on the rules_nodejs "core" package, not `build_bazel_rules_nodejs`.
- To control the TS version, a user could trivially define their own `@typescript` dependency before calling our dependencies helper.

This won't type-check interesting user code yet, because there's no dynamic linker so typescript will fail to resolve 3p libraries the user depends on.


js/repositories.bzl
```diff
@@ -44,3 +47,11 @@ def rules_js_dependencies():
         #url = "https://github.com/aspect-build/bazel-lib/archive/refs/tags/v{}.tar.gz".format(versions.bazel_lib),
         url = "https://github.com/aspect-build/bazel-lib/archive/{}.tar.gz".format(versions.bazel_lib),
     )
+
+    maybe(
+        http_archive,
+        name = "typescript",
+        build_file = "//ts:BUILD.typescript",
+        sha256 = "70d5d30a8ee92004e529c41fc05d5c7993f7a4ddea33b4c0909896936230964d",
+        urls = ["https://registry.npmjs.org/typescript/-/typescript-4.6.3.tgz"],
```
jbedard, line 56: This means the version is locked with the version of rules_js?

alexeagle: No, take a look at the commit description, already preemptive answered that 😜

jbedard: Genius 🤦‍♂️

kormide (approved):

ts/BUILD.typescript
```diff
@@ -0,0 +1,42 @@
...
+
+nodejs_package(
+    name = "npm_typescript",
+    src = "_npm_typescript",
+    package_name = "typescript",
+    visibility = ["//visibility:public"],
+    deps = [],
+)
+
+write_file(
```
jbedard, line 20: Why is this required?

ts/BUILD.typescript
```diff
@@ -0,0 +1,42 @@
...
+    src = "package",
+    out = "typescript",
+)
+
+nodejs_package(
+    name = "npm_typescript",
+    src = "_npm_typescript",
+    package_name = "typescript",
+    visibility = ["//visibility:public"],
+    deps = [],
```
jbedard, line 17: Is this required? 🤷

alexeagle: no, the empty deps line is not meaningful, just came from copying this out of a generated file, will remove

ts/BUILD.typescript
alexeagle: the new `nodejs_binary` here doesn't support `$(rootpath)` expansions so we can't point it directly at the `external/npm_typescript/.../tsc.js` file.

Could be removed in the future without change to user API

ts/private/ts_project.bzl
```diff
@@ -0,0 +1,177 @@
+"""ts_project implementation
+
+Nearly identical to the ts_project wrapper macro in rules_nodejs.
```
jbedard, line 3: Is there a 1-line summary of what is different from rules_nodejs that could be added here?

alexeagle: yeah good suggestion

jbedard (approved):

PR #30, opened 2022-04-08
feat: simple postinstall implementation

kormide:


js/private/postinstall.bzl
```diff
@@ -0,0 +1,27 @@
+load("@aspect_bazel_lib//lib:run_binary.bzl", "run_binary")
+load("@aspect_bazel_lib//lib:copy_directory.bzl", "copy_directory")
+
+def postinstall(name, src, package_name):
```
alexeagle, line 4: Is this macro worth the extra indirection, vs. in-line it, just print both targets in the generated build file?

js/private/postinstall.js
```diff
@@ -0,0 +1,57 @@
+const {mkdirSync, copyFileSync, readdirSync, readFileSync, existsSync, statSync} = require('fs');
```
alexeagle, line 1: Did you look at existing implementation of lifecycle in npm cli?

js/private/postinstall.bzl
kormide: Good point. This started out as a rule rather than a macro, but given how simple it is it makes sense to inline it.

js/private/postinstall.js
kormide: I didn't think to check. I'll do that. I realize the implementation will need to become a lot more complex, but just wanted to do a simple pass first.

js/npm_import.bzl
```diff
@@ -27,23 +27,32 @@ def _npm_import_impl(repository_ctx):
...
     repository_ctx.file("BUILD.bazel", """
-load("@aspect_bazel_lib//lib:copy_directory.bzl", "copy_directory")
 load("@aspect_rules_js//js:nodejs_package.bzl", "nodejs_package")
+load("@aspect_rules_js//js/private:postinstall.bzl", "postinstall")
 
-# Turn a source directory into a TreeArtifact for RBE-compat
-copy_directory(
-    # The default target in this repository
-    name = "_{name}",
+postinstall(
```
kormide, line 47: TODO: repository rule should short circuit this if there's no postinstall.

kormide:
I inlined the postinstall steps and short-circuited the inclusion of those steps. Going to look into what the npm cli does now.

js/private/postinstall.js
kormide: So the npm cli ends up calling a spawn like we do here, but they also add node_modules' `.bin` dirs to the path, and they have some windows-specific logic around escaping arguments and how the path gets set, whether `cmd.exe` gets called, etc.

My plan was to work on adding `.bin`s to the PATH next. Are you okay with this simpler version going in first?

kormide:
TODO: I should add a test or something to the example that exercises the postinstall in `qar`.

js/private/postinstall.js
alexeagle: yes, just wanted you to get some context from their impl to inform TODOs, comments we could add here about design, etc.

I'd at least suggest adding some pointers to what you found, so that when we read this code later we can compare with npm

js/npm_import.bzl
```diff
@@ -27,23 +27,50 @@ def _npm_import_impl(repository_ctx):
...
+            result = repository_ctx.execute(["mv", "extract_tmp/%s" % nested_folder, "extract_tmp/%s" % new_nested_folder])
+        if result.return_code:
+            fail("failed to rename nested package folder that had spaces: \nSTDOUT:\n%s\nSTDERR:\n%s" % (result.stdout, result.stderr))
+        nested_folder = new_nested_folder
+
+    package_path = "extract_tmp" + _sep(repository_ctx) + nested_folder
+    has_posinstall = _has_postinstall(repository_ctx, package_path)
+    build_content = ""
+
+    if has_posinstall:
```
alexeagle, line 47: nit: spelling

js/npm_import.bzl
```diff
@@ -159,3 +190,13 @@ translate_package_lock = repository_rule(
     implementation = lib.implementation,
     attrs = lib.attrs,
 )
+
+def _has_postinstall(repository_ctx, package_path):
```
alexeagle, line 194: there is also preinstall
https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts
but I don't think we need to cover everything yet

alexeagle (commented):
looks good, I think @gregmagolan should give the approve though

PR #31, opened 2022-04-12
test: add e2e tests for node_package (along with translate_package_lock and npm_import)

gregmagolan:


.bazelrc
```diff
@@ -1,10 +1,5 @@
-# Bazel settings that apply to this repository.
-# Take care to document any settings that you expect users to apply.
-# Settings that apply only to CI are in .github/workflows/ci.bazelrc
-
-# Recommended when using copy_directory, see
-# https://github.com/aspect-build/bazel-lib/blob/main/docs/copy_directory.md
-startup --host_jvm_args=-DBAZEL_TRACK_SOURCE_DIRECTORIES=1
+# import common bazelrc shared with e2e workspaces
+import %workspace%/.bazelrc.common
```
alexeagle, line 2: did you add a link to that .bazelrc.common in some docs? does it end up in users file somehow?

.bazelrc.common
```diff
@@ -0,0 +1,26 @@
+# Bazel settings that apply to this repository.
```
alexeagle, line 1: may as well include the other standards like strict action env

.bazelrc.common
```diff
@@ -0,0 +1,26 @@
...
+#   Bazel needs to create symlinks to build the runfiles tree.
+#   Creating symlinks on Windows requires one of the following:
+#       1. Bazel is run with administrator privileges.
+#       2. The system version is Windows 10 Creators Update (1703) or later
+#          and developer mode is enabled.
+build --enable_runfiles
+
+# Turn off legacy external runfiles
+# This prevents accidentally depending on this feature, which Bazel will remove.
+build --nolegacy_external_runfiles
```
alexeagle, line 26: not all users can turn this on since it doesn't work with some rules. I'd recommend not telling users to do this - we should just enforce it on ourselves

WORKSPACE
```diff
@@ -38,38 +38,36 @@ nodejs_register_toolchains(
...
 
 # Manually import a package using explicit coordinates.
 # Just a demonstration of the syntax de-sugaring.
 npm_import(
+    name = "example_npm_deps__acorn-8.4.0",
     integrity = "sha512-ULr0LDaEqQrMFGyQ3bhJkLsbtrQ8QibAseGZeaSUiT/6zb9IvIkomWHJIvgvwad+hinRAgsI51JcWk2yvwyL+w==",
-    package = "acorn",
-    version = "8.4.0",
-    deps = [],
+    package_name = "acorn",
```
alexeagle, line 48: these renames don't seem any better to me

WORKSPACE
```diff
@@ -38,38 +38,36 @@ nodejs_register_toolchains(
...
-# Read the package-lock.json file to automate creation of remaining npm_import rules
-translate_package_lock(
-    name = "npm_deps",
-    package_lock = "//example:package-lock.json",
+load("@aspect_rules_js//js:translate_pnpm_lock.bzl", "translate_pnpm_lock")
+
+# Read the pnpm-lock.json file to automate creation of remaining npm_import rules
+translate_pnpm_lock(
+    name = "example_npm_deps",
+    # yq -o=json -I=2 '.' pnpm-lock.yaml > pnpm-lock.json
```
alexeagle, line 58: yq landed in bazel-lib so we could adopt now?

e2e/js_package/.bazelrc
```diff
@@ -0,0 +1,2 @@
+# import common bazelrc shared with e2e workspaces
+import %workspace%/../../.bazelrc.common
```
alexeagle, line 2: oh that answers my earlier question. maybe a comment over there about where else the bazelrc.common is used

e2e/js_package/.npmrc
```diff
@@ -0,0 +1,2 @@
+node-linker=pnp
```
alexeagle, line 1: any comments about how this file participates in the example? do we read it? does npm read it?

e2e/js_package/WORKSPACE
```diff
@@ -0,0 +1,46 @@
...
+load("@rules_nodejs//nodejs:repositories.bzl", "nodejs_register_toolchains")
+
+nodejs_register_toolchains(
+    name = "node16",
+    node_version = "16.9.0",
+)
+
+load("@aspect_rules_js//js:translate_pnpm_lock.bzl", "translate_pnpm_lock")
+
+translate_pnpm_lock(
```
alexeagle, line 19: we probably need to update the .github/workflows/workspace_snippet ?

js/private/translate_pnpm_lock.bzl
```diff
@@ -0,0 +1,364 @@
...
+    if link_package == ".":
+        link_package = rctx.attr.pnpm_lock.package
+
+    direct_dependencies = lockfile.get("dependencies")
+    packages = lockfile.get("packages")
+
+    repositories_bzl = [
+        """load("@aspect_rules_js//js:npm_import.bzl", "npm_import")""",
+        "",
+        "def npm_repositories():",
```
alexeagle, line 265: needs `# buildifier: disable=function-docstring` or else actually add a docstring

Otherwise I get yelled at when I vendor the resulting generated starlark file

js/private/npm_utils.bzl
```diff
@@ -0,0 +1,46 @@
...
+    "Make a bazel friendly name from a package name and a version that can be used in repository and target names"
+    escaped = name.replace("/", "_").replace("@", "at_")
+    version_escaped = _normalize_version(version)
+    return "%s_%s" % (escaped, version_escaped)
+
+def _normalize_version(version):
+    "Make a bazel friendly version information"
+
+    # 21.1.0_rollup@2.70.2 becomes 21.1.0_rollup_2.70.2
+    return version.replace("@", "_")
```
stoiky, line 13: I have another example where in case of 2 peer dependencies they are concatenated with a `+`

e.g.
```
/mobx-react/5.4.3_mobx@5.10.1+react@17.0.1
/@babel/eslint-parser/7.17.0_@babel+core@7.14.3+eslint@7.12.1
```

thesayyn: Good catch. Probably the case of two peer dependencies.

thesayyn: Pushed a new commit for this.

js/private/npm_import.bzl
```diff
@@ -0,0 +1,248 @@
...
+        msg = "tar %s failed: \nSTDOUT:\n%s\nSTDERR:\n%s" % (dirname, result.stdout, result.stderr)
+        fail(msg)
+
+    rctx.file("BUILD.bazel", "exports_files([\"{dir}\"])".format(dir = dirname))
+
+    deps = []
+    for dep in rctx.attr.deps:
+        dep_split = dep.split("@")
+        dep_name = "@".join(dep_split[:-1])
+        dep_version = dep_split[-1]
```
stoiky, line 149: This doesn't work in the context of peer dependencies that have multiple `@`s (more than 2)

now:
```
============
Original: @babel/plugin-syntax-class-static-block@7.14.5__babel_core_7.14.3
Name: @babel/plugin-syntax-class-static-block
Version: 7.14.5__babel_core_7.14.3
============
Original: eslint-config-prettier@6.15.0_eslint@7.12.1
Name: eslint-config-prettier@6.15.0_eslint
Version: 7.12.1
```

Should be
```
============
Original: eslint-config-prettier@6.15.0_eslint@7.12.1
Name: eslint-config-prettier
Version: 6.15.0_eslint@7.12.1
============
Original: @typescript-eslint/parser@4.13.0_eslint@7.12.1+typescript@4.5.4
Name: @typescript-eslint/parser
Version: 4.13.0_eslint@7.12.1+typescript@4.5.4
```

Maybe something like:
```
        scope = ""
        if dep.startswith("@"):
            scope = "@"
            dep = dep[1:]
        dep_split = dep.partition("@")
        dep_name = scope + dep_split[0]
        dep_version = dep_split[2]
```

js/private/npm_utils.bzl
```diff
@@ -0,0 +1,47 @@
+"Utility functions for npm rules"
+
+def _bazel_name(name, version):
```
mrmeku, line 3: This function works as a blacklist of certain symbols that we don't want to be in a bazel name. I think this function should be written to work as a whitelist instead since the number of valid characters that are valid for bazel is very small and well known. 

As one example implementation, you could iterate over the characters of the name. If the character is valid keep it, if it matches a hard coded set of transformations you maintain, transform it, otherwise replace it with a default accepted character

js/run_js_binary.bzl
```diff
@@ -0,0 +1,146 @@
...
+        silent_on_success = True,
+        copy_srcs_to_bin = True,
+        **kwargs):
+    """Wrapper around @aspect_bazel_lib run_binary that adds convienence attributes for using a js_binary tool.
+
+    This rule does not require Bash `native.genrule`.
+
+    Args:
+        name: Target name
+        tool: The tool to run in the action.
```
mrmeku, line 31: Consider renaming to binary to be more in line with the doc string and self documenting

js/run_js_binary.bzl
```diff
@@ -0,0 +1,146 @@
...
+        _copy_to_bin(
+            name = copy_to_bin_name,
+            srcs = srcs,
+            tags = kwargs.get("tags"),
+        )
+        extra_srcs = [":%s" % copy_to_bin_name]
+
+    # Automatically add common and useful make variables to the environment for js_binary targets
+    # under rules_js
+    extra_env = {
```
mrmeku, line 104: I'm a bit surprised these env variables aren't provided by the upstream run_binary rule. Should they be?

js/private/js_binary.sh.tpl
```diff
@@ -0,0 +1,260 @@
...
+function normalize_windows_path {
+    # Apply the followings paths transformations to normalize paths on Windows
+    # -process driver letter
+    # -convert path separator
+    # -lowercase everything
+    echo $(sed -e 's#^\(.\):#/\L\1#' -e 's#\\#/#g' -e 's/[A-Z]/\L&/g' <<< "$1")
+    return
+}
+
+# Find our runfiles as ${PWD}/${RUNFILES_DIR} is not always correct.
```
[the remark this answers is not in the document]
gregmagolan: `RUNFILES_DIR` needs a bit of mutation to be useful. It varies between relative and absolute pat for one and we need it to be absolute. I think we can trim down the bash here tho and find the minimal transformation. Will leave a TODO.

.bazelrc
gregmagolan: broken out for use in e2e workspaces; wasn't targeting users with this one

.bazelrc.common
gregmagolan: yeah; we should align all of our rulesets and create tooling to keep things like this consistent across repos

WORKSPACE
gregmagolan: @kormide is on it

js/repositories.bzl
```diff
@@ -33,6 +35,16 @@ def rules_js_dependencies():
         ],
     )
 
+    # TEMP TEMP >>> rules_nodejs overide
+    maybe(
+        http_archive,
+        name = "rules_nodejs",
+        sha256 = "48146434180db3f5be9be0890d58cf3250cc81acc652a04816aea0c0d06cfbd9",
+        strip_prefix = "rules_nodejs-cd48e24da0f44b9f49cb4b0254a8747b987970fe",
+        url = "https://github.com/gregmagolan/rules_nodejs/archive/cd48e24da0f44b9f49cb4b0254a8747b987970fe.tar.gz",
```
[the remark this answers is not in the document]
gregmagolan: its got changes to ts_project lib; we can fix it upstream to make it conditional depending on which rule set is using it.... or we could just fork it

js/test/BUILD.bazel.bak
```diff
@@ -0,0 +1,3 @@
+load(":translate_pnpm_lock_tests.bzl", "translate_pnpm_lock_tests")
```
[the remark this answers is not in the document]
gregmagolan: disabled these tests during development; should be re-enabled before landing

ts/BUILD.typescript
```diff
@@ -38,9 +31,15 @@ js_binary(
...
+    name = "ts_project_options_validator_js",
+    src = "@rules_nodejs//nodejs/private:ts_project_options_validator.js",
+    out = "ts_project_options_validator.js",
+)
+
 js_binary(
     name = "validator",
     data = ["npm_typescript"],
-    entry_point = "@rules_nodejs//nodejs/private:ts_project_options_validator.js",
+    entry_point = ":ts_project_options_validator.js",
```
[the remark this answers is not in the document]
gregmagolan: super problematic to have an entry point in some random external repository where node_modules resolution doesn't work from

js/run_js_binary.bzl
gregmagolan: Hmmm. That diverges from the upstream `run_binary` which this is meant to wrap so might be more confusing that way. Not sure. Wtdy @alexeagle .

It would be good to be able to assert that the tool is indeed a js_binary as well.

js/private/npm_import.bzl
gregmagolan: Thanks for the suggestion. I've some time today finally to understand the proper shape for peer deps in this rule set. Will reply back here when I find it.

js/run_js_binary.bzl
alexeagle: yeah we should stay similar to run_binary.

This isn't JS-specific really, except the placeholder naming

gregmagolan: we'll use this wrapper for the generated "bin" build actions

dragonsinth:
@kormide when I try to use this (44ed021) from my own repository, it dies trying to find `yq`:

```
ERROR: /Users/scottb/src/mn/projects/fullstory/WORKSPACE:137:20: fetching translate_pnpm_lock rule //external:npm: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_scottb/befbf4792f1e89f7a96261537dac474e/external/aspect_rules_js/js/private/translate_pnpm_lock.bzl", line 252, column 33, in _impl
		lockfile = _process_lockfile(rctx)
	File "/private/var/tmp/_bazel_scottb/befbf4792f1e89f7a96261537dac474e/external/aspect_rules_js/js/private/translate_pnpm_lock.bzl", line 164, column 34, in _process_lockfile
		result = rctx.execute([yq_bin(rctx, rctx.attr.yq_repository), "-o=json", ".", rctx.path(rctx.attr.pnpm_lock)])
	File "/private/var/tmp/_bazel_scottb/befbf4792f1e89f7a96261537dac474e/external/aspect_rules_js/js/private/yq.bzl", line 17, column 34, in yq_bin
		content = rctx.read(rctx.path(Label("@%s_host//:index.bzl" % yq_repository)))
Error in path: Unable to load package for @yq_host//:index.bzl: The repository '@yq_host' could not be resolved: Repository '@yq_host' is not defined
```

Am I supposed to do some additional setup in my own WORKSPACE that isn't covered by `js_dependencies()`?

kormide:
> @kormide when I try to use this ([44ed021](https://github.com/aspect-build/rules_js/commit/44ed021daada2d42aaeb2e3524369589b12e7ee7)) from my own repository, it dies trying to find `yq`:
> 
> ```
> ERROR: /Users/scottb/src/mn/projects/fullstory/WORKSPACE:137:20: fetching translate_pnpm_lock rule //external:npm: Traceback (most recent call last):
> 	File "/private/var/tmp/_bazel_scottb/befbf4792f1e89f7a96261537dac474e/external/aspect_rules_js/js/private/translate_pnpm_lock.bzl", line 252, column 33, in _impl
> 		lockfile = _process_lockfile(rctx)
> 	File "/private/var/tmp/_bazel_scottb/befbf4792f1e89f7a96261537dac474e/external/aspect_rules_js/js/private/translate_pnpm_lock.bzl", line 164, column 34, in _process_lockfile
> 		result = rctx.execute([yq_bin(rctx, rctx.attr.yq_repository), "-o=json", ".", rctx.path(rctx.attr.pnpm_lock)])
> 	File "/private/var/tmp/_bazel_scottb/befbf4792f1e89f7a96261537dac474e/external/aspect_rules_js/js/private/yq.bzl", line 17, column 34, in yq_bin
> 		content = rctx.read(rctx.path(Label("@%s_host//:index.bzl" % yq_repository)))
> Error in path: Unable to load package for @yq_host//:index.bzl: The repository '@yq_host' could not be resolved: Repository '@yq_host' is not defined
> ```
> 
> Am I supposed to do some additional setup in my own WORKSPACE that isn't covered by `js_dependencies()`?

Yeah you'll need to add this to your workspace after `js_dependencies`:
```
load("//js:configure.bzl", "js_configure")

js_configure()
```

dragonsinth:
Ah thanks I was still doing `nodejs_register_toolchains`.  Interestingly, it looks like you have to use the default toolchain name or translate will be unhappy.

kormide:
> Ah thanks I was still doing `nodejs_register_toolchains`. Interestingly, it looks like you have to use the default toolchain name or translate will be unhappy.

Hm, sounds like a bug. Which toolchain name were you overriding?

gregmagolan:
> Ah thanks I was still doing `nodejs_register_toolchains`. Interestingly, it looks like you have to use the default toolchain name or translate will be unhappy.

You should be able to specify a different toolchain name to translate_pnpm_lock via the `yq_repository` attribute. It defaults to `yq`.

dragonsinth:
> > Ah thanks I was still doing `nodejs_register_toolchains`. Interestingly, it looks like you have to use the default toolchain name or translate will be unhappy.
> 
> Hm, sounds like a bug. Which toolchain name were you overriding?

For some reason, we were using `node16` in our `nodejs_register_toolchains`, so I tried to call:

```
js_configure(node_name = "node16", node_version = "16.13.0")
```
But the implementation of `translate_pnpm_lock` really expects you to leave it as the default name.  It just took me a minute to figured it out and remove the argument.

gregmagolan:
> But the implementation of `translate_pnpm_lock` really expects you to leave it as the default name. It just took me a minute to figure it out and remove the argument.

Oh I see, it is the node toolchain that you're tripping over. You can specify a non-default with the `node_repository` attribute on `translate_pnpm_lock`. It defaults to `nodejs`.

dragonsinth:
Ah, makes sense.  Sorry I've been muddling my way through a bit today.  I know the doc isn't quite up to where this branch is yet.

js/private/npm_import.bzl
gregmagolan: Sorted out. The correct shape for peer deps in now landed on main branch.

gregmagolan:
Replaced by #44

PR #34, opened 2022-04-13
fix: add startup selector to --host_jvm_args=-DBAZEL_TRACK_SOURCE_DIRECTORIES=1

gregmagolan:


.bazelrc
```diff
@@ -4,7 +4,7 @@
 
 # Recommended when using copy_directory, see
 # https://github.com/aspect-build/bazel-lib/blob/main/docs/copy_directory.md
---host_jvm_args=-DBAZEL_TRACK_SOURCE_DIRECTORIES=1
+startup --host_jvm_args=-DBAZEL_TRACK_SOURCE_DIRECTORIES=1
```
alexeagle, line 7: filed https://github.com/bazelbuild/bazel/issues/15245 cuz I couldn't find any other tracking bug for the root cause here

alexeagle (approved):

PR #38, opened 2022-04-19
feat: postinstall

kormide:
Uses pnpm's lifecycle package.

Some of the changes here are formatting changes caused by installing the pre-commit.


js/private/npm_utils.bzl
```diff
@@ -3,11 +3,11 @@
 def _bazel_name(name, version):
     "Make a bazel friendly name from a package name and a version that can be used in repository and target names"
     escaped = name.replace("/", "_").replace("@", "at_")
-    return "%s_%s" % (escaped, version)
+    return "%s_%s" % (escaped, version.replace("@", "at_").replace("+", "_"))
 
 def _versioned_name(name, version):
     "Make a developer-friendly name for a package name and version"
-    return "%s@%s" % (name, version)
+    return "%s@%s" % (name, version.replace("@", "at_").replace("+", "_"))
```
kormide, line 10: Using peer deps (the pnpm logger package) revealed a new case where invalid repository names were being created. This fixes it, but you might want to double check that what I'm doing here is in line with your intention for these methods.

.bazelrc.common
```diff
@@ -24,3 +24,12 @@ build --enable_runfiles
...
 build --nolegacy_external_runfiles
+
+# Turn on --incompatible_strict_action_env which was on by default
+# in Bazel 0.21.0 but turned off again in 0.22.0. Follow
+# https://github.com/bazelbuild/bazel/issues/7026 for more details.
+# This flag is needed to so that postinstall scripts can be executed
+# on the host.
+# See https://github.com/angular/angular/issues/27514.
+build --incompatible_strict_action_env
+run --incompatible_strict_action_env
```
alexeagle, line 35: Run is inherited from build

WORKSPACE
```diff
@@ -44,30 +44,42 @@ load("@aspect_rules_js//js:npm_import.bzl", "npm_import")
...
 )
 
 load("@aspect_rules_js//js:translate_pnpm_lock.bzl", "translate_pnpm_lock")
 
+translate_pnpm_lock(
+    name = "aspect_rules_js_lifecycle_npm_deps",
+    pnpm_lock = "//js/private/lifecycle:pnpm-lock.json",
+    # Avoid a circular dependency on the lifecycle hooks package,
+    # which is used when skip_lifecycle_hooks = False
+    skip_lifecycle_hooks = True,
```
alexeagle, line 60: Usually Booleans should have a true value in the positive case. So it should be enable_...=False

js/private/npm_import.bzl
```diff
@@ -75,31 +75,44 @@ def is_windows_os(rctx):
...
+        src = "{nodejs_package_src}",
         package_name = "{package_name}",
         package_version = "{package_version}",
-        visibility = ["//visibility:public"],{maybe_indirect}{maybe_deps}
+        deps = {deps},
+        visibility = ["//visibility:public"],{maybe_indirect}
+    )
+"""
+
+_RUN_LIFECYCLE_HOOKS_TMPL = """run_binary(
```
alexeagle, line 97: Nit: you can start the multiple line string with either a comment or a backslash so the nesting looks right

alexeagle (commented):
Cool, will let Greg approve

WORKSPACE
kormide: Nice, that makes the logic look cleaner as well.

example/BUILD.bazel
```diff
@@ -312,4 +312,33 @@ diff_test(
...
+# The @aspect-test/qar package's postinstall creates a data.json file
+
+write_file(
+    name = "write10",
+    out = "case10.js",
+    content = ["""
+const fs = require("fs");
+const path = require("path");
+const packageDir = process.argv[2];
+const content = JSON.parse(fs.readFileSync(path.join(packageDir, "data.json")));
```
gregmagolan, line 329: would be simpler to depend on `package("@aspect-test/qar"),` and do a
```
const content = require('@aspect-test/qar/data.json')
```
here

js/private/BUILD.bazel
```diff
@@ -44,8 +44,8 @@ bzl_library(
     srcs = ["npm_import.bzl"],
     visibility = ["//js:__subpackages__"],
     deps = [
-        "@aspect_bazel_lib//lib:repo_utils",
```
gregmagolan, line 47: ahh. looks like I don't have formatting setup. I'll get the formatting right on the branch so these don't show up on this diff

js/private/lifecycle/lifecycle-hooks.js
```diff
@@ -0,0 +1,78 @@
...
+    if (argv.length !== 2) {
+        console.error('Usage: node lifecycle-hooks.js [packageDir] [outputDir]')
+        process.exit(1)
+    }
+    const packageDir = argv[0]
+    const outputDir = argv[1]
+
+    copyPackageContents(packageDir, outputDir)
+
+    const packageJson = JSON.parse(
```
gregmagolan, line 22: This could be simplified to `const packageJson = require(path.join(packageDir, 'package.json'))`

js/private/lifecycle/lifecycle-hooks.js
```diff
@@ -0,0 +1,78 @@
...
+        readFileSync(path.join(packageDir, 'package.json'))
+    )
+
+    if (
+        packageJson.scripts?.preinstall ||
+        packageJson.scripts?.install ||
+        packageJson.scripts?.postinstall
+    ) {
+        const nodeModulesPath = path.join(outputDir, '..', 'node_modules')
```
gregmagolan, line 31: how does this pathing work out that node_modules is `$(outputDir)/../node_modules`

js/private/lifecycle/lifecycle-hooks.js
```diff
@@ -0,0 +1,78 @@
...
+        // If the package we're running postinstall on has no deps, then node_modules
+        // won't exist in the sandbox. The lifecycle functions provided by pnpm require
+        // it to exist, so create an empty folder if needed.
+        if (!existsSync(nodeModulesPath)) {
+            mkdirSync(nodeModulesPath)
+        }
+
+        // Runs preinstall, install, and postinstall hooks
+        await runPostinstallHooks({
+            pkgId: packageJson.name + '/' + packageJson.version,
```
gregmagolan, line 42: no guarantee that package.json has `name` and `version` fields so we should handle it gracefully if they don't. version can be defaulted to `0.0.0` (which I do for first party deps elsewhere). not sure what to do if name doesn't exist. what does `runPostinstallHooks` do with `pkgId`? is it load bearing somehow?

js/private/npm_import.bzl
```diff
@@ -75,31 +75,45 @@ def is_windows_os(rctx):
...
+        deps = {deps},
+        visibility = ["//visibility:public"],{maybe_indirect}
+    )
+"""
+
+_RUN_LIFECYCLE_HOOKS_TMPL = """# Run lifecycle hooks on the package
+    run_binary(
+        name = "{namespace}__{bazel_name}_postinstall",
+        srcs = ["@{rctx_name}//:{dir}"] + {deps},
+        args = [ "../../../$(execpath @{rctx_name}//:{dir})", "../../../$(@D)"],
```
gregmagolan, line 101: can you comment on why we prefix the paths with `../../..`?

js/private/npm_import.bzl
```diff
@@ -234,15 +256,33 @@ _ATTRS = {
...
     attrs = _ATTRS,
 )
+
+def _has_lifecycle_hooks(rctx, package_path):
+    pkg_json_path = package_path + _sep(rctx) + "package.json"
+    pkg_json = json.decode(rctx.read(pkg_json_path))
+    return "scripts" in pkg_json and (
+        "preinstall" in pkg_json["scripts"] or
+        "install" in pkg_json["scripts"] or
+        "postinstall" in pkg_json["scripts"]
```
gregmagolan, line 282: thinking ahead to when we will add user postinstall steps to npm_import, my thinking was we'd use a `bazel_postinstall` of `rules_js_postinstall` script so we don't have to worry about appending to an existing postinstall. but... pnpm lifecycle hooks tooling won't run that at all. so I think we have two options,

1) find a way to get pnpm lifecycle tooling to also run our user postinstall script

2) if there is a user specified postinstall and an existing postinstall, then rename the existing one to "npm_postinstall" and add the "bazel_postinstall" and then insert a "postinstall" that runs both in some standard and platform agnostic way: `"postinstall": "npm run npm_postinstall && npm run bazel_postinstall"`?

gregmagolan: although, it might be that the `&&` in a`"postinstall": "existing commands && user commands"` is understood by pnpm lifecycle hooks and is already platform agnostic so lets try that first before overcomplicating

gregmagolan: can you add custom user postinstall commands via `translate_pnpm_lock` and `npm_import` to this diff to prove it out?

js/private/lifecycle/lifecycle-hooks.js
kormide: TODO: Leave comment here explaining

js/private/lifecycle/lifecycle-hooks.js
```diff
@@ -0,0 +1,78 @@
...
+            pkgRoot: path.resolve(outputDir),
+            rawConfig: {
+                stdio: 'inherit',
+            },
+            rootModulesDir: path.resolve(nodeModulesPath),
+            unsafePerm: true,
```
kormide, line 48: TODO: double check what this does and comment

js/private/lifecycle/lifecycle-hooks.js
kormide: This actually isn't a real property. It's in their usage which is just out of date.

kormide:
Updated to run custom postinstall scripts. I brought in the `yq` toolchain to inject custom postinstall scripts into `package.json` which are then run by the lifecycle-hooks script.

Ready for review.

kormide:
I forgot about `postinstall_deps`. Since this is getting large, can I do that in a follow-up pr?

gregmagolan:
> I forgot about `postinstall_deps`. Since this is getting large, can I do that in a follow-up pr?

Sounds good. That part we know will work so lets do it as a follow-up.

WORKSPACE
```diff
@@ -50,6 +54,18 @@ npm_import(
     package_version = "8.4.0",
 )
 
+translate_pnpm_lock(
```
gregmagolan, line 57: suppose we'll need to make these available to user workspaces via a bzl file since they'll need to be initialized downstream

gregmagolan: we don't run the e2e workspaces yet in CI so you have to do it manually, but you can tease out the error by running
`bazel build ...` in `/Users/greg/aspect/rules/rules_js/e2e/node_package`

```
$ bazel build ...
ERROR: /Users/greg/aspect/rules/rules_js/e2e/node_package/BUILD.bazel:3:13: error loading package '@aspect_rules_js//js/private/lifecycle': Unable to find package for @aspect_rules_js_lifecycle_npm_deps//:package.bzl: The repository '@aspect_rules_js_lifecycle_npm_deps' could not be resolved: Repository '@aspect_rules_js_lifecycle_npm_deps' is not defined. and referenced by '//:npm__at_aspect-test_qar_4.0.0_postinstall'
ERROR: Analysis of target '//:npm__at_aspect-test_qar_4.0.0_postinstall' failed; build aborted: Analysis failed
```

WORKSPACE
```diff
@@ -62,10 +78,14 @@ translate_pnpm_lock(
     },
     # yq -o=json -I=2 '.' pnpm-lock.yaml > pnpm-lock.json
     pnpm_lock = "//example:pnpm-lock.json",
+    postinstall = {
+        "@aspect-test/qar": "echo 'moo' > cow.txt",
+        "@aspect-test/qar@4.0.0": "echo 'mooo' >> cow.txt",
```
gregmagolan, line 83: 😂  "moomooo"

example/BUILD.bazel
```diff
@@ -417,3 +417,57 @@ diff_test(
...
+write_file(
+    name = "write12",
+    out = "case12.js",
+    content = ["""
+const fs = require('fs')
+const path = require('path')
+const content = fs.readFileSync(path.join(process.argv[2], 'cow.txt'), 'utf-8')
+console.log(content)
+if (!/^moo\\s+mooo\\s*$/.test(content)) {
+    console.error("expected file to contain cow sounds")
```
gregmagolan, line 460: 😂

js/private/lifecycle/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
...
+load("//js:js_binary.bzl", "js_binary")
+load("@aspect_rules_js_lifecycle_npm_deps//:node_modules.bzl", "node_modules")
+load("@aspect_rules_js_lifecycle_npm_deps//:package.bzl", "package")
+
+node_modules()
+
+js_binary(
+    name = "lifecycle-hooks",
+    data = [
+        package("@pnpm/lifecycle"),
```
gregmagolan, line 12: You can make this `@aspect_rules_js_lifecycle_npm_deps//@pnpm/lifecycle` and avoid using the `package` function

js/private/lifecycle/lifecycle-hooks.js
```diff
@@ -0,0 +1,87 @@
...
+
+    if (
+        packageJson.scripts?.preinstall ||
+        packageJson.scripts?.install ||
+        packageJson.scripts?.postinstall ||
+        packageJson.scripts?._rules_js_postinstall
+    ) {
+        // The postinstalled copy of the package always one directory beneath the
+        // node_modules folder that npm_import copies into a package's output folder.
+        const nodeModulesPath = path.join(outputDir, '..', 'node_modules')
```
gregmagolan, line 35: perhaps `path.resolve(path.join(outputDir, '..', 'node_modules'))` here so `existsSync` and `mkdirSync` are using the resolve path instead of the path with a `..` segment?

js/private/lifecycle/lifecycle-hooks.js
```diff
@@ -0,0 +1,87 @@
...
+    )
+
+    if (
+        packageJson.scripts?.preinstall ||
+        packageJson.scripts?.install ||
+        packageJson.scripts?.postinstall ||
+        packageJson.scripts?._rules_js_postinstall
+    ) {
+        // The postinstalled copy of the package always one directory beneath the
+        // node_modules folder that npm_import copies into a package's output folder.
```
gregmagolan, line 34: nit: `// node_modules folder that the node_package dependency targets which are in the same bazel package output.`

js/private/npm_import.bzl
```diff
@@ -234,15 +265,57 @@ _ATTRS = {
...
+    "enable_lifecycle_hooks": attr.bool(
+        doc = """If true, runs lifecycle hooks declared in this package and the custom postinstall script if one exists.""",
+        default = True,
+    ),
     "experimental_reference_deps": attr.bool(
         doc = """Experimental reference deps allow dep to support circular deps between npm packages.
         This feature depends on dangling symlinks, however, which is still experimental in bazel,
         has issues with "host" and "exec" configurations, and does not yet work with remote exection.""",
     ),
+    "_yq_repository": attr.string(
```
gregmagolan, line 277: shouldn't be private incase the user has a different yq repository name

js/private/translate_pnpm_lock.bzl
```diff
@@ -303,6 +314,12 @@ def _impl(rctx):
         patch_args = rctx.attr.patch_args.get(name, [])[:]
         patch_args.extend(rctx.attr.patch_args.get(versioned_name, []))
 
+        postinstall = rctx.attr.postinstall.get(name)
+        if not postinstall:
+            postinstall = rctx.attr.postinstall.get(versioned_name)
+        elif rctx.attr.postinstall.get(versioned_name):
+            postinstall = "%s && %s" % (postinstall, rctx.attr.postinstall.get(versioned_name))
```
gregmagolan, line 321: hmm. maybe this will be conceptually simpler and more robust if we use different script names for each level
`_rules_js_postinstall_name`, `_rules_js_postinstall_versioned_name`, `_rules_js_postinstall_global` and we run then in the order `global`, `name`, `versioned_name` in the rule

WORKSPACE
gregmagolan: Not sure how to handle the `register_yq_toolchains`. Do we do this automatically for users in a setup macro or do we ask users to register explicitly. @alexeagle wtyt?

js/private/translate_pnpm_lock.bzl
kormide: Hm I'm mixed on this one. That would mean that each `npm_import` needs to take three postinstall attributes which seems less clean, especially if someone is calling it manually. I don't feel strongly though. Wdyt?

kormide:
Addressed feedback but there is one error when building the `e2e` project that I'm still tracking down.

WORKSPACE
```diff
@@ -5,14 +5,28 @@ workspace(
...
+
+load("@aspect_rules_js//js:npm_import.bzl", "npm_import", "translate_pnpm_lock")
+
+translate_pnpm_lock(
+    name = "aspect_rules_js_lifecycle_npm_deps",
+    # Avoid a circular dependency on the lifecycle hooks package,
+    # which is used when enable_lifecycle_hooks = True
+    enable_lifecycle_hooks = False,
+    pnpm_lock = "@aspect_rules_js//js/private/lifecycle:pnpm-lock.json",
+)
```
kormide, line 29: We vendor the files from the repo this creates in `//js/private/lifecycle`, and in `js_configure` we load the `npm_repositories` to install the lifecycle deps.

js/private/lifecycle/BUILD.bazel
```diff
@@ -0,0 +1,28 @@
...
+# Vendor files from translate_pnpm_lock in WORKSPACE so that the transitive
+# close of the lifecycle deps can be set up in js_configure()
+write_source_files(
+    name = "lifecycle_repo_vendored_files",
+    files = {
+        "repositories.bzl": "@aspect_rules_js_lifecycle_npm_deps//:repositories.bzl",
+        "node_modules.bzl": "@aspect_rules_js_lifecycle_npm_deps//:node_modules.bzl",
+        "package.bzl": "@aspect_rules_js_lifecycle_npm_deps//:package.bzl",
+    },
+)
```
kormide, line 28: These are the files we need to vendor.

js/private/translate_pnpm_lock.bzl
```diff
@@ -268,9 +279,10 @@ def _impl(rctx):
     node_modules_bzl_file = "node_modules.bzl"
     node_modules_header_bzl = []
     node_modules_bzl = [
-        """def node_modules():
+        """# buildifier: disable=function-docstring
```
kormide, line 282: I had to add this because we vendor this file and buildifier complains.

js/configure.bzl
```diff
@@ -0,0 +1,37 @@
...
+    nodejs_register_toolchains(
+        name = node_name,
+        node_version = node_version,
+    )
+    register_yq_toolchains(
+        name = yq_name,
+        version = yq_version,
+    )
+
+    # TODO: does this belong here?
```
gregmagolan, line 37: nit: stray comment

example/BUILD.bazel
```diff
@@ -427,3 +427,57 @@ diff_test(
...
+    process.exit(1);
+}
+"""],
+)
+
+js_test(
+    name = "test12",
+    args = ["$(location %s)" % package_dir("@aspect-test/qar")],
```
gregmagolan, line 478: nit: use `$(rootpath)` here as location is deprecated

js/private/translate_pnpm_lock.bzl
gregmagolan: Oh. You can differentiate the scope without having different npm_import attributes. But this seems better to discuss as a follow-up as it might be that `&&` is actually very robust and simpler.

gregmagolan (approved):
amazing! 🌮

js/private/translate_pnpm_lock.bzl
gregmagolan: Oh, I see what you mean in `npm_import`. That a good point. The alternative would be to make it a string_list but that adds complication. Guess we just need some test coverage for `&&` on different platforms.

PR #39, opened 2022-04-20
refactor: rename nodejs_binary, nodejs_test, nodejs_package to js_binary, js_test and js_package for brevity, disambiguation and to match rule set name

gregmagolan:


thesayyn (approved):

alexeagle:
SGTM as well

PR #43, opened 2022-04-25
vendor pnpm/lifecycle as a minified bundle rather than make users fet…

alexeagle:
…ch its deps


kormide (changes requested):
Can you remove the `translate_package_lock` from `WORKSPACE`?

kormide (approved):
:+1:

PR #44, opened 2022-04-25
feat: create a pnpm-style symlinked node_modules trees structure to r…

alexeagle:
…esolving node_modules


js/private/js_binary.sh.tpl
```diff
@@ -0,0 +1,275 @@
+#!{bash}
```
alexeagle, line 1: @gregmagolan can we get shellcheck running on this file? will probably catch bugs

js/private/yq.bzl
```diff
@@ -0,0 +1,24 @@
+"Utilities for using the yq toolchain in a repository rule"
```
alexeagle, line 1: @kormide does this file belong in bazel-lib instead?

js/private/yq.bzl
```diff
@@ -0,0 +1,24 @@
...
+        rctx: the repository context
+        yq_repository: the name of the yq toolchain repository
+
+    Returns:
+        Path to the yq binary
+    """
+
+    # Parse the resolved host platform from yq host repo //:index.bzl
+    content = rctx.read(rctx.path(Label("@%s_host//:index.bzl" % yq_repository)))
+    search_str = "host_platform=\""
```
alexeagle, line 18: gross! we shouldn't do this - instead the `yq_register_toolchains(name = "yq")` should produce a repository `@yq` which is for the host platform

kormide: I think this is a workaround for an issue @gregmagolan found when we tried making a target pointing to the resolved executable.

js/private/yq.bzl
kormide: Yeah this should go in bazel-lib.

js/private/yq.bzl
alexeagle: yeah probably that `rctx.path` doesn't follow symlinks - but we figured out a way around it in rules_python without ever trying to predict the repo name

kormide: Are you referring to https://github.com/bazelbuild/rules_python/blob/ae7a2677b3003b13d45bc9bfc25f1425bed5b407/python/pip_install/pip_repository.bzl#L45?

js/private/js_binary.sh.tpl
gregmagolan: `js_binary.tpl.sh`?

js/private/yq.bzl
gregmagolan: Instead of a repository with aliases we could just create a `@yq_host` or `@yq` repository that is a duplicate of `@yq_<platform>`. This is a bit of duplication on disk _but_ it makes the path to the yq for the host platform simple.

js/private/js_binary.sh.tpl
alexeagle: will do in a later PR, since this file isn't actually shell

gregmagolan (approved):
🌮

PR #48, opened 2022-04-26
test: add e2e tests for node_package (along with translate_package_lock and npm_import)

gregmagolan:
Adds a clean e2e test for the symlinked node_modules structure produced by nodejs_package in concert with translate_package_lock and npm_import. Also covers the case of a node_package and js_binary in an external repository (rules_foo).


.bazelignore
```diff
@@ -0,0 +1,2 @@
+e2e/node_package
```
alexeagle, line 1: I've instead followed the pattern of listing the nested workspaces in the root WORKSPACE to prevent traversing into them. Not sure which is better, maybe they're equivalent. Should we be consistent?

e2e/node_package/pnpm-lock.yaml
```diff
@@ -0,0 +1,76 @@
+lockfileVersion: 5.3
```
alexeagle, line 1: could you add a `.gitattributes` file so GH stops asking us to review these generated files?

alexeagle (approved):
nice!

e2e/node_package/rules_foo/npm_repositories.bzl
```diff
@@ -0,0 +1,100 @@
+"@generated by @aspect_rules_js//js/private:translate_pnpm_lock.bzl from pnpm lock file @rules_foo@rules_foo//foo:pnpm-lock.yaml"
```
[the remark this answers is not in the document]
gregmagolan: 🌮

.bazelignore
gregmagolan: I prefer here myself so WORKSPACE is cleaner

PR #50, opened 2022-04-26
Improve error message when we can't find nodejs toolchain

alexeagle:


kormide (approved):

alexeagle:
ugh it messes up the docs :(

alexeagle:
we don't depend on this anymore :)

PR #53, opened 2022-04-26
fix: don't instantiate __lc target of package unless there is a lifecycle hook

gregmagolan:
Plus some cleanup


alexeagle (approved):
cool, I think I saw an error go by yesterday related to `__lc` so I'm hoping this happens to fix mine too

PR #57, opened 2022-04-27
chore: cleanup & shellcheck of launcher bash

gregmagolan:


BUILD.bazel
```diff
@@ -10,13 +9,3 @@ gazelle(
     name = "gazelle",
     gazelle = "gazelle_bin",
 )
-
-bzl_library(
```
alexeagle, line 14: should be a `#gazelle:exclude` comment or it will come back

alexeagle (approved):

PR #60, opened 2022-04-27
Example of workspace packages

stoiky:


alexeagle:
I think we captured this now, we have a pnpm-workspace.yaml in the repo root and each example is its own npm package

PR #61, opened 2022-04-27
feat: create macros for packages with bin entries

thesayyn:


thesayyn:
There are two ways that one can use to call macros; 

```
load("@npm//rollup:index.bzl", "rollup")
rollup.rollup(kwargs)
```

```
load("@npm_rollup_2.70.2//:index.bzl", "rollup")
rollup(kwargs)
```

Using the macro through the aliased repository is unusual but there is no other way to alias function from another repository without using structs. We could have simply reexported the functions if we had that information inside translate_pnpm.

js/private/npm_import.bzl
```diff
@@ -427,5 +483,13 @@ def _has_lifecycle_hooks(pkg_json):
...
 
-def _parse_bins(pkg_json):
-    return pkg_json.get("bin", {})
+def _get_bin_entries(rctx, package, pkg_json):
+    if "bin" in pkg_json:
+        bin = pkg_json["bin"]
+
+        # https://docs.npmjs.com/cli/v7/configuring-npm/package-json#bin
+        if type(bin) != "dict":
+            name = pnpm_utils.binary_name(package)
```
gregmagolan, line 492: oh.. nice catch on the non-dict case

js/private/npm_import.bzl
```diff
@@ -427,5 +483,13 @@ def _has_lifecycle_hooks(pkg_json):
         "_rules_js_postinstall" in pkg_json["scripts"]
     )
 
-def _parse_bins(pkg_json):
-    return pkg_json.get("bin", {})
+def _get_bin_entries(rctx, package, pkg_json):
```
gregmagolan, line 486: rctx not used

example/BUILD.bazel
```diff
@@ -8,7 +8,8 @@ load("@example_npm_deps__acorn-8.4.0//:node_package.bzl", node_package_acorn = "
 load("//js:defs.bzl", "js_binary", "js_test")
 load("//js:node_package.bzl", "node_package")
 load("//js:run_js_binary.bzl", "run_js_binary")
-
+load("@example_npm_deps__rollup_2.70.2//:index.bzl", "rollup")
+load("@example_npm_deps//uvu:index.bzl", "uvu")
```
gregmagolan, line 12: 😎

example/BUILD.bazel
```diff
@@ -25,6 +26,46 @@ node_package(
 ###########################
 # Fixtures for tests in this file
 
+# Generated bin test as a build action
+rollup(
+    name = "rollup_help",
+    args = ["--help"],
+    # do not call node_package macro for rollup as we did it earlier in the BUILD file.
+    node_package = False,
```
gregmagolan, line 34: hmm. this seems like a smell here to stamp out a node_package or not

gregmagolan: yeah; looking closer we can drop that. users will have to stamp out the node_package with node_modules() typically and you should be able to use `rollup` here in any package so stamping everywhere wouldn't work anyway

gregmagolan:
Chatted with Alex and we settled on 

```
load("@npm//rollup:package_json.bzl", rollup_bin = "bin")
rollup_bin.rollup(kwargs) # a run_js_binary
rollup_bin.rollup_test(kwargs) # a js_test
rollup_bin.rollup_binary(kwargs) # a js_binary
```

gregmagolan (approved):
🌮

PR #69, opened 2022-04-28
feat: use @aspect_bazel_lib_host repo to determine label to yq binary

gregmagolan:


gregmagolan:
Going with another approach

PR #70, opened 2022-04-29
refactor: use standard mechanism for finding yq binary

kormide:


js/private/npm_import.bzl
```diff
@@ -404,9 +404,9 @@ _ATTRS = {
         doc = """If true, runs lifecycle hooks declared in this package and the custom postinstall script if one exists.""",
         default = True,
     ),
-    "yq_repository": attr.string(
-        doc = """The basename for the yq toolchain repository from @aspect_bazel_lib.""",
-        default = "yq",
+    "yq": attr.string(
```
kormide, line 407: Used a string because if label this will fail on windows.

gregmagolan: `attr.label` should be fine. The label here isn't evaluated in such a way to make sure it points to a real file.

js/private/repo_toolchains.bzl
```diff
@@ -0,0 +1,37 @@
...
+
+    # Parse the resolved host platform from node host repo //:index.bzl
+    content = rctx.read(rctx.path(Label("@%s_host//:index.bzl" % rctx.attr.node_repository)))
+    search_str = "host_platform=\""
+    start_index = content.index(search_str) + len(search_str)
+    end_index = content.index("\"", start_index)
+    host_platform = content[start_index:end_index]
+
+    # Return the path to the node binary
+    return rctx.path(Label("@%s_%s//:bin/node%s" % (rctx.attr.node_repository, host_platform, ".exe" if repo_utils.is_windows(rctx) else "")))
```
kormide, line 37: Once rnj is updated, it would be nice if we could just make a dumb function that takes the repo name and doesn't know about node or yq.

js/private/npm_import.bzl
kormide: Oh cool

js/private/translate_pnpm_lock.bzl
```diff
@@ -148,33 +147,22 @@ _ATTRS = {
         doc = """The basename for the node toolchain repository from @build_bazel_rules_nodejs.""",
         default = "nodejs",
     ),
-    "yq_repository": attr.string(
-        doc = """The basename for the yq toolchain repository from @aspect_bazel_lib.""",
-        default = "yq",
+    "yq": attr.string(
+        doc = """The label to the yq binary to use. If executing on a windows host, the .exe extension will be appended if missing.""",
```
gregmagolan, line 151: lets also check for `.cmd` and `.bat` which are also valid binary extensions on Windows

js/private/repo_toolchains.bzl
```diff
@@ -0,0 +1,37 @@
...
+def yq_path(rctx):
+    """Find the path to the host resolved yq binary from a repository rule.
+
+    Args:
+        rctx: The repository rule context
+
+    Returns:
+        Path to yq
+    """
+    return rctx.path(Label("{yq}{maybe_ext}".format(
```
gregmagolan, line 14: You can use
```
result = rctx.path(rctx.attr.yq)
```
without issue on Windows even if that files is not valid and then append .exe to `result`

js/private/translate_pnpm_lock.bzl
kormide: You mean check for the existence of `.exe`, then `.bat` if it doesn't exist, and so forth?

js/private/repo_toolchains.bzl
kormide: Actually, maybe I'll just do that now and I'll leave the node path out of this until rnj is ready.

js/private/repo_toolchains.bzl
kormide: It looks like you can't add a string to a Path.

js/private/repo_toolchains.bzl
alexeagle: yeah but I think they're necessarily different, because we need to add the _host on it

`rctx.path(Label("@{node}_host//:bin/node{maybe_ext}".format(`

alexeagle: I guess they don't have to be though. You could change translate_pnpm_lock.bzl to use

```
"node": attr.string(
        doc = """The label to the node binary to use.""",
        default = "@nodejs_host//:bin/node",
    ),
    "yq": attr.string(
        doc = """The label to the yq binary to use. If executing on a windows host, the .exe extension will be appended if missing.""",
        default = "@yq//:yq",
    ),
```

and then the logic is the same

kormide: Mm, yeah.

kormide:
Updated

js/private/translate_pnpm_lock.bzl
gregmagolan: Yeah, if windows and extension is already `.exe`, `.bat` or `.cmd` then don't append `.exe`

kormide: Oh, I thought you meant run several `rctx.execute`s to find out which one actually exists.

js/private/repo_toolchains.bzl
gregmagolan: right, rctx.path returns a Path object. cool cool

js/private/repo_toolchains.bzl
```diff
@@ -0,0 +1,36 @@
+"Helpers to locate toolchains from within repository rules"
+
+load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
+
+def yq_path(rctx):
+    """Find the path to the host resolved yq binary from a repository rule.
+
```
gregmagolan, line 7: nit: document here that `.exe` is append on windows if extension is not one of `.exe`, `.bat` and `.cmd`

kormide: Documented below in return value.

js/private/repo_toolchains.bzl
```diff
@@ -26,13 +28,5 @@ def node_path(rctx):
...
-    content = rctx.read(rctx.path(Label("@%s_host//:index.bzl" % rctx.attr.node_repository)))
-    search_str = "host_platform=\""
-    start_index = content.index(search_str) + len(search_str)
-    end_index = content.index("\"", start_index)
-    host_platform = content[start_index:end_index]
-
-    # Return the path to the node binary
-    return rctx.path(Label("@%s_%s//:bin/node%s" % (rctx.attr.node_repository, host_platform, ".exe" if repo_utils.is_windows(rctx) else "")))
+    str_label = str(rctx.attr.node)
+    return rctx.path(Label(str_label + _ext(rctx, str_label)))
```
kormide, line 32: I think the docstring for `node_path` is out of date now.

gregmagolan (approved):
🌮

PR #75, opened 2022-04-29
refactor: rename node_package to link_node_package

gregmagolan:
and `node_modules()` macro to `link_node_packages()` as this is more accurate and is future proof for linking Plug'n'Play which doesn't have a `node_modules` folder


gregmagolan:
This was meant to be a draft still as its based off of https://github.com/aspect-build/rules_js/pull/48. Going to land that first and then rebase.

alexeagle:
looks like a stack on #48 - I'll review after the rebase

.gitattributes
```diff
@@ -1,3 +1,3 @@
 docs/*.md linguist-generated=true
 pnpm-lock.yaml linguist-generated=true
-e2e/node_package/rules_foo/npm_repositories.bzl linguist-generated=true
+e2e/link_node_package/rules_foo/npm_repositories.bzl linguist-generated=true
```
alexeagle, line 3: wonder if we should use a filename convention for vendored repositories.bzl files

alexeagle (approved):
love it. I think keeping the term "link" will help users understand the transition here as well.

PR #76, opened 2022-04-29
feat: add js_package rule

gregmagolan:
This is intended to make the first party deps API a little cleaner.

There is still the outstanding problem of first party deps that depend on other first party deps and 3rd party deps and that all want to get linked in DAG and how to express that. That will be handled in a follow-up.


e2e/link_js_package/BUILD.bazel
```diff
@@ -1,3 +1,11 @@
 load("@npm//:link_js_packages.bzl", "link_js_packages")
+load("@aspect_rules_js//js:link_js_package.bzl", "link_js_package")
 
 link_js_packages()
+
+link_js_package(
+    name = "linked_e2e_lib",
+    src = "//lib",
+    package = "@e2e/lib",
```
gregmagolan, line 9: this `package` attr is optional if `src` is specified. will remove

e2e/link_js_package/lib/BUILD.bazel
```diff
@@ -0,0 +1,11 @@
+load("@aspect_rules_js//js:js_package.bzl", "js_package")
+
+js_package(
+    name = "lib",
+    srcs = [
```
gregmagolan, line 5: this is using `copy_to_directory` impl under the hood and inherits all of its attributes and features

example/BUILD.bazel
```diff
@@ -23,7 +23,6 @@ link_js_packages()
 link_js_package(
     name = package("@mycorp/mylib").name,
     src = "//example/lib",
-    package = "@mycorp/mylib",
```
gregmagolan, line 26: no longer needed since package name comes from js_package (via JsPackageInfo)

example/lib/BUILD.bazel
```diff
@@ -1,10 +1,11 @@
-load("@aspect_bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
+load("//js:js_package.bzl", "js_package")
```
alexeagle, line 1: I think it ought to come from `defs.bzl` with `js_binary` and `js_test` WDYT? Is it likely to have extra `load` starlark deps that those rules don't?

js/private/js_package.bzl
```diff
@@ -0,0 +1,102 @@
...
+    ),
+    "version": attr.string(
+        # TODO: validate that version matches in an action if src is set
+        doc = "Must match the `version` field in the `package.json` file for this package.",
+        default = "0.0.0",
+    ),
+    "_windows_constraint": attr.label(default = "@platforms//os:windows"),
+}, copy_to_directory_lib.attrs)
+
+def _impl(ctx):
```
alexeagle, line 42: TODO for a validator action that checks the package name matches package.json if there is one?

js/private/js_package.bzl
```diff
@@ -0,0 +1,102 @@
...
+A DeclarationInfo is also provided so that the target can be used as an input to rules that expect one such as ts_project."""
+
+_ATTRS = dicts.add({
+    "src": attr.label(
+        doc = "A source directory or output directory to use for this package. For specifying a list of files, use `srcs` instead.",
+        allow_single_file = True,
+    ),
+    "package": attr.string(
+        # TODO: validate that name matches in an action if src is set
+        doc = "Must match the `name` field in the `package.json` file for this package.",
```
alexeagle, line 31: if there's a package.json in the folder already, can you have `package_from = ":package.json"` like we did for `ts_version`?

js/private/link_js_package.bzl
```diff
@@ -14,17 +15,16 @@ _LinkJsPackageInfo = provider(
...
         "virtual_store_directory": "the TreeArtifact of this node package's virtual store location",
-        "linked_js_package_dir": "the symlink of this package at the root of the node_modules if this is a direct npm dependency",
     },
 )
 
 _ATTRS = {
     "src": attr.label(
-        allow_single_file = True,
-        doc = """A source directory or TreeArtifact containing the package files.
+        doc = """A js_package target or a target that provides a JsPackageInfo.
```
alexeagle, line 23: nit: they're not mutually exclusive so "or any other target that provides"

e2e/link_js_package/lib/package.json
```diff
@@ -0,0 +1,3 @@
+{
+    "name": "@e2e/lib"
```
alexeagle, line 2: nit: "private": True so it's clear this name doesn't exist on npm

example/lib/BUILD.bazel
```diff
@@ -1,10 +1,11 @@
...
 
-copy_to_directory(
+js_package(
     name = "lib",
     srcs = [
         "index.js",
         "package.json",
     ],
+    package = "@mycorp/mylib",
```
alexeagle, line 9: ideally the fact that package.json is in srcs would be enough for us to read this, but I guess you want to put it in the provider which means you can't use an action to read the value :(

js/private/js_package.bzl
alexeagle: oh you have todo above already...

js/private/js_package.bzl
```diff
@@ -0,0 +1,102 @@
...
+        JsPackageInfo(
+            label = ctx.label,
+            package = ctx.attr.package,
+            version = ctx.attr.version,
+            directory = directory,
+        ),
+    ])
+    return providers
+
+js_package = rule(
```
alexeagle, line 97: should it export a struct as well, for custom rule composition?

js/private/link_js_package.bzl
```diff
@@ -45,67 +45,70 @@ Can be left unspecified to allow for circular deps between `link_js_package`s.
...
+        
+Must be specified is `src` is not set. If both `src` and `package` are set, must match the JsPackageInfo package name.""",
     ),
     "version": attr.string(
-        # TODO: validate that version matches in an action if src is set
-        doc = "Must match the `version` field in the `package.json` file for this package.",
-        default = "0.0.0",
+        doc = """The package version being linked.
+
+Must be specified is `src` is not set. If both `src` and `version` are set, must match the JsPackageInfo version.""",
```
alexeagle, line 55: nit: is -> if

js/private/link_js_package.bzl
```diff
@@ -45,67 +45,70 @@ Can be left unspecified to allow for circular deps between `link_js_package`s.
         providers = [_LinkJsPackageInfo],
     ),
     "package": attr.string(
-        # TODO: validate that name matches in an action if src is set
-        doc = "Must match the `name` field in the `package.json` file for this package.",
-        mandatory = True,
+        doc = """The package name being linked.
+        
+Must be specified is `src` is not set. If both `src` and `package` are set, must match the JsPackageInfo package name.""",
```
alexeagle, line 50: what's the use case for linking a package that doesn't have JsPackageInfo? is there an example of that already?

js/private/js_package.bzl
gregmagolan: not as trivial as that since the package.json we want is in the tree artifact and we're not in a repository rule context where we can open a package.json file and set the package & version attributes of a build action

js/private/link_js_package.bzl
gregmagolan: inside npm_import.bzl there is a complex pattern of "reference" link_js_packages to break circular deps; I should add that not specifying a src isn't recommended outside of there since you shouldn't write 1p deps that depend on each other.

js/private/link_js_package.bzl
gregmagolan: 🌮

js/private/link_js_package.bzl
gregmagolan: added to docstring

example/lib/BUILD.bazel
gregmagolan: yeah; no way to read it out of the package.json in a rule to get it into a provider

gregmagolan: the other direction of generating a default package.json if the user hasn't provided one we can tho however, which may be the more common pattern

e2e/link_js_package/lib/package.json
gregmagolan: 🌮

js/private/link_js_package.bzl
gregmagolan: :taco:

example/lib/BUILD.bazel
gregmagolan: It depends on DeclarationInfo from rules_nodejs core

js/private/js_package.bzl
gregmagolan: sure; I set a common pattern for js_binary, js_test, js_package and link_js_package to all do this; these are the foundation rules here

thesayyn (approved):
SGTM

example/lib/BUILD.bazel
jbedard: If both exist like in this case will or should we set the package name within the package.json?

gregmagolan: In theory we could have the validation action spit out a whole new tree artifact with the mutated package json. Tho that behaviour seems surprising. Instead I think we should auto-populate the package with the gazelle plug-in. Verify action should run as well similar to how we verify tsconfig settings in ts_project

PR #78, opened 2022-05-03
fix: get rid of _lc/node_modules tree and run lifecycle hooks in main node_modules tree

gregmagolan:
There is no way to do lifecycle hooks with circular deps between them without having an entire node_modules tree per package that has a lifecycle hook. This doesn't scale. For the set of packages that have lifecycle hooks, we're going to require a DAG between them.

We may need to add options to break circular deps between packages with lifecycle hooks so you can specify which order they should run in .

### DIFF
--- a/js/private/link_node_package.bzl
+++ b/js/private/link_node_package.bzl
@@ -140,13 +140,17 @@ def _impl(ctx):
                 direct_files.append(bin_out)
 
     if ctx.file.src:
-        # output the package as a TreeArtifact to its virtual store location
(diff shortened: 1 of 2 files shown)


alexeagle (approved):
Now that you forbid cycles between packages that need lifecycle hooks, what is the error message for users who run into that problem?

gregmagolan:
> Now that you forbid cycles between packages that need lifecycle hooks, what is the error message for users who run into that problem?

You get the classic bazel "you have a cycle" error message

PR #82, opened 2022-05-04
ci: assert we don't fetch unused dependencies

alexeagle:


gregmagolan:
I'll pre-factor the fix for this by adding manual tags to all the generated links so that we're lazy even with `bazel build //...`

PR #88, opened 2022-05-06
feat: support pnpm lockfile 5.4 which was introduced in pnpm 7.0.0

gregmagolan:
pnpm 7.0.0 was released five days ago;

also support `importers` top-level field as a pre-factor for supporting pnpm workspaces

### DIFF
--- a/js/private/translate_pnpm_lock.js
+++ b/js/private/translate_pnpm_lock.js
@@ -71,11 +71,14 @@ async function main(argv) {
         process.exit(1)
     }
 
-    // TODO: support a range: lockVersion >= 5.3 and lockVersion < 6.0
-    const expectedLockVersion = '5.3'
-    if (lockVersion != expectedLockVersion) {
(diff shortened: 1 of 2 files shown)


js/private/translate_pnpm_lock.js
```diff
@@ -90,11 +94,33 @@ async function main(argv) {
     const dev = !!process.env.TRANSLATE_PACKAGE_LOCK_DEV
     const noOptional = !!process.env.TRANSLATE_PACKAGE_LOCK_NO_OPTIONAL
 
+    let importers = {}
```
kormide, line 97: nit: don't need an assignment here

thesayyn (approved):
SGTM

js/private/translate_pnpm_lock.js
gregmagolan: ahh, good point 🌮

kormide (approved):

PR #96, opened 2022-05-11
Dot dot issue

stoiky:
Just an example of a workspace without a package in the workspace root
Similar to what Rush does


stoiky:
fixed with https://github.com/aspect-build/rules_js/pull/97

PR #98, opened 2022-05-11
fix: don't double copy 3rd party packages when linking with no lifecycle hook

gregmagolan:
This speeds up linking time on after a bazel clean by approx 35%

### DIFF
--- a/js/private/js_package.bzl
+++ b/js/private/js_package.bzl
@@ -58,9 +58,13 @@ def _impl(ctx):
 
     if ctx.attr.src:
         if ctx.file.src.is_source:
-            # copy the source directory to a TreeArtifact
-            directory = ctx.actions.declare_directory(ctx.attr.name)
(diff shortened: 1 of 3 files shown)


js/private/js_package.bzl
```diff
@@ -58,9 +58,13 @@ def _impl(ctx):
 
     if ctx.attr.src:
         if ctx.file.src.is_source:
-            # copy the source directory to a TreeArtifact
-            directory = ctx.actions.declare_directory(ctx.attr.name)
-            copy_directory_action(ctx, ctx.file.src, directory, is_windows = is_windows)
+            if hasattr(ctx.attr, "provide_source_directory") and ctx.attr.provide_source_directory:
```
alexeagle, line 61:
```suggestion
            if getattr(ctx.attr, "provide_source_directory", False):
```

alexeagle (approved):

js/private/js_package.bzl
gregmagolan: 🧙

PR #100, opened 2022-05-12
yamlark

kormide:
DO NOT MERGE, still testing

Can't test if this works yet because main seems to be broken.

The parser is kind of disgusting, but it works 🤷‍♂️ .

### DIFF
--- a/js/private/pnpm_yaml.bzl
+++ b/js/private/pnpm_yaml.bzl
@@ -1,0 +1,285 @@
+"Parse pnpm lock files into starlark"
+
+_STATE = struct(
+    # Read a property key
+    PROPERTY = 0,
+    # Parse an object's value
+    PARSE_VALUE = 1,
+    # We parsed a value with a colon, double check that it's a property and not a string
+    CONFIRM_PROPERTY = 2,
+    # We are parsing a sequence
+    PARSE_SEQUENCE = 3,
+    PARSE_BRACKET_SEQUENCE = 4,
+    # Handle cases after closing a curly brace
+    AFTER_OBJECT_CLOSE = 5,
+)
+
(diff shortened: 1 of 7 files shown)


js/private/test/repositories_checked.bzl
```diff
@@ -182,15 +182,6 @@ def npm_repositories():
...
             "fs.realpath": ["1.0.0"],
-            "inflight": ["1.0.6"],
-            "inherits": ["2.0.4"],
-            "minimatch": ["3.1.2"],
-            "once": ["1.4.0"],
-            "path-is-absolute": ["1.0.1"],
-            "wrappy": ["1.0.2"],
-            "brace-expansion": ["1.1.11"],
-            "balanced-match": ["1.0.2"],
-            "concat-map": ["0.0.1"],
```
kormide, line 193: It's not clear to me why using my yaml parser resulted in some changes to this file.

kormide:
It looks like I need to support sequences. Found a case.

kormide:
Almost there, but there's still a case I'm missing.

js/private/pnpm_yaml.bzl
```diff
@@ -0,0 +1,219 @@
...
+    elif _is_float(value):
+        return float(value)
+    elif _is_bool(value):
+        return _to_bool(value)
+    else:
+        return value.strip("'")
+
+def _get_current_object(result, stack):
+    object = result
+    for i in range(len(stack)):
```
alexeagle, line 197: nit: there's also enumerate, which gives back a list of tuples of (item, index)

kormide:
Oh no.... I did something stupid and I lost my work on supporting sequences 🙀 .

PR #103, opened 2022-05-15
yamlark++

kormide:
@alexeagle @gregmagolan


kormide:
Verified that it's producing the same result as previously for our example and e2e lockfiles.

kormide:
I added a third commit to remove the remaining uses of yq to do the postinstall substitution.

WORKSPACE
```diff
@@ -18,12 +18,6 @@ nodejs_register_toolchains(
     node_version = "16.9.0",
 )
 
-load("@aspect_bazel_lib//lib:repositories.bzl", "DEFAULT_YQ_VERSION", "register_yq_toolchains")
-
-register_yq_toolchains(
```
alexeagle, line 23: after a release here, we'll get to simplify the e2e/workspace/WORKSPACE file for all the downstream repos too :)

js/private/yaml.bzl
```diff
@@ -0,0 +1,413 @@
...
+    kns_states = _get_key_and_sequence_states(stack)
+    return len(kns_states) > 0 and kns_states[-1]["id"] == _STATE.SEQUENCE and kns_states[-1]["flow"]
+
+def _in_flow_map(stack):
+    kns_states = _get_key_and_sequence_states(stack)
+    return len(kns_states) > 0 and kns_states[-1]["id"] == _STATE.KEY and kns_states[-1]["flow"]
+
+def _pop_higher_indented_states(stack, indent):
+    remove = []
+    for state in stack:
```
alexeagle, line 409: nit: seems like instead of two iterations to collect removals and then apply, you could do one iteration where you construct a list to keep and return that?

js/private/yaml.bzl
```diff
@@ -0,0 +1,413 @@
...
+
+def _to_bool(value):
+    if value == "true":
+        return True
+    elif value == "false":
+        return False
+    fail("Cannot convert scalar %s to a starlark boolean" % value)
+
+def _parse_key(key):
+    if key.startswith("'"):
```
alexeagle, line 388: I guess we've never seen a specimen of pnpm-lock where there's also some escaped quoting within the string value?

alexeagle (approved):
nice job, very tidy and satisfying!

js/private/yaml.bzl
kormide: So the issue here is that I wanted to do an in-place removal, and starlark complains if you try to remove an element inside of a loop.

js/private/yaml.bzl
kormide: I don't think so, but this should only remove surrounding quotations. I'm not sure if:
```
'foo: bar
```
is valid yaml, but maybe I should do a `startswith` and `endswith` check before stripping the quotes.

kormide: The online parsers I checked seem to puke when a key starts with a quote but allow a key to end with a quote.

PR #106, opened 2022-05-18
refactor: make nodejs_binary amenable to custom rules

alexeagle:
Seems that it was broken following #22 in some rewrite

### DIFF
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -139,7 +139,7 @@ _NODE_OPTION = """NODE_OPTIONS+=(\"{value}\")"""
 def _target_tool_short_path(path):
     return ("../" + path[len("external/"):]) if path.startswith("external/") else path
 
-def _bash_launcher(ctx, entry_point_path, args):
+def _bash_launcher(ctx, entry_point_path, log_prefix_rule_set, log_prefix_rule, fixed_args):
     bash_bin = ctx.toolchains["@bazel_tools//tools/sh:toolchain_type"].path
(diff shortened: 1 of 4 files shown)


js/private/js_binary.bzl
```diff
@@ -213,7 +213,7 @@ def _create_launcher(ctx):
 
     output_data_files = copy_files_to_bin_actions(ctx, ctx.files.data, is_windows = is_windows)
 
-    bash_launcher = _bash_launcher(ctx, entry_point_path, ctx.attr.args)
```
alexeagle, line 216: @gregmagolan something is wrong here, this last argument to `_bash_launcher` is never used, so usage of args is broken

gregmagolan: Ohh. I see. We are currently depending on bazel to pass arguments to the binary using the standard `args` method which works when running the binary `bazel run` and `bazel test`. In `nodejs_binary` we had a `templated_args` that allowed you to bake the args into the launcher script but we don't have that feature in `js_binary` currently. It seems like a smell to me since other rule sets don't allow you do bake arguments in.

gregmagolan: That call should just be
```
bash_launcher = _bash_launcher(ctx, entry_point_path)
```
and there doesn't have to be anything special done with `ctx.attr.args` for arguments to work with a binary or test rule

gregmagolan: Sorted it out. I added in fixed_args but only for rules that extend from js_binary_lib and not into the public API

gregmagolan (approved):
🌮

PR #112, opened 2022-05-19
Fix package_json.bzl file for packages with dashes in their names

alex-torok:
### Summary

When generating `package_json.bzl` files, replace `-` symbols with `_` symbols to ensure that the generated starlark function names are valid functions. Previously, packages with `-` in their name (like `webpack-bundle-analyzer` were being given invalid function names in the `package_json.bzl` files.

`rules_nodejs` does the same thing:
https://github.com/bazelbuild/rules_nodejs/blob/d69c6461aa3272bed90132b3b0c7376a4b274920/internal/npm_install/generate_build_file.ts#L1349-L1351

### Test Plan
Run unit tests with `bazel test //js/private/test/...`

Introduce a new `write_source_file` test that asserts that the `package_json.bzl` file for `webpack-bundle-analyzer` is properly generated with function names like `webpack_bundle_analyzer{,_test,_binary}` and that the exported `bin` struct is configured properly.

I chose this package since it was already available in `@example_npm_deps`, but could swap it to something else if desired.


alexeagle (approved):
Amazing, thank you!
Exciting to see usage of this repo already 😎 

Could you make the CLA bot happy?

alex-torok:
> Could you make the CLA bot happy?

Sorry, I didn't realize there was a CLA. I'll need to get internal company approval before I can sign, which may take a bit.

Issue #113, opened 2022-05-19
generated bin doesn't work in subpackage


alexeagle:
`third_party/npm/BUILD`
```
load("@npm//npm-check:package_json.bzl", "bin")

exports_files(["package.json"])

bin.npm_check_binary(
    name = "check",
    data = [
        "//third_party/npm:package.json",
    ],
    args = [
        "--no-color",
        "--no-emoji",
        "--save-exact",
        "--skip-unused",
        "third_party/npm",
    ],
)
```

Then you get a bad error message about "must be a treeartifact" but actually the label does'nt exist at all

```
alexeagle@system76-pc:~/Projects/net_$ bazel build third_party/npm:all
ERROR: /home/alexeagle/Projects/net_/third_party/npm/BUILD:5:21: in directory_path rule //third_party/npm:check__entry_point: 
Traceback (most recent call last):
	File "/home/alexeagle/.cache/bazel/_bazel_alexeagle/b027b6729610d26a644911de4dac9747/external/aspect_bazel_lib/lib/private/directory_path.bzl", line 17, column 13, in _directory_path
		fail("directory attribute must be created with Bazel declare_directory (TreeArtifact)")
Error in fail: directory attribute must be created with Bazel declare_directory (TreeArtifact)
ERROR: Analysis of target '//third_party/npm:check__entry_point' failed; build aborted: Analysis of target '//third_party/npm:check__entry_point' failed
FAILED: Build did NOT complete successfully (6 packages loaded, 8 targets configured)
    currently loading: third_party/python ... (4 packages)
    Fetching @go_sdk; fetching

alexeagle@system76-pc:~/Projects/net_$ bazel query --output=build //third_party/npm:check__entry_point
# /home/alexeagle/Projects/net_/third_party/npm/BUILD:5:21
directory_path(
  name = "check__entry_point",
  generator_name = "check",
  generator_function = "npm_check_binary",
  generator_location = "third_party/npm/BUILD:5:21",
  directory = "//third_party/npm:jsp__npm-check__5.9.2__dir",
  path = "bin/cli.js",
)

# Rule check__entry_point instantiated at (most recent call last):
#   /home/alexeagle/Projects/net_/third_party/npm/BUILD:5:21                                                                       in <toplevel>
#   /home/alexeagle/.cache/bazel/_bazel_alexeagle/b027b6729610d26a644911de4dac9747/external/npm__npm-check__5.9.2/package_json.bzl:36:20 in npm_check_binary
[...]

PR #119, opened 2022-05-23
Add example showing mocha integration

jfirebaugh:
This ports a macro we were using with rules_nodejs to rules_js. However, it's failing with:

```
$ bazel build //example/mocha:test
ERROR: /Users/john/figma/rules_js/example/mocha/BUILD.bazel:3:11: in directory_path rule //example/mocha:test__entry_point: 
	File "/private/var/tmp/_bazel_johnfirebaugh/3cc5be9735d30b6128426fd19d172550/external/aspect_bazel_lib/lib/private/directory_path.bzl", line 17, column 13, in _directory_path
		fail("directory attribute must be created with Bazel declare_directory (TreeArtifact)")
Error in fail: directory attribute must be created with Bazel declare_directory (TreeArtifact)
```

If this is user error, can the error message be improved so that the resolution is more obvious? Otherwise, feel free to fix and merge as a useful example. I'm sending this PR from a personal fork so "edits by maintainers" is enabled.


jfirebaugh:
I think the error here is https://github.com/aspect-build/rules_js/issues/113.

alexeagle:
the fix for #113 is merged, so I added a bit more fixes here to get the test working.
I'm not sure the best way to tell mocha the paths to all the spec files. I gave it `package_name/*test.js`

jfirebaugh:
That looks good to me and is similar to what we do.

Hopefully I did the right thing to fix the remaining CI failures (ran `bazel run //js/private/test:write_translate_pnpm_lock_{0,1}`).

alexeagle:
I think we should fixup #128 so the mocha example is a separate package.json under the pnpm workspace

alexeagle (approved):
I'll move the examples folder around in a follow-up

alexeagle (approved):

alexeagle:
@gregmagolan this was green at an earlier snapshot, but after rebasing and changing to the pnpm workspace directory I'm struggling to fix this

```
ERROR: TypeError: Could not load reporter "mocha-multi-reporters":

 Error: Cannot find module '/home/runner/.cache/bazel/_bazel_runner/8d53428c79e05e77d1e[55](https://github.com/aspect-build/rules_js/runs/6644021024?check_suite_focus=true#step:5:56)522fa21a0fd/sandbox/linux-sandbox/245/execroot/aspect_rules_js/bazel-out/k8-fastbuild/bin/example/mocha/_test_launcher.sh.runfiles/aspect_rules_js/mocha-multi-reporters'
```

alexeagle:
I refactored the examples/ folder so this should be a nested pnpm package with its own package.json adding the dep on mocha

gregmagolan:
The failure with the require of `mocha-multi-reporters` is that mocha is requiring it directly from its code and not from a user  `config.js` file (which other test frameworks such as jest do).

Without workspaces, this worked accidentally since the require from the pnpm virtual store `<bazel-bin>/example/node_modules/.aspect_rules_js/mocha_x.x.x/node_modules/mocha/some/file.js` worked since there was a `<bazel-bin>/example/node_modules/mocha-multi-reporters`.

With workspaces, the virtual store is moved up one level to `<bazel-bin>/node_modules/.aspect_rules_js` so mocha from the virtual store can no longer require `mocha-multi-reporters`. This is strictly correct since `mocha` doesn't have a peer dep on `mocha-multi-reporters`.

We can fix it by either hoisting `mocha-multi-reporters` to the root `package.json` or by injecting `mocha-multi-reporters` as a peerDep with a `.pnpmfile.cjs` hook:

```
function readPackage(pkg, context) {
    if (pkg.name === 'mocha') {
        pkg.peerDependencies = {
            ...pkg.peerDependencies,
            'mocha-multi-reporters': '*',
        }
        context.log('Adding mocha-multi-reporters peerDependency to mocha')
    }

    return pkg
}

module.exports = {
    hooks: {
        readPackage,
    },
}
```

gregmagolan:
Greened up. Just needs a rebase now to get over the conflicts. But it might be easier to just make a new PR that introduces the `examples/mocha` folder and adds a `examples/mocha/package.json` to the workspace since we're now going with that pattern at head.

Issue #121, opened 2022-05-23
Generated bin should have the package in data by default


alexeagle:
Users have to manually repeat the package in `data` right now. For example, the following example fails with this confusing error
```
==================== Test output for //example:lint:
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module 'v8-compile-cache'
Require stack:
- /home/alexeagle/.cache/bazel/_bazel_alexeagle/581b2ac03dd093577e8a6ba6b6509be5/execroot/aspect_rules_js/bazel-out/k8-fastbuild/bin/example/node_modules/.aspect_rules_js/eslint@8.16.0/node_modules/eslint/bin/eslint.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:94:18)
    at Object.<anonymous> (/home/alexeagle/.cache/bazel/_bazel_alexeagle/581b2ac03dd093577e8a6ba6b6509be5/execroot/aspect_rules_js/bazel-out/k8-fastbuild/bin/example/node_modules/.aspect_rules_js/eslint@8.16.0/node_modules/eslint/bin/eslint.js:13:1)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:79:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/alexeagle/.cache/bazel/_bazel_alexeagle/581b2ac03dd093577e8a6ba6b6509be5/execroot/aspect_rules_js/bazel-out/k8-fastbuild/bin/example/node_modules/.aspect_rules_js/eslint@8.16.0/node_modules/eslint/bin/eslint.js'
  ]
}
```

unless you uncomment the data line

```starlark
load("@npm//:defs.bzl", "link_js_packages")
load("@npm//eslint:package_json.bzl", "bin")

link_js_packages()

bin.eslint_test(
    name = "lint",
#     data = ["@npm//eslint"],
)
```

[...]

PR #122, opened 2022-05-26
fix: add js_package of npm package to the data of the generated js_binaries for its bins

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/121

### DIFF
--- a/example/BUILD.bazel
+++ b/example/BUILD.bazel
@@ -626,7 +626,6 @@ uvu_test(
     ],
     data = [
         ":acorn_spec",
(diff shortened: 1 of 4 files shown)


example/BUILD.bazel
```diff
@@ -626,7 +626,6 @@ uvu_test(
     ],
     data = [
         ":acorn_spec",
-        "@npm//uvu",
```
alexeagle, line 629: 😍

alexeagle (approved):

PR #128, opened 2022-05-26
refactor: use pnpm workspaces for main WORKSPACE pnpm-lock.yaml file

gregmagolan:
This is a pre-factor to bringing in node patches for symlink guards so we can bring devDependencies such at typscript into the root package.json.


js/private/test/defs_checked.bzl
```diff
@@ -1,4 +1,4 @@
-"@generated by @aspect_rules_js//js/private:translate_pnpm_lock.bzl from pnpm lock file @//example:pnpm-lock.yaml"
+"@generated by @aspect_rules_js//js/private:translate_pnpm_lock.bzl from pnpm lock file @//:pnpm-lock.yaml"
```
alexeagle, line 1: it's getting annoying how long this "golden" file is - maybe we should have a separate pnpm workspace in an internal testing folder to generate something more stable. 
for example @jfirebaugh just tripped on it https://github.com/aspect-build/rules_js/pull/119#issuecomment-1139053668

pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
+    - 'example'
```
alexeagle, line 2: now or later: I think we'll want to pluralize "examples" and have a package.json under each, like the mocha one in https://github.com/aspect-build/rules_js/pull/119

alexeagle (approved):

pnpm-workspace.yaml
gregmagolan: sounds like a follow-up

PR #129, opened 2022-05-26
feat: add additional verbose output to js_binary launcher to indicate cwd

gregmagolan:


js/private/js_binary.sh.tpl
```diff
@@ -129,6 +129,9 @@ else
         printf "\nERROR: %s: BAZEL_BINDIR must be set in environment when not running out of runfiles so js_binary can run out of the output tree on build actions\n" "$LOG_PREFIX" >&2
         exit 1
     fi
+    if [ "${JS_BINARY__VERBOSE:-}" ]; then
```
f0rmiga, line 132:
```suggestion
    if [ -n "${JS_BINARY__VERBOSE:-}" ]; then
```

f0rmiga: I'm surprised it works without `-n`!

f0rmiga (approved):

js/private/js_binary.sh.tpl
gregmagolan: `-n` is the implicit default :) 

```
[ -n STRING ] or [ STRING ] | True if the length of "STRING" is non-zero.
```

https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html#sect_07_01_02

f0rmiga: TIL

PR #130, opened 2022-05-27
feat: add verbose attribute to run_js_binary

gregmagolan:


js/run_js_binary.bzl
```diff
@@ -21,6 +21,7 @@ def run_js_binary(
         exit_code_out = None,
         silent_on_success = True,
         copy_srcs_to_bin = True,
+        verbose = False,
```
alexeagle, line 24: it feels wrong to permanently turn verbosity on or off, vs. a custom flag setting?

gregmagolan: it can be turned on via an environment variable; this attribute and a sibling one in js_binary are just convenience attributes to set the environment variable for a single target.

gregmagolan: you could also run `bazel build ... --define= JS_BINARY__VERBOSE=1` to add verbosity for all js_binary targets

gregmagolan: actually, I thought of a better pattern for this and for the equivalent attribute in js_binary. I'll open another PR.

alexeagle: Cool feels like something we never got quite right

Issue #114, opened 2022-05-19
js rules should be hermetic and not follow symlinks out of the sandbox


jfirebaugh:
I noticed some of my uses of `js_binary` were accidentally escaping the sandbox through `__dirname`. For consistency with `nodejs_binary` and per previous discussions in rules_nodejs (e.g. https://github.com/bazelbuild/rules_nodejs/issues/109), `--preserve-symlinks` should be the default.

alexeagle:
Good catch, that seems obviously wrong, sending PR with repro

jfirebaugh:
It looks like there may be side effects with pnpm-style package linking. I'm still debugging this, but locally setting `node_options = ["--preserve-symlinks", "--preserve-symlinks-main"]` causes requires of transitive dependencies to start failing.

alexeagle:
yeah I found the same thing, I think pnpm does rely on some amount of walking symlinks into the CAS.
Will leave for Greg to ponder.

jfirebaugh:
My 2¢, if it turns out pnpm isolated linking is fundamentally incompatible with `--preserve-symlinks`: I would prefer hermeticity with hoisted modules over non-hermeticity and isolated modules.

alexeagle:
We have the sandbox guards in rules_nodejs/package/patches which is probably needed to improve hermeticity. Some packages follow symlinks regardless of what require does.

gregmagolan:
AFAIK `--preserve-symlinks` comes with issues when used with pnpm style symlinked node_modules structure. I haven't investigated it so it might work in some cases however I'd rather not flip it to default. My plan to bring in the symlink guards node patches from rules_nodejs and use them when running in the sandbox.

gregmagolan:
It's near to top of my list to bring in here next. Should be a fairly quick fix which I could get done over the long weekend here.

gregmagolan:
Started on it today, https://github.com/aspect-build/rules_js/pull/133.

But I'll need to tweak the node patches so they just guard the sandbox but still allow symlinks to be resolved within the sandbox. I'll save it for another day when I have a fresh brain.

alex-torok:
I just got bit by this - it appears that even the entrypoint script for `js_test` targets is being resolved out of the sandbox. I've put up an example and a writeup at https://github.com/alex-torok/rules_js/pull/1.

[...]

PR #133, opened 2022-05-27
feat: add execroot & runfiles symlink guards to js_binary

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/114

The fs patches here have been modified from their original implementation in rules_nodejs https://github.com/bazelbuild/rules_nodejs/blob/173a0252f273de48a6a237518e9cb0d9854cb8e6/packages/node-patches/src/fs.ts.

The have been improved to allow following symlinks within the sandbox, which is required for finding deps in a symlinks node_modules structure. Since the sandbox (on macos) is a tree of symlinks that all point back to their counterparts outside of the sandbox, the old implementation prevented following symlinks entirely (akin to passing `--preserve-symlinks) since every follow was an escape from the guarded sandbox.

In this implementation, if the readlink escape is a symlink to a symlink then we readlink the escaped symlink. For lstat we do the same. For realpath, if we the relative path from the link to the realpath can be remapped into the escaped root then it is.

In rules_js, the guards are setup to protect the execroot (which protects the sandbox when there is one and protects from escaping into the source tree when there isn't) and to protect the runfiles tree for the js_binary in the build, test or run target since that is a symlink tree as well. This is simpler than what was done in rules_nodejs which was to protect all of the dynamically linked node_modules trees.


js/private/node-patches/src/fs.ts
```diff
@@ -0,0 +1,671 @@
...
+            }
+        }
+        return str
+    }
+
+    // =========================================================================
+    // fs.readlink
+    // =========================================================================
+
+    // tslint:disable-next-line:no-any
```
alexeagle, line 327: may as well toss the tslint lines that are no longer doing anything

alexeagle (commented):
let's add the tests too, otherwise it's too tempting to leave it for later...

js/private/node-patches/src/fs.ts
gregmagolan: roger roger

gregmagolan:
> let's add the tests too, otherwise it's too tempting to leave it for later...

Done. Brought unit tests in from rules_nodejs for the patches

e2e/npm_link_package/src/main.js
```diff
@@ -15,7 +15,7 @@ console.log(c.id())
 const fp = require('@e2e/lib')
 console.log('--@e2e/lib--')
 console.log(fp.id())
-const rulesFooA = require('../external/rules_foo/foo/node_modules/@aspect-test/a')
+const rulesFooA = require('../../rules_foo/foo/node_modules/@aspect-test/a')
```
alexeagle, line 18: Interesting change needed

alexeagle (approved):

e2e/npm_link_package/src/main.js
gregmagolan: yea, the path here is runfiles relative for this test. no legacy external runfiles :)

PR #134, opened 2022-05-27
run e2e tests on fasterci

apesternikov:


.fasterci/config.yaml
```diff
@@ -16,3 +16,31 @@ workflows:
                     - //...
                 test_targets:
                     - //...
+          - name: e2e/link_js_package
```
alexeagle, line 19: hey, thanks - could we use the same matrix syntax as in the github actions configuration?

alexeagle: this is what I'm referring to: https://github.com/aspect-build/rules_js/blob/main/.github/workflows/ci.yaml#L19-L25

gregmagolan:
I hit a failure on unit tests on FasterCI that was consistent and only seen there and not our github actions CI. Turned off FasterCI so it wouldn't break master. See https://github.com/aspect-build/rules_js/commit/cd7dafad8b3c86ff7ff337c50259721949dfdea2

.fasterci/config.yaml
apesternikov: Using yaml refs is recommended way to achieve this. see https://github.com/adobe/rules_gitops/blob/fd76291f513fce4fc637c22f8055b3a49ad82fe7/.fasterci/config.yaml#L32 for example.

PR #136, opened 2022-05-29
refactor: reorganize example into multiple examples

alexeagle:


gregmagolan (approved):
🌮 

Guess we sort out how to number test-cases now that they are spreadout in a follow-up and keep this one as just moving them around.

PR #138, opened 2022-05-30
refactor: cleanup pnpm workspaces

gregmagolan:


gregmagolan:
Rolling this into a public API cleanup now

PR #139, opened 2022-05-31
feat: simplify public API link_js_package and package label helpers

gregmagolan:
BREAKING CHANGE

This unwinds the `dot_dot` package name segments in favor of always using the link package name for consistency and readability where the `@npm//path/to/link/package/alias:alias` refers to a predictable and consistent `path/to/link/package/node_modules` link folder.

First party `link_js_package` targets now look like this.

```
link_js_package(
    name = "mycorp_mylib_link",
    src = "//examples/lib",
    deps = [
        link_js_package_dep("acorn", "8.4.0"),
    ],
)
```

`src` & `deps` are only set when linking into the `root_package` (the package where the virtual store is linked) which defaults to the root package.

When linking into a package that is not the root package the call is simply

```
link_js_package(
    name = "mycorp_mylib_link",
)
```

`name` must match between the two so that the virtual store targets match up

Where a manual link_js_package is used, a "{name}" alias is created by default so you can dep on `:mycorp_mylib_link`. The directory artifact alias of that is "{name}__dir" which in the case above would be ":mycorp_mylib_link__dir".

The above also works for a manual npm_import 3rd party package with the link_js_package macro loaded from the "links" repository generated by npm_import.

```
load("@acorn__8.4.0__links//:link_js_package.bzl", link_acorn = "link_js_package")
link_acorn(
    name = "acorn_link",
)
```

will create alias targets `:acorn_link` and `:acorn_link__dir`.

No more `package` and `package_dir` helpers are needed with this new approach and they are now removed.


BUILD.bazel
```diff
@@ -1,11 +1,28 @@
...
 # bazel-bin/examples/node_modules as well as the virtual store since this package.json
 # is the root of the pnpm workspace
 link_js_packages()
 
+# Link the acorn package, which was fetched separately with npm_import from /WORKSPACE, to the
+# virtual store in bazel-bin/node_modules
+link_acorn()
+
+# Linking a first party dependency to the virtual store in bazel-bin/node_modules; `direct` is set to
+# False so we don't link this first party dependency as a direct dependency here
```
alexeagle, line 16: this comment wasn't enough for me to understand what direct=False is doing

BUILD.bazel
```diff
@@ -1,11 +1,28 @@
...
+link_js_package(
+    name = "@mycorp/mylib",
+    src = "//examples/lib",
+    direct = False,
+    deps = [
+        link_js_package_dep("acorn", "8.4.0"),
```
alexeagle, line 22: maybe comment about the meaning of the second argument... what happens without that?

e2e/pnpm_workspace_dot_dot/app/a/BUILD.bazel
```diff
@@ -1,16 +1,16 @@
 load("@aspect_rules_js//js:defs.bzl", "js_binary", "js_test")
 load("@npm//:defs.bzl", "link_js_packages")
-load("@npm//dot_dot/app/a/@aspect-test/a:package_json.bzl", aspect_test_a_bin = "bin")
```
alexeagle, line 3: goodbye dot_dot I never even knew you

examples/BUILD.bazel
```diff
@@ -1,58 +1,13 @@
...
-# Trivial test fixture: the shortest legal JS program
-write_file(
-    name = "write_one",
-    out = "one.js",
-    content = ["1"],
-    visibility = ["//examples:__subpackages__"],
-)
-
 # For using acorn as our test fixture, this is
 # the serialized AST for the that shortest legal JS program
```
alexeagle, line 7: nit: "the that" - and now that one.js isn't created here the comment doesn't make much sense

WORKSPACE
```diff
@@ -69,7 +69,14 @@ npm_repositories()
 npm_import(
     name = "acorn__8.4.0",
     integrity = "sha512-ULr0LDaEqQrMFGyQ3bhJkLsbtrQ8QibAseGZeaSUiT/6zb9IvIkomWHJIvgvwad+hinRAgsI51JcWk2yvwyL+w==",
+    # Packages where to link this npm dependency as a direct dependency
+    link_paths = [
```
alexeagle, line 73: but it's not sufficient, since those packages also need a link_* call in their BUILD file? should there be a comment here that this allows the link but doesn't create it?

examples/js_binary/BUILD.bazel
```diff
@@ -1,18 +1,26 @@
...
+
 ####################################################
 # Test case 4
 # Show that a js_binary can be used with run_js_binary
 # because everything it needs to run is in the runfiles
 
 js_binary(
     name = "bin",
-    data = [package("acorn", "examples")],
+    data = [":acorn_link"],
```
alexeagle, line 23: yeah this is way better just to refer to the target above, feels like Bazel idiom again

js/private/pnpm_utils.bzl
```diff
@@ -91,14 +91,14 @@ pnpm_utils = struct(
     friendly_name = _friendly_name,
     virtual_store_name = _virtual_store_name,
     strip_peer_dep_version = _strip_peer_dep_version,
-    # Prefix namespace to use for generated js_binary targets and aliases
-    js_package_target_namespace = "jsp__",
```
alexeagle, line 95: looks ugly breaking change for my downstream projects, but now's the time

pnpm-workspace.yaml
```diff
@@ -1,3 +1,3 @@
 packages:
-    - 'examples'
```
alexeagle, line 2: but there's still a package.json in this folder right? should it be deleted too?

js/private/pnpm_utils.bzl
gregmagolan: indeed

BUILD.bazel
gregmagolan: Updated to
```
    # `direct` set to False as an example of *not* also linking this first-party dependency as a
    # direct dependency in the package at bazel-bin/node_modules/@mycorp/mylib.
    direct = False,
```

e2e/pnpm_workspace_dot_dot/app/a/BUILD.bazel
gregmagolan: 👋

WORKSPACE
gregmagolan: Remote this altogether from the example; it is simpler allow npm_imports with the default of `link_paths = []` to be linked anywhere depending on the `direct` attribute of `link_js_package`. This makes the behaviour of linking a manual npm_import match the behaviour of linking a first party `js_package`.

alexeagle (approved):
yum 🚢

PR #152, opened 2022-06-03
Refactor public API into near final shape for 1.0 launch

gregmagolan:
BREAKING CHANGES


js/private/js_library.bzl
```diff
@@ -0,0 +1,106 @@
...
+                new_target_name = file.basename.replace(".", "_"),
+                this_package = "%s//%s" % (ctx.label.workspace_name, ctx.label.package),
+                this_target = ctx.label,
+            )
+            fail(msg)
+
+    output_srcs = copy_files_to_bin_actions(ctx, ctx.files.srcs, is_windows = is_windows)
+
+    for src in output_srcs:
+        all_files.append(src)
```
alexeagle, line 53: nit: no need to create `all_files` as its always identical to `ctx.files.srcs`

js/defs.bzl
```diff
@@ -7,19 +7,13 @@ load(
     _js_test = "js_test",
 )
 load(
-    "//js/private:js_package.bzl",
```
alexeagle, line 10: include the typical `load` statement in the module doc at the top?

npm/defs.bzl
```diff
@@ -0,0 +1,22 @@
+"""Rules for fetching and linking npm dependencies and packaging and linking first-party deps"""
```
alexeagle, line 1: Include the typical load statement here, so it appears in the docs?

js/private/js_library.bzl
```diff
@@ -0,0 +1,106 @@
+"""Expose some files with DeclarationInfo, like filegroup but can be a dep of ts_project
```
alexeagle, line 1: I'm not quite convinced such a thing should be named `js_library` but we can merge and then discuss renaming again

e2e/bzlmod/BUILD.bazel
```diff
@@ -1,14 +1,12 @@
...
-link_js_packages()
+link_all_npm_packages(name = "node_modules")
 
 js_test(
     name = "test",
     data = [
-        "//:direct_link__chalk",
-        # these don't work yet:
-        # "@npm//chalk",
+        ":node_modules/chalk",
```
alexeagle, line 9: these references look much better than the `@npm//...` ones for 1p :)

e2e/pnpm_workspace/lib/b/BUILD.bazel
```diff
@@ -1,9 +1,9 @@
-load("@aspect_rules_js//js:defs.bzl", "js_package")
-load("@npm//:defs.bzl", "link_js_packages")
+load("@aspect_rules_js//npm:defs.bzl", "npm_package")
+load("@npm//:defs.bzl", "link_all_npm_packages")
 
-link_js_packages()
+link_all_npm_packages(name = "node_modules")
 
-js_package(
+npm_package(
```
alexeagle, line 6: what did we decide about a `package_json` attribute to call out that they ought to be 1:1? do in a followup?

examples/macro/mocha.bzl
```diff
@@ -20,8 +20,8 @@ def mocha_test(name, srcs, args = [], data = [], env = {}, **kwargs):
         data = data + [
             "_%s_srcs" % name,
             "//examples/macro:mocha_reporters.json",
-            "@npm//examples/macro/mocha-multi-reporters",
-            "@npm//examples/macro/mocha-junit-reporter",
+            "//examples/macro:node_modules/mocha-multi-reporters",
```
alexeagle, line 23: the ordering is so much more clear with the 1p package coming first and the node_modules address second 😍

scripts/in_each_e2e.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
alexeagle, line 1: Is this just to be silly? It's still dead code right? :)

alexeagle (approved):

scripts/in_each_e2e.sh
gregmagolan: I actually use this locally quite a bit with `bazel test ... && ./scripts/in_each_e2e.sh bazel test ...` but I can put in my. home folder if you think it is not useful

e2e/pnpm_workspace/lib/b/BUILD.bazel
gregmagolan: we were considering it; under the hood its just syntax sugar but lets chat about it in the public API review

examples/js_library/one/BUILD.bazel
```diff
@@ -0,0 +1,11 @@
+load("@aspect_rules_js//js:defs.bzl", "js_library")
+
+js_library(
+    name = "one",
+    srcs = [
+        "one.d.ts",
+        "one.js",
+        "package.json",
```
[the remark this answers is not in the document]
gregmagolan: yup, I intentionally didn't include an index.d.ts

js/private/js_library.bzl
gregmagolan: 👍

js/defs.bzl
gregmagolan: ✅

Issue #150, opened 2022-06-02
Help a user who forgot to update their pnpm-lock.yaml


alexeagle:
We could have some repository rule to re-run `pnpm install --lockfile-only` and verify that the users file is up-to-date.

Otherwise you might add a first-party dep to your package.json and then Bazel doesn't see that change.

cgrindel:
@alexeagle Is this addressed by #157?

gregmagolan:
Resolved with `update_pnpm_lock` feature in `npm_translate_lock`

PR #157, opened 2022-06-03
Add example for bazel integration to update lockfile

alex-torok:
This PR shows an example of adding a method to `bazel run` a target to update a `pnpm-lock.yaml` file along with a test target that asserts that a `pnpm-lock.yaml` file is not out-of-date.

The test target exploits a bit of how `pnpm` works so that it can check if the lockfile is out-of-date without needing to download any packages (keeping runtime of the test at ~0.5s)

I noticed #150, and was just working on a method to integrate this into bazel. I think there are some extra considerations that would need to be taken (particularly if pnpm workspaces are used versus a single package.json file).

Maybe this could be of use to land as an example, or maybe it could be polished up and added to the library as a fully supported feature. I just wanted to show how I was approaching the problem of an out-of-date lockfile in our repo.


examples/assert_lockfile_up_to_date/assert_lockfile_frozen.js
```diff
@@ -0,0 +1,60 @@
...
+const fs = require('fs');
+const child_process  = require('child_process');
+const readline = require('readline');
+
+const FIX_COMMAND = 'bazel run //examples/assert_lockfile_up_to_date:update_pnpm_lockfile'
+
+// https://github.com/pnpm/pnpm/blob/8fa95fd868d1afac6f7454ac17f77f8a017d2110/packages/headless/src/index.ts#L166
+const OUTDATED_LOCKFILE_ERR_CODE = "ERR_PNPM_OUTDATED_LOCKFILE"
+
+// pnpm does not offer a no-install method to assert that a lockfile
```
alexeagle, line 11: would be nice to file this upstream, then point to that issue here, so there's some chance of unwinding the hack later (and to learn if anyone else has a nicer hack)

alexeagle (commented):
I think with a README added about how you use this, it would be nice to land as-is

alex-torok:
I've added a README that explains a bit about how this works along with a discussion about limitations and future work.

examples/assert_lockfile_up_to_date/assert_lockfile_frozen.js
```diff
@@ -0,0 +1,61 @@
+const path = require('path');
+const fs = require('fs');
+const child_process  = require('child_process');
+const readline = require('readline');
+
+const FIX_COMMAND = 'bazel run //examples/assert_lockfile_up_to_date:update_pnpm_lockfile'
```
alexeagle, line 6: opt: maybe https://docs.aspect.build/aspect-build/bazel-lib/v0.12.0/docs/write_source_files-docgen.html would be a nice way to express the assertion and provides a built-in fix command

alexeagle (approved):
Thanks! Looks great. Do you want to merge like this, or try out the `write_source_files` rule?

alex-torok:
> Thanks! Looks great. Do you want to merge like this, or try out the `write_source_files` rule?

I intentionally avoided using `write_source_files` -- I tried explaining it a bit in the README, but maybe didn't do a great job. If this used `write_source_files`, then the test for checking if the lockfile is synced would require running the action that downloads and installs packages from npm. (It would have a data dependency on the output pnpm-lock.yaml file).

alexeagle:
Ah, okay thanks for explaining (again)

PR #160, opened 2022-06-03
fix: don't generate scoped target if there are no direct deps in that scope

gregmagolan:


alexeagle (approved):
does it fix some bug?

gregmagolan:
> does it fix some bug?

No. Just cleared out some useless generated targets.

PR #163, opened 2022-06-03
fix: bazel-lib version in MODULE.bazel

kormide:
Diverged from `repositories.bzl`.


kormide:
Error should resolve itself when https://github.com/aspect-build/bazel-central-registry/pull/4 is merged.

alexeagle (approved):

Issue #166, opened 2022-06-07
bug: `github.com` packages not supported


vpanta:
We have a transitive dependency on `aframe` which requires other packages [via a github reference](https://github.com/aframevr/aframe/blob/2c4509aa1a9a3c447c08be9facfea0c7d6a9e156/package.json#L40). In our `pnpm-lock.json` that looks like:

```yaml
  /aframe/1.3.0:
    resolution: {integrity: sha512-f6OQaDf49SzdSxVskJPvPWldOwHpYMsGttmQHhU6FRiASsnyULKG1nqZom9pDuS3fFYEzQVqRMakQ2AhXeLkFg==}
    engines: {node: '>= 4.6.0', npm: '>= 2.15.9'}
    dependencies:
      custom-event-polyfill: 1.0.7
      debug: github.com/ngokevin/debug/ef5f8e66d49ce8bc64c6f282c15f8b7164409e3a
      deep-assign: 2.0.0
      document-register-element: github.com/dmarcos/document-register-element/8ccc532b7f3744be954574caf3072a5fd260ca90
      ...
```
(caveat: we don't use pnpm ourselves, but this comes from `pnpm import` on our Node `package-lock.json`)

With `rules_js@0.11.0`, `bazel sync` gets the following error:
```
Repository rule npm_translate_lock defined at:
  /private/var/tmp/_bazel_vasilios/96c9699731d9b99e83af4f7d36096194/external/aspect_rules_js/npm/npm_import.bzl:32:37: in <toplevel>
ERROR: An error occurred during the fetch of repository 'npm':
   Traceback (most recent call last):
        File "/private/var/tmp/_bazel_vasilios/96c9699731d9b99e83af4f7d36096194/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 319, column 33, in _impl
                lockfile = _process_lockfile(rctx)
        File "/private/var/tmp/_bazel_vasilios/96c9699731d9b99e83af4f7d36096194/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 145, column 43, in _process_lockfile
                return translate_to_transitive_closure(lockfile, rctx.attr.prod, rctx.attr.dev, rctx.attr.no_optional)
        File "/private/var/tmp/_bazel_vasilios/96c9699731d9b99e83af4f7d36096194/external/aspect_rules_js/npm/private/transitive_closure.bzl", line 86, column 17, in translate_to_transitive_closure
[...]

PR #169, opened 2022-06-09
fix: support github.com style npm dependency versions in pnpm lock files

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/166


npm/private/npm_translate_lock.bzl
```diff
@@ -308,10 +308,11 @@ def _gen_npm_imports(lockfile, attr):
             package = name,
             patch_args = patch_args,
             patches = patches,
-            pnpm_version = pnpm_version,
```
alexeagle, line 311: Looks like breaking downstream for vendored files, can we avoid that?

alexeagle (approved):

npm/private/npm_translate_lock.bzl
gregmagolan: Shouldn't be breaking as it is an internal variable. Generated .bzl and BUILD files should not be affected.

Issue #86, opened 2022-05-05
Npm alias protocol does not resolve


stoiky:
As per https://yarnpkg.com/features/protocols a package name may have an alias to another package in the form of:
```
"react-intl-cdo": "npm:react-intl@2.8.0", 
```

This in `pnpm-lock.yaml` transforms to `react-intl-cdo: /react-intl/2.8.0_prop-types@15.7.2+react@17.0.1` and will fail inside `js/private/translate_pnpm_lock.js` with the error below because the `pnpmName` will be `react-intl-cdo//react-intl/2.8.0_prop-types@15.7.2+react@17.0.1`

```
(node:54092) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'dependencies' of undefined
    at gatherTransitiveClosure 
(/Users/mstoichi/code/github.com/rules_js/js/private/translate_pnpm_lock.js:52:34)
    at main (/Users/mstoichi/code/github.com/rules_js/js/private/translate_pnpm_lock.js:188:9)
```

gregmagolan:
I'll pick this one up next week

kzadorozhny:
`v0.11.0` fails with error:
```
ERROR: An error occurred during the fetch of repository 'npm':
   Traceback (most recent call last):
	File "/private/var/tmp/_bazel_zadorozh/a141ef13fb1568f0df4f5afbc3b9310b/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 319, column 33, in _impl
		lockfile = _process_lockfile(rctx)
	File "/private/var/tmp/_bazel_zadorozh/a141ef13fb1568f0df4f5afbc3b9310b/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 145, column 43, in _process_lockfile
		return translate_to_transitive_closure(lockfile, rctx.attr.prod, rctx.attr.dev, rctx.attr.no_optional)
	File "/private/var/tmp/_bazel_zadorozh/a141ef13fb1568f0df4f5afbc3b9310b/external/aspect_rules_js/npm/private/transitive_closure.bzl", line 114, column 34, in translate_to_transitive_closure
		gather_transitive_closure(
	File "/private/var/tmp/_bazel_zadorozh/a141ef13fb1568f0df4f5afbc3b9310b/external/aspect_rules_js/npm/private/transitive_closure.bzl", line 34, column 36, in gather_transitive_closure
		package_info = packages[utils.pnpm_name(name, version)]
[...]

PR #171, opened 2022-06-09
feat: support 'npm:' style package.json aliased npm deps

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/86

**BREAKING CHANGES**

Some breaking refactoring was needed to support aliases linking for 3rd party deps which the `npm:` style dep requires.

* npm_import link_packages attribute is now a dict is package to comma separated aliases to link as. If aliases are "" then it will link as the package name. For example, here is the `@plotly/regl` package which is liked both as its package name and as an alias `regl` in the package `npm/private/test`,

```
    npm_import(
        name = "npm__at_plotly_regl__2.1.2",
        root_package = "",
        link_workspace = "",
        link_packages = {
            "npm/private/test": "@plotly/regl,regl",
        },
        package = "@plotly/regl",
        version = "2.1.2",
```

this comes from the `npm/private/test/package.json` deps:

```
        "regl": "npm:@plotly/regl@2.1.2",
        "@plotly/regl": "2.1.2",
```

* npm_import's generated `npm_link_package` function renamed to `npm_link_imported_package` to differentiate it from the public API `npm_link_package` function and the name should now just be the link root name `node_modules` and not `node_modules/@foo/bar` since the function can make multiple link aliases now. For example,

```
load("@acorn__8.4.0__links//:defs.bzl", npm_link_acorn = "npm_link_imported_package")
npm_link_acorn(name = "node_modules")
```

* removes `npm_link_package_dep` helpers since virtual store targets are now have predictable names. See example in root BUILD file:

```
# Linking a first-party dependency to the virtual store in bazel-bin/node_modules/.aspect_rules_js and
# as a direct link to bazel-bin/node_modules/@mycorp/mylib
npm_link_package(
    name = "node_modules/@mycorp/mylib",
    src = "//examples/lib",
    # We dep on the virtual store link targets which are layed out in the root package, typically
    # the package where the pnpm-lock.yaml file is located.
    deps = {
        # For 3rd party packages that come from npm_translate_package lock, the name starts with the
        # name passed to npm_link_all_packages macro post-fixed with `/<package>/store/<version>`
        # where `package` is the package name (include @scope segment if any) and the `version` is
        # the specific version of the package that comes from the pnpm-lock.yaml file.
        "//:node_modules/uuid/store/8.3.2": "uuid",
        # For 3rd party package that come directly from an npm_import, the name starts with the name
        # passed to that import's npm_link_package macro post-fixed with `/store/<version>` where
        # `version` matches the `version` attribute in the npm_import of the package.
        "//:node_modules/acorn/store/8.4.0": "acorn",
    },
)
```

I am planning follow-up non-breaking re-factoring of npm_import and npm_translate_lock code to make it more readable now that we are close to 1.0.


BUILD.bazel
```diff
@@ -14,22 +12,34 @@ npm_link_all_packages(name = "node_modules")
 # Link the acorn package, which was fetched separately with npm_import from /WORKSPACE, to the
 # virtual store in bazel-bin/node_modules/.aspect_rules_js and as a direct link to
 # bazel-bin/node_modules/acorn
-npm_link_acorn(name = "node_modules/acorn")
+# buildifier: disable=duplicated-name
+npm_link_acorn(name = "node_modules")
+
+# Do the same for the pnpm package
+# buildifier: disable=duplicated-name
+npm_link_pnpm(name = "node_modules")
```
alexeagle, line 20: hmm this is pretty confusing to have two rules named the same. I think the editor plugin and maybe buildifier also will yell at you that you've made a mistake

e2e/bzlmod/BUILD.bazel
```diff
@@ -1,10 +1,11 @@
 load("@aspect_rules_js//js:defs.bzl", "js_test")
 load("@npm//:defs.bzl", "npm_link_all_packages")
-load("@npm_meaning-of-life__links//:defs.bzl", npm_link_meaning_of_life = "npm_link_package")
+load("@npm_meaning-of-life__links//:defs.bzl", npm_link_meaning_of_life = "npm_link_imported_package")
 
 npm_link_all_packages(name = "node_modules")
 
-npm_link_meaning_of_life(name = "node_modules/meaning-of-life")
+# buildifier: disable=duplicated-name
```
alexeagle, line 7: yeah... this seems undesirable, I think we need to fix it.

could it be


```
load("@npm_meaning-of-life__links//:defs.bzl", meaning_of_life = "npm_link_imported_package")
npm_link_all_packages(
    name = "node_modules",
    imported_links = [
        meaning_of_life,
    ]
)
```
??

pnpm-workspace.yaml
```diff
@@ -1,5 +1,5 @@
 packages:
+    - 'examples/lib'
```
alexeagle, line 2: why not `examples/*`

e2e/bzlmod/BUILD.bazel
gregmagolan: Yea, it's not great but OTOH its only a problem with manual npm_imports.

Users could also do,
```
npm_link_meaning_of_life()
```
since `node_modules` is default anyway.

But I also like your suggestion.

pnpm-workspace.yaml
gregmagolan: `examples/assert_lockfile_up_to_date` has a package.json but it shouldn't be a workspace

alexeagle: there's exclude for that
```
packages:
  # all packages in subdirs of packages/ and components/
  - 'packages/**'
  - 'components/**'
  # exclude packages that are inside test directories
  - '!**/test/**'
```

alexeagle: yeah we should probably add that assertion too huh

e2e/bzlmod/BUILD.bazel
gregmagolan: I think your way reads better even for uncommon cases; and it has the added benefit of being able to include the manual npm_imports that are linked direct in a given package to the `:node_modules` catch-all target for that package

gregmagolan: I'll try it out

BUILD.bazel
gregmagolan: fixed up

e2e/bzlmod/BUILD.bazel
gregmagolan: fixed up

mattem: I'm not clear what the outcome was, but leaving out `name` from macros makes buildifier mad, also I use this "query" a bunch, so this would yield missing results:

```
buildozer "print name" //...:all
```

alexeagle: outcome is to have one target with a `name` field, so all good there

PR #172, opened 2022-06-10
feat: import single packages as bzlmod extension

kormide:
I need this for some of the bzlmod e2e additions to the other repos.


alexeagle (approved):
awesome!

PR #174, opened 2022-06-10
fix: string substitution error on windows

kormide:


alexeagle (approved):
yay CI even if it's downstream

PR #183, opened 2022-06-10
fix: failing windows test due to wrong expected line endings

kormide:
The test compares `js_run_binary` `stdout` which always seems to produce unix line endings, and a `write_file` which by default uses the platform's line ending format. I'm not sure which one of the two to change.

### DIFF
--- a/examples/js_binary/BUILD.bazel
+++ b/examples/js_binary/BUILD.bazel
@@ -301,6 +301,7 @@ write_file(
         "        [--tokenize] [--locations] [---allow-hash-bang] [--allow-await-outside-function] [--compact] [--silent] [--module] [--help] [--] [infile]",
         "",
     ],
+    newline = "unix",
 )
 
 diff_test(


alexeagle (approved):

alexeagle:
I don't care that much which we fix... but it does feel more principled that `js_run_binary` should produce platform line endings.

kormide:
> I don't care that much which we fix... but it does feel more principled that `js_run_binary` should produce platform line endings.

I agree that would be the more principled fix. I'll do that instead.

kormide:
Updated to output stdout/err to platforms-specific line ending. I did the transformation after rather than piping output in the node command because it makes it difficult to get the error code of the child process later (I was trying to use `PIPESTATUS` but couldn't get it working). In any case, I think it's cleaner to do this after.

js/private/js_binary.sh.tpl
```diff
@@ -390,6 +390,20 @@ if [ "${JS_BINARY__EXPECTED_EXIT_CODE:-}" ]; then
     fi
 fi
 
+# Use platform line endings in stdout/stderr files
+if [ "$(is_windows)" -eq "1" ]; then
+    SED_WINDOWS_LINE_ENDINGS="s/$/\\r/"
+    TMPFILE=$(mktemp)
+    if [ "${STDOUT_CAPTURE:-}" ]; then
+        cat "$STDOUT_CAPTURE" | sed "$SED_WINDOWS_LINE_ENDINGS" >"$TMPFILE"
```
alexeagle, line 398: This feels strange vs just printing the right thing in the first place. Are you sure we can't fix at the spot the \n was printed?

kormide: I attempted to put this sed pipe above in call to node when we write to stdout/err in the first place, but I had difficulty getting the exit code of the node process. Normally you'd use `PIPESTATUS` get the error code of something earlier in the pipe, but the fact that we run the node process in the background with `&` makes that tricker and I couldn't figure it out.

alexeagle: that doesn't seem like the spot where the printing happens though. shouldn't the program inside the node process write the correct thing?

kormide: If the stdout/stderr files are depended upon by non-test targets, then it would break hermeticity to change the line-endings like this, right?

kormide: I think that would make it fail on linux/osx. If we just want to fix the test, then telling `write_file` that the expected stdout file should have unix line endings would be the most straightforward.

kormide: The original test in question was `test8`, which is no longer in the diff.

alexeagle: let's go with the other solution then, I'd really like to avoid adding this complexity here

alexeagle (approved):

Issue #164, opened 2022-06-06
bug: aliased transitive dependency names do not work


alex-torok:
We have a dependency on plotly.js, which uses an aliased version of one of its dependencies:

https://github.com/plotly/plotly.js/blob/e0a5f13a4c9b8e2f74204499168d12b2d0bab285/package.json#L110

When adding a dependency like this, the pnpm lockfile has a version marker that matches the pnpm name:
```yaml
  /plotly.js/2.12.1:
    resolution: {integrity: sha512-XbZ3w3jSl+ihZPMXWwblmjinqbzdd+vT+3XKZpyffxmlMmTp/3vY1ewpQDuNyufUoiT0o0ekCqa80Qlzi7jwlA==}
    dependencies:
      ...
      probe-image-size: 7.2.3
      regl: /@plotly/regl/2.1.2
      regl-error2d: 2.0.12
```

The issue presents itself first as an error with iterating through the deps in transitive_dependencies.bzl:

```
ERROR: An error occurred during the fetch of repository 'npm':
   Traceback (most recent call last):
        File "/home/alex.torok/.cache/bazel/_bazel_alex.torok/080e6abb1a3d3304a3cf19fc0f4ef9c1/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 319, column 33, in _impl
                lockfile = _process_lockfile(rctx)
        File "/home/alex.torok/.cache/bazel/_bazel_alex.torok/080e6abb1a3d3304a3cf19fc0f4ef9c1/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 145, column 43, in _process_lockfile
                return translate_to_transitive_closure(lockfile, rctx.attr.prod, rctx.attr.dev, rctx.attr.no_optional)
        File "/home/alex.torok/.cache/bazel/_bazel_alex.torok/080e6abb1a3d3304a3cf19fc0f4ef9c1/external/aspect_rules_js/npm/private/transitive_closure.bzl", line 114, column 34, in translate_to_transitive_closure
                gather_transitive_closure(
        File "/home/alex.torok/.cache/bazel/_bazel_alex.torok/080e6abb1a3d3304a3cf19fc0f4ef9c1/external/aspect_rules_js/npm/private/transitive_closure.bzl", line 34, column 36, in gather_transitive_closure
                package_info = packages[utils.pnpm_name(name, version)]
Error: key "regl//@plotly/regl/2.1.2" not found in dictionary
```

[...]

PR #195, opened 2022-06-13
fix: support transitive npm aliased deps in pnpm lockfile

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/164

```
.../rules_js/bazel-bin/node_modules/.aspect_rules_js/plotly.js@2.12.1/node_modules/regl -> /private/var/tmp/_bazel_greg/7aed427a991d86f99332ab79b9b11780/execroot/aspect_rules_js/bazel-out/darwin-fastbuild/bin/node_modules/.aspect_rules_js/@plotly+regl@2.1.2/node_modules/@plotly/regl
```


.bazelignore
```diff
@@ -1,2 +1,7 @@
 e2e/
-examples/node_modules/
+examples/assert_lockfile_up_to_date/node_modules/
```
alexeagle, line 2: nit: could add a test that this file and pnpm-workspace.yaml agree on the node_modules folder locations

alexeagle (approved):

.bazelignore
gregmagolan: agreed; it is annoying that it is can easily fall out of date; users should do the same

gregmagolan: will make an issue to do as a follow-up

gregmagolan: https://github.com/aspect-build/rules_js/issues/196

PR #204, opened 2022-06-14
temporary workaround for fasterci 

apesternikov:
the build is broken here:
https://github.com/aspect-build/rules_js/pull/195/checks?check_run_id=6868676346


alexeagle (approved):
I checked that this works on my mac.

```
     --no-same-owner
             (x mode only) Do not extract owner and group IDs.  This is the reverse of --same-owner and the default
             behavior if tar is run as non-root.

     --no-same-permissions
             (x mode only) Do not extract full permissions (SGID, SUID, sticky bit, ACLs, extended attributes or
             extended file flags).  This is the reverse of -p and the default behavior if tar is run as non-root.
```
seems like this is the default behavior though?

PR #210, opened 2022-06-15
fix: broken tests caused by platform incompatible postinstall scripts

kormide:
Fixes a couple of the diff tests.


alexeagle (approved):

alexeagle:
did we already fix the reason they got broken in the first place? could it happen again?

kormide:
> did we already fix the reason they got broken in the first place? could it happen again?

The failures were just caused by not testing the postinstall scripts on Windows, both in the `@aspect-test` packages as well as the custom ones here. So long as we have a windows ci we should catch them. The incompatibilities are an effect of how npm runs postinstalls and the platform differences, not from rules_js.

gregmagolan (approved):

PR #216, opened 2022-06-16
feat: add out_dirs, mnemonic, progress_message and execution_requirements to js_run_binary rule and remote output_dir

gregmagolan:


js/private/js_run_binary.bzl
```diff
@@ -51,16 +52,21 @@ def js_run_binary(
...
-            If you output a directory, there can only be one output, which will be a
-            directory named the same as the target.
-
         outs: Output files generated by the action.
 
             These labels are available for `$(location)` expansion in `args` and `env`.
 
+        out_dirs: Output directories generated by the action.
+
+            These labels are _not_ available for `$(location)` expansion in `args` and `env` sincew
```
kormide, line 61: nit: spelling

js/private/js_run_binary.bzl
```diff
@@ -30,6 +30,9 @@ def js_run_binary(
         silent_on_success = True,
         copy_srcs_to_bin = True,
         log_level = None,
+        mnemonic = "JsRunBinary",
```
alexeagle, line 33: maybe sth like "UnknownJsBinary" to suggest that it's something a user ought to be setting

npm/private/npm_translate_lock.bzl
```diff
@@ -390,9 +390,6 @@ To disable this warning, set `warn_on_unqualified_tarball_url` to False in your
         ))
     return result
 
-def _join(*elements):
```
alexeagle, line 393: again, I think this is broken because theirs doesn't discard empty segments like when workspace name is "" because it's the local one

npm/private/npm_import.bzl
```diff
@@ -532,8 +534,8 @@ def _impl_links(rctx):
         npm_package_target = npm_package_target,
         npm_package_target_lc = npm_package_target_lc,
         lc_deps = starlark_codegen_utils.to_dict_attr(lc_deps, 2, quote_key = False),
-        lifecycle_build_target = str(rctx.attr.lifecycle_build_target),
-        lifecycle_target_name = lifecycle_target_name,
+        has_lifecycle_build_target = str(rctx.attr.lifecycle_build_target),
+        lifecycle_output_dir = lifecycle_output_dir,
```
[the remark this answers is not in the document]
kormide: n/m, didn't see context

kormide (approved):

npm/private/npm_translate_lock.bzl
gregmagolan: I think it does reading the impl, https://github.com/bazelbuild/bazel-skylib/blob/main/lib/paths.bzl#L88

The only way it could go terribly wrong is if you pass `None` values to it instead of empty strings.

js/private/js_run_binary.bzl
gregmagolan: hmmm. not sure on that one. maybe if we set a good mnemonic for all of the generated "bins" then that would make sense otherwise it would be everywhere.

gregmagolan: lets leave it as a follow-up

PR #222, opened 2022-06-17
feat: add public_hoist_packages attribute to npm_translate_lock to emulate .npmrc public-hoist-pattern[]

gregmagolan:
Turns out pnpm does not encode what it hoists in its lockfile. It is a link time decision in the implementation.

This lays the foundation for supporting https://pnpm.io/npmrc#public-hoist-pattern and https://pnpm.io/npmrc#shamefully-hoist.

Follow-ups to this would be to support globs to get closer to what users can do with `.npmrc`. For example,

`.npmrc`:
```
public-hoist-pattern[]=*eslint*
public-hoist-pattern[]=*prettier*
public-hoist-pattern[]=*plugin*
```

could be emulated with

```
    public_hoist_packages = {
        "*eslint*": [""],
        "*prettier*": [""],
        "*plugin*": [""],
    },
```

and

`.npmrc`:
```
shamefully-hoist=true
```

with

```
    public_hoist_packages = {
        "*": [""],
    },
```

Once we can do that then we could just let users use the `.npmrc` file as the source of truth instead:

```
npm_translate_lock(
    name = "npm",
    npmrc = "//:.npmrc",
    ...
)
```

NB: `.npmrc` is an INI file https://en.wikipedia.org/wiki/INI_file


.npmrc
```diff
@@ -1,3 +1,4 @@
 # Set a custom registry for a scope that is picked up by pnpm when resolving packages;
 # This affects the lockfile format and is here to cover this case.
 @types:registry=https://registry.yarnpkg.com
+public-hoist-pattern[]=ms
```
alexeagle, line 4: comment that bazel / rules_js won't actually see this yet, it only has effect for `pnpm i`

examples/npm_deps/BUILD.bazel
```diff
@@ -210,3 +210,24 @@ uvu_test(
         ":uvu_spec",
     ],
 )
+
+#######################################
+# Use case 6: depend on a hoisted npm package that wasn't a direct dependency package.json
```
alexeagle, line 215: would be nice to say which direct dependency has the `ms` dependency

npm/private/npm_translate_lock.bzl
```diff
@@ -331,6 +339,18 @@ def _gen_npm_imports(lockfile, attr):
                     else:
                         link_packages[link_package].append(dep_package)
 
+        # check if this package should be hoisted via public_hoist_packages
+        public_hoist_packages = []
+        public_hoist_packages = attr.public_hoist_packages.get(name, [])[:]
+        public_hoist_packages.extend(attr.public_hoist_packages.get(friendly_name, []))
+        if unfriendly_name:
+            public_hoist_packages.extend(attr.patches.get(unfriendly_name, []))
+        for public_hoist_package in public_hoist_packages:
```
alexeagle, line 348: nice short code change is a good sign that the abstractions will endure more feature sprawl :)

alexeagle (approved):

examples/npm_deps/BUILD.bazel
gregmagolan: ✅

npm/private/npm_translate_lock.bzl
gregmagolan: 🌮

.npmrc
gregmagolan: ✅

PR #224, opened 2022-06-17
fix: improve error handling if there are conflicting public hoisted packages

gregmagolan:
For example,

```
    public_hoist_packages = {
        "ms": ["examples/npm_deps"],
    },
```

produces error messaage:

```
ERROR: /Users/greg/aspect/rules/js/rules_js/WORKSPACE:46:19: fetching npm_translate_lock rule //external:npm: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_greg/7aed427a991d86f99332ab79b9b11780/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 649, column 21, in _impl
		fail(msg)
Error in fail: 

Invalid public hoist configuration with multiple packages to hoist to `examples/npm_deps/node_modules/ms`: ["ms@0.7.3", "ms@2.0.0", "ms@2.1.2", "ms@2.1.3"]

Trying selecting a specific version of `ms` to hoist in public_hoist_packages. For example `ms@0.7.3`:

    public_hoist_packages = {
        "ms@0.7.3": ["examples/npm_deps"]
    }
```

and 

```
    public_hoist_packages = {
        "ms@0.7.3": ["examples/npm_deps"],
        "ms@2.1.3": ["examples/npm_deps"],
    },
```

produces,

```
ERROR: no such package '@npm//': 

Invalid public hoist configuration with multiple packages to hoist to `examples/npm_deps/node_modules/ms`: ["ms@0.7.3", "ms@2.1.3"]

Check the public_hoist_packages attribute for duplicates.
```

### DIFF
--- a/npm/private/npm_translate_lock.bzl
+++ b/npm/private/npm_translate_lock.bzl
@@ -442,6 +442,51 @@ def _verify_node_modules_ignored(rctx, importer_paths, bazelignore):
             missing_ignores.append(expected)
     return missing_ignores
 
+def _check_for_conflicting_public_links(npm_imports, public_hoist_packages):
+    if not public_hoist_packages:
+        return
+    all_public_links = {}
+    for (i, _import) in enumerate(npm_imports):
+        for link_package, link_names in _import.link_packages.items():
+            if link_package not in all_public_links:
+                all_public_links[link_package] = {}
+            for link_name in link_names:
+                if link_name not in all_public_links[link_package]:
+                    all_public_links[link_package][link_name] = []
+                all_public_links[link_package][link_name].append("{}@{}".format(_import.package, _import.version))
+    for link_package, link_names in all_public_links.items():
+        for link_name, link_packages in link_names.items():
+            if len(link_packages) > 1:
+                if link_name in public_hoist_packages:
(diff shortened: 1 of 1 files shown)


npm/private/npm_translate_lock.bzl
```diff
@@ -442,6 +442,49 @@ def _verify_node_modules_ignored(rctx, importer_paths, bazelignore):
             missing_ignores.append(expected)
     return missing_ignores
 
+def _check_for_conflicting_public_links(npm_imports, public_hoist_packages):
+    all_public_links = {}
+    for (i, _import) in enumerate(npm_imports):
+        for link_package, link_names in _import.link_packages.items():
+            if link_package not in all_public_links:
+                all_public_links[link_package] = {}
+            for link_name in link_names:
```
alexeagle, line 451: O(n^3) complexity - we should stay vigilant about this becoming slow at some point? it's an "eager fetch" evaluation for ALL bazel users in the workspace. Seems like an easy fix here is just to short-circuit and exit the function at the beginning if `public_hoist_packages` is empty

gregmagolan: good call; link_packages and link_names should be small in most cases so this should loop should not grow exponentially in large sets of npm deps but still better to not loop if there are no public host packages

alexeagle (approved):

PR #230, opened 2022-06-20
chore: simplify 1p dep example

gregmagolan:


examples/lib/BUILD.bazel
```diff
@@ -6,6 +6,9 @@ npm_package(
         "index.js",
         "package.json",
     ],
-    package = "@mycorp/mylib",
+    # `package = "@mycorp/mylib"` does not need to be specified since the name
+    # that this package is linked as is set by the pnpm lockfile. If it is
```
alexeagle, line 10: should/could it print a warning/error if you *do* specify package in a place where the value will be ignored?

alexeagle (approved):

examples/lib/BUILD.bazel
gregmagolan: will make an issue to do as a follow-up

Issue #220, opened 2022-06-17
Problem with resolving dependencies under webpack


alex-torok:
I'm hitting an issue with webpack resolving dependencies. I'm struggling to get a working reproduction going in the `rules_js` repo and am hoping that posting my findings so far will be enough to help diagnose the issue and see if it is related to rules_js.

Apologies if this is long-winded, but I wanted to provide as much context as possible.

The project that I'm bringing rules_js into is currently based on rules_nodejs and uses some of our own custom macros around the autogenerated `webpack_cli` (from the `bin` field of its `package.json`) to kick off a webpack build. I'm attempting to make the migration to rules_js easier for teams at my company by setting up dependency targets that choose rules_js or rules_nodejs npm targets based on a build configuration, and the different autogenerated `webpack_cli` macros based on a flag. With this, I apply a config transition on the `webpack_cli` target outputs to get my proxy targets to choose rules_js or rules_nodejs npm targets.

I'm not sure if relevant, but our repo is configured to run bazel actions in a tmpfs with `build --sandbox_base=/dev/shm` in the `.bazelrc`, which is why many of the paths in the below logs have that prefix.

I have the webpack action running with rules_js, but webpack is showing errors with resolving transitive npm dependencies.

<details>
  <summary>Click to see the full webpack error, but I'll summarize below</summary>
  
```
ModuleNotFoundError: Module not found: Error: Can't resolve '@material-ui/styles' in '/dev/shm/bazel-sandbox.c232cb948e685633b81332a1d7238c336adde8c94b02204f66763727dc68ee39/linux-sandbox/2218/execroot/__main__/bazel-out/k8-opt-ST-1db56f8cc3dd/bin/node_modules/@material-ui/core/esm/styles'
    at /home/alex.torok/.cache/bazel/_bazel_alex.torok/f8d954fb8d563619a1a7548575ac13e0/execroot/__main__/bazel-out/k8-opt-ST-1db56f8cc3dd/bin/node_modules/.aspect_rules_js/webpack@4.40.2_webpack-cli@3.3.8/node_modules/webpack/lib/Compilation.js:925:10
[...]

PR #237, opened 2022-06-22
fix: improved symlink guards for sandbox, runfiles & execroot

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/220

Symlink guards improved & new unit test cases added


examples/webpack_cli/index.js
```diff
@@ -0,0 +1,36 @@
+const {
```
alexeagle, line 1: where did the example come from? maybe attribution in the README

npm/private/npm_translate_lock.bzl
```diff
@@ -598,9 +597,11 @@ load("@aspect_rules_js//npm/private:npm_linked_packages.bzl", "npm_linked_packag
                     }
 
     if fp_links:
-        defs_bzl_header.append("""load("@aspect_rules_js//npm/private:npm_link_package.bzl",
+        defs_bzl_header.append("""load(
+    "@aspect_rules_js//npm/private:npm_link_package.bzl",
```
alexeagle, line 601: nit: unrelated refactoring?

js/private/node-patches/src/fs.ts
```diff
@@ -549,73 +454,270 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
...
+                    const stat = fs.statSync(origRealpathSync(f))
                     patchDirent(v, stat)
                 }
             }
         }
     }
+
+    function nextHop(loc: string, cb: (next: string | false) => void): void {
+        let nested = []
+        const oneHop = (maybe, cb: (next: string | false) => void) => {
```
alexeagle, line 496: this `cb` is shadowing the one from the outer function, maybe we should have a lint for that?

alexeagle (approved):
wow lots of new code, too much to review line-by-line in detail!

npm/private/npm_translate_lock.bzl
gregmagolan: to satisfy lint; not sure how it got out of sync; but I'll split out the example into its own commit with proper attribution and then this change will go with that.

examples/webpack_cli/index.js
gregmagolan: 👍

js/private/node-patches/src/fs.ts
gregmagolan: in this case it is intentional since it is the same cb.
```
const oneHop = (maybe, cb: (next: string | false) => void) => {...}
```
is followed by
```
oneHop(loc, cb)
```
which just passed the `cb` passed to the function forward

PR #242, opened 2022-06-23
fix: use camel case in generated bin mnemonics

alex-torok:
https://github.com/aspect-build/rules_js/pull/218 added the ability to use the bin name as the default mnemonic for generated package_json.bzl entries. Some generated bin entries have underscores in their name, which is not allowed in action mnemonics. 

Without this change, you can get:
```
Error in run: mnemonic must only contain letters and/or digits, and have non-zero length, was: "webpack_cli"
```

Use a CamelCase version of the bin name instead.

### DIFF
--- a/npm/private/npm_import.bzl
+++ b/npm/private/npm_import.bzl
@@ -379,11 +379,12 @@ def _impl(rctx):
                 """load("@aspect_rules_js//js:defs.bzl", _js_binary = "js_binary", _js_run_binary = "js_run_binary", _js_test = "js_test")""",
             ]
             for name in bins:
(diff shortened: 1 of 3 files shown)


alex-torok:
Oops, just realized spaces are a no-go too - `Error in run: mnemonic must only contain letters and/or digits, and have non-zero length, was: "webpack cli"` I'll see if I can CamelCase the mnemonic.

gregmagolan (approved):
Sweet. Thanks for the fix. I hit this yesterday as well and Alex started fixing it and then got busy with other things.

PR #243, opened 2022-06-23
ci: auto-publish to bcr on release

kormide:


.github/workflows/bcr/presubmit.yml
```diff
@@ -0,0 +1,10 @@
+bcr_test_module:
+    module_path: 'e2e/bzlmod'
+    matrix:
+        platform: ['debian10', 'macos', 'ubuntu2004', 'windows']
+    tasks:
+        run_tests:
+            name: 'Run test module'
+            platform: ${{ platform }}
+            test_targets:
+                - '//...'
```
kormide, line 10: I couldn't find a spec for `presubmit.yml`. If we don't include build and test targets, does it automatically use `//...`?

.github/workflows/bcr/source.template.json
```diff
@@ -0,0 +1,5 @@
+{
+    "integrity": "**leave this alone**",
```
alexeagle, line 2: do we actually have to have exactly this content in the value?

.github/workflows/bcr/presubmit.yml
alexeagle: take a look in Yun's https://github.com/meteorcloudy/bzlmod-examples

.github/workflows/bcr/source.template.json
kormide: Nope

.github/workflows/bcr/presubmit.yml
kormide: I took a look but he has the same example everywhere: https://github.com/meteorcloudy/bzlmod-examples/blob/main/registry/another/modules/A/1.1/presubmit.yml

alexeagle (approved):
Is it good to have so much duplication between the two bcr*.yml files vs. having both workflows defined in one?

kormide:
> Is it good to have so much duplication between the two bcr*.yml files vs. having both workflows defined in one?

It's not great. The publishing to our own bcr was sort of an after thought, and I didn't want to mix them in case we bring the real bcr one into bazel-contrib.

kormide:
> > Is it good to have so much duplication between the two bcr*.yml files vs. having both workflows defined in one?
> 
> It's not great. The publishing to our own bcr was sort of an after thought, and I didn't want to mix them in case we bring the real bcr one into bazel-contrib.

Oh and there would be a lot of repetition anyway since the create pr plugin commits the working directory and publishes the pr all in one. We'd have to redo most of the steps to form the bcr entry in the working directory to call it again.

PR #246, opened 2022-06-25
util: add basic ini parser

jbedard:
Is this about what we're looking for? A basic skylark method that can be used anywhere... 🤷 

idk if we have a standard for where to put private utils like this?

### DIFF
--- a/npm/private/test/BUILD.bazel
+++ b/npm/private/test/BUILD.bazel
@@ -4,6 +4,7 @@ load(":transitive_closure_tests.bzl", "transitive_closure_tests")
 load(":translate_lock_tests.bzl", "translate_lock_tests")
 load(":yaml_tests.bzl", "yaml_tests")
 load(":utils_tests.bzl", "utils_tests")
+load(":ini_test.bzl", "ini_tests")
 
 # gazelle:exclude *_checked.bzl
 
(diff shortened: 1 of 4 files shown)


alexeagle:
Is there anything JS-specific about this? I would have imagined it belongs in bazel-lib so that we could use it in all our rulesets.

(Note, the yaml parser is JS-specific because we only bothered to implement the subset that pnpm-lock files cover)

jbedard:
This is also only a subset which npmrc requires, but I agree it should be in bazel-lib if/when it is a full parser. See the comment specifying what is not supported.

alexeagle (approved):
move to `npm/private` next to yaml.bzl then, and just call `ini.bzl`

jbedard:
Note we'll also have a dumb partial glob implementation. I think that could also go into `npm/private`?

npm/private/test/BUILD.bazel
```diff
@@ -4,13 +4,15 @@ load(":transitive_closure_tests.bzl", "transitive_closure_tests")
...
 load(":utils_tests.bzl", "utils_tests")
+load(":ini_test.bzl", "init_tests")
 
 # gazelle:exclude *_checked.bzl
 
 npm_link_all_packages(name = "node_modules")
 
 # Unit tests
 utils_tests(name = "test_utils")
+init_tests(name = "test_ini")
```
alexeagle, line 15: init rather than ini ?

jbedard: I keep doing that 😡

Thank you lol

alexeagle (approved):

PR #252, opened 2022-06-27
fix: remove DefaultInfo.files from js_library

### LINKED ISSUE #251: js_library should not use files attribute of DefaultInfo
js_library is meant to be a general purpose means of gathering output artifacts from various sources which contribute to a javascript library. These assets can include various modules from node_modules, images, css assets in addition to javascript sources. The set of files that can be gathered is very very large. 

The files attribute of DefaultInfo represents the default outputs to build when this target is specified on the bazel command line. That being the case, every file inside files will appear on the command line / be uploaded to the build event stream as an important output file. This is not desirable behavior since the file set is so large. 

In practice this has caused annoyances due to command line output being noisy with thousands of files and has cause build event streams to crash due to the misuse of the number of files deemed important.

mrmeku:
Fixes #251

### DIFF
--- a/js/private/js_library.bzl
+++ b/js/private/js_library.bzl
@@ -86,7 +86,7 @@ target in {file_basename}'s package and add that target to the deps of {this_tar
 
     return [
         DefaultInfo(
-            files = depset(transitive = files_depsets),
+            files = depset(direct = output_srcs),
             runfiles = runfiles.merge_all(deps_runfiles),
         ),
         declaration_info(


gregmagolan (approved):
🌮 

Thanks for the fix Mr. Muller.

matthewjh:
@alexeagle @gregmagolan @alex-torok feels like we should have an e2e test that breaks without this?

PR #255, opened 2022-06-27
docs: add usage guide

alexeagle:


CONTRIBUTING.md
```diff
@@ -31,11 +31,13 @@ directory:
...
 ```sh
 OVERRIDE="--override_repository=aspect_rules_js=$(pwd)/rules_js"
-echo "build $OVERRIDE" >> ~/.bazelrc
-echo "query $OVERRIDE" >> ~/.bazelrc
+echo "common:override $OVERRIDE" >> ~/.bazelrc
 ```
 
-This means that any usage of `@aspect_rules_js` on your system will point to this folder.
+This means that any usage of `@aspect_rules_js` on your system will point to this folder, if you
+pass the `--config=override` flag to Bazel.
```
JiaLiPassion, line 38: Should we name the name of this config to `dev` or `local`?

CONTRIBUTING.md
```diff
@@ -31,11 +31,13 @@ directory:
...
+(We don't want to enable this for all builds, since you'll later forget this was done and wonder
+why you don't reproduce the same behavior as others.)
 
 ## Releasing
```
JiaLiPassion, line 42: Also not related to this PR, is there any way to do a local `test` before commit and push?

docs/README.md
```diff
@@ -0,0 +1,320 @@
...
+-   Check for [known issues](https://github.com/aspect-build/rules_js/issues)
+-   Pay for support, provided by <https://aspect.dev>.
+
+## Installation
+
+From the release you wish to use:
+<https://github.com/aspect-build/rules_js/releases>
+copy the WORKSPACE snippet into your `WORKSPACE` file.
+
+Note that [bzlmod] is experimentally supported as well.
```
JiaLiPassion, line 18: Should we add a `snippet` to describe how to use `bzlmod` ?

docs/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Node.js
+
+rules_js depends on rules_nodejs version 5.0 or greater.
+
+Installation is included in the `WORKSPACE` snippet you pasted from the Installation instructions above.
+
+**API docs:**
+
+-   Installing and choosing the version of Node.js:
```
JiaLiPassion, line 48: In my understanding, the `nodejs` toolchain is installed in the `WORKSPACE` snippet above, we don't need to `install` nodejs ourselves. Is that correct?

docs/README.md
```diff
@@ -0,0 +1,320 @@
...
+Create such a file if you don't have one. You could install pnpm on your machine, or use `npx` to run it.
+For example, this command creates a lockfile with minimal installation needed:
+
+```shell
+$ npx pnpm install --lockfile-only
+```
+
+Next, you'll typically use `npm_translate_lock` to translate the Yaml file to Starlark, which Bazel extensions understand.
+The `WORKSPACE` snippet you pasted above already contains this code.
+
```
JiaLiPassion, line 73: Should we post two screenshots or code here to display the `pnpm lock` and the `Starlark` version side by side, it may be eaiser for the user to understand.

docs/README.md
```diff
@@ -0,0 +1,320 @@
...
+1. You don't need to install pnpm on your machine to build and test with Bazel.
+1. We re-use pnpm's resolver, by consuming the `pnpm-lock.yaml` file it produces.
+1. We use Bazel's downloader API to fetch package tarballs and extract them to external repositories.
+   To modify the URLs Bazel uses to download packages (for example, to fetch from Artifactory), read
+   <https://blog.aspect.dev/configuring-bazels-downloader>.
+1. We re-use the [`@pnpm/lifecycle`](https://www.npmjs.com/package/@pnpm/lifecycle) package to perform postinstall steps.
+   (These run as cacheable Bazel actions.)
+1. Finally, you link the `node_modules` tree by adding a `npm_link_package` or `npm_link_all_packages` in your `BUILD` file,
+   which populates a tree under `bazel-bin/[path/to/package]/node_modules`.
```
JiaLiPassion, line 84: I believe this is the most important part of the rulejs core design logic, maybe we can add more information with sequence to describe the flow with more details in the future PR.

docs/README.md
thesayyn: maybe link to `e2e/bzlmod` for the full example.

docs/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+You can also write your own custom rule, though this is an advanced topic and not covered in this documentation.
+
+### Documenting your macros and custom rules
+
+You can use [stardoc] to produce API documentation from Starlark code.
+We recommend producing Markdown output, and checking those `.md` files into your source repository.
+This makes it easy to browse them at the same revision as the sources.
+
+You'll need to create `bzl_library` targets for your starlark files.
```
JiaLiPassion, line 274: Use `Starlark` (Capital first letter) to keep consistant in the doc.

docs/README.md
```diff
@@ -0,0 +1,320 @@
...
+    ],
+)
+
+stardoc_with_diff_test(
+    name = "tsc-docs",
+    bzl_library_target = ":tsc",
+)
+
+update_docs(name = "docs")
+```
```
JiaLiPassion, line 300: Do we have a workable example in the `bazel_examples` repo? If we do have, we can add a link here.

docs/js_library.md
```diff
@@ -1,11 +1,25 @@
 <!-- Generated with Stardoc: http://skydoc.bazel.build -->
 
-Expose some files with DeclarationInfo, like filegroup but can be a dep of ts_project.
+`js_library` is similar to [`filegroup`](https://docs.bazel.build/versions/main/be/general.html#filegroup); there are no Bazel actions to run.
```
JiaLiPassion, line 3: Just a question, is there any differences between `js_library` and `filegroup` rule?

alexeagle:
> 1. Tools installation.
Added the bazelisk installation note. Bazelisk installs Bazel, so there's no need to mention that, nor any version info.

> 2. Since we are using the 5.x of Bazel in the bazelrc, should we also update the pre-commit-buildifier version?
Sure, updated to latest in .pre-commit-config.yaml. However if you got some version skew, that seems like a bug - pre-commit should install the version we ask for regardless what you setup on your machine...

CONTRIBUTING.md
alexeagle: sure, `bazel test ...` like usual.

docs/README.md
alexeagle: that's right. I'll just label this one "choosing the version..."

docs/README.md
alexeagle: I don't think users would understand the content of either file actually. They're both opaque blobs you never read, and GitHub and other code review tools don't even show you the changes. I did add a section here about how you could check in the generated file though, rather than compute on-the-fly. Do you think it's understandable?

docs/README.md
alexeagle: added a section below about how we use the bazel-bin folder for srcs, genfiles, and deps

PR #249, opened 2022-06-27
feat: add an example for js_image_layer

thesayyn:
Closes #116

PR #256, opened 2022-06-28
refactor: ask the runfiles API to create our symlinks

alexeagle:
This lets them be propagated to a container_image layer which also wants to have them, just via the binary target

Pulled this file from https://github.com/aspect-build/rules_js/pull/249/


gregmagolan (approved):
🌮

PR #260, opened 2022-06-28
util: add package glob util

jbedard:


npm/private/pkg_glob.bzl
```diff
@@ -0,0 +1,72 @@
...
+
+    Returns:
+        True if the package matches the glob expression
+    """
+    exp_i = 0
+    pkg_i = 0
+
+    # HACK: assume "|" will never be in the expression and use it to
+    # split on * while keeping * as an array entry.
+    exp_parts = exp.replace("*", "|*|").strip("|").split("|")
```
gregmagolan, line 25: hmm. this might blow up if a user passes "**"; should probably fail if there is a "**" and say that this isn't valid syntax for this function

gregmagolan (approved):
🌮

Issue #167, opened 2022-06-07
bug: `file:` package protocol not supported


vpanta:
Admittedly, this might be more WAI then a bug, and we can likely work around it, but filing as a bug in case y'all decide it's a necessary feature.

We have local packages which we reference via a `file:` protocol. When we have them in our `pnpm-lock.yaml` and execute a `bazel sync` with `rules_js@0.11.0`, we get the following error:
```
ERROR: An error occurred during the fetch of repository 'npm':
   Traceback (most recent call last):
        File "/private/var/tmp/_bazel_vasilios/96c9699731d9b99e83af4f7d36096194/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 319, column 33, in _impl
                lockfile = _process_lockfile(rctx)
        File "/private/var/tmp/_bazel_vasilios/96c9699731d9b99e83af4f7d36096194/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 145, column 43, in _process_lockfile
                return translate_to_transitive_closure(lockfile, rctx.attr.prod, rctx.attr.dev, rctx.attr.no_optional)
        File "/private/var/tmp/_bazel_vasilios/96c9699731d9b99e83af4f7d36096194/external/aspect_rules_js/npm/private/transitive_closure.bzl", line 86, column 17, in translate_to_transitive_closure
                fail("unsupported package path " + package_path)
Error in fail: unsupported package path file:packages-dist/admin-bunsen
```

Similar to #166, the error here is in recognition of the listed `packages`, which in this case begin with `file:` and have a `resolution.directory` entry. We can/will prune these from the `pnpm-lock.yaml`, as we'd prefer local packages to reference each other via local `bazel` targets. But it is possibly worth an explicit comment that local packages are not allowed, or some feature allowing them to be ignored when processing `pnpm-lock.yaml`?

gregmagolan:
I'll think about this one but I think you are correct that we should either prune them or support it

PR #262, opened 2022-06-28
feat: support file: links in package.json & pnpm lockfile

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/167

NB: packages references by `file:` links cannot themselves `file:` or `workspace:` links. You'll get a friendly error message if you try. For that pattern just switch to pnpm workspaces and drop the `file:` links altogether and use `workspace:` links between packages.

### DIFF
--- a/js/private/js_run_binary.bzl
+++ b/js/private/js_run_binary.bzl
@@ -183,7 +183,7 @@ def js_run_binary(
         for log_level_env in _js_binary_envs_for_log_level(log_level):
             extra_env[log_level_env] = "1"
 
-    if len(outs) + len(out_dirs) < 1:
+    if not stdout and not stderr and not exit_code_out and (len(outs) + len(out_dirs) < 1):
         # run_binary will produce the actual error, but we want to give an additional JS-specific
         # warning message here. Note that as a macro, we can't tell the name of the rule provided
         # by the users BUILD file (e.g. for "typescript_bin.tsc(outs = [])" we'd wish to say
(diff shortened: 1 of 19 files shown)


js/private/js_run_binary.bzl
```diff
@@ -183,7 +183,7 @@ def js_run_binary(
         for log_level_env in _js_binary_envs_for_log_level(log_level):
             extra_env[log_level_env] = "1"
 
-    if len(outs) + len(out_dirs) < 1:
+    if not stdout and not stderr and not exit_code_out and (len(outs) + len(out_dirs) < 1):
```
alexeagle, line 186: ah thanks, I saw the logspam...

npm/private/npm_translate_lock.bzl
```diff
@@ -566,7 +580,52 @@ load("@aspect_rules_js//npm/private:npm_linked_packages.bzl", "npm_linked_packag
             fail("expected dict of dependencies in processed importer '%s'" % import_path)
         link_package = _link_package(root_package, import_path)
         for dep_package, dep_version in dependencies.items():
-            if dep_version.startswith("link:"):
+            if dep_version.startswith("file:"):
```
alexeagle, line 583: wish we had some smaller tests for this as well, rather than just the e2e - we do have some starlark unit tests for this file already, anything to add there?

alexeagle (approved):

npm/private/npm_translate_lock.bzl
gregmagolan: we have the golden tests as well which are pretty good if you know how to read them; but yes, some more unit tests in the future. this large function also needs to be broken up into smaller ones that can be unit tested individually

PR #265, opened 2022-06-29
bug: js_library with 'deps' no longer pulls deps into sandbox output tree

alex-torok:
After updating to the latest release of rules_js, I hit an issue where `js_library` targets no longer pull the files from their `deps` into the sandbox's bin directory. I bisected the problem to #252, which modified the `DefaultInfo` for `js_library`.

I'm not sure if I'm not following the correct usage pattern for `js_library`, but also noticed that there are no cases of `js_library` with `deps` on a `:node_modules/*` target within the `examples/` or `e2e/` directories.

This PR reproduces what I'm seeing by building on top of the existing `webpack_cli` example. You can see the build failure with `bazel build //examples/webpack_cli:bundle`

Should `js_library` work in the way showcased here? If not, what is the recommended way to express a library target with srcs and a dependency on an npm library that gets consumed by a `js_run_binary` target?

### DIFF
--- a/examples/webpack_cli/BUILD.bazel
+++ b/examples/webpack_cli/BUILD.bazel
@@ -48,5 +48,4 @@ bin.webpack_cli(
     ],
     log_level = "debug",
     chdir = package_name(),
-    silent_on_success = False,
 )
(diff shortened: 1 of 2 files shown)


alex-torok:
fyi @gregmagolan

alexeagle:
I think your usage looks right and we should probably revert that change and merge this.

alexeagle:
@mrmeku I re-read your description on #252 and I guess we'll be back to command-line spam and crashing BEP stream when we revert. But I don't think runfiles is sufficient to propagate 3p dependencies to an action like webpack, and so far @gregmagolan has resisted adding any additional providers that would let us propagate these outside of DefaultInfo.

alexeagle:
https://github.com/aspect-build/bazel-examples/pull/40 - we also broke our Angular example with that change, so I'm reverting it now, sorry @mrmeku

alexeagle (approved):

PR #275, opened 2022-07-06
docs: remove stale reference to non-linked 3p targets

alexeagle:


npm/private/npm_translate_lock.bzl
```diff
@@ -92,8 +83,6 @@ write_source_files(
 ```
 
 Then in `WORKSPACE`, load from that checked-in copy or instruct your users to do so.
-In this case, the aliases are not created, so you get only the `npm_import` behavior
```
alexeagle, line 95: I'm not sure if we still need something here to warn "hey the @npm_deps repo won't exist in this case"

thesayyn: I think we still need that since we have the bin aliases.

npm/private/npm_translate_lock.bzl
```diff
@@ -54,23 +56,12 @@ However it causes Bazel to eagerly evaluate the `npm_translate_lock` rule for ev
 even if the user didn't ask for anything JavaScript-related.
```
thesayyn, line 56: L53-L54 still mentions the aliases.

npm/private/npm_translate_lock.bzl
thesayyn: https://github.com/aspect-build/rules_js/blob/506d26d59b356514d0a0a38ea33fc7d6eacc6677/npm/private/npm_translate_lock.bzl#L808-L839 we still have bin aliases.

Issue #272, opened 2022-07-05
npm_translate_lock: accept package-lock.json / yarn.lock files and run `pnpm import`


alexeagle:
From @jfirebaugh, a suggestion that I like a lot:

- Users don't have to be re-trained, they continue to use npm/yarn locally to install packages and update the lockfile
- Bazel, via rules_js, translates the lockfile as one more subprocess spawn in our repository rule
- This means that Bazel behavior can be different from local, due to hoisting bugs and undeclared dependency edges

We just need to ensure that `pnpm import` is fast enough and hermetic/deterministic.

Effectively, under Bazel you're accepting the pnpm migration, but end-users don't have to know about that yet. So this has different tradeoffs from @gregmagolan 's idea of running `npm install`/`yarn install` as a transition to rules_js.

gregmagolan:
Done in #276

PR #276, opened 2022-07-06
npm_translate_lock: accept package-lock.json and run `pnpm import`

jfirebaugh:
Implements the package.json/package-lock.json portion of #272 (I didn't include yarn.lock support).


npm/private/npm_translate_lock.bzl
```diff
@@ -506,21 +497,72 @@ Check the public_hoist_packages attribute for duplicates.
...
+            executable = False,
+        )
+        rctx.file(
+            "package-lock.json",
+            content = rctx.read(rctx.attr.npm_package_lock),
+            executable = False,
+        )
+        rctx.execute([
+            rctx.path(Label("@nodejs_host//:bin/node")),
+            rctx.path(Label("@pnpm//:package/bin/pnpm.cjs")),
```
jfirebaugh, line 539: I was somewhat surprised that this seemed to work without adding `npm_import(name = "pnpm", package = "pnpm", ...)` anywhere. Is there already a `@pnpm` repository available?

alexeagle: yes, it was added here: https://github.com/aspect-build/rules_js/pull/157/files#diff-5493ff8e9397811510e780de47c57abb70137f1afe85d1519130dc3679d60ce5R78

But since that's not a dependency we expose to users, this code will fail under some other workspace, such as in the e2e/ folders.

WORKSPACE
```diff
@@ -91,3 +91,13 @@ npm_import(
     root_package = "",
     version = "6.32.19",
 )
+
+npm_translate_lock(
+    name = "npm_translate_package_lock",
+    npm_package_json = "//examples/npm_translate_package_lock:package.json",
```
alexeagle, line 97: this is undesirable, we don't want to invalidate repo rules just because a non-dependency-related change was made to package.json - can we avoid reading it?

alexeagle (commented):
API shape change seems about right to me

npm/private/npm_translate_lock.bzl
```diff
@@ -506,21 +497,72 @@ Check the public_hoist_packages attribute for duplicates.
...
+            str(rctx.attr.pnpm_lock.workspace_name),
+            str(rctx.attr.pnpm_lock),
+        )
+    else:
+        if rctx.attr.npm_package_json == None or rctx.attr.npm_package_lock == None:
+            fail("npm_translate_lock requires either pnpm_lock or npm_package_json/npm_package_lock attributes.")
+
+        rctx.file(
+            "package.json",
+            content = rctx.read(rctx.attr.npm_package_json),
```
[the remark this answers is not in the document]
jfirebaugh: Looks like this is what is read from it: https://github.com/pnpm/pnpm/blob/2a34b21ced6e4d033659405cec63b182649e8238/packages/plugin-commands-installation/src/getOptionsFromRootManifest.ts

I don't know exactly what all of these options do, but it seems reasonable to omit support for them and supply an empty package.json. Those that need them can fall back to manual `pnpm import` + `pnpm_lock` attribute.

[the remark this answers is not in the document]
jfirebaugh: Nope, doesn't work. It reads other stuff too, and if package.json is empty it doesn't generate a pnpm-lock.yaml.

Sorry, I don't think there's a way around this.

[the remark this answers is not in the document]
alexeagle: okay well there are a couple options:
- just document that when you use this variant, it's going to cause invalidations. It's not as bad as rules_nodejs - we won't re-run a whole `npm install` when it's invalidated, so maybe the perf loss isn't that noticeable
- get fancy and run `jq` on the package.json to derive one that only has the subset of bits pnpm reads - but I don't think this would work to avoid invalidations anyway since it's still one big repository rule. And it's brittle, pnpm could read more in their next version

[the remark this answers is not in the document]
alexeagle: I documented that passing `package_json` causes invalidations.

npm/private/npm_translate_lock.bzl
alexeagle: I changed `npm_translate_lock` to a wrapper macro so that it can `maybe()` fetch the `@pnpm` repository in case one of the alternate lockfiles is provided.

npm/private/npm_translate_lock.bzl
```diff
@@ -511,22 +519,92 @@ Check the public_hoist_packages attribute for duplicates.
...
+        fail("npm_translate_lock with yarn_lock or npm_package_lock attribute also requires that the package_json attribute is set")
+    if rctx.attr.pnpm_lock:
+        count += 1
+
+        # don't allow a pnpm lock file that isn't in the root directory of a bazel package
+        if paths.dirname(rctx.attr.pnpm_lock.name):
+            fail("pnpm-lock.yaml file must be at the root of a bazel package")
+        if rctx.attr.package_json:
+            fail("The package_json attribute should not be used with pnpm_lock.")
+    if count != 1:
```
gregmagolan, line 538: clever assertions with count

gregmagolan (approved):
Wow. That turned out nice and clean. 🧙

PR #277, opened 2022-07-07
test: add e2e test containing transitive npm_package deps

jbedard:
This adds an extra indirect dependency to the `npm_link_package` test to ensure indirect deps are passed along to runtime.

The second commit adds an esm version of that test. Basically copy and pasted and switched to esm which requires some extra node flags and minor other differences. Currently the file structure and files are very similar so a visual diff between the two nicely shows the differences.


alexeagle:
Hey do you still want to finish this one?

jbedard:
Now that we have Greg's changes I'll give it a shot 👍

jbedard:
@gregmagolan this is passing now and ready

gregmagolan (approved):
🌮

PR #281, opened 2022-07-12
feat: add patch_node_fs attribute to js_binary and js_run_binary for easy opt-out

gregmagolan:


js/private/js_run_binary.bzl
```diff
@@ -127,6 +128,17 @@ def js_run_binary(
 
             See https://docs.bazel.build/versions/main/be/common-definitions.html#common.tags for useful keys.
 
+        patch_node_fs_api: Patch the to Node.js `fs` API (https://nodejs.org/api/fs.html) for this node program
```
alexeagle, line 131: naming nit: "api" doesn't seem necessary here, the user isn't coding against it. WDYT about just `patch_node_fs`?

js/private/test/node-patches/BUILD.bazel
```diff
@@ -17,9 +17,7 @@ TESTS = [
             "//js/private/node-patches/src:compile",
         ],
         entry_point = t,
-        env = {
-            "JS_BINARY__DISABLE_NODE_PATCHES": "1",
```
alexeagle, line 21: breaking I guess, I think I used this somewhere and will find out :)

alexeagle (approved):

js/private/test/node-patches/BUILD.bazel
gregmagolan: it wasn't advertised anywhere and I didn't use it anywhere else so should be fine

js/private/js_run_binary.bzl
gregmagolan: sgtm

js/private/js_binary.bzl
```diff
@@ -168,6 +168,19 @@ _ATTRS = {
         values = _LOG_LEVELS.keys(),
         default = "error",
     ),
+    "patch_node_fs": attr.bool(
```
jbedard, line 171: Should we try to put the word "sandbox" in the name since "patch" normally implies a bug-fix?

jbedard, line 171: The child process patch will also have a flag? `patch_node_child_process`?

gregmagolan: sandbox wouldn't be accurate either since the guards prevent leaving execroot, runfiles and the sandbox. I create a pretty verbose docstring on this one. most users will never need to touch it but it does leave an easy opt-out if needed.

gregmagolan: I think the child process fs patches should just be implied if this is true since they are the same fs patches applied to spawn processes; just more a more robust implementation that covers the child spawn case when we get to it for 1.1

gregmagolan: `node_fs_guards` perhaps but I like "patch" as it is technically what we are doing by changing the behavior of the node fs code

jbedard: Ah yeah, you're correct... it's the same patches but just properly applied to sub-processes as well

jbedard: I guess I'd worry that `fs` will become overloaded and will mean patching a lot more then the node "fs" library like the name implies 🤷

gregmagolan: Hopefully we won't be patching much more in node so we won't to worry about it :)

alexeagle: 🐒 -patch

PR #283, opened 2022-07-13
feat: use declare_symlink in npm_link_package when --experimental_allow_unresolved_symlinks is set

gregmagolan:
Once Bazel fully supports the unresolved symlink feature (https://github.com/bazelbuild/bazel/issues/10298) which is currently being worked on by @fmeum split across multiple PRs to bazel such as https://github.com/bazelbuild/bazel/pull/15963 then setting `--experimental_allow_unresolved_symlinks` should be sufficient to allow for remote-execution to work with rules_js.


npm/private/npm_link_package.bzl
```diff
@@ -208,18 +229,25 @@ def _impl_store(ctx):
...
             if dep_virtual_store_directory:
                 for dep_alias in dep_aliases:
                     # "node_modules/{virtual_store_root}/{virtual_store_name}/node_modules/{package}"
                     dep_symlink_path = paths.join("node_modules", utils.virtual_store_root, virtual_store_name, "node_modules", dep_alias)
-                    dep_symlink = ctx.actions.declare_file(dep_symlink_path)
-                    ctx.actions.symlink(
-                        output = dep_symlink,
-                        target_file = dep_virtual_store_directory,
-                    )
+                    if ctx.attr.allow_unresolved_symlinks:
```
jbedard, line 238: Is this block worth putting in a util method?

.bazelrc.common
```diff
@@ -42,3 +42,6 @@ build:openbsd --nolegacy_external_runfiles
 # on the host.
 # See https://github.com/angular/angular/issues/27514.
 build --incompatible_strict_action_env
+
+# TODO: enable once this is supported in Bazel
+# build --experimental_allow_unresolved_symlinks
```
jbedard, line 47: I guess we need a separate CI job to test this?

npm/private/npm_link_package.bzl
```diff
@@ -208,18 +229,25 @@ def _impl_store(ctx):
...
+                        dep_symlink = ctx.actions.declare_symlink(dep_symlink_path)
+                        ctx.actions.symlink(
+                            output = dep_symlink,
+                            target_path = relative_file(dep_virtual_store_directory.path, dep_symlink.path),
```
jbedard, line 242: Why does it have to be relative in this case? Can it not also be relative in the other?

.bazelrc.common
gregmagolan: Yea, and it would anyway right now since it is not yet supported by Bazel in a way that works with rules_js even without RBE in the 6.0.0 pre-release

npm/private/npm_link_package.bzl
gregmagolan: `ctx.actions.symlink` takes a File and not a path and it decides on the symlink path itself. Bazel lays out symlinks with absolute paths with that style.

.bazelrc.common
jbedard: So the change to support it hasn't even been published in a pre-release yet?

npm/private/npm_link_package.bzl
gregmagolan: good call. refactored into util method

npm/private/npm_link_package.bzl
```diff
@@ -163,8 +174,36 @@ If unset, the package name of the src npm_link_package_store is used.
...
 
+def _make_symlink(ctx, symlink_path, target_file):
+    files = []
+    if ctx.attr.allow_unresolved_symlinks:
+        symlink = ctx.actions.declare_symlink(symlink_path)
+        ctx.actions.symlink(
+            output = symlink,
+            target_path = relative_file(target_file, symlink.path),
+        )
+        files.append(target_file)
```
jbedard, line 197: Interesting, I didn't notice previously (before the helper func) that the new `declare_symlink` also requires the target file. That makes sense, but why did the old one not require it? Because it was absolute instead of relative 🤔

jbedard (approved):

npm/private/npm_link_package.bzl
gregmagolan: The new `declare_symlink` takes a target path so the file itself has to be explicitly added as an input. That why they call the feature unresolved/dangling symlinks since you can symlink to something that is not an input.

The old method takes a `target_file` which is automatically added to the inputs transitively via the symlink file

.bazelrc.common
gregmagolan: Not yet. There are a number of fixes needed. One landed and the rest are in PRs.

fmeum:
@gregmagolan Note that `--experimental_allow_unresolved_symlinks` will probably be renamed when stabilized. I also think that Starlark API changes are necessary for it to work as desired, so the graceful migration implemented here may not work.

Issue #282, opened 2022-07-12
Failed to parse valid yaml in npm_translate_lock.bzl


arobinson:
I’m having an issue with moving from my POC application to our actual application. I’m getting this error while parsing the pnpm lock file:
```
(15:38:58) ERROR: /Users/andrew/development/work/client/.../WORKSPACE.bazel:32:19: fetching npm_translate_lock rule //external:npm: Traceback (most recent call last):
        File "/Users/andrew/.cache/bazel/f26303f99f5aa5dc2ecc7fa834780eb5/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 509, column 33, in _impl
                lockfile = _process_lockfile(rctx)
        File "/Users/andrew/.cache/bazel/f26303f99f5aa5dc2ecc7fa834780eb5/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 164, column 37, in _process_lockfile
                lockfile = utils.parse_pnpm_lock(rctx.read(rctx.path(rctx.attr.pnpm_lock)))
        File "/Users/andrew/.cache/bazel/f26303f99f5aa5dc2ecc7fa834780eb5/external/aspect_rules_js/npm/private/utils.bzl", line 49, column 23, in _parse_pnpm_lock
                return _parse_yaml(lockfile_content)
        File "/Users/andrew/.cache/bazel/f26303f99f5aa5dc2ecc7fa834780eb5/external/aspect_rules_js/npm/private/yaml.bzl", line 42, column 31, in parse
                _handle_PARSE_NEXT(state, input, stack, starlark)
        File "/Users/andrew/.cache/bazel/f26303f99f5aa5dc2ecc7fa834780eb5/external/aspect_rules_js/npm/private/yaml.bzl", line 141, column 30, in _handle_PARSE_NEXT
                _set_result_value(starlark, stack, _parse_scalar(state["buffer"]))
        File "/Users/andrew/.cache/bazel/f26303f99f5aa5dc2ecc7fa834780eb5/external/aspect_rules_js/npm/private/yaml.bzl", line 346, column 24, in _set_result_value
                curr_result.append(value)
Error: 'string' value has no field or method 'append'
(15:38:58) ERROR: no such package '@npm//': 'string' value has no field or method 'append'
```
To debug this, I added a print statement where it was dying:
```
        if type(curr_result) == "dict":
            curr_result[kns_states[-1]["key"]] = value
        else:
            print("curr_result", curr_result, "value", value)
[...]

PR #284, opened 2022-07-13
feat: parse literal mulitline yaml strings

kormide:
fyi: @arobinson

Addresses https://github.com/aspect-build/rules_js/issues/282.

### DIFF
--- a/npm/private/test/yaml_tests.bzl
+++ b/npm/private/test/yaml_tests.bzl
@@ -126,6 +126,81 @@ def _parse_basic_test_impl(ctx):
 }
 """))
 
+    # Literal multiline strings (|), strip (-) and keep (+)
+    asserts.equals(env, {"foo":"bar\n" }, parse("""\
+foo: |
+    bar
+"""))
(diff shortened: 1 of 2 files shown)


npm/private/test/yaml_tests.bzl
```diff
@@ -126,6 +126,81 @@ def _parse_basic_test_impl(ctx):
 }
 """))
 
+    # Literal multiline strings (|), strip (-) and keep (+)
+    asserts.equals(env, {"foo":"bar\n" }, parse("""\
+foo: |
+    bar
```
gregmagolan, line 132: would a trailing spaces after `bar` be kept?

kormide: Yes, they should be. I'll add a test case for that.

kormide: Added some trailing spaces to the third test case after `baz`. You can't see those spaces in the github diff, but you can see it in the expected value.

gregmagolan: 🌮

Issue #285, opened 2022-07-13
launching js_binary .sh fails when launched from sh_binary


jbedard:
It seems invoking the `sh_binary` does not generate the `.runfiles` of the `js_library`. Invoking or building the `js_library` first creates the `.runfiles`, then the `sh_binary` works going forward.

Note `sh_test` does correctly generate the `.runfiles`.

Repro: https://github.com/jbedard/rules_js/commit/264a0596e62f3e3012dd79347288750577cfa1b4
```
bazel run //examples/js_binary:test10
FATAL: aspect_rules_js[js_binary]: RUNFILES environment variable is not set

bazel run //examples/js_binary:bin10-js_binary
bazel run //examples/js_binary:test10
```
(the second `test10` works after running the `js_binary` correctly)

Another example: https://github.com/pjjw/rules_nodejs_1451/tree/7a816e786de504d7687d3a0dd0ae97729749710c

gregmagolan:
From this thread on Bazel slack: https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1657694352729829

PR #286, opened 2022-07-14
fix: determine the RUNFILES directory when running a js_binary in an sh_binary

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/285

Without the fix commit,

```
$ bazel run //examples/js_binary:bin10
INFO: Analyzed target //examples/js_binary:bin10 (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //examples/js_binary:bin10 up-to-date:
  bazel-bin/examples/js_binary/launch_case10.sh
  bazel-bin/examples/js_binary/bin10
INFO: Elapsed time: 0.139s, Critical Path: 0.00s
INFO: 2 processes: 2 internal.
INFO: Build completed successfully, 2 total actions
INFO: Build completed successfully, 2 total actions
FATAL: aspect_rules_js[js_binary]: RUNFILES environment variable is not set
```

with the commit

```
$ bazel run //examples/js_binary:bin10
INFO: Analyzed target //examples/js_binary:bin10 (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //examples/js_binary:bin10 up-to-date:
  bazel-bin/examples/js_binary/launch_case10.sh
  bazel-bin/examples/js_binary/bin10
INFO: Elapsed time: 0.191s, Critical Path: 0.00s
INFO: 2 processes: 2 internal.
INFO: Build completed successfully, 2 total actions
INFO: Build completed successfully, 2 total actions

```

### DIFF
--- a/examples/js_binary/BUILD.bazel
+++ b/examples/js_binary/BUILD.bazel
@@ -327,3 +327,52 @@ diff_test(
 aspect_test_a_bin.bin_a_test(
     name = "aspect_bin_a_test",
 )
+
+####################################################
+# Use case 10
+# Show launching js_binary() indirectly from a sh_binary()
+
+write_file(
+    name = "write10_js",
+    out = "case10.js",
+    content = ["""\
+        console.log("js_binary: ", __filename);
+        process.exit(0);
+"""],
+)
+js_binary(
+    name = "bin10-js_binary",
+    entry_point = "case10.js",
+)
+
+write_file(
+    name = "write10_launch_sh",
+    out = "launch_case10.sh",
+    content = ["""\
+        echo "sh launcher: $0"
+
+        # --- begin runfiles.bash initialization v2 ---
+        # Copy-pasted from the Bazel Bash runfiles library v2.
+        set -uo pipefail; f=bazel_tools/tools/bash/runfiles/runfiles.bash
+        source "${RUNFILES_DIR:-/dev/null}/$f" 2>/dev/null || \
+        source "$(grep -sm1 "^$f " "${RUNFILES_MANIFEST_FILE:-/dev/null}" | cut -f2- -d' ')" 2>/dev/null || \
+        source "$0.runfiles/$f" 2>/dev/null || \
+        source "$(grep -sm1 "^$f " "$0.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+        source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+        { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
+        # --- end runfiles.bash initialization v2 ---
+
+        $(rlocation aspect_rules_js/examples/js_binary/bin10-js_binary.sh) -h
+"""],
+)
+
+sh_binary(
+    name = "test10",
+    srcs = [":launch_case10.sh"],
+    data = [
+        ":bin10-js_binary",
+    ],
+    deps = [
+        "@bazel_tools//tools/bash/runfiles",
+    ],
+)
--- a/examples/js_binary/BUILD.bazel
+++ b/examples/js_binary/BUILD.bazel
@@ -269,7 +269,7 @@ js_test(
     log_level = "debug",
 )
 
-# bazel run //examples:test10_binary
+# bazel run //examples:case7_binary
 js_binary(
     name = "case7_binary",
     args = ["dummy"],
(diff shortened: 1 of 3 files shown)


examples/js_binary/BUILD.bazel
```diff
@@ -327,3 +327,52 @@ diff_test(
...
+        source "$(grep -sm1 "^$f " "$0.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+        source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+        { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
+        # --- end runfiles.bash initialization v2 ---
+
+        $(rlocation aspect_rules_js/examples/js_binary/bin10-js_binary.sh) -h
+"""],
+)
+
+sh_binary(
```
jbedard, line 369: Note that this is a `sh_binary` and not a `sh_test` like we probably want. But the bug doesn't occur with `sh_test`. Any idea why the bug wouldn't occur when testing?

I think if we leave it as `sh_binary` then we need to add another rule (which I forget) to run it as a test?

js/private/js_binary.sh.tpl
```diff
@@ -138,13 +138,12 @@ function is_windows {
...
 # Example:
-# C:/Users/XUser/_bazel_XUser/7q7kkv32/execroot/A/b/C -> /c/users/xuser/_bazel_xuser/7q7kkv32/execroot/a/b/c
+# C:/Users/XUser/_bazel_XUser/7q7kkv32/execroot/A/b/C -> /c/Users/XUser/_bazel_XUser/7q7kkv32/execroot/A/b/C
 function normalize_windows_path {
     # Apply the followings paths transformations to normalize paths on Windows
     # -process driver letter
     # -convert path separator
-    # -lowercase everything
-    sed -e 's#^\(.\):#/\L\1#' -e 's#\\#/#g' -e 's/[A-Z]/\L&/g' <<< "$1"
+    sed -e 's#^\(.\):#/\L\1#' -e 's#\\#/#g' <<< "$1"
```
jbedard, line 146: How did windows capitalization come into play with this issue?

js/private/js_binary.sh.tpl
```diff
@@ -173,15 +172,24 @@ function normalize_windows_path {
...
-        RUNFILES=${RUNFILES_MANIFEST_FILE%/MANIFEST}
+        NORMALIZED_RUNFILES_MANIFEST_FILE="$RUNFILES_MANIFEST_FILE"
+    fi
+    if [[ "${NORMALIZED_RUNFILES_MANIFEST_FILE}" == *.runfiles_manifest ]]; then
+        # Newer versions of Bazel put the manifest besides the runfiles with the suffix .runfiles_manifest.
+        # For example, the runfiles directory is named my_binary.runfiles then the manifest is beside the
+        # runfiles directory and named my_binary.runfiles_manifest
+        RUNFILES=${NORMALIZED_RUNFILES_MANIFEST_FILE%_manifest}
+    elif [[ "${NORMALIZED_RUNFILES_MANIFEST_FILE}" == */MANIFEST ]]; then
+        # Older versions of Bazel put the manifest file named MANIFEST in the runfiles directory
```
jbedard, line 188: Do you know what version switched it? Can we put a comment so we can potentially remove the `elif` in the future?

js/private/js_binary.sh.tpl
gregmagolan: The replacement of `/MANIFEST` is simpler if we don't lowercase it. Not sure why it was there in the first place; that code came from rules_nodejs

examples/js_binary/BUILD.bazel
gregmagolan: yea, it is sh_binary specific since sh_test sets the TEST_SRCDIR env var so RUNFILES is determined that way

gregmagolan: if you run the sh_binary through a genrule it is not reproducible either as the environment is setup differently in that case. only `bazel run` directly on the `sh_binary` will reproduce

js/private/js_binary.sh.tpl
gregmagolan: Not sure and doesn't seem worthwhile to do a binary search for this one

examples/js_binary/BUILD.bazel
jbedard: > yea, it is sh_binary specific since sh_test sets the TEST_SRCDIR env var so RUNFILES is determined that way

Not really related to this PR... but can RUNFILES be determined the same everywhere? Doing it different for tests seems odd 🤔

jbedard (approved):

examples/js_binary/BUILD.bazel
gregmagolan: Yeah, that code could get simplified in the future. Tricky to do right now as we don't have great coverage for different platforms and all of the corner cases.

PR #288, opened 2022-07-15
fix: handle transitive file: dependencies

### LINKED ISSUE #268: Bug: `file:` resolutions aren't supported
Hey there,

If you have `resolutions` in your `package.json` like so:

```json
{
  "resolutions": {
    "ts-log": "file:local/ts-log"
  }
}
```

And then you have a package that uses – say – `ts-log` (as per the above), like `@graphql-codegen/cli`, a build will fail because of the `:` in the path.

```ERROR: app/BUILD.bazel:15:22: //:.aspect_rules_js/node_modules/@graphql-codegen/cli/2.6.4_@types+node@18.0.0/pkg: invalid label ':.aspect_rules_js/node_modules/ts-log/file:local/ts-log/ref' in dict key element: invalid target name '.aspect_rules_js/node_modules/ts-log/file:local/ts-log/ref': target names may not contain ':'```

I'm aware #167 was recently fixed, but I'm seeing this in `rules_js-1.0.0-rc.1` which AFAICT was released afterwards.

Thanks!

### PR DESCRIPTION
**Pull request by gregmagolan**
Fixes #268

I added support for direct "file:" links but I did not cover the case of transitive "file:" links via resolutions. This PR addresses that corner case.

Multiple commits intentionally; do not squash on land.

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -13,5 +13,10 @@
                 }
             }
         }
+    },
+    "// resolutions.semver-max&is-odd": "// these are for the npm/private/test/package.json but pnpm has a bug where resolutions need to be specified in the root package.json but the paths are still relative to the package.json where the resolution applies",
```
jbedard, line 17: 🤔 

So what if `semver-max` was also used in another location where this relative path doesn't work?

gregmagolan: Yep. Its an annoying pnpm bug

jbedard: Is there a link to an issue?

gregmagolan: I didn't look to see if it had been reported yet

gregmagolan: They have a few of these "it has to be done in the root package.json" issues for other fields besides resolutions

jbedard:
I don't really know this code enough to confidently review this. Overall LGTM from what I do understand, but I think @alexeagle should also look before merging...

npm/private/npm_import.bzl
```diff
@@ -472,17 +472,25 @@ def _impl_links(rctx):
...
-            version = store_version,
-            virtual_store_root = utils.virtual_store_root,
-        )
-        ref_deps[dep_store_target_ref] = ref_deps[dep_store_target_ref] + [dep_name] if dep_store_target_ref in ref_deps else [dep_name]
+            if dep_version.startswith("/"):
+                store_package, store_version = utils.parse_pnpm_name(dep_version[1:])
+            else:
+                store_package = dep_name
+                store_version = dep_version
+            dep_store_target = """":{virtual_store_root}/{{}}/{package}/{version}/ref".format(link_root_name)""".format(
```
alexeagle, line 488: oh god the format() calls are nested... 🩸 👁️ 
and it starts with four double quotes

maybe using `":{virtual_store_root}/%s/{package}/{version}/ref" % link_root_name` syntax for the inner interpolation would help readability?

package.json
```diff
@@ -12,6 +12,11 @@
                     "mocha-multi-reporters": "*"
                 }
             }
+        },
+        "// overrides.semver-max&is-odd": "// these are for the npm/private/test/package.json but pnpm has a bug where resolutions need to be specified in the root package.json but the paths are still relative to the package.json where the resolution applies",
```
alexeagle, line 16: we should file

alexeagle (approved):

jbedard (approved):

npm/private/npm_import.bzl
gregmagolan: not a bad idea, will save that for refactoring :)

package.json
gregmagolan: @jbedard Feel like doing the honors?

PR #299, opened 2022-07-19
test: add node process spawning tests

### LINKED ISSUE #236: Verify that a js_binary that spawns a node child process gets the toolchain-managed node interpreter
This may have been carried over from rules_nodejs but we aren't sure there's a test to verify it.

jbedard:
Closes #236

Currently all pass. Let me know if you can think of more cases or think any of these are unnecessary/overkill.


gregmagolan (approved):
🌮 nice

PR #300, opened 2022-07-19
fix: only include direct DeclarationInfo dependencies in js_library DeclarationInfo

jbedard:


thesayyn:
According to https://github.com/bazelbuild/rules_nodejs/blob/c962f12e06d99230f17d82a732958f3c00c9438e/nodejs/private/providers/declaration_info.bzl#L49-L51  and https://github.com/bazelbuild/rules_nodejs/blob/c962f12e06d99230f17d82a732958f3c00c9438e/nodejs/private/providers/declaration_info.bzl#L30 this change seems to be right.

thesayyn (approved):

gregmagolan (approved):
🌮

PR #310, opened 2022-07-26
fix: apply node fs patched symlink guards in spawned child node processes

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/311

### DIFF
--- a/js/private/test/node-patches/spawn.js
+++ b/js/private/test/node-patches/spawn.js
@@ -2,6 +2,8 @@ const child_process = require('child_process')
 
 const printbinjs = process.argv[2]
 const printbinsh = process.argv[3]
+const printdepthjs = process.argv[4]
+const printdepthsh = process.argv[5]
 
 // We don't want to bring jest into this repo so we just fake the describe and it functions here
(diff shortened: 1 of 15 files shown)


js/private/test/node-patches/spawn.js
```diff
@@ -97,34 +97,25 @@ describe('child_process', async () => {
...
-    if (process.platform == 'linux') {
-        // TODO: enable this case on macos; it doesn't have realpath installed
-        await it('should launch patched node via child_process.execFileSync()', () => {
-            testSync({ stdout: child_process.execFileSync(printbinsh) })
-        })
-    }
+    await it('should launch patched node via child_process.execFileSync()', () => {
+        testSync({ stdout: child_process.execFileSync(printbinsh) })
+    })
 })
```
jbedard, line 121: Is there an easy way to also assert that the process is patched?

gregmagolan: came up with a way; added assertions

jbedard (approved):

PR #320, opened 2022-07-28
fix: generate package_json.bzl only when pkg has bins

thesayyn:
this change allows us to do `bazel fetch @npm//:all`


gregmagolan (approved):
🌮

alexeagle:
Ha, glad I left the code behind in that comment

PR #322, opened 2022-07-28
chore: update to aspect_bazel_lib 1.8.0

gregmagolan:


jbedard (approved):

jbedard:
What's the reason for the docs diff?

gregmagolan:
> What's the reason for the docs diff?

`npm_package` uses `copy_to_directory` attributes via composition and those changed with the upgrade

PR #330, opened 2022-08-01
refactor: pre-factors for introducing JsInfo and related fixes and features

gregmagolan:


npm/private/npm_link_package_store.bzl
```diff
@@ -0,0 +1,93 @@
...
+load("@bazel_skylib//lib:paths.bzl", "paths")
+load("@rules_nodejs//nodejs:providers.bzl", "DeclarationInfo")
+load(":utils.bzl", "utils")
+load(":npm_linked_package_info.bzl", "NpmLinkedPackageInfo")
+load(":npm_package_store_info.bzl", "NpmPackageStoreInfo")
+
+_DOC = """Links an npm package that is backed by an npm_package_store into a node_modules tree as a direct dependency.
+
+This is used in conjunction with the npm_package_store rule that outputs an npm package into the
+node_modules/.apsect_rules_js virtual store in a pnpm style symlinked node_modules structure.
```
alexeagle, line 12: nit: apsect

npm/private/npm_link_package_store.bzl
```diff
@@ -0,0 +1,93 @@
...
+    transitive_files_depset = depset(files, transitive = [ctx.attr.src[NpmPackageStoreInfo].transitive_files])
+
+    providers = [
+        DefaultInfo(
+            files = transitive_files_depset,
+        ),
+        NpmLinkedPackageInfo(
+            label = ctx.label,
+            link_package = ctx.label.package,
+            package = ctx.attr.src[NpmPackageStoreInfo].package,
```
alexeagle, line 72: nit: maybe pull store_info out as a variable and point out it's guaranteed to exist by the constraint on the src attr

npm/private/npm_link_package_store.bzl
```diff
@@ -0,0 +1,93 @@
...
+        ),
+        DeclarationInfo(
+            declarations = depset(files, transitive = [ctx.attr.src[DeclarationInfo].declarations]),
+            transitive_declarations = depset(files, transitive = [ctx.attr.src[DeclarationInfo].transitive_declarations]),
+        ),
+    ]
+    if OutputGroupInfo in ctx.attr.src:
+        providers.append(ctx.attr.src[OutputGroupInfo])
+
+    return providers
```
alexeagle, line 86: we could have some analysis unit tests for the providers being right-shaped, seem useful?

npm/private/utils.bzl
```diff
@@ -83,6 +84,24 @@ def _virtual_store_name(name, version):
     escaped_version = version.replace("/", "+")
     return "%s@%s" % (escaped_name, escaped_version)
 
+def _make_symlink(ctx, symlink_path, target_file):
+    files = []
+    if ctx.attr.allow_unresolved_symlinks:
```
alexeagle, line 89: I'm confused about how the attr here is about unresolved symlinks, but the logic applies to resolved symlinks as well. Should something be renamed?

npm/private/npm_package_store_info.bzl
```diff
@@ -0,0 +1,15 @@
+"NpmPackageStoreInfo provider"
+
+NpmPackageStoreInfo = provider(
+    doc = "Provides an npm package store",
```
alexeagle, line 4: this is just repeating the name, maybe say a bit more?

npm/private/npm_linked_package_info.bzl
```diff
@@ -0,0 +1,14 @@
+"NpmLinkedPackageInfo provider"
+
+NpmLinkedPackageInfo = provider(
```
alexeagle, line 3: nit: did you consider putting more than one provider in the same file?

alexeagle (approved):

npm/private/utils.bzl
gregmagolan: It comes from `experimental_allow_unresolved_symlinks`. I think the name comes from the fact that you can have a symlink to something that is not an input. The thing you're symlinking to is not carried transitively into the sandbox or it might just not exist in or out of the sandbox at all if you pass in a path to a non-existant file.

npm/private/npm_linked_package_info.bzl
gregmagolan: My preference is to have one file per provider. Makes the bar for adding a new provider a bit higher.

npm/private/npm_package_store_info.bzl
gregmagolan: ✅ 
```
    doc = """Provides an npm package store
    
    This is a npm package within the virtual store of a pnpm-style symlinked node_modules tree.
    See https://pnpm.io/symlinked-node-modules-structure for more information about
    symlinked node_modules trees.""",
```

npm/private/npm_link_package_store.bzl
gregmagolan: ✅

npm/private/npm_link_package_store.bzl
gregmagolan: ✅

npm/private/npm_link_package_store.bzl
gregmagolan: yup; those are useful. lets do as a follow-up

PR #331, opened 2022-08-01
feat: add lifecycle_hooks_execution_requirements attribute

### LINKED ISSUE #270: Feature: allow for a way to add tags for the lifecycle hook actions
Some `npm` packages want to download prebuild binaries during their lifecycle actions. See for example https://sharp.pixelplumbing.com/install#prebuilt-binaries.

When running `bazel` with `--sandbox_default_allow_network=false` these downloads will fail. It would be great to be able to add `tags` to the generated `npm_import()` calls.

JiaLiPassion:
Close #270

add `lifecycle_hooks_execution_requirements` to `npm_translate_lock` and
`npm_import` rule.

 execution requirements applied to the preinstall, install and postinstall lifecycle hooks on npm packages.
 The execution requirements can be defined per package by package name or globally using "*".
 For example:

```
 lifecycle_hooks_execution_requirements: {
     "*": ["requires_network"],
     "@foo/bar": ["no_sandbox"],
 }
```


npm/private/npm_import.bzl
```diff
@@ -615,6 +617,11 @@ def _impl_links(rctx):
             msg = "lifecycle_hooks_env contains invalid key value pair '%s', required '=' separator not found" % env
             fail(msg)
 
+    well_known_tags = propagate_well_known_tags(rctx.attr.lifecycle_hooks_execution_requirements)
```
gregmagolan, line 620: don't think we need to go through propagate_well_known_tags here since the property is execution_requirements and not tags the user is setting

npm/private/npm_translate_lock.bzl
```diff
@@ -146,6 +146,16 @@ _ATTRS = {
         }
         """,
     ),
+    "lifecycle_hooks_execution_requirements": attr.string_list_dict(
+        doc = """Execution requirements applied to the preinstall, install and postinstall lifecycle hooks on npm packages.
+        The execution requirements can be defined per package by package name or globally using "*".
+        For example:
+        lifecycle_hooks_execution_requirements: {
+            "*": ["requires_network"],
```
gregmagolan, line 154: `requires-network` and `no-sandbox` are the values bazel expects right?

gregmagolan:
Neat. Thanks for the PR @JiaLiPassion

npm/private/npm_import.bzl
JiaLiPassion: Thank you for the review, so I just read the doc here https://github.com/aspect-build/bazel-lib/blob/main/docs/run_binary.md, it seems the `execution_requirements` accept the `common args` defined here https://docs.bazel.build/versions/main/be/common-definitions.html#common.tags, so you mean we don't need to filter out the values are not belong to the `common args`?

npm/private/npm_translate_lock.bzl
JiaLiPassion: In my understanding, the use case is here https://github.com/aspect-build/rules_js/issues/270, so the user run 
`bazel --nosandbox_default_allow_network` but still want `npm_import` use network, so by setting the `execution_requirements`, the user can run `bazel --nosandbox_default_allow_network`  and the `npm_import` can still access network for the specified packages.

npm/private/npm_import.bzl
gregmagolan: Correct. Since we're not pulling these from `tags` where there might be values that are not meant for execution requirements then we don't need to filter anything. The user knows that execution requirements values are expected.

npm/private/npm_translate_lock.bzl
gregmagolan: Oh. I was referring to the `_` vs the `_`. I believe the values should have `-` to be correct.

npm/private/npm_import.bzl
JiaLiPassion: Got it!

npm/private/npm_translate_lock.bzl
JiaLiPassion: Oh, yeah, you are right, thanks.

gregmagolan (approved):
🌮

PR #334, opened 2022-08-01
fix: js_binary doesn't need to copy files to bin for runfiles inputs

gregmagolan:
We technically don't need to copy anything to bin that is passed to runfiles since bazel flattens that tree. Not copying things to bin in js_binary has some interesting side-effects that are seen. Copying was hiding some bugs/bad behavior.

### DIFF
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -17,7 +17,6 @@ js_binary(
 
 load("@aspect_bazel_lib//lib:paths.bzl", "BASH_RLOCATION_FUNCTION")
 load("@aspect_bazel_lib//lib:windows_utils.bzl", "create_windows_native_launcher_script")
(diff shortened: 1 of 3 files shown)


js/private/js_binary.bzl
```diff
@@ -327,20 +326,18 @@ def _create_launcher(ctx, log_prefix_rule_set, log_prefix_rule, fixed_args = [])
 
         # Copy entry and data files that are not already in the output tree to the output tree.
```
alexeagle, line 327: Update comment?

alexeagle (approved):

js/private/js_binary.bzl
gregmagolan: ✅

gregmagolan: Nice catch

PR #343, opened 2022-08-02
fix: add missing file needed for gh actions bcr workflow

kormide:


.github/workflows/create-bcr-entry.mjs
```diff
@@ -0,0 +1,224 @@
...
+
+  const [projectPath, bcrPath, ownerSlashRepo, tag] = argv;
+  const version = getVersionFromTag(tag);
+
+  const moduleName = getModuleName(resolve(projectPath, "MODULE.bazel"));
+  const bcrTemplatesPath = resolve(projectPath, ".github/workflows/bcr");
+  const bcrEntryPath = resolve(bcrPath, "modules", moduleName);
+  const bcrVersionEntryPath = resolve(bcrEntryPath, version);
+
+  if (!existsSync(bcrEntryPath)) {
```
jbedard, line 43: If the method is async any reason not to use the `fs/promises` API?

jbedard:
Is there anything rules_js specific in this script or is it copied from somewhere?

kormide:
> Is there anything rules_js specific in this script or is it copied from somewhere?

This is just copied from `bazel-lib`. This is a temporary solution until the bcr app is up and running.

.github/workflows/create-bcr-entry.mjs
kormide: I'd rather not have copies of this file diverge between repos, and it's temporary, so I'm not too concerned about minor improvements like this.

jbedard: Yeah, if this is just copied then don't change anything. I didn't realize it was copied until looking through a bit more.

jbedard (approved):

Issue #341, opened 2022-08-02
node patches are not applied when the entry point is a mjs file


jbedard:
When the entry point is a mjs file the node patches are not applied correctly. It seems `--preserve-symlinks-main` fixes it which should potentially be enabled by default.

~~The js entry_point file requires preserving the symlink just like any other symlinked file in the sandbox. `--preserve-symlinks-main` should be enabled for all launched node processes just like `--preserve-symlinks`~~.

https://nodejs.org/api/cli.html#--preserve-symlinks-main (added in node v10.2.0)

gregmagolan:
don't want `--preserve-symlinks` on as it breaks node_modules resolution in rules_js; the symlink guards keep processes in the sandbox and their runfiles. but the `--preserve-symlinks-main` option on by default might be correct since `.mjs` entry points skirt around the symlink guards and leave the sandbox currently

PR #352, opened 2022-08-03
feat: set --preserve-symlinks-main in js_binary by default and add preserve_symlinks_main attribute for opt-out

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/341, the issue of .mjs entry files escaping the runfiles and sandbox

### DIFF
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -155,6 +155,19 @@ _ATTRS = {
         """,
         default = True,
     ),
+    "preserve_symlinks_main": attr.bool(
+        doc = """When True, the --preserve-symlinks-main flag is passed to node.
+
+        This prevents node from following entry script out of runfiles and the sandbox which can happen for some entry
+        files such as `.mjs` where the fs node patches, which guard the runfiles and sandbox, are not used.
+
+        This flag was added in Node.js v10.2.0 (released 2018-05-23). If your node toolchain is configured to use a
+        Node.js version older than this you'll need to set this attribute to False.
+
+        See https://nodejs.org/api/cli.html#--preserve-symlinks-main for more information.
+        """,
+        default = True,
+    ),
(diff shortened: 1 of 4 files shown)


js/private/test/shellcheck_launcher.sh
```diff
@@ -338,7 +338,7 @@ fi
 
 # Gather node options
 NODE_OPTIONS=()
-
+NODE_OPTIONS+=("--preserve-symlinks-main")
```
alexeagle, line 341: is there also a cmd.exe change?

alexeagle (approved):

js/private/test/shellcheck_launcher.sh
gregmagolan: These is no cmd.exe. We carry a bash depedency on Windows.

js/private/js_binary.bzl
```diff
@@ -155,6 +155,19 @@ _ATTRS = {
...
         default = True,
     ),
+    "preserve_symlinks_main": attr.bool(
+        doc = """When True, the --preserve-symlinks-main flag is passed to node.
+
+        This prevents node from following entry script out of runfiles and the sandbox which can happen for some entry
+        files such as `.mjs` where the fs node patches, which guard the runfiles and sandbox, are not used.
+
+        This flag was added in Node.js v10.2.0 (released 2018-05-23). If your node toolchain is configured to use a
+        Node.js version older than this you'll need to set this attribute to False.
```
jbedard, line 165: "If your node toolchain is configured to use a Node.js version older than this you'll need to set this attribute to False."

to 

"If your node toolchain is configured to use an older Node.js version this attribute must be set to False."

?

gregmagolan: You version is better but its merged already 🤷

Issue #345, opened 2022-08-02
macro example has a hidden require failure


jfirebaugh:
The output is hidden by default since the tests themselves pass despite this require failure, but it shows up with `--test_output=all`:

```
~/figma/rules_js $ bazel test --test_output=all //examples/macro:test
INFO: Analyzed target //examples/macro:test (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
Target //examples/macro:test up-to-date:
  bazel-bin/examples/macro/test.sh
INFO: Elapsed time: 0.216s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
PASSED: //examples/macro:test (see /private/var/tmp/_bazel_johnfirebaugh/3cc5be9735d30b6128426fd19d172550/execroot/aspect_rules_js/bazel-out/darwin-fastbuild/testlogs/examples/macro/test/test.log)
INFO: From Testing //examples/macro:test
==================== Test output for //examples/macro:test:
Failed to load reporter: "mocha-junit-reporter" or "/private/var/tmp/_bazel_johnfirebaugh/3cc5be9735d30b6128426fd19d172550/sandbox/darwin-sandbox/6/execroot/aspect_rules_js/bazel-out/darwin-fastbuild/bin/examples/macro/test.sh.runfiles/aspect_rules_js/mocha-junit-reporter": Error: Cannot find module 'mocha-junit-reporter'
Require stack:
- /private/var/tmp/_bazel_johnfirebaugh/3cc5be9735d30b6128426fd19d172550/sandbox/darwin-sandbox/6/execroot/aspect_rules_js/bazel-out/darwin-fastbuild/bin/examples/macro/test.sh.runfiles/aspect_rules_js/node_modules/.aspect_rules_js/mocha-multi-reporters@1.5.1_mocha@10.0.0/node_modules/mocha-multi-reporters/lib/MultiReporters.js
- /private/var/tmp/_bazel_johnfirebaugh/3cc5be9735d30b6128426fd19d172550/sandbox/darwin-sandbox/6/execroot/aspect_rules_js/bazel-out/darwin-fastbuild/bin/examples/macro/test.sh.runfiles/aspect_rules_js/node_modules/.aspect_rules_js/mocha-multi-reporters@1.5.1_mocha@10.0.0/node_modules/mocha-multi-reporters/index.js
[...]

PR #353, opened 2022-08-03
Mocha tests are escaping the sandbox

jfirebaugh:
The attached test fails with the following (snipping the error from #345):

```
~/figma/rules_js $ bazel test --test_output=all //examples/macro:test
INFO: Analyzed target //examples/macro:test (1 packages loaded, 6 targets configured).
INFO: Found 1 test target...
FAIL: //examples/macro:test (see /private/var/tmp/_bazel_johnfirebaugh/3cc5be9735d30b6128426fd19d172550/execroot/aspect_rules_js/bazel-out/darwin-fastbuild/testlogs/examples/macro/test/test.log)
INFO: From Testing //examples/macro:test:
==================== Test output for //examples/macro:test:

  mocha
    ✔ integrates with Bazel
    1) is sandboxed


  1 passing (4ms)
  1 failing

  1) mocha
       is sandboxed:
     AssertionError [ERR_ASSERTION]: The input did not match the regular expression /examples\/macro\/test\.sh\.runfiles/. Input:

'/Users/john/figma/rules_js/examples/macro'

      at Context.<anonymous> (/Users/john/figma/rules_js/examples/macro/test.js:9:16)
      at processImmediate (node:internal/timers:464:21)



================================================================================
```

As seen in the assertion failure, `__dirname` points into the source tree, indicating that this test has escaped the sandbox. In a real repository, this leads to require failures because from the source tree, tests can't resolve relative paths to .js outputs compiled from TypeScript.

### DIFF
--- a/examples/macro/test.js
+++ b/examples/macro/test.js
@@ -4,4 +4,8 @@ describe('mocha', () => {
     it('integrates with Bazel', () => {
         assert(true)
     })
+
+    it('is sandboxed', () => {
+        assert.match(__dirname, /examples\/macro\/test\.sh\.runfiles/)
+    })
 })


gregmagolan:
Indeed they are. Mocha seems to be skirting around the node fs patches somehow.

The `mocha-junit-reports` error at least you can fix with

```
diff --git a/package.json b/package.json
index 9841b67..08414c2 100644
--- a/package.json
+++ b/package.json
@@ -11,6 +11,11 @@
                 "peerDependencies": {
                     "mocha-multi-reporters": "*"
                 }
+            },
+            "mocha-multi-reporters": {
+                "peerDependencies": {
+                    "mocha-junit-reporter": "*"
+                }
             }
         },
```

gregmagolan:
We optimized js_binary to not copy data files to bin, which works if programs say in their runfiles, but it fails if they skirt the fs patches and leave since the symlinks are followed back to the source tree.

gregmagolan:
We'll have to look into what mocha is doing internally. I just hit the same with the Angular architect test rule, which uses mocha under the hood. Probably the same issue.

gregmagolan:
Put this PR up. https://github.com/aspect-build/rules_js/pull/354. It won't prevent mocha from leaving the sandbox (we have to dig further into mocha to figure that one out), but it will keep it from following symlinks all the way to the source tree. Setting that flag will land mocha in the output tree where it can still see all of its inputs.

gregmagolan:
Setting that attribute to True on the test target,

```
mocha_test(
    name = "test",
    srcs = ["test.js"],
    copy_data_to_bin = True,
)
```

keeps that test process in the output-tree,

```
  1) mocha
       is sandboxed:
     AssertionError [ERR_ASSERTION]: The input did not match the regular expression /examples\/macro\/test\.sh\.runfiles/. Input:

'/private/var/tmp/_bazel_greg/7aed427a991d86f99332ab79b9b11780/execroot/aspect_rules_js/bazel-out/darwin-fastbuild/bin/examples/macro'

      at Context.<anonymous> (/private/var/tmp/_bazel_greg/7aed427a991d86f99332ab79b9b11780/execroot/aspect_rules_js/bazel-out/darwin-fastbuild/bin/examples/macro/test.js:9:16)
      at processImmediate (node:internal/timers:464:21)
```

gregmagolan:
This one fixes the reporter resolution failure, https://github.com/aspect-build/rules_js/pull/355

gregmagolan:
I tracked this one down yesterday. It is happening as mocha is using the esm `import()` to load script internally, which it turns out, has a usage of `realPathSync` within the node code path that isn't getting to our fs realPathSync monkey patches.

In mocha, https://github.com/mochajs/mocha/blob/84b2f846148b180d6e1af088f77358a85c81d1ba/lib/nodejs/esm-utils.js#L7

via 

https://github.com/mochajs/mocha/blob/84b2f846148b180d6e1af088f77358a85c81d1ba/lib/cli/run-helpers.js#L94

which ends up in `resolve.js` in node where there is a `realPathSync` call

https://github.com/nodejs/node/blob/4c5b96b376aa778cbe651362c29a26e0d4c32ccd/lib/internal/modules/esm/resolve.js#L310

It is unclear why that code path doesn't get the fs monkey paches and the cjs require loader does.

gregmagolan:
Issued opened for underlying cause https://github.com/aspect-build/rules_js/issues/362

PR #354, opened 2022-08-03
feat: add opt-in copy_files_to_bin attribute to js_binary

gregmagolan:
Work-around for issue observed in Mocha (https://github.com/aspect-build/rules_js/pull/353) where it skirts around the fs patches and leaves its runfiles/sandbox.

Also seen in angular architect test rule, which uses Mocha under the hood: https://github.com/aspect-build/bazel-examples/blob/a76be559cd89a645308ff0f2e6a02a82c780010f/angular/defs.bzl#L133. Could also be mocha in that case.


js/private/js_binary.bzl
```diff
@@ -168,6 +169,19 @@ _ATTRS = {
         """,
         default = True,
     ),
+    "copy_data_to_bin": attr.bool(
+        doc = """When True, data files and the entry_point file are copied to the Bazel output tree before being passed
+        as inputs to runfiles.
+
+        This default to False. It should only be needed if the program manages to skirt the node js patches and leave
```
jbedard, line 176: "node js" => "Node.js"?

I've noticed we have a mix of those and just "node" throughout or docs. We should probably try to standardize those but I don't think "node js" adds any value to "node" (while "Node.js" seems like a more formal way of saying "node"? 🤷).

js/private/js_binary.bzl
```diff
@@ -168,6 +169,19 @@ _ATTRS = {
...
+        its runfiles and/or sandbox. In that case, setting this to True will prevent the program from following symlinks
+        into the source tree and it will end up in the output tree instead where node_modules and other inputs the
+        program may need will be.
+
+        Mocha, for example, has been observed to escape its runfiles & sandbox: https://github.com/aspect-build/rules_js/pull/353.
```
jbedard, line 181: Do we actually understand why? It would be nice to have a full explanation.

I'm also curious if things like mocha are doing something "wrong" and wonder if it would be worth reporting it to them. I think we (or at least I) need to understand what's going on here more before thinking about that.

jbedard (approved):

js/private/js_binary.bzl
gregmagolan: We don't yet. One of us needs to take the time to see what mocha is doing under the hood.

js/private/js_binary.bzl
gregmagolan: Whoops. Was supposed to be "node fs patches"

PR #357, opened 2022-08-03
chore: adjust bcr template files

kormide:


.github/workflows/bcr/presubmit.yml
```diff
@@ -6,7 +6,5 @@ bcr_test_module:
     run_tests:
       name: "Run test module"
       platform: ${{ platform }}
-      test_flags:
-      - "--enable_runfiles"
```
kormide, line 10: I added this yesterday to match what's already in the bcr, but we shouldn't need this as runfiles is already enabled in our `.bazelrc`.

jbedard (approved):

PR #366, opened 2022-08-06
Fix bazel caches in GitHub Actions

tetsuok:
Currently,  caching outputs of `bazel test` seems not working when I looked at "Post Mount bazel caches" step:
``` 
Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
```

When I removed double quotes from cache paths, the warning was gone.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,8 +36,8 @@ jobs:
               uses: actions/cache@v3
               with:
                   path: |
-                      "~/.cache/bazel"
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
Thanks! I bet we copy pasted this mistake in more places...

kormide:
> Thanks! I bet we copy pasted this mistake in more places...

Yeah we do. I was playing with fixing it here too but the tests don't seem to be cached: https://github.com/aspect-build/bazel-lib/pull/209

PR #367, opened 2022-08-07
feat: support pnpm shared-workspace-lockfile=false mode

gregmagolan:
This is nearly supported from existing constructs except that transitive deps of first-party workspace links are not picked up because the information is split across lock files.

Should be relatively easy to create an `npm_translate_locks` repository rule or macro to make this easy for the end user. Just have to figure out how to pass transitive deps for 1st party workspace links around when the data is split across lock files.


gregmagolan:
cc @alexeagle @jondlm @sushain97

jbedard:
@gregmagolan this is old and untouched so I'm closing for now...

PR #377, opened 2022-08-10
feat: always make a @pnpm workspace

alexeagle:
This allows teams to document how users can always have Bazel manage their pnpm version, rather than use something non-hermetic for their local developer workflows.


mattem (approved):
Thanks ❤️ 
Slight nit in that the label is still undocumented imo, ie I have to know the implementation of `npm_translate_lock` (or know to look at that rules docs) in order to discover the `@pnpm` workspace.

alexeagle:
yeah and worse, there's no convenient way to execute pnpm. `npm_import` only gives a `@pnpm//:package` label, after a few minutes of trying things I don't see any way to invoke the bazel-managed pnpm from my terminal.

Maybe @gregmagolan or @jbedard would like to add some commits here... or chat with me about whether we are missing something. I'd imagine `npm_import(extra_build_content="js_binary(...)")`

gregmagolan:
A js_binary that uses the entry point directly from the package sources could work. Without a linking and running out of the virtual store, it would only work for packages with no transitive dependencies.

gregmagolan:
It would work for pnpm since it has no transitives: https://unpkg.com/browse/pnpm@7.9.0/package.json.

I can run it manually with
```
bazel run @nodejs_host//:bin/node $(bazel info output_base)/external/pnpm/package/bin/pnpm.cjs
```
just fine

gregmagolan:
`extra_build_content` for npm_import would be a good solution for this special case

alexeagle:
okay I'll just document that method for now, thanks

PR #380, opened 2022-08-11
refactor: cleanup bzl_library in js_image

alexeagle:
Makes it more clear what our direct dependencies are, vs transitives from python/pkg


thesayyn (approved):
LGTM. hopefully, this doesn't get destroyed by gazelle.

thesayyn:
Filed #381 to remove the dependency on rules_python.

PR #384, opened 2022-08-12
feat: add expand_template rule that support stamp var substitutions; not yet part of the public API

gregmagolan:
We have an `expand_template` rule in bazel-lib but it doesn't support stamp vars and making it do so requires some runtime. If we could do what I did here in `js/private/js_expand_template.js` in bash then bazel-lib would be a better home for this rule. That seems difficult so I put up the rule here so I could use the node runtime to load the stable & volatile status files & perform the replacements.

Don't love the naming `js_expand_template` but `js_` is the rule namespace we've used for everything else. Maybe we drop the prefix for this rule since it is not js specific except for using a js program to perform the replacements. Open to ideas.


js/private/js_expand_template.js
```diff
@@ -0,0 +1,87 @@
...
+ * STABLE_BUILD_SCM_VERSION 6.0.0-beta.6+12.sha-83c699d.with-local-changes
+ * BUILD_TIMESTAMP 1520021990506
+ *
+ * Parsing regex is created based on Bazel documentation describing the status file schema:
+ *   The key names can be anything but they may only use upper case letters and underscores. The
+ *   first space after the key name separates it from the value. The value is the rest of the line
+ *   (including additional whitespace).
+ */
+function _parseStatusFile(p) {
+    if (!p) return []
```
jbedard, line 16: Should this not be `{}`?

I assume that means this `if` is never hit and we can just delete the line?

js/private/js_expand_template.js
```diff
@@ -0,0 +1,87 @@
...
+    const results = {}
+    const statusFile = fs.readFileSync(p, { encoding: 'utf-8' })
+    for (const match of `\n${statusFile}`.matchAll(/^([A-Z_]+) (.*)/gm)) {
```
jbedard, line 19: Is it a requirement that it's all caps?

Can the name/value be separated by more then one space? I think `\s+` would work?

jbedard, line 19: Why the newline at the start?

js/private/js_expand_template.js
```diff
@@ -0,0 +1,87 @@
...
+        statusRegexprs.forEach((r) => {
+            const [regexp, replacement] = r
+            value = value.replace(regexp, replacement)
+        })
+        substitutions[key] = value
+    }
+
+    const regexprs = []
+    for (const key of Object.keys(substitutions)) {
+        regexprs.push([new RegExp(key, 'g'), substitutions[key]])
```
jbedard, line 73: Can `key` contain regex special chars? Do we need to escape it?

js/private/js_expand_template.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    let content = fs.readFileSync(template, { encoding: 'utf-8' })
+    regexprs.forEach((r) => {
+        const [regexp, replacement] = r
```
jbedard, line 78: I think you can do `regexprs.forEach(([regexp, replacement]) => {` 🤷

jbedard:
I think I'd vote to drop the prefix. Other then the implementation and .bzl file location the API itself has nothing js specific so lets not make it look like it does? Then we can potentially move it to bazel-lib without changing more then just import statements etc.

js/private/js_expand_template.js
gregmagolan: Yea; it can be removed; guess that code is wrong in rules_nodejs where I borrowed it from

js/private/js_expand_template.js
gregmagolan: Don't think it is a requirement. Nice catch.

gregmagolan: Good question. Code borrowed from rules_nodejs. Will investigate.

js/private/js_expand_template.js
gregmagolan: Using RegExp probably just wrong in this context. It should just be straight string matching.

gregmagolan: Will update.

gregmagolan:
> I think I'd vote to drop the prefix. Other then the implementation and .bzl file location the API itself has nothing js specific so lets not make it look like it does? Then we can potentially move it to bazel-lib without changing more then just import statements etc.

Yeah. That makes sense. Could also land it and not make it public API either yet.

jbedard:
> Yeah. That makes sense. Could also land it and not make it public API either yet.

Ideally we could do it in bazel-lib and make that one public, so I'd vote to hide this for now until we at least give it a try in bazel-lib.

thesayyn:
> We have an `expand_template` rule in bazel-lib but it doesn't support stamp vars and making it do so requires some runtime.

What if we use bazel lib's expander with a configurable `expander` attribute so that bazel-lib's starlark code is shared across rulesets and each ruleset provides its expander?

js/private/js_expand_template.js
gregmagolan: 👍

js/private/js_expand_template.js
gregmagolan: 👍

js/private/js_expand_template.js
gregmagolan: ✅

js/private/js_expand_template.js
gregmagolan: It's required for the matching algo to find unmatched lines with
```
        // Lines which go unmatched define an index value of `0` and should be skipped.
        if (match.index === 0) {
            continue
        }
```

gregmagolan: Fixed ✅

gregmagolan:
> 

Not a lot to share and the docstrings change with the additional feature of support stamp vars. When not stamping, we can use `ctx.actions.expand_template` and when stamping the path is different and takes an action `ctx.actions.run`. The bazel-lib impl is simply
```
        ctx.actions.expand_template(
            template = ctx.file.template,
            output = ctx.outputs.out,
            substitutions = substitutions,
            is_executable = ctx.attr.is_executable,
        )
```

.bazelrc
```diff
@@ -1,6 +1,9 @@
 # import common bazelrc shared with e2e workspaces
 import %workspace%/.bazelrc.common
 
+# Mock versioning command to test the --stamp behavior
+build --workspace_status_command="echo BUILD_SCM_VERSION 1.2.3"
```
jbedard, line 5: Is there any other way to do this? Having test data within the root bazelrc isn't the greatest imo :/

We could do `build:stamp-test` or something like that, but that means the tests need to do `--config=stamp-test` which might be too much of a hassle? Can it go in a bazelrc in a subfolder? IDK if there's any solution worth doing at this time 🤷

js/private/expand_template.bzl
```diff
@@ -0,0 +1,112 @@
+"expand_template rule"
+
+load("@aspect_bazel_lib//lib:expand_make_vars.bzl", _expand_locations = "expand_locations", _expand_variables = "expand_variables")
+load("@aspect_bazel_lib//lib:stamping.bzl", "STAMP_ATTRS", "maybe_stamp")
+load("@bazel_skylib//lib:dicts.bzl", "dicts")
+
+def _expand_substitutions(ctx, substitutions):
+    result = {}
+    for k, v in substitutions.items():
+        result[k] = " ".join([_expand_variables(ctx, e, outs = [ctx.outputs.out], attribute_name = "substitutions") for e in _expand_locations(ctx, v, ctx.attr.data).split(" ")])
```
jbedard, line 10: buildifier didn't wrap this line? 🤔

jbedard (approved):

.bazelrc
gregmagolan: Moved to `.bazelrc.testing` to keep things clean

js/private/expand_template.bzl
gregmagolan: guess not; wrapped in myself

Issue #386, opened 2022-08-15
`node-gyp rebuild` fails while running the lifecycle hooks for package aws-lambda-ric


bhavitsharma:
Installing the package `aws-lambda-ric@2.0.0` fails when its postinstall script runs the script `node-gyp rebuild`. 
The error is:

```
 File "/home/devbox/.cache/bazel/_bazel_devbox/d1bf528b9e29083fb94e8c7ce57bbdd1/sandbox/linux-sandbox/1633/execroot/relicx/bazel-out/k8-fastbuild/bin/ts/no
de_modules/.aspect_rules_js/node-gyp@8.1.0/node_modules/node-gyp/gyp/pylib/gyp/generator/make.py", line 749, in Write
    self.fp = open(output_filename, "w")
FileNotFoundError: [Errno 2] No such file or directory: './build/../../../../../../../../../../../../../../execroot/relicx/bazel-out/k8-fastbuild/bin/ts/nod
e_modules/.aspect_rules_js/node-addon-api@3.2.1/node_modules/node-addon-api/nothing.target.mk'
```

The installation works fine for other packages dependent on `node-gyp`. 
Manually running `pnpm install` correctly installs the packages.
I can't use `lifecycle_hooks_exclude` since we have a run-time dependency on this package.
I would appreciate any help!

bhavitsharma:
https://github.com/aws/aws-lambda-nodejs-runtime-interface-client
aws-lambda-ric

gregmagolan:
Can you try setting the execution requirements for that lifecycle hook so it runs out of the sandbox? The path in the error message is suspicious.

In you `npm_translate_lock`, you can configure this for a specific package by name or name@version,

```
    lifecycle_hooks_execution_requirements = {
        "aws-lambda-ric": ["no-sandbox"],
    },
```

```
    lifecycle_hooks_execution_requirements = {
        "aws-lambda-ric@2.0.0": ["no-sandbox"],
    },
```

or for all packages,

```
    lifecycle_hooks_execution_requirements = {
        "*": ["no-sandbox"],
    },
```

bhavitsharma:
Wow, This worked. Thanks a lot, @gregmagolan, for your prompt response! Appreciate it.
Two questions:
1. Can the `no-sandbox` have undesired effects while running other ts builds in sandbox mode? 
[...]

PR #389, opened 2022-08-16
fix: run lifecycle hooks outside of sandbox by default

gregmagolan:
At one point while benchmarking I had intended to make this the default. Running outside of the sandbox is a better default for performance since the output TreeArtifact doesn't need to be copied out of the sandbox after the action completes. The symlink node_modules structure already provides strict dependencies as well by virtue of the layout.

This would also prevent issues such as https://github.com/aspect-build/rules_js/issues/386 when some package lc hooks follow symlinks out of the sandbox and fail.


jbedard:
Is there any way to test this?

npm/npm_import.bzl
```diff
@@ -225,11 +236,15 @@ def npm_import(
             At best omitting this field will make your build non-hermetic.
 
             It is optional to make development easier but should be set before shipping.
+
```
jbedard, line 239: Are these formatting changes correct? The other arguments above the diff don't have the blank line...

npm/private/npm_translate_lock.bzl
```diff
@@ -181,6 +181,17 @@ _ATTRS = {
         }
         """,
     ),
+    "lifecycle_hooks_no_sandbox": attr.bool(
+        doc = """If True, the "no-sandbox" execution requirements is added
+        to all lifecycle hooks.
+
+        Equivalent to adding `"*": ["requires-network"]` to lifecycle_hooks_execution_requirements.
```
jbedard, line 188: Shouldn't this be `"no-sandbox"`?

jbedard (approved):

npm/private/npm_translate_lock.bzl
gregmagolan: Indeed. Nice catch.

npm/npm_import.bzl
gregmagolan: Whoops. Cleaned up.

bhavitsharma:
Love it! Thanks again for the quick resolution!

PR #390, opened 2022-08-16
feat: add node_modules/.bin binary files support to npm_translate_lock, npm_import and npm_link_package_store

gregmagolan:
This will be most useful for integration tests that want to run `node_modules/.bin` tools directly.

Only generating bash binaries for now. Windows can use sh_test.

For now the bin entries must be entered manually.

In the future, this field may be automatically populated by translate_lock_file from information in the pnpm lock file. That feature is currently blocked on https://github.com/pnpm/pnpm/issues/5131.

### DIFF
--- a/npm/npm_import.bzl
+++ b/npm/npm_import.bzl
@@ -30,23 +30,243 @@ load("//npm/private:versions.bzl", "PNPM_VERSIONS")
 load("//npm/private:utils.bzl", _utils = "utils")
 load("//npm/private:npm_translate_lock.bzl", _npm_translate_lock_lib = "npm_translate_lock")
 
-npm_translate_lock_rule = repository_rule(
-    doc = _npm_translate_lock_lib.doc,
+_npm_translate_lock = repository_rule(
     implementation = _npm_translate_lock_lib.implementation,
     attrs = _npm_translate_lock_lib.attrs,
 )
 
 LATEST_PNPM_VERSION = PNPM_VERSIONS.keys()[-1]
 
-def npm_translate_lock(name, npm_package_lock = None, yarn_lock = None, pnpm_version = LATEST_PNPM_VERSION, **kwargs):
-    """Wrapper macro around [npm_translate_lock_rule](#npm_translate_lock_rule)
+def npm_translate_lock(
+        name,
+        pnpm_lock = None,
+        package_json = None,
+        npm_package_lock = None,
+        yarn_lock = None,
+        patches = {},
+        patch_args = {},
+        custom_postinstalls = {},
+        prod = False,
+        public_hoist_packages = {},
+        dev = False,
+        no_optional = False,
+        lifecycle_hooks_exclude = [],
+        run_lifecycle_hooks = True,
+        lifecycle_hooks_envs = {},
+        lifecycle_hooks_execution_requirements = {},
+        bins = {},
+        lifecycle_hooks_no_sandbox = True,
+        verify_node_modules_ignored = None,
+        warn_on_unqualified_tarball_url = True,
+        pnpm_version = LATEST_PNPM_VERSION):
+    """Repository rule to generate npm_import rules from pnpm lock file or from a package.json and yarn/npm lock file.
 
-    This macro creates a "pnpm" repository.
-    rules_js currently only uses this repository
+    The pnpm lockfile format includes all the information needed to define npm_import rules,
+    including the integrity hash, as calculated by the package manager.
+
+    For more details see, https://github.com/pnpm/pnpm/blob/main/packages/lockfile-types/src/index.ts.
+
+    Instead of manually declaring the `npm_imports`, this helper generates an external repository
+    containing a helper starlark module `repositories.bzl`, which supplies a loadable macro
+    `npm_repositories`. This macro creates an `npm_import` for each package.
+
+    The generated repository also contains BUILD files declaring targets for the packages
+    listed as `dependencies` or `devDependencies` in `package.json`, so you can declare
+    dependencies on those packages without having to repeat version information.
+
+    Bazel will only fetch the packages which are required for the requested targets to be analyzed.
+    Thus it is performant to convert a very large pnpm-lock.yaml file without concern for
+    users needing to fetch many unnecessary packages.
+
+    **Setup**
+
+    In `WORKSPACE`, call the repository rule pointing to your pnpm-lock.yaml file:
+
+    ```starlark
+    load("@aspect_rules_js//npm:npm_import.bzl", "npm_translate_lock")
+
+    # Read the pnpm-lock.yaml file to automate creation of remaining npm_import rules
+    npm_translate_lock(
+        # Creates a new repository named "@npm_deps"
+        name = "npm_deps",
+        pnpm_lock = "//:pnpm-lock.yaml",
+        # Recommended attribute that also checks the .bazelignore file
+        verify_node_modules_ignored = "//:.bazelignore",
+    )
+    ```
+
+    Next, there are two choices, either load from the generated repo or check in the generated file.
+    The tradeoffs are similar to
+    [this rules_python thread](https://github.com/bazelbuild/rules_python/issues/608).
+
+    1. Immediately load from the generated `repositories.bzl` file in `WORKSPACE`.
+    This is similar to the
+    [`pip_parse`](https://github.com/bazelbuild/rules_python/blob/main/docs/pip.md#pip_parse)
+    rule in rules_python for example.
+    It has the advantage of also creating aliases for simpler dependencies that don't require
+    spelling out the version of the packages.
+    However it causes Bazel to eagerly evaluate the `npm_translate_lock` rule for every build,
+    even if the user didn't ask for anything JavaScript-related.
+
+    ```starlark
+    # Following our example above, we named this "npm_deps"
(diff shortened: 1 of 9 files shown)


npm/private/npm_link_package_store.bzl
```diff
@@ -34,6 +35,17 @@ If unset, the package name of the src npm_package_store is used.
 If set, takes precendance over the package name in the src npm_package_store.
 """,
     ),
+    "bins": attr.string_dict(
+        doc = """Dictionary of `node_modules/.bin` binary files to create mapped to their node entry points.
```
kormide, line 39: I think spelling out what the keys and values are here would be helpful. Or adding an sample.I wouldn't know what to put unless I saw an example.

npm/private/npm_link_package_store.bzl
```diff
@@ -46,13 +58,25 @@ If set, takes precendance over the package name in the src npm_package_store.
     ),
 }
 
+_BIN_TMPL = """#!/bin/sh
+basedir=$(dirname "$(echo "$0" | sed -e 's,\\\\,/,g')")
+exec node "$basedir/{bin_path}" "$@"
+"""
+
```
kormide, line 65: Unrelated to this PR, but do we need a windows `.cmd` equivalent for these like I did [here](https://github.com/angular/angular/pull/47135/files#diff-2bb203c749eb393e1596af31197ef437a63f87fd6ac87dc6222b89f70fee37cfR239-R243)?

npm/private/npm_translate_lock.bzl
```diff
@@ -164,21 +164,50 @@ _ATTRS = {
...
+
+        Binary files to create are declared as key/value pairs of the form "bin_name=bin_path".
+
+        For example:
+
+        ```
+        bins: {
+            "@foo/bar": ["foo=./foo.js", "bar=./bar.js"],
+        }
+        ```
```
kormide, line 210: Ah, I see you have an example here.

npm/private/npm_link_package_store.bzl
gregmagolan: As a follow-up probably yes. For downstream work I'm going to use sh_test which works on Windows.

npm/private/test/BUILD.bazel
```diff
@@ -48,3 +48,14 @@ filegroup(
...
     ],
 )
+
+sh_test(
+    name = "bin_test",
+    srcs = ["bin_test.sh"],
+    args = ["$(NODE_PATH)"],
+    data = [
+        ":node_modules/typescript",
+        "//:node_modules/typescript",
```
kormide, line 58: Is this testing something related to having two packages with conflicting bins? Perhaps a clarifying comment?

kormide (approved):
:taco:

npm/private/npm_link_package_store.bzl
gregmagolan: Added more examples

npm/private/test/BUILD.bazel
gregmagolan: Good call. Added comment.

npm/npm_import.bzl
```diff
@@ -225,13 +237,35 @@ def npm_import(
...
+        bins: Dictionary of `node_modules/.bin` binary files to create mapped to their node entry points.
+
+            This is typically derived from the "bin" attribute in the package.json
+            file of the npm package being linked.
+
+            For example:
+
+            ```
+            bins: {
+                "foo": "./foo.js",
```
jbedard, line 261: Shouldn't the value be in an array? And have the `=`?

Is there a better package name then "foo" that clearly indicates that that is the package? Even just `pkg-foo`?

npm/private/npm_link_package_store.bzl
```diff
@@ -76,6 +120,7 @@ def _impl(ctx):
     providers = [
         DefaultInfo(
             files = depset(files),
+            runfiles = ctx.runfiles(transitive_files),
```
jbedard, line 123: This was missing before? Is that related to this PR?

npm/npm_import.bzl
jbedard: Oh I see, this is the `npm_import`, and it does align with the `npm_translate_lock` 👍

gregmagolan: Yeah, unfortunately there is no `string_list_dict_dict` so the equals is a work-around in npm_translate_lock.

gregmagolan: Tho I might do the translate from string_list_dict_dict in wrapper macro instead so the API is consistent

npm/private/npm_link_package_store.bzl
gregmagolan: It related to the sh_test added. It lets you pass a linked npm package to the data of a generic rule, such as sh_test, and get those files in the runfiles.

jbedard: That seems like a fix/feature on its own that could land separately? Or just it's own commit 🤷 

Is there an easy test we can add for this?

gregmagolan: I'll add a simple sh_test and move it to it own commit

npm/private/npm_translate_lock.bzl
```diff
@@ -6,203 +6,26 @@ load(":utils.bzl", "utils")
...
-        This points to a `.bazelignore` file to verify that all nested node_modules directories
-        pnpm will create are listed.
-
-        See https://github.com/bazelbuild/bazel/issues/8106
-        """,
-    ),
-    "warn_on_unqualified_tarball_url": attr.bool(
-        default = True,
-    ),
+    "pnpm_lock": attr.label(),
```
jbedard, line 10: This means bazel has no reference to the label docs?

npm/npm_import.bzl
gregmagolan: ✅

npm/private/npm_translate_lock.bzl
gregmagolan: Yeah. The docs have been hoisted to the macro now. Same as npm_import already does.

jbedard: Bazel no longer knows of the doc strings though. Is that ok? Does bazel use or provide those in any way?

gregmagolan: stardocs docs are still generated correct from the macro docstring summary and Args. Other than that the docstrings are not load bearing anywhere.

gregmagolan:
Will be really nice when the pnpm lock file has bin entries and we can generate these automatically

npm/npm_import.bzl
```diff
@@ -30,23 +30,243 @@ load("//npm/private:versions.bzl", "PNPM_VERSIONS")
...
+            For example:
+
+            ```
+            lifecycle_hooks_execution_requirements: {
+                "*": ["requires-network"],
+                "@foo/bar": ["no-sandbox"],
+            }
+            ```
+
+        bins: Binary files to create in `node_modules/.bin` for packages in this lock file.
```
jbedard, line 234: Is it worth specifying that the paths are relative to the package directory and not node_modules or node_modules/.bin? Maybe that's obvious enough 🤷

npm/npm_import.bzl
```diff
@@ -57,11 +277,39 @@ def npm_translate_lock(name, npm_package_lock = None, yarn_lock = None, pnpm_ver
...
-    npm_translate_lock_rule(
+    # convert bins to a string_list_dict to satisfy attr type in repository rule
+    bins_string_list_dict = {}
+    if type(bins) != "dict":
+        fail("Expected bins to be a dict")
+    for key, value in bins.items():
+        if type(value) != "dict":
+            fail("Expected values in bins to be a dicts")
+        for value_key, value_value in value.items():
+            if key not in bins_string_list_dict:
```
jbedard, line 288: This could be up a line outside the loop

npm/npm_import.bzl
```diff
@@ -30,23 +30,243 @@ load("//npm/private:versions.bzl", "PNPM_VERSIONS")
...
-npm_translate_lock_rule = repository_rule(
-    doc = _npm_translate_lock_lib.doc,
+_npm_translate_lock = repository_rule(
     implementation = _npm_translate_lock_lib.implementation,
     attrs = _npm_translate_lock_lib.attrs,
 )
 
 LATEST_PNPM_VERSION = PNPM_VERSIONS.keys()[-1]
 
-def npm_translate_lock(name, npm_package_lock = None, yarn_lock = None, pnpm_version = LATEST_PNPM_VERSION, **kwargs):
```
jbedard, line 41: Does this mean things like `tags` and `testonly` can no longer be specified? Is that ok or intentional?

jbedard (approved):

npm/npm_import.bzl
gregmagolan: Should be obvious enough; it is also mentioned that the values are derived from the package.json of the npm package.

npm/npm_import.bzl
gregmagolan: Nice catch

npm/npm_import.bzl
gregmagolan: They never could in the first place since these are repository rules :)

jbedard: Perfect 👍

PR #391, opened 2022-08-16
fix: use less RAM via depsets instead of lists

vpanta:
When testing out these rules (and `aspect-build/rules_ts`) with a large package dependency tree in our monorepo, our bazel servers were OOMing pretty heavily. I finally got it to build by giving the heap 12GB of room (and it took most of it).

Looking at the heap dump, it seemed like the majority of the space was eaten up by `StarlarkLists` existing under the `JsInfo` and `NpmPackageStoreInfo` providers. One example had ~29,000,000 elements but only ~7000 unique. Another had ~15,000,000 elements with still just ~7000 unique.

A rules dump from Bazel also showed the following for `npm_*` rules:
```
RULE                             COUNT     ACTIONS          BYTES         EACH  
npm_package_store_internal       3,380       6,244     17,596,040        5,205  
npm_package_internal             1,685           0      1,311,080          778  
npm_link_package_store             170         170    370,766,376    2,180,978  
npm_package_store                   95       1,309        524,544        5,521  
npm_package                         90         129      2,127,552       23,639  
```

After this, our servers did not OOM, but build completed, and the rules dump has the following values for `npm_*` rules:
```
RULE                             COUNT     ACTIONS          BYTES         EACH  
npm_package_store_internal       3,380       6,244     17,564,960        5,196  
npm_package_internal             1,685           0      2,621,816        1,555  
npm_link_package_store             167         167        524,376        3,139  
npm_package_store                   92         978        524,264        5,698  
npm_package                         87         113      1,959,624       22,524  
```

A quick visual interpretation of the diff here:
Screen Shot 2022-08-16 at 1 57 09 PM
50% would be no change. I'm unsure why `npm_package_internal` nearly doubled, although I'd expect that to be just because the build actually completed (so they were fully populated), whereas you can see the `npm_link_package_store` is now negligible compared to before.

### DIFF
--- a/js/private/js_info.bzl
+++ b/js/private/js_info.bzl
@@ -3,26 +3,26 @@
 JsInfo = provider(
     doc = "Encapsulates information provided by rules in rules_js and derivative rule sets",
     fields = {
-        "declarations": "A list of declaration files produced by the target",
-        "npm_linked_packages": "A list of NpmLinkedPackageInfo providers that are dependencies of this target",
-        "npm_package_stores": "A list of NpmPackageStoreInfo providers from npm dependencies of this target to use as direct dependencies when linking downstream npm_package targets with npm_link_package",
-        "sources": "A list of source files produced by the target",
-        "transitive_declarations": "A list of declaration files produced by the target and the target's transitive deps",
-        "transitive_npm_linked_packages": "A list of NpmLinkedPackageInfo providers that are dependencies of this target and the target's transitive deps",
-        "transitive_npm_package_stores": "A list of NpmPackageStoreInfo providers from npm dependencies of this target and the target's transitive dependencies to use as direct dependencies when linking downstream npm_package targets with npm_link_package",
-        "transitive_sources": "A list of source files produced by the target and the target's transitive deps",
+        "declarations": "A depset of declaration files produced by the target",
+        "npm_linked_packages": "A depset of NpmLinkedPackageInfo providers that are dependencies of this target",
+        "npm_package_stores": "A depset of NpmPackageStoreInfo providers from npm dependencies of this target to use as direct dependencies when linking downstream npm_package targets with npm_link_package",
+        "sources": "A depset of source files produced by the target",
+        "transitive_declarations": "A depset of declaration files produced by the target and the target's transitive deps",
+        "transitive_npm_linked_packages": "A depset of NpmLinkedPackageInfo providers that are dependencies of this target and the target's transitive deps",
+        "transitive_npm_package_stores": "A depset of NpmPackageStoreInfo providers from npm dependencies of this target and the target's transitive dependencies to use as direct dependencies when linking downstream npm_package targets with npm_link_package",
+        "transitive_sources": "A depset of source files produced by the target and the target's transitive deps",
     },
 )
 
 def js_info(
-        declarations = [],
-        npm_linked_packages = [],
-        npm_package_stores = [],
-        sources = [],
-        transitive_declarations = [],
-        transitive_npm_linked_packages = [],
-        transitive_npm_package_stores = [],
-        transitive_sources = []):
+        declarations = None,
+        npm_linked_packages = None,
+        npm_package_stores = None,
+        sources = None,
+        transitive_declarations = None,
+        transitive_npm_linked_packages = None,
+        transitive_npm_package_stores = None,
+        transitive_sources = None):
     """Construct a JsInfo
 
     Args:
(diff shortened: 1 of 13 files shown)


js/private/js_library_helpers.bzl
```diff
@@ -27,39 +27,41 @@ def gather_transitive_sources(sources, targets):
...
+        target[JsInfo].transitive_declarations
         for target in targets
-        if JsInfo in target and hasattr(target[JsInfo], "transitive_declarations")
-        for item in target[JsInfo].transitive_declarations
+        if JsInfo in target
     ]
-    transitive_declarations = [file for file in transitive_declarations if file]
-    return transitive_declarations
+    if type(declarations) == "list":
+      return depset(declarations, transitive = transitive_deps)
```
jbedard, line 62: Should we do `depset([], transitive = [depset(declarations)] + transitive_deps)` so it is equivelent to the `else`?

I'm not sure if it really matters but have this slight difference could cause a difference downstream?

vpanta: That's definitely doable. I think the only difference would be in the ordering of the files when pulled out (one treats it as siblings instead of parents). While waiting on approving the CLA, I'll see if anyone has a good reason to do this (or not to do this).

js/private/js_info.bzl
```diff
@@ -38,13 +38,31 @@ def js_info(
     Returns:
         A JsInfo provider
     """
+    # These are for legacy-users of js_info, who might pass in lists instead of depsets.
```
alexeagle, line 41: Yeah so about this being a breaking change to a public API...

Thanks for including this for compatibility. I think it helps, but it's not sufficient. Old producing rules can work with newer rules_js, but old consuming rules cannot. I'm thinking through implications.

An org might have some custom rules in a private repo. A given version of those rules consumes only lists, another version only depsets. Then the end-user has to select the version of those rules that works with the rules_js version they install. Seems tricky to explain, and not obvious to users since the error message you'd get is just `Error: 'list' value has no field or method 'to_list'`.

The improvement to resource usage in this PR might justify the breaking nature of it. It's hard to say since we don't know how many usages of the public API there are.

Any ideas how to make it optional/less breaking?

vpanta: No worries, I wasn't sure the best way to introduce it across the board. Namely: if this would be something that we'd need to make in lock-step to all the repos at once, or if we make this change more incrementally. I can re-approach it with that in mind though!

alexeagle: We don't really have the lock-step option anymore now that we shipped 1.0.0, at least in theory.

vpanta: So going over the codebases, I think all consumption of these providers from other Aspect repositories is currently through the `js_lib_helpers` struct. It looks like the main point of incompatibility is that those methods will return `depset`s instead of `list`s. So I've updated that API to use methods which can accept either lists or depsets, and return lists.

I don't know how y'all want to deal with the incompatibility, but my approach is aiming at:

1. Release a new `js_lib_depset_helpers` struct, and keep the `js_lib_helpers` struct for lists-users.
   - Repositories wanting the latest after this is added can change their code to work with `js_lib_depset_helpers` instead.
2. When y'all are ready to break things, make `js_lib_helpers` and `js_lib_depset_helpers` be the same struct.
3. And when you're ready to cleanup, remove `js_lib_depset_helpers`.

I did a GitHub-wide search for `*.bzl` using `JsInfo`, but only came across your repos, users of `bazelbuild/rules_nodejs`, and a `rivethealth/rules_javascript` (a separate attempt at JS rules for bazel), so this approach seems like it shouldn't cause too much tumult?

js/private/js_library_helpers.bzl
vpanta: Made this change.

js/private/js_info.bzl
alexeagle: That seems like a good compromise to me. Anyone who wrote their own rules in the last couple weeks and is using the helpers will be okay. Anyone interacting directly with the provider will have to figure out the version skew issue.

Of course the sooner we make and release the change the easier the migration is, since fewer things in the world will have taken a dependency on the lists version.

gregmagolan:
@vpanta Can you please try the change in https://github.com/aspect-build/rules_js/pull/411 in your repo and measure memory usage? It attempts to resolve this issue without any API changes.

gregmagolan:
Replaced by https://github.com/aspect-build/rules_js/pull/413

PR #392, opened 2022-08-16
fix: provide runfiles of all transitives from npm_link_package_store

gregmagolan:
Pre-factor from https://github.com/aspect-build/rules_js/pull/392


jbedard (approved):
🌮

PR #394, opened 2022-08-18
chore: move lifecycle bundling into Bazel

alexeagle:
This ensures that our bundle is up-to-date as the version is bumped

note: the lifecycle/min/index.min.js file is different from before, which is clearly undesirable. I pinned to the same versions and run an identical command line. I'm still unable to explain why.


gregmagolan:
e2e test failure?

alexeagle:
Yeah it exposes a dev dependency into a package that users load. I need to find a tricky way to put this in some other package, while still letting write_source_file do its thing.

alexeagle:
It's been a year, so it looks like I'm not going to find time to finish this.

Issue #396, opened 2022-08-19
/npm/pnpm-lock.yaml (No such file or directory)


uhthomas:
Given

##### WORKSPACE

```starlark
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "aspect_rules_js",
    sha256 = "538049993bec3ee1ae9b1c3cd669156bca04eb67027b222883e47b0a2aed2e67",
    strip_prefix = "rules_js-1.0.0",
    url = "https://github.com/aspect-build/rules_js/archive/refs/tags/v1.0.0.tar.gz",
)

load("@aspect_rules_js//js:repositories.bzl", "rules_js_dependencies")

rules_js_dependencies()

load("@rules_nodejs//nodejs:repositories.bzl", "DEFAULT_NODE_VERSION", "nodejs_register_toolchains")

nodejs_register_toolchains(
    name = "nodejs",
    node_version = DEFAULT_NODE_VERSION,
)

load("@aspect_rules_js//npm:npm_import.bzl", "npm_translate_lock")

npm_translate_lock(
    name = "npm",
    package_json = "//:package.json",
    verify_node_modules_ignored = "//:.bazelignore",
    yarn_lock = "//:yarn.lock",
)

load("@npm//:repositories.bzl", "npm_repositories")

npm_repositories()
```

Error:

```
❯ bazel build //...
Starting local Bazel server and connecting to it...
INFO: Repository npm instantiated at:
  /Users/thomas/code/sourcegraph/WORKSPACE:23:19: in <toplevel>
  /private/var/tmp/_bazel_thomas/371526a2a92600fa64bb909f1dd7a076/external/aspect_rules_js/npm/npm_import.bzl:55:28: in npm_translate_lock
Repository rule npm_translate_lock_rule defined at:
  /private/var/tmp/_bazel_thomas/371526a2a92600fa64bb909f1dd7a076/external/aspect_rules_js/npm/npm_import.bzl:32:42: in <toplevel>
ERROR: An error occurred during the fetch of repository 'npm':
   Traceback (most recent call last):
        File "/private/var/tmp/_bazel_thomas/371526a2a92600fa64bb909f1dd7a076/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 597, column 37, in _impl
                lockfile = _process_lockfile(rctx, "pnpm-lock.yaml")
[...]

PR #398, opened 2022-08-19
fix: add error handling for failure of pnpm import

alexeagle:
Otherwise we just skip past this and give the user an unhelpful error that pnpm-lock.yaml isn't found

Partial fix for #396

### DIFF
--- a/npm/private/npm_translate_lock.bzl
+++ b/npm/private/npm_translate_lock.bzl
@@ -432,11 +432,14 @@ def _impl(rctx):
         else:
             fail("rules_js internal validation error, please file an issue")
 
-        rctx.execute([
+        result = rctx.execute([
             rctx.path(Label("@nodejs_host//:bin/node")),
             rctx.path(Label("@pnpm//:package/bin/pnpm.cjs")),
             "import",
         ])
+        if result.return_code:
+            msg = "pnpm import exited with status %s: \nSTDOUT:\n%s\nSTDERR:\n%s" % (result.return_code, result.stdout, result.stderr)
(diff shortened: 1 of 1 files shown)


npm/private/npm_translate_lock.bzl
```diff
@@ -432,11 +432,14 @@ def _impl(rctx):
...
             fail("rules_js internal validation error, please file an issue")
 
-        rctx.execute([
+        result = rctx.execute([
             rctx.path(Label("@nodejs_host//:bin/node")),
             rctx.path(Label("@pnpm//:package/bin/pnpm.cjs")),
             "import",
         ])
+        if result.return_code:
+            msg = "pnpm import failed: \nSTDOUT:\n%s\nSTDERR:\n%s" % (result.stdout, result.stderr)
```
jbedard, line 441: worth putting the error code in?

alexeagle: sure, maybe we should do that everywhere else too

jbedard: I assume the error code of external tools might be meaningful sometimes 🤷

jbedard (approved):

jbedard:
I assume error cases like this would be a pain to test? Could we easily test it with a `test.sh` script and maybe a bad lock file?

alexeagle:
I think the general solution here is the starlark linter should tag uses of rctx.execute where the return value is discarded

alexeagle:
https://github.com/bazelbuild/buildtools/issues/1085

PR #400, opened 2022-08-22
chore: add CI on RBE

alexeagle:


npm/private/npm_import.bzl
```diff
@@ -619,7 +619,10 @@ def _impl_links(rctx):
             msg = "lifecycle_hooks_env contains invalid key value pair '%s', required '=' separator not found" % env
             fail(msg)
 
-    lifecycle_hooks_execution_requirements = {}
+    lifecycle_hooks_execution_requirements = {
+        # Workaround engflow actions having no DNS despite requires-network tag
+        "no-remote-exec": "1",
```
gregmagolan, line 624: you can set the lifecycle hooks for the specific offending packages using `rctx.attr.lifecycle_hooks_execution_requirements` in `npm_translate_lock`

alexeagle: hard to know which ones they are though. the figma package suggests that node-gyp always talks to nodejs.org

alexeagle: okay I'll just repeat until I identify them all, for this PR at least

.bazelversion
```diff
@@ -1,4 +1,4 @@
-5.3.0
+rolling
```
gregmagolan, line 1: not sure about using rolling here; it might be bad for contributors if things suddenly fail due to a new release

.github/workflows/ci.yaml
```diff
@@ -18,6 +18,9 @@ jobs:
 
         strategy:
             matrix:
+                config:
+                    - 'local'
+                    - 'rbe'
```
gregmagolan, line 23: These is one additional matrix case that would be good to test on CI which is the local but with `--experimental_allow_unresolved_symlinks` case. This should be the default at some point in the future as it will allow the link code to be simplified quite a bit.

gregmagolan (approved):
🌮 ^2

PR #402, opened 2022-08-23
feat: demonstrate stamping the version into package.json

alexeagle:


examples/npm_package/packages/pkg_a/BUILD.bazel
```diff
@@ -1,16 +1,29 @@
...
 
+jq(
+    name = "package",
+    srcs = ["package.json"],
+    filter = "|".join([
+        # Don't directly reference $STAMP as it's only set when stamping
+        # This 'as' syntax results in $stamp being null in unstamped builds.
+        "$ARGS.named.STAMP as $stamp",
+        # Provide a default using the "alternative operator" in case $stamp is null.
+        ".version = ($stamp.STABLE_BUILD_VERSION // \"0.0.0\")",
```
gregmagolan, line 16: perhaps link to relevant jq documentation for users that are not familiar with the tool?

gregmagolan (approved):

examples/npm_package/packages/pkg_a/BUILD.bazel
```diff
@@ -1,16 +1,29 @@
+load("@aspect_bazel_lib//lib:jq.bzl", "jq")
 load("@aspect_rules_js//npm:defs.bzl", "npm_package")
 load("@aspect_rules_js//js:defs.bzl", "js_library")
 load("@npm//:defs.bzl", "npm_link_all_packages")
 
 npm_link_all_packages(name = "node_modules")
 
+jq(
```
kormide, line 8: Do you think this will be common enough that it's worth wrapping in a macro supplied by rules_js?

alexeagle: that's what I'm wondering here - I secretly sent this as a strawman API proposal. At first I thought it could look like

```
npm_package(
    ...
    package_json = "package.json",
    version_stamp_key = "STABLE_BUILD_VERSION",
)
```

but I think it might be load-bearing in the dependency graph that the package.json comes along with the js_library so fields like `main` can be read at the use site without having to go through an npm_package?

alexeagle: the other problem with a macro is which file to put it in it since it adds a dependency on a jq toolchain which not all rules_js users have setup... I think this means it must be a new starlark file rather than adding a `load` statement to an existing one

alexeagle: oh I'm wrong there, the toolchain is only needed at analysis time, not loading time, so putting into existing file is fine

PR #411, opened 2022-08-25
fix: reduce memory usage from duplicate elements in rules_js providers

gregmagolan:
As reported in https://github.com/aspect-build/rules_js/pull/391, memory is pilling up in providers lists due to duplicate items. In particular from `npm_link_package_store` rules likely due to the `transitive_files` list.

This attempts to resolve the issue without any breaking API changes by de-duping all lists in JsInfo, NpmPackageInfo, NpmPackageStoreInfo & NpmLinkedPackageInfo providers as well as in the js_library_helpers gather_* functions.

For rules_js 2.0 we should consider switching some or all of these to depsets. The NpmPackageInfo, NpmPackageStoreInfo & NpmLinkedPackageInfo should probably have been depsets from the start. JsInfo we chose lists so that it would be easier to iterate over files in rules without having to collapse a deeply nested depset which can be expensive.

### DIFF
--- a/js/private/js_info.bzl
+++ b/js/private/js_info.bzl
@@ -1,5 +1,7 @@
 "JsInfo provider"
 
+load("@bazel_skylib//lib:sets.bzl", "sets")
+
 JsInfo = provider(
     doc = "Encapsulates information provided by rules in rules_js and derivative rule sets",
     fields = {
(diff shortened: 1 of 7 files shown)


js/private/js_info.bzl
```diff
@@ -14,6 +14,26 @@ JsInfo = provider(
...
+        js_info_partial: A partial JsInfo with some but not necessarily all fields populated
+
+    Returns:
+        A JsInfo provider
+    """
+    return JsInfo(
+        declarations = js_info_partial.declarations if hasattr(js_info_partial, "declarations") else [],
+        npm_linked_packages = js_info_partial.npm_linked_packages if hasattr(js_info_partial, "npm_linked_packages") else [],
+        npm_package_stores = js_info_partial.npm_package_stores if hasattr(js_info_partial, "npm_package_stores") else [],
+        sources = js_info_partial.sources if hasattr(js_info_partial, "sources") else [],
```
alexeagle, line 30: getattr takes a second default param which would tighten this up

alexeagle (commented):
Do you think the change to depset would be too breaking, even with vpanta approach to accepting both forms?
This seems to solve memory use at the expense of runtime performance?
We should probably gather memory high watermark as a benchmark in the tests

gregmagolan:
> Do you think the change to depset would be too breaking, even with vpanta approach to accepting both forms? This seems to solve memory use at the expense of runtime performance? We should probably gather memory high watermark as a benchmark in the tests

We're not flattening any deeply nested depsets and using them strictly as flat sets here to de-dup lists so calling to_list() on these is not as bad as a depset that is many levels deep with transitives. But there will still be some runtime performance.

I measured memory in rules_js and there was an improvement but I need a larger set of npm deps to see a significant difference. Perhaps I'll pull up the npm benchmarks and measure analysis time difference and memory usage there.

gregmagolan:
If this change resolves the memory issue and adds minimal drag then the JsInfo provider direct elements may still better as lists to make it easier to iterate over them in rules. transitive lists could be changed depsets since they will grow in size in a big graph and are not likely to be iterated over. But if the drag is minimal here then we an make those changes in 2.0 so we don't have to support two modes in the helpers.

alexeagle:
When we do change it for 2.0 I think we may have to accept both lists and depsets to make it possible for ppl to update (either rules_js first or their custom rules first)

we do have direct_blah and transitive_blah on these, doesn't the transitive one get built up from a tree traversal? I'm not seeing the spot where you'd usually have a depset for that, though, maybe it's fine like you say.

gregmagolan:
I measured this change in our npm benchmarks repo and it will be too much of a performance hit. We'll probably just switch to depsets for 1.1.0 and make it slightly breaking for any custom rules that are depending on lists.

PR #413, opened 2022-08-26
fix: use depsets instead of lists in rules_js providers

gregmagolan:
Improves performance and reduced memory usage in large graphs. Forked from https://github.com/aspect-build/rules_js/pull/391

Performance gain on linking observed. 45s to link on the npm benchmark vs the current 55s.

This is going to be breaking for downstream rule sets but non-breaking for end-users. This seems acceptable for 1.1 in order to fix the memory issues and get the performance gains. There shouldn't be that many custom rules built on top of rules_js yet so better to take the breaking downstream rule set change now rather than later.


js/private/js_info.bzl
```diff
@@ -3,48 +3,71 @@
...
-        "npm_package_stores": "A list of NpmPackageStoreInfo providers from npm dependencies of this target to use as direct dependencies when linking downstream npm_package targets with npm_link_package",
-        "sources": "A list of source files produced by the target",
-        "transitive_declarations": "A list of declaration files produced by the target and the target's transitive deps",
-        "transitive_npm_linked_packages": "A list of NpmLinkedPackageInfo providers that are dependencies of this target and the target's transitive deps",
-        "transitive_npm_package_stores": "A list of NpmPackageStoreInfo providers from npm dependencies of this target and the target's transitive dependencies to use as direct dependencies when linking downstream npm_package targets with npm_link_package",
-        "transitive_sources": "A list of source files produced by the target and the target's transitive deps",
+        "declarations": "A depset of declaration files produced by the target",
+        "npm_linked_package_files": "A depset of files that are that are in npm linked package dependencies of this target",
```
jbedard, line 7:
```suggestion
        "npm_linked_package_files": "A depset of files in npm linked package dependencies of this target",
```
(or at least remove the dupe)

js/private/js_info.bzl
```diff
@@ -3,48 +3,71 @@
...
         transitive_declarations: See JsInfo documentation
+        transitive_npm_linked_package_files: See JsInfo documentation
         transitive_npm_linked_packages: See JsInfo documentation
-        transitive_npm_package_stores: See JsInfo documentation
         transitive_sources: See JsInfo documentation
 
     Returns:
         A JsInfo provider
     """
+    if type(declarations) == "list":
```
jbedard, line 44: Do you think we'll always have this or drop it in 2.0? If drop then lets add a comment?

js/private/js_library.bzl
```diff
@@ -143,23 +143,19 @@ target in {file_basename}'s package and add that target to the deps of {this_tar
...
             sources.append(file)
 
-    # sources
-    sources.extend([
-        item
+    # sources as depset
+    sources = depset(sources, transitive = [
+        target[JsInfo].sources
         for target in targets
-        if JsInfo in target and hasattr(target[JsInfo], "sources")
```
jbedard, line 150: Would `JsInfo.sources` being `None` and keeping this `if` be of any value? Or just annoying?

js/private/js_library_helpers.bzl
```diff
@@ -27,39 +27,39 @@ def gather_transitive_sources(sources, targets):
...
     """Gathers transitive sources from a list of direct sources and targets
 
     Args:
-        declarations: Direct sources which should be included in `transitive_declarations`
+        declarations: list or depset of direct sources which should be included in `transitive_declarations`
         targets: List of targets to gather `transitive_declarations` from `JsInfo`
 
     Returns:
-        List of transitive sources
+        Depset of transitive sources
```
jbedard, line 53:
```suggestion
        depset of transitive sources
```
? 🤔

js/private/js_library_helpers.bzl
```diff
@@ -69,104 +69,91 @@ def gather_npm_linked_packages(srcs, deps):
...
-def gather_npm_package_stores(targets):
+def gather_npm_package_store_deps(targets):
     """Gathers NpmPackageStoreInfo providers from the list of targets
 
     Args:
         targets: the list of targets to gather from
 
     Returns:
-        A `struct(direct, transitive)` of direct and transitive npm package stores gathered
+        Depset of npm package stores gathered
```
jbedard, line 117:
```suggestion
        A `depset` of npm package stores gathered
```
Similar to the last suggestion. Just seems like when referring to a type we shouldn't change the case of anything...

js/private/js_info.bzl
gregmagolan: its useful for downstream rules if you just have a list of files you can pass them in without wrapping them in a depset; otoh, downstream rules could write `depset(files).to_list()` and it would be a perf hit that would be hard to catch in the code

js/private/js_library.bzl
gregmagolan: Nice catch. It is worth keeping the hasattr as a defence in depth

npm/private/npm_package.bzl
```diff
@@ -125,38 +125,34 @@ def _impl(ctx):
...
-        for item in target[JsInfo].transitive_npm_package_stores
+        if JsInfo in target and hasattr(target[JsInfo], "npm_package_store_deps")
     ]
 
     copy_to_directory_action(
         ctx,
         srcs = ctx.attr.srcs,
         dst = dst,
-        additional_files = additional_files,
+        additional_files = additional_files.to_list(),
```
alexeagle, line 155: can we avoid to_list here by fixing copy_to_directory_actions?

npm/private/npm_package_store.bzl
```diff
@@ -158,6 +158,8 @@ def _impl(ctx):
     transitive_files = []
     direct_ref_deps = {}
 
+    npm_package_store_deps = ctx.attr.src[NpmPackageInfo].npm_package_store_deps.to_list() if ctx.attr.src else []
```
alexeagle, line 161: we can avoid the to_list here by keeping these as depsets. You're just making them transitives of a later depset it seems

gregmagolan: This one we're iterating on below so we need to convert. This code path is only for manually linked 1st party deps; it is an empty depset in the happy path pnpm-lock file path

npm/private/npm_package.bzl
gregmagolan: Don't think we can here since copy_to_directory needs to iterate over all files it gets and get their individual paths. This is only for 1st party deps; 3rd party deps from pnpm lock don't use `npm_package`

alexeagle (approved):

js/private/js_info.bzl
gregmagolan: 👍

js/private/js_library_helpers.bzl
gregmagolan: 👍

Issue #423, opened 2022-08-31
https resolutions in package.json causes failures


paullewis:
If I have the following `package.json`:

```
{
  "name": "repro",
  "version": "1.0.0",
  "resolutions": {
    "jsonify": "http://localhost/jsonify-1.0.0.tgz"
  },
  "dependencies": {
    "json-stable-stringify": "1.0.1"
  }
}
```

The build will fail with this error:

```
ERROR: BUILD.bazel:13:22: in deps attribute of npm_package_store_internal rule //:.aspect_rules_js/node_modules/json-stable-stringify/1.0.1/pkg: target '//:.aspect_rules_js/node_modules/jsonify/@localhost/jsonify-1.0.0.tgz/ref' does not exist. Since this rule was created by the macro 'npm_link_all_packages', the error might have been caused by the macro implementation
```

(Here I'm mimicking an npm registry with `localhost` but it applies no matter what the origin is.)

AFAICT there's an issue with the paths in `external/npm__json-stable-stringify__1.0.1__links`, and in particular the `ref_deps` doesn't seem to account for the fact we've got a tarball:

```
ref_deps = {
    ":.aspect_rules_js/{}/jsonify/@localhost/jsonify-1.0.0.tgz/ref".format(link_root_name): "jsonify",
}
```

gregmagolan:
I took a minute to try to repo and besides another issue with `#` not being escaped, I'm able to specify `http://` versions and pnpm overrides (should be equivalent to resolutions). https://github.com/aspect-build/rules_js/pull/424

What the pnpm lockfile entry look like for jsonify for you?

paullewis:
The whole pnpm lock file is:

```
lockfileVersion: 5.4

overrides:
  jsonify: http://localhost/jsonify-1.0.0.tgz

specifiers:
  json-stable-stringify: 1.0.1

dependencies:
  json-stable-stringify: 1.0.1

packages:

  /json-stable-stringify/1.0.1:
    resolution: {integrity: sha1-mnWdOcXy/1A/1TAGRu1EX4jE+a8=, tarball: https://github.com/substack/json-stable-stringify/archive/refs/tags/1.0.1.tar.gz}
    dependencies:
      jsonify: '@localhost/jsonify-1.0.0.tgz'
    dev: false

  '@localhost/jsonify-1.0.0.tgz':
    resolution: {tarball: http://localhost/jsonify-1.0.0.tgz}
    name: '@spotify-internal/jsonify'
[...]

PR #424, opened 2022-08-31
fix: handle http:// package.json references with mismatching package name

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/423

Also escape # characters when generating bazel name from pnpm lockfile values for `http://` references such as

```
"inline-fixtures": "https://registry.yarnpkg.com/inline-fixtures/-/inline-fixtures-1.1.0.tgz#5f9edf644ac02a1322012dbbb9f0a4739831d662",
```


package.json
```diff
@@ -2,8 +2,9 @@
     "private": true,
     "devDependencies": {
         "@types/node": "16.11.54",
-        "inline-fixtures": "1.1.0",
-        "typescript": "4.8.2"
+        "inline-fixtures": "https://registry.yarnpkg.com/inline-fixtures/-/inline-fixtures-1.1.0.tgz#5f9edf644ac02a1322012dbbb9f0a4739831d662",
+        "typescript": "4.8.2",
+        "json-stable-stringify": "1.0.1"
```
jbedard, line 7: What is this used for? Did this reproduce the problem?

gregmagolan: I did now, https://github.com/aspect-build/rules_js/issues/423#issuecomment-1233130904

gregmagolan: Will put this back to DRAFT and include the fix for name issue in #423

jbedard: It would be nice to have a more explicit test. Or at least a comment with a link to the issue so no one sees this "unused dependency" and deletes it, or upgrades it to a version no longer reproducing the problem etc.

Maybe we could have an e2e test for tests like this? Somewhere to put package/lock related tests instead of putting them in this root package.json?

gregmagolan: I have a spot already. Will hoist them to npm/private/test/package.json

npm/private/npm_import.bzl
```diff
@@ -33,7 +33,7 @@ def npm_imported_package_store(
     ref_deps = {ref_deps}
 
     has_lifecycle_build_target = {has_lifecycle_build_target}
-    store_target_name = "{virtual_store_root}/{{}}/{package}/{version}".format(link_root_name)
```
jbedard, line 36: Was there previously any reason that this was *not* `virtual_store_name`?

jbedard (approved):

npm/private/npm_import.bzl
gregmagolan: Legacy reasons I believe when the virtual store name needed to be friendly. It doesn't any more so the target name can match the directory name which is much better.

PR #425, opened 2022-08-31
ci: run PR workflow on main repo and not forks

jbedard:
... can this PR test the ability for forked PRs to run? 🤔 (I'll assume not)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,11 +5,15 @@ on:
     # Triggers the workflow on push or pull request events but only for the main branch
     push:
         branches: [main]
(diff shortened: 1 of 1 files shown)


alexeagle:
huh, this PR triggered tests though https://github.com/aspect-build/bazel-lib/pull/231

alexeagle:
I checked that bazel-lib still triggers tests when run from a fork
https://github.com/aspect-build/bazel-lib/pull/232
so I think you're right, your change here isn't live until we merge it, which is probably desirable so external ppl can't break in

alexeagle (approved):

PR #428, opened 2022-09-02
 ci: exclude rbe jobs if no engflow secret

jbedard:


.github/workflows/ci.yaml
```diff
@@ -16,15 +16,32 @@ permissions:
   contents: read
 
 jobs:
+    # Dynamically generate a bit of JSON depending on whether our action has access to repository secrets.
```
alexeagle, line 19: Need to revert that line above that changed to pull_request_target too

jbedard: Updated. I assuming keeping the `permissions` is ok?

alexeagle: Yeah I don't think that matters, though I'd remove just to keep things more uniform

alexeagle (approved):

PR #430, opened 2022-09-02
fix: do not run coverage reporter if already reported

### LINKED ISSUE #337: Feature request: custom coverage reporting
I've managed to get Jest running via:

```bazel
load("@npm//:jest/package_json.bzl", jest_bin = "bin")

jest_bin.jest_test(
    name = "src_tests",
    args = [
        "--roots=%s" % native.package_name(),
        "--config=$(rootpath //:copy_jest_config)",
    ],
    data = [
        ":src",
        "//:copy_jest_config",
    ],
)
```

Due to some quirks of Jest it does require that a config is sent over (hence the target in the `data` attribute). However, Jest also wants to be the one to generate coverage reports. So in the `jest.config.js` I have the following:

```javascript
module.exports = {
  cache: false,
  collectCoverage: true,
  coverageDirectory: process.env.COVERAGE_DIR,
  collectCoverageFrom: undefined,
  coverageReporters: ['text', ['lcovonly', {file: 'lcov.info'}]],
  modulePathIgnorePatterns: [],
  testMatch: ['**/*.test.js'],
  testPathIgnorePatterns: ['/node_modules/'],
};
```

Which means that Jest is now happily outputting `lcov.info` to the `COVERAGE_DIR`. However, the coverage reporter for `js_test` _seems_ to kick in and overwrite the `lcov.info` file based on the V8 data (`coverage.js`, I think?).

Would it be possible to either add an attribute indicating a path to find existing lcov data (as in this case, where I can generate the lcov via Jest into the `COVERAGE_DIR` directly), or skip the generation of `lcov.info` if something else has already written the file?

[Note: I edited this description once I figured out it was the `lcov.info` file I needed to have Jest create.]

thesayyn:
fixes #337

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -71,5 +71,7 @@ jobs:
                   # Bazelisk will download bazel to here
                   XDG_CACHE_HOME: ~/.cache/bazel-repo
               working-directory: ${{ matrix.folder }}
+              # Coverage does not work properly with RBE. See: bazelbuild/bazel#4685
+              if: ${{ matrix.config == 'local' }}
               run: |
                   bazel --bazelrc=$GITHUB_WORKSPACE/.github/workflows/ci.bazelrc --bazelrc=.bazelrc coverage --config=${{ matrix.config }} --instrument_test_targets //...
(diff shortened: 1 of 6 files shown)


thesayyn:
RBE failing due to https://github.com/bazelbuild/bazel/issues/4685

cgrindel:
@thesayyn Do we need RBE to declare victory on this PR? If so, why?

thesayyn:
> @thesayyn Do we need RBE to declare victory on this PR? If so, why?

No. just fixed the branch. It is good to go once green.

gregmagolan (approved):
🌮 ^2

Issue #435, opened 2022-09-06
gather_transitive_closure of npm_translate_lock doesn't handle `link:` entries


gregmagolan:
I recall running into this a while ago but had bigger fish to fry at the time. At this point all the other known issues with parsing the pnpm lock file are resolved so should tackle this one.

TL;DR of the issue is that the `gather_transitive_closure` doesn't consider the `importers` in the lock file so when it encounters a `link:` reference in an npm package such as this one,

```
  /@sourcegraph/extension-api-classes/1.1.0_sohmsbidf4ixb2vnv3mpdkpiau:
    resolution: {integrity: sha512-7VHKaqcDlHqsG8gNWeAJ+deOvhYJyFkrceSUOf6fkYZ1t+0rk2BaSPu6YyylMm7jmy/Kybq5DnOW8Jsymiwnmg==}
    peerDependencies:
      sourcegraph: '>=24.6.0'
    dependencies:
      '@sourcegraph/extension-api-types': 2.1.0_sohmsbidf4ixb2vnv3mpdkpiau
      sourcegraph: link:client/extension-api
    dev: false
```

(which comes via the peerDependency on `sourcegraph` in the npm package https://unpkg.com/browse/@sourcegraph/extension-api-classes@1.1.0/package.json)

it doesn't know how to resolve it to the first-party package found in the importers.

For more context, this issue was originally reported in this Bazel Slack thread on the #javascript channel: https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1662477599465439?thread_ts=1662471165.169249&cid=CEZUUKQ6P

gregmagolan:
Fixed by #441

PR #441, opened 2022-09-09
fix: handle link: dependencies in pnpm-lock.yaml packages

gregmagolan:
- Add failing test case
- Fix issue #435

### DIFF
--- a/npm/private/transitive_closure.bzl
+++ b/npm/private/transitive_closure.bzl
@@ -40,7 +40,12 @@ def gather_transitive_closure(packages, no_optional, direct_deps, transitive_clo
             if version in transitive_closure[name]:
                 continue
             transitive_closure[name].insert(0, version)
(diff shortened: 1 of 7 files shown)


npm/private/transitive_closure.bzl
```diff
@@ -40,7 +40,12 @@ def gather_transitive_closure(packages, no_optional, direct_deps, transitive_clo
             if version in transitive_closure[name]:
                 continue
             transitive_closure[name].insert(0, version)
-            package_info = packages[package_key]
+            if package_key.startswith("link:"):
+                # we don't need to drill down through first-party links for the transitive closure since there are no cycles
+                # allowed in first-party links
+                continue
```
alexeagle, line 46: nit: `continue` statements can be hard to follow, looks like you could just invert the condition and avoid that?

alexeagle (approved):

PR #445, opened 2022-09-11
feat: add npmrc attribute to npm_translate_lock to be available for pnpm import

gregmagolan:
Needed so that you can set https://pnpm.io/npmrc#strict-peer-dependencies to false in cases where there are expected and acceptable missing peer dependencies when calling `pnpm import` in `npm_translate_lock`


jbedard (approved):

jbedard:
Can we add some type of test for this?

gregmagolan:
> Can we add some type of test for this?

Yup. Was planning to before landing. Added to e2e tests.

PR #454, opened 2022-09-15
Mark intermediary targets in package_json.bzl as manual

alex-torok:
I noticed that building and testing a directory with some targets tagged manual still led to npm packages being fetched. This was due to the intermediary `*__entry_point` and `*__js_binary` targets not being tagged manual.

Update `npm/private/npm_import.bzl` to tag those targets as manual and add a test that calls each of the methods of the exported `bin` struct to ensure that the only non-manual-tagged targets are the ones we asked for from the macros.

The functional change in this PR is the smallest part - most of it is new tests to ensure there are no regressions. This was the first time I've used a starlark testing library, so seeking any feedback on if the tests should be written differently. One area that I think may be worth changing would be to put the tests into their own package so that I don't need to do filtering based on the prefix of the target names that I defined.

I also needed to update the skylib dev dependency to get access to `loadingtest` in `@bazel_skylib//lib:unittest.bz`.

### DIFF
--- a/js/dev_repositories.bzl
+++ b/js/dev_repositories.bzl
@@ -35,9 +35,8 @@ def rules_js_dev_dependencies():
     # see https://github.com/bazelbuild/bazel-skylib/issues/250
     http_archive(
         name = "bazel_skylib",
-        sha256 = "07b4117379dde7ab382345c3b0f5edfc6b7cff6c93756eac63da121e0bbcc5de",
-        strip_prefix = "bazel-skylib-1.1.1",
-        urls = ["https://github.com/bazelbuild/bazel-skylib/archive/refs/tags/1.1.1.tar.gz"],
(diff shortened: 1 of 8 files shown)


gregmagolan (approved):
:taco:

Thanks for the fix. Nice unit tests.

I believe the npm package will still be fetched just from the load statement of the package_json.bzl.

We have an issue up on pnpm to include details of the bins in the lock file. Might have make that change ourselves upstream since we are the only use case that needs it. If we have that information in the lock file then we could generate the package_json.bzl without fetching the npm package. As it stands, we need to read the package.json file from the downloaded package to know what bin entries exist.

alex-torok:
> I believe the npm package will still be fetched just from the load statement of the package_json.bzl.

Ahh, that's correct. I noticed this and got distracted with the opportunity of fixing it and learning starlark unit tests and didn't check to see that it would actually solve the underlying problem 😆.

PR #455, opened 2022-09-15
fix: bzlmod bazel-lib version not in sync with WORKSPACE

kormide:


kormide:
I should write up a github action that will check for these.

jbedard:
Sounds like this should be a renovate bot feature?

jbedard (approved):

PR #457, opened 2022-09-16
feat: supply a @pnpm//:pnpm binary

### LINKED ISSUE #247: Support migration without installing pnpm, node, etc. on host
I assume rules_js has access to node and pnpm binaries through bazel targets. Could `main/docs/migrate.md` be updated to mention the `bazel run` command that's equivalent to `pnpm install` based on a local pnpm installation?

Part of the beauty of bazel is there is often no need to install random development tools on a workstation, so I prefer `bazel run` wherever possible.

alexeagle:
This lets us document how to run Bazel's managed node and pnpm, without leaking out to the system node.

Requires a new feature on npm_import so that we can inject more content to the BUILD file.

Fixes #247

### DIFF
--- a/npm/private/npm_import.bzl
+++ b/npm/private/npm_import.bzl
@@ -415,6 +415,9 @@ def _impl(rctx):
         version = rctx.attr.version,
     ))
 
+    if rctx.attr.extra_build_content:
+        rctx_files["BUILD.bazel"].append("\n" + rctx.attr.extra_build_content)
+
     if bins:
         virtual_store_name = utils.virtual_store_name(rctx.attr.package, rctx.attr.version)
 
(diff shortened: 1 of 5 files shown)


alexeagle:
Note, this is following up from a TODO: https://github.com/aspect-build/rules_js/pull/377#issuecomment-1211151735

npm/private/npm_import.bzl
```diff
@@ -415,6 +415,8 @@ def _impl(rctx):
         version = rctx.attr.version,
     ))
 
+    rctx_files["BUILD.bazel"].append(rctx.attr.extra_build_content)
```
gregmagolan, line 418: perhaps
```
if rctx.attr.extra_build_content:
```
then append a new line first

gregmagolan (approved):
🌮

npm/private/npm_import.bzl
alexeagle: The _JS_PACKAGE_TMPL happens to end with newline, so we don't need one, but you're right we may as well be defensive here

alexeagle:
Also updated to allow `extra_build_content=["list", "of", "lines"]` like `write_file`

PR #458, opened 2022-09-16
fix: required for to use rules_js with Bazel HEAD without --allow_experimental_unresolved_symlinks

gregmagolan:
Testing against bazel HEAD this morning I ran into this failure building rules_js

```
ERROR: /Users/greg/aspect/rules/js/rules_js/BUILD.bazel:10:22: in npm_link_package_store rule //:node_modules/@types/node: 
	File "/Users/greg/aspect/rules/js/rules_js/npm/private/npm_link_package_store.bzl", line 94, column 31, in _impl
		files = utils.make_symlink(ctx, root_symlink_path, virtual_store_directory)
	File "/Users/greg/aspect/rules/js/rules_js/npm/private/utils.bzl", line 106, column 28, in _make_symlink
		ctx.actions.symlink(
Error in symlink: symlink() with "target_file" directory param requires that "output" be declared as a directory (did you mean to use declare_directory() instead of declare_file()?)
```

This is due to a BREAKING CHANGE in Bazel that requires you to use `declare_directory` if you're making a symlink to a directory with `ctx.actions.symlink` (and not using experimental unresolved symlinks `declare_symlink`).

Unfortunately, this breaking change is _not_ backward compatible as if I make the fix for Bazel head I break rules_js for Bazel 5.3.0:

```
ERROR: /Users/greg/aspect/rules/js/rules_js/BUILD.bazel:10:22: Creating symlink bazel-out/darwin-fastbuild/bin/node_modules/.aspect_rules_js/@registry.yarnpkg.com+inline-fixtures+-+inline-fixtures-1.1.0.tgz#5f9edf644ac02a1322012dbbb9f0a4739831d662/node_modules/tmp failed: failed to create symbolic link 'node_modules/.aspect_rules_js/@registry.yarnpkg.com+inline-fixtures+-+inline-fixtures-1.1.0.tgz#5f9edf644ac02a1322012dbbb9f0a4739831d662/node_modules/tmp' to 'node_modules/.aspect_rules_js/tmp@0.0.33/node_modules/tmp' due to I/O error: /private/var/tmp/_bazel_greg/7aed427a991d86f99332ab79b9b11780/execroot/aspect_rules_js/bazel-out/darwin-fastbuild/bin/node_modules/.aspect_rules_js/@registry.yarnpkg.com+inline-fixtures+-+inline-fixtures-1.1.0.tgz#5f9edf644ac02a1322012dbbb9f0a4739831d662/node_modules/tmp (File exists)
```

which means we also need a conditional on the Bazel version here.

### DIFF
--- a/npm/private/utils.bzl
+++ b/npm/private/utils.bzl
@@ -99,13 +99,23 @@ def _make_symlink(ctx, symlink_path, target_file):
         )
         files.append(target_file)
     else:
-        symlink = ctx.actions.declare_file(symlink_path)
+        if _is_at_least_bazel_6() and target_file.is_directory:
+            # BREAKING CHANGE in Bazel 6 requires you to use declare_directory if your target_file
+            # in ctx.actions.symlink is a directory artifact
+            symlink = ctx.actions.declare_directory(symlink_path)
+        else:
+            symlink = ctx.actions.declare_file(symlink_path)
         ctx.actions.symlink(
             output = symlink,
             target_file = target_file,
         )
     files.append(symlink)
     return files
+
+def _is_at_least_bazel_6():
+    # Hacky way to check if the we're using at least Bazel 6. Would be nice if there was a ctx.bazel_version instead.
+    # native.bazel_version only works in repository rules.
+    return "apple_binary" not in dir(native)
 
 utils = struct(
     bazel_name = _bazel_name,


alexeagle:
We can detect the bazel version at runtime and just be conditional on which behavior we use, right?

gregmagolan:
> We can detect the bazel version at runtime and just be conditional on which behavior we use, right?

Yup. Your check on native worked.

PR #475, opened 2022-09-26
feat: enable npm authentication via .npmrc

pedrobarco:
This PR enables users to authenticate with npm via `.npmrc` and download packages from private registries.

Special thanks to @perarnet for sharing a snippet in bazel's slack channel, which inspired this new approach.


npm/private/npm_import.bzl
```diff
@@ -689,8 +690,35 @@ _ATTRS = dicts.add(_COMMON_ATTRS, {
...
     "url": attr.string(),
+    "npmrc": attr.label(),
 })
 
+def _get_npm_auth(rctx, download_url):
+    # Read token from npmrc label
+    if rctx.attr.npmrc:
+        npmrc_path = rctx.path(rctx.attr.npmrc)
+        npmrc = rctx.read(npmrc_path).splitlines()
+        for line in npmrc:
```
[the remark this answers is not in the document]
gregmagolan: Hmm. One problem with doing this load and parsing logic in `npm_import` and passing the `npmrc` file forward from `npm_translate_lock` is that every `npm_import` will do duplicate work. Probably better to do the `parse_ini` in `npm_translate_lock` and have `npm_import` take an `npm_auth` attribute that takes the dict that this function returns. You could use a string attribute and pass the `json.encode`d representation of the dict to `npm_import` and then just `json.decode` in that repository rule.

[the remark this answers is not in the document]
pedrobarco: Awesome, I have followed your approach but I was not able to use the `json.encode`d dict since it was not useful to have it like so in `npm_import` because I would have to parse the token value again if it was an environment variable (duplicate work). So I've managed to refactor this attribute to be a string and just pass the "final" token value to `npm_import`.

npm/private/npm_translate_lock.bzl
```diff
@@ -408,6 +408,7 @@ def _impl(rctx):
     root_package = None
     link_workspace = None
     lockfile_description = None
+    npmrc = rctx.attr.npmrc
```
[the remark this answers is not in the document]
pedrobarco: Done! Let me know if it's clear enough 😄 d2181f2

npm/private/npm_translate_lock.bzl
```diff
@@ -131,6 +132,27 @@ def _gather_values_from_matching_names(keyed_lists, *names):
...
+            # Parse environment variable from config
+            token = npmrc[_NPM_TOKEN_KEY]
+
+            # A token can be a reference to an environment variable
+            if token.startswith("$"):
+                # ${NPM_TOKEN} -> NPM_TOKEN
+                # $NPM_TOKEN -> NPM_TOKEN
+                token = token.removeprefix("$").removeprefix("{").removesuffix("}")
+                if token in rctx.os.environ.keys() and rctx.os.environ[token]:
+                    token = rctx.os.environ[token]
```
gregmagolan, line 153: should this fail if the environment variable does not exist? what does npm do in that case?

npm/private/npm_import.bzl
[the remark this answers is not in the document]
gregmagolan: nice. that's great.

npm/private/npm_translate_lock.bzl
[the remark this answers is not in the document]
gregmagolan: lgtm

npm/private/npm_translate_lock.bzl
pedrobarco: I'd say it shouldn't, just like npm. I've tried running `npm`, `yarn` and `pnpm` with no `${NPM_TOKEN}`:

### npm

```
$ npm I
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/<my-private-package> - Not found
npm ERR! 404 
npm ERR! 404  '<my-private-package>' is not in this registry.
```
`npm` only fails when downloading a dependency from a private registry without providing authentication so it doesn't fail when replacing the environment variable. So, if I'm working on a package that does not have to download private dependencies or if I already have those dependencies cached I shouldn't need to fail, as it works as expected.

### yarn

```
$ yarn
error An unexpected error occurred: "Failed to replace env in config: ${NPM_TOKEN}".
```
`yarn` fails if the environment variable doesn't exist. I don't think this is the right approach, as explained above.

### pnpm

```
$ bazel run -- @pnpm//:pnpm -C $PWD
 WARN  Issue while reading "/path/to/my/.npmrc". Failed to replace env in config: ${NPM_TOKEN}
...
 ERR_PNPM_FETCH_404  GET https://registry.npmjs.org/<my-private-package>: Not Found - 404
```
`pnpm` does exactly the same as `npm` but provides the same warning as `yarn`. Essentially it's a mix of both approaches.

### rules_js

```
$ bazel build //<private-package-test>
ERROR: /Users/pbc02/repos/<my-repo>/BUILD.bazel:7:22: no such package '<my-private-package>':
java.io.IOException: Error downloading [https://registry.npmjs.org/<my-private-pacakge>] to /private/var/tmp/_bazel_pbc02/8c60ffac9d59542e1012c70fa68cf734/external/<my-private-package>: GET returned 404 Not Found
```
`rules_js` does exactly the same as `npm`. So I'd say we're on the right track. 
What do you think we should do 😄?

gregmagolan: Cool. Thanks for checking all three.

Since rules_js is tied closely to pnpm lets print out a warning with similar text. 

```
WARNING: Issue while reading "/path/to/my/.npmrc". Failed to replace env in config: ${NPM_TOKEN}.
```

Seems useful for users to see if the build fails due to authentication because of the missing environment variable.

We have a few spots where we do this that you can use as a reference,

https://github.com/aspect-build/rules_js/blob/ffed0b0a827868180f8df9e4eb9ba6d8d792c390/js/private/js_run_binary.bzl#L267

https://github.com/aspect-build/rules_js/blob/ffed0b0a827868180f8df9e4eb9ba6d8d792c390/npm/private/npm_translate_lock.bzl#L268

Don't think this warning requires a disable flag.

pedrobarco: Fixed with ea9a75f

gregmagolan (approved):
🌮

npm/private/npm_translate_lock.bzl
```diff
@@ -131,6 +132,34 @@ def _gather_values_from_matching_names(keyed_lists, *names):
...
 
+def _get_npm_auth(rctx):
+    _NPM_TOKEN_KEY = "//registry.npmjs.org/:_authtoken"
+    token = None
+
+    # Read token from npmrc label
+    if rctx.attr.npmrc:
+        npmrc_path = rctx.path(rctx.attr.npmrc)
+        npmrc = parse_ini(rctx.read(npmrc_path))
+        if _NPM_TOKEN_KEY in npmrc:
```
dancmeyers, line 143: Unless I'm reading something very wrong, this only allows authentication to the npmjs registry itself. What about arbitrary private repositories like AWS CodeArtifact or similar?

pedrobarco: Yap, unfortunately this only allows authentication to the npmjs registry. 
Do you have any suggestion on how we could support 3rd party registries and private repositories? It has to be agnostic to the registry so I'm worried that we would need to calculate the appropriate token for each package rather than having a single token for all packages.

pedrobarco: I've created #499 to address this issue. 
Let me know what you think about it.

PR #482, opened 2022-09-30
feat: add missing fixed_env attribute to create_launcher helper for downstream rules to use

gregmagolan:
@mrmeku's work on rules_cypress showed that there is a need for this

The API change here is not breaking and only affects downstream rules


js/private/js_binary.bzl
```diff
@@ -227,9 +228,9 @@ _NODE_OPTION = """JS_BINARY__NODE_OPTIONS+=(\"{value}\")"""
 def _target_tool_short_path(path):
     return ("../" + path[len("external/"):]) if path.startswith("external/") else path
 
-def _bash_launcher(ctx, entry_point_path, log_prefix_rule_set, log_prefix_rule, fixed_args, is_windows):
+def _bash_launcher(ctx, entry_point_path, log_prefix_rule_set, log_prefix_rule, fixed_args, fixed_env, is_windows):
     envs = []
-    for (key, value) in ctx.attr.env.items():
+    for (key, value) in dicts.add(fixed_env, ctx.attr.env).items():
```
jbedard, line 233: I think `fixed_env` is a reference, so this will modify the `fixed_env`? Or does the static `dicts.add` clone it?

jbedard: Oh nm, I found the docs 👍

jbedard (approved):

Issue #483, opened 2022-09-30
npm_import and lifecycle-hooks occasional fail due to package.json short-reads


jbedard:
For example:
```
SyntaxError: Unexpected end of JSON input
    at makeBins (/home/circleci/.cache/bazel/_bazel_circleci/.../bin/external/aspect_rules_js/npm/private/lifecycle/lifecycle-hooks.sh.runfiles/aspect_rules_js/npm/private/lifecycle/min/index.min.js:73:42097)
    at async main (/home/circleci/.cache/bazel/_bazel_circleci/.../bin/external/aspect_rules_js/npm/private/lifecycle/lifecycle-hooks.sh.runfiles/aspect_rules_js/npm/private/lifecycle/min/index.min.js:73:43024)
    at async /home/circleci/.cache/bazel/_bazel_circleci/.../bin/external/aspect_rules_js/npm/private/lifecycle/lifecycle-hooks.sh.runfiles/aspect_rules_js/npm/private/lifecycle/min/index.min.js:73:44054
```

Kartones:
Also seeing the issue with v1.4.0, despite the retry introduced in that release:
```
Error parsing @swc/core-win32-arm64-msvc/package.json: SyntaxError: Unexpected end of JSON input

""""

""""
Retrying...
Error: Error parsing @swc/core-win32-arm64-msvc/package.json: SyntaxError: Unexpected end of JSON input

""""

""""
at makeBins ([redacted]/aspect_rules_js/npm/private/lifecycle/lifecycle-hooks.sh.runfiles/aspect_rules_js/npm/private/lifecycle/min/index.min.js:73:42314)
at async makeBins ([redacted]/aspect_rules_js/npm/private/lifecycle/lifecycle-hooks.sh.runfiles/aspect_rules_js/npm/private/lifecycle/min/index.min.js:73:41777)
at async main ([redacted]/aspect_rules_js/npm/private/lifecycle/lifecycle-hooks.sh.runfiles/aspect_rules_js/npm/private/lifecycle/min/index.min.js:73:43267)
at async [redacted]/aspect_rules_js/npm/private/lifecycle/lifecycle-hooks.sh.runfiles/aspect_rules_js/npm/private/lifecycle/min/index.min.js:73:44297
```

It doesn't looks as package dependant, seen it 3 times this week each with different packages.

gregmagolan:
I've seen it locally on macos as well. What is strange is that it looks like in your logs above
```
[...]

PR #484, opened 2022-09-30
fix: retry package.json load+parse on error, log more detailed error message

jbedard:
See https://github.com/aspect-build/rules_js/issues/483


npm/private/lifecycle/lifecycle-hooks.js
```diff
@@ -34,9 +34,29 @@ async function makeBins(nodeModulesPath, scope, segmentsUp) {
             'package.json'
         )
         if (fs.existsSync(packageJsonPath)) {
-            const packageJson = JSON.parse(
-                await fs.promises.readFile(packageJsonPath)
-            )
+            let packageJson = ''
+
+            // Catch invalid JSON errors and throw a more detailed Error.
+            // See bad quotes in puppeteer@8.0.0 for an example of bad json, or
```
gregmagolan, line 40: Were the bad qoutes just a red herring and the actual failure was the short read?

gregmagolan (approved):

npm/private/lifecycle/lifecycle-hooks.js
jbedard: I'm not 100% sure, so maybe I'll just drop this comment for now.

PR #485, opened 2022-10-01
chore: configure renovate to group PRs

jbedard:
The main thing is group all non-major updates into single PRs.

I'm still tempted to group updates into smaller PRs instead of one giant one. Things like "all npm dev dependencies", "all npm non-dev dependencies", "all bazel deps", "all java deps" etc. Is that better then a single PR for everything? Or is less PRs better no matter what gets mixed in?

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,12 +1,22 @@
 {
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
 
-    "extends": ["config:base", ":preserveSemverRanges", "schedule:weekly"],
+    "extends": [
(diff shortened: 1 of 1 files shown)


thesayyn:
> I'm still tempted to group updates into smaller PRs instead of one giant one.

this might be better but let's try the giant one first.

thesayyn (approved):
LGTM

cgrindel (approved):

jbedard:
I've added the "cleanup" and "prioritize" labels.

Issue #155, opened 2022-06-03
Demonstrate how to publish to npm


alexeagle:
Should be able to build a package and publish it to some registry

gregmagolan:
Should be fairly easy to add a `.publish` run target to `npm_package` (same as pkg_npm has in rules_nodejs).

To make it more useful it should come with stamping support. The shape of `stamping` in `npm_package` needs a little thought

alexeagle:
yeah I keep wanting to do something with stamping - but really the minimum necessary thing is just a `jq` filter to replace the `version` in package.json, so maybe we want to add stamping support to the `jq` rule in aspect_bazel_lib and then at most we'd need to make `npm_package` macro expand to include one of those

alexeagle:
Stamping of package.json is done so this is unblocked for someone

alexeagle:
note from one user, if you `pnpm publish` in the output tree today, the .git is tree directories up and still find by pnpm when it check for a dirty git state. You can use `--no-git-checks` to avoid that.

gregmagolan:
> how to ship this change into the [official doc](https://docs.aspect.build/rules/aspect_rules_js/docs/npm_package)? it is still about rules_js v1.26.x while I saw this feature is available in v1.30.0

I'll put up a PR to the docsite to update to the latest version of rules_js. One day we'll have some more automation for that :)

PR #498, opened 2022-10-06
feat: add include_npm attribute to js_binary & js_test

gregmagolan:
Some tools (such as vsce) assume npm is on the PATH (https://github.com/angular/vscode-ng-language-service/pull/1754).

npm on the PATH will also make implementing https://github.com/aspect-build/rules_js/issues/155 much easier than the rules_nodejs equivalent in pkg_npm.


alexeagle (approved):
🌮

PR #499, opened 2022-10-06
(OLD) feat: support multiple npm registries and tokens via .npmrc

pedrobarco:
This PR adds the ability to support multiple registries and auth tokens via `.npmrc`

### DIFF
--- a/npm/private/npm_import.bzl
+++ b/npm/private/npm_import.bzl
@@ -345,7 +345,8 @@ _PACKAGE_JSON_BZL_FILENAME = "package_json.bzl"
 def _impl(rctx):
     # scoped packages contain a slash in the name, which doesn't appear in the later part of the URL
     package_name_no_scope = rctx.attr.package.rsplit("/", 1)[-1]
-    download_url = rctx.attr.url
+    download_url = rctx.attr.url if rctx.attr.url else utils.npm_registry_download_url(rctx.attr.package, rctx.attr.version)
+
     auth = {
         download_url: {
             "type": "pattern",
--- a/npm/private/npm_import.bzl
+++ b/npm/private/npm_import.bzl
@@ -345,12 +345,7 @@ _PACKAGE_JSON_BZL_FILENAME = "package_json.bzl"
 def _impl(rctx):
     # scoped packages contain a slash in the name, which doesn't appear in the later part of the URL
     package_name_no_scope = rctx.attr.package.rsplit("/", 1)[-1]
-    download_url = rctx.attr.url if rctx.attr.url else "{0}{1}/-/{2}-{3}.tgz".format(
-        utils.npm_registry_url,
-        rctx.attr.package,
-        package_name_no_scope,
-        utils.strip_peer_dep_version(rctx.attr.version),
-    )
+    download_url = rctx.attr.url
     auth = {
         download_url: {
             "type": "pattern",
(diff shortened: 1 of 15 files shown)


npm/private/npm_translate_lock.bzl
```diff
@@ -766,6 +780,11 @@ load("@aspect_rules_js//npm/private:npm_package_store.bzl", _npm_package_store =
         maybe_bins = ("""
         bins = %s,""" % starlark_codegen_utils.to_dict_attr(_import.bins, 2)) if len(_import.bins) > 0 else ""
 
+        if _import.package.startswith("@"):
+            scope = _import.package.split("/")[0]
```
gregmagolan, line 784: nit: you can specify maxsplit as the 2nd param
```
scope = _import.package.split("/", 1)[0]
```
if you know ahead of time you only ever want to split once

npm/private/npm_translate_lock.bzl
```diff
@@ -133,32 +133,47 @@ def _gather_values_from_matching_names(keyed_lists, *names):
     return result
 
 def _get_npm_auth(rctx):
-    _NPM_TOKEN_KEY = "//registry.npmjs.org/:_authtoken"
-    token = None
+    _NPM_TOKEN_KEY = ":_authtoken"
```
gregmagolan, line 136: nice. this looks right. can you add a decent docstring to this function with some examples of what these values might look like in `.npmrc`? will help the next person who touches this code

npm/private/npm_translate_lock.bzl
```diff
@@ -766,6 +823,11 @@ load("@aspect_rules_js//npm/private:npm_package_store.bzl", _npm_package_store =
         maybe_bins = ("""
         bins = %s,""" % starlark_codegen_utils.to_dict_attr(_import.bins, 2)) if len(_import.bins) > 0 else ""
 
+        if _import.package.startswith("@"):
+            scope = _import.package.split("/", 1)[0]
+            maybe_npm_auth = ("""
+        npm_auth = "%s",""" % npm_auth[scope]) if scope in npm_auth else ""
```
gregmagolan, line 829: This looks like it will break for npm packages that don't have a scope. If the package name doesn't start with `@` the scope will be undefined.

Thinking about this some more, it seems more robust to not even consider scope at this point and just consider the registry. If you return a `registry: token` dict from `_get_npm_auth` instead then you can just check if the registry `url` of the npm_import has a token to use. `url` might not be set, in which case you can assume it default to the npm registry `https://registry.npmjs.org` (which it does in private/npm_import.bzl)

npm/private/npm_import.bzl
```diff
@@ -345,11 +345,7 @@ _PACKAGE_JSON_BZL_FILENAME = "package_json.bzl"
 def _impl(rctx):
     # scoped packages contain a slash in the name, which doesn't appear in the later part of the URL
     package_name_no_scope = rctx.attr.package.rsplit("/", 1)[-1]
-    download_url = rctx.attr.url if rctx.attr.url else "https://registry.npmjs.org/{0}/-/{1}-{2}.tgz".format(
-        rctx.attr.package,
-        package_name_no_scope,
-        utils.strip_peer_dep_version(rctx.attr.version),
-    )
+    download_url = rctx.attr.url
```
gregmagolan, line 348: this needs to be
```
download_url = rctx.attr.url if rctx.attr.url else utilsnpm_registry_download_url(rctx.attr.package, rctx.attr.version)
```
for the cases where url is not set explicitely

PR #500, opened 2022-10-07
chore: add e2e/npm_translate_lock_auth e2e test for coverage of new .npmrc auth token feature

gregmagolan:
With auth token not set, e2e will fail with
```
WARNING: Download from https://registry.npmjs.org/@aspect-priv-npm/a/-/a-1.0.0.tgz failed: class java.io.FileNotFoundException GET returned 404 Not Found
ERROR: An error occurred during the fetch of repository 'npm__at_aspect-priv-npm_a__1.0.0':
   Traceback (most recent call last):
	File "/home/runner/.cache/bazel/_bazel_runner/79bc8f71d716975c4b639fae2dec5381/external/aspect_rules_js/npm/private/npm_import.bzl", line 362, column 18, in _impl
		rctx.download(
Error in download: java.io.IOException: Error downloading [https://registry.npmjs.org/@aspect-priv-npm/a/-/a-1.0.0.tgz] to /home/runner/.cache/bazel/_bazel_runner/79bc8f71d716975c4b639fae2dec5381/external/npm__at_aspect-priv-npm_a__1.0.0/package.tgz: GET returned 404 Not Found
ERROR: /home/runner/work/rules_js/rules_js/e2e/npm_translate_lock_auth/WORKSPACE:29:17: fetching _npm_import rule //external:npm__at_aspect-priv-npm_a__1.0.0: Traceback (most recent call last):
	File "/home/runner/.cache/bazel/_bazel_runner/79bc8f71d716975c4b639fae2dec5381/external/aspect_rules_js/npm/private/npm_import.bzl", line 362, column 18, in _impl
		rctx.download(
Error in download: java.io.IOException: Error downloading [https://registry.npmjs.org/@aspect-priv-npm/a/-/a-1.0.0.tgz] to /home/runner/.cache/bazel/_bazel_runner/79bc8f71d716975c4b639fae2dec5381/external/npm__at_aspect-priv-npm_a__1.0.0/package.tgz: GET returned 404 Not Found
[...]

PR #501, opened 2022-10-07
chore: add minimal root package workspace files to e2e tests so that you can run pnpm install in them

gregmagolan:
pnpm does a funny thing when you call pnpm install. If it finds a workspace file anywhere up the tree then it does the install for that workspace instead of your current working directory. Simple workspace file with just the root package in subdirectories solves the problem.

lock file generated when there isn't a workspace.yaml and when there is one with just a . package is identical. No workspace file basically implies a workspaces with a single importer at the root.

See discussion here https://github.com/aspect-build/rules_js/pull/500#discussion_r990489875 for more context


alexeagle:
I looked around the pnpm docs and can't find any way to inhibit that wrong path search behavior either...

alexeagle (approved):

PR #502, opened 2022-10-08
DNL: add @aspect-build/a npm package from private GitHub registry to e2e/npm_translate_lock_auth

gregmagolan:
DNL. This will get integrated into another PR


gregmagolan:
Landed with https://github.com/aspect-build/rules_js/pull/503

PR #503, opened 2022-10-08
feat: support multiple npm registries and tokens via .npmrc

pedrobarco:
This PR adds the ability to support multiple registries and auth tokens via .npmrc.

This branch was cloned from #499 and addresses #502. 

It is being developed in an upstream branch because the e2e test won't run on my fork since it requires a secret which isn't available for forks.

### DIFF
--- a/npm/private/npm_translate_lock.bzl
+++ b/npm/private/npm_translate_lock.bzl
@@ -171,7 +171,7 @@ def get_npm_auth(npmrc, npmrc_path, environ):
         environ: A map of environment variables with their values.
 
     Returns:
-        A list with a tokens dict and a registries dict.
+        A tuple with a tokens dict and a registries dict.
     """
 
     _NPM_TOKEN_KEY = ":_authtoken"
@@ -338,7 +338,7 @@ def _gen_npm_imports(lockfile, root_package, attr, registries = {}):
                 if not registry:
                     (scope, _) = utils.parse_package_name(name)
                     registry = "https://{}".format(registries[scope]) if scope in registries else utils.npm_registry_url
-                url = registry + tarball
+                url = "{0}/{1}".format(registry.removesuffix("/"), tarball)
 
         result.append(struct(
             custom_postinstall = custom_postinstall,
--- a/npm/private/npm_translate_lock.bzl
+++ b/npm/private/npm_translate_lock.bzl
@@ -133,6 +133,43 @@ def _gather_values_from_matching_names(keyed_lists, *names):
     return result
 
 def _get_npm_auth(rctx):
+    """Parses npm tokens from `.npmrc`.
+
+    - creates a tokens dict: {registry: token}
+    - creates a scopes dict: {scope: registry}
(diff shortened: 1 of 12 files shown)


npm/private/npm_translate_lock.bzl
```diff
@@ -277,36 +330,14 @@ def _gen_npm_imports(lockfile, root_package, attr):
...
-re-generating the lockfile should generate a fully qualified tarball URL for this package.
-
-To disable this warning, set `warn_on_unqualified_tarball_url` to False in your
-`npm_translate_lock` repository rule.
-====================================================================================================
-
-""".format(name, version, tarball))
+                    (scope, _) = utils.parse_package_name(name)
+                    registry = "https://{}".format(registries[scope]) if scope in registries else utils.npm_registry_url
                 url = registry + tarball
```
gregmagolan, line 341: should make sure that `registry` ends with `/` here when concatenating the strings `url = registry + tarball` or else inject `/` separator if between them if it doesn't

npm/private/utils.bzl
```diff
@@ -117,6 +117,26 @@ def _is_at_least_bazel_6():
     # native.bazel_version only works in repository rules.
     return "apple_binary" not in dir(native)
 
+def _parse_package_name(package):
+    # Parse a @scope/name string and return a [scope, name] list
```
gregmagolan, line 121: nit: "return a (scope, name) tuple"

npm/private/npm_translate_lock.bzl
```diff
@@ -132,35 +133,87 @@ def _gather_values_from_matching_names(keyed_lists, *names):
...
-        if _NPM_TOKEN_KEY in npmrc:
-            # Parse environment variable from config
-            token = npmrc[_NPM_TOKEN_KEY]
+    Args:
+        npmrc: The `.npmrc` file.
+        npmrc_path: The file path to `.npmrc`.
+        environ: A map of environment variables with their values.
+
+    Returns:
+        A list with a tokens dict and a registries dict.
```
gregmagolan, line 174: nit: A tuple with a tokens dict and a registries dict

gregmagolan:
Looks great. Just a few minor comments and this is good to land. Will cut a minor release after it does.

npm/private/utils.bzl
pedrobarco: My bad, I've had it like you suggested but I found a similar function that was using the same description: https://github.com/aspect-build/rules_js/blob/724c377bfed9db82881b4174b9652700f9c426c7/npm/private/utils.bzl#L33-L39

Should I change its description too?

npm/private/npm_translate_lock.bzl
pedrobarco: Same as above. On it ⛵

npm/private/utils.bzl
gregmagolan: Yes please. Nice catch. Looks like that one got out of sync.

gregmagolan (approved):
🌮 🚀

PR #505, opened 2022-10-08
chore: deprecate unused npm_translate_lock warn_on_unqualified_tarball_url param

gregmagolan:


alexeagle:
Oops we never wired it up?

alexeagle (approved):

gregmagolan:
> Oops we never wired it up?

It was wired but got removed in #503. Turns out it is not just older versions of pnpm that leave unqualified tarball URLs. Latest version does as well and we deal with it appropriately now.

PR #506, opened 2022-10-08
fix: js_image bazel lib

thesayyn:


gregmagolan (approved):
What changed?

thesayyn:
> What changed?

for some reason I couldn't quite understand, the regex does not work. looks like a false positive.

PR #530, opened 2022-10-18
fix: support generic npmrc registry config

jbedard:
This means we can no longer set the default registry at the time we write the files, instead we have to set it at the time we read the `.npmrc` and pass it throughout all the methods.


gregmagolan (approved):
🌮

PR #535, opened 2022-10-20
fix: fix async/await logic in lifecycle hooks copyPackageContents

gregmagolan:
🤦 Fixes https://github.com/aspect-build/rules_js/issues/483


npm/private/lifecycle/lifecycle-hooks.js
```diff
@@ -151,12 +151,14 @@ async function main(args) {
...
 async function copyPackageContents(packageDir, destDir) {
     const contents = await fs.promises.readdir(packageDir)
-    contents.forEach(async (file) => {
-        await copyRecursive(
-            path.join(packageDir, file),
-            path.join(destDir, file)
-        )
-    })
+    await Promise.all(
+        contents.map(async (file) => {
```
jbedard, line 155: I think we can just do:
```
await Promise.all(contents.map(file => copyRecursive(...)))
```
Since `Promise.all` handles all the async we just need to pass an array of promises...

gregmagolan: Nice catch 👍

jbedard (approved):

PR #536, opened 2022-10-20
fix: remove package.json parsing retry logic

jbedard:
I've kept the try/catch with the better error message for now. Think it's worth keeping that?


npm/private/lifecycle/lifecycle-hooks.js
```diff
@@ -34,26 +34,16 @@ async function makeBins(nodeModulesPath, scope, segmentsUp) {
...
-                packageJson = JSON.parse(json)
+                packageJson = JSON.parse(
+                    await fs.promises.readFile(packageJsonPath)
+                )
             } catch (e) {
-                console.error(
-                    `Error parsing ${packageName}/package.json: ${e}\n\n""""\n${json}\n""""\nRetrying...`
+                // Catch and throw a more detailed error message.
+                throw new Error(
+                    `Error reading ${packageName}/package.json: ${e}\n\n""""\n${json}\n""""`
```
gregmagolan, line 45: json will be undefined now

gregmagolan: You can keep the readFile out of the try/catch, it will throw and the error is sound.
try/catch around the json.parse is sufficient. Perhaps change the error message to `Error parsing`?

gregmagolan (approved):
🌮

PR #538, opened 2022-10-21
fix: make js_binary more portable by removing dependency on bazel_tools/tools/bash/runfiles/runfiles.bash

gregmagolan:


js/private/utils.bzl
```diff
@@ -0,0 +1,115 @@
+"Utility functions for js rules"
```
alexeagle, line 1: utils.bzl is a slipperly slope to containing anything. How about bash.bzl?

alexeagle: or, does it make sense for it to live near the BASH_RLOCATION_FUNCTION function in bazel_lib?

alexeagle (approved):

js/private/utils.bzl
gregmagolan: bash.bzl sgtm; this snippet is not fully generalized yet as it depends on a logf_fatal function being available. could be hoisted to bazel-lib in the future once generalized all the way

PR #540, opened 2022-10-21
feat: add js_devserver rule

gregmagolan:


js/private/js_binary.bzl
```diff
@@ -261,6 +261,7 @@ def _bash_launcher(ctx, entry_point_path, log_prefix_rule_set, log_prefix_rule,
         "JS_BINARY__TARGET_CPU": "$(TARGET_CPU)",
         "JS_BINARY__TARGET": "$(TARGET)",
         "JS_BINARY__WORKSPACE": "$(WORKSPACE)",
+        "JS_BINARY__PACKAGE": ctx.label.package,
```
alexeagle, line 264: possible that it's simpler to parse this from JS_BINARY__TARGET at the use-site?

js/private/js_devserver.bzl
```diff
@@ -0,0 +1,91 @@
+"Implementation details for js_devserver rule"
```
alexeagle, line 1: I guess you're still working on this - rename to js_run_devserver everywhere?

js/private/js_devserver.js
```diff
@@ -0,0 +1,161 @@
+const path = require('path')
+const fs = require('fs')
+const os = require('os')
+const child_process = require('child_process')
+
+const RUNFILES_ROOT = path.join(
+    process.env.RUNFILES,
```
alexeagle, line 7: JS_BINARY__RUNFILES?

js/private/js_devserver.js
```diff
@@ -0,0 +1,161 @@
...
+            )
+            await mkdirp(path.dirname(dst))
+            if (exists) {
+                await fs.promises.unlink(dst)
+            }
+            await fs.promises.symlink(src, dst)
+        } else if (lstat.isDirectory()) {
+            await mkdirp(dst)
+            const contents = await fs.promises.readdir(src)
+            await Promise.all(
```
alexeagle, line 50: seems like this code is more synchronous than required?

js/private/js_devserver.js
```diff
@@ -0,0 +1,161 @@
...
+            console.log(
+                `Syncing symlink ${src.slice(RUNFILES_ROOT.length + 1)}`
+            )
+            await mkdirp(path.dirname(dst))
+            if (exists) {
+                await fs.promises.unlink(dst)
+            }
+            await fs.promises.symlink(src, dst)
+        } else if (lstat.isDirectory()) {
+            await mkdirp(dst)
```
jbedard, line 48: I think this can be dropped since the recursive call will do it.

js/private/js_devserver.js
```diff
@@ -0,0 +1,161 @@
...
+            const contents = await fs.promises.readdir(src)
+            await Promise.all(
+                contents.map((entry) =>
+                    syncRecursive(path.join(src, entry), path.join(dst, entry))
+                )
+            )
+        } else {
+            console.log(`Syncing file ${src.slice(RUNFILES_ROOT.length + 1)}`)
+            await mkdirp(path.dirname(dst))
```
jbedard, line 57: This could be in an `else` and only done if the file doesn't already exist? Probably same with the symlink case.

js/private/js_devserver.js
```diff
@@ -0,0 +1,161 @@
...
+        process.exit(1)
+    }
+}
+
+async function sync(files, rootdir, sync = false) {
+    for (const file of files) {
+        const src = path.join(RUNFILES_ROOT, file)
+        const dst = path.join(rootdir, file)
+        if (sync) {
+            await syncRecursive(src, dst)
```
jbedard, line 76: I think this can be done concurrently for all `files`?

```
const syncedFiles = files.map(file = > syncRecurse(
    path.join(RUNFILES_ROOT, file),
    path.join(rootdir, file)
);

if (sync) {
   await Promise.all(syncedFiles);
}
```
?

js/private/js_devserver.js
```diff
@@ -0,0 +1,161 @@
...
+
+        proc.on('close', (code) => {
+            console.error(`child tool process exited with code ${code}`)
+            resolve()
+            process.exit(code)
+        })
+
+        process.stdin.on('data', (chunk) => {
+            const chunkString = chunk.toString()
+            if (chunkString.indexOf('IBAZEL_BUILD_COMPLETED SUCCESS') !== -1) {
```
jbedard, line 124: What does the string look like such that we can't use `===`?

Could use `chunkString.includes('...')`

js/private/js_run_devserver.js
```diff
@@ -0,0 +1,165 @@
...
+            resolve()
+            process.exit(code)
+        })
+
+        process.stdin.on('data', (chunk) => {
+            const chunkString = chunk.toString()
+            if (chunkString.indexOf('IBAZEL_BUILD_COMPLETED SUCCESS') !== -1) {
+                console.log('IBAZEL_BUILD_COMPLETED SUCCESS')
+                // Re-parse the config file to get the latest list of data files to copy
+                sync(
```
jbedard, line 129: Since we aren't doing `await`... what if another ibazel `IBAZEL_BUILD_COMPLETED` comes in while we are in the middle of a sync? Then we'll have 2 concurrent syncs that might conflict?

js/private/js_run_devserver.js
```diff
@@ -0,0 +1,165 @@
...
+
+async function syncRecursive(src, dst) {
+    try {
+        const lstat = await fs.promises.lstat(src)
+        const last = synced.get(src)
+        if (!lstat.isDirectory() && last && lstat.mtimeMs == last) {
+            // this file is already up-to-date
+            return
+        }
+        synced.set(src, lstat.mtimeMs)
```
jbedard, line 36: This never gets cleaned so it will grow forever if files are added/removed. Do we even need it? It seems like it just avoids an `lstat(dst)` which can replace the `existsSync` below anyway...?

```
const [srcLstat, destLstat] = await Promise.all([
   fs.promises.lstat(src),
   fs.promises.lstat(dst).catch(() => null), //<-- .catch() required for file not existing case?
])
```
if (!srcLstat.isDirectory() && destLstat && srcLstat.mtimeMs === destLstat.mtimeMs) return;

const exists = !!destLstat;

js/private/js_devserver.bzl
gregmagolan: yup. about to push that change

js/private/js_binary.bzl
gregmagolan: Personally I prefer to have in easily available for all rules in an env variable. I'll hoist the commits to change the envs into its own PR and we can discuss the merits there.

js/private/js_devserver.js
gregmagolan: 👍

js/private/js_devserver.js
gregmagolan: Yeah. I need to be initially synchronous for the initial sync but then it can be more asynchronous for incremental syncs. I'll think about how to make that work.

js/private/js_devserver.js
```diff
@@ -0,0 +1,161 @@
...
+            }
+        })
+    })
+}
+
+;(async () => {
+    let rootdir
+    try {
+        rootdir = path.join(
+            await fs.promises.mkdtemp(path.join(os.tmpdir(), 'js_devserver-')),
```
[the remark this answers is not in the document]
gregmagolan: only tests get one

js/private/js_devserver.js
gregmagolan: I don't think there is any guarantee that the chunks in stdin stream won't contain multiple messages from ibazel

js/private/js_run_devserver.js
gregmagolan: It is bounded by the set of files in config.data_files. We do need the map of file => mtimeMS to know which files have changed since the last sync. On every sync we go through the full set of data_files and copy over any files that have changed or are new since the last sync.

js/private/js_devserver.js
jbedard: Too synchronous because it does one level of the tree at a time before recursing a level deeper? 🤔

js/private/js_run_devserver.js
jbedard: Can't we just add `async` here so each `IBAZEL_BUILD_COMPLETED` waits for the previous to complete?

gregmagolan: Yeah. That is probably best and simples. And then figure out how to do any many parallel promises as possible in sync itself.

js/private/js_run_devserver.js
jbedard: But `config.data_files` can continue to change and `synced` never removes anything. If I just rename a file the old one will never be removed, although that might be small enough who cares? 🤷

jbedard: > We do need the map of file => mtimeMS to know which files have changed since the last sync.

Can't we do `lstat` to find which files have changed since the last sync? It gives you all the same information but avoids a) the map, b) the `existsSync` down a line. However that `lstat` would then be *before* the `if (changed)` where the `existsSync` currently is *after*.

gregmagolan: Yes, technically if you keep adding files to your BUILD graph then it grows unbounded... tho it is only during the course of an ibazel session and your output tree would grow unbounded as well in that case. It wouldn't have hard to check what files are no longer present and unlink them from the tmpdir and remove then from the set tho. Might add that before landing for completeness. I wonder if the runfiles tree also grows unbounded? Assuming it does and it is not wiped clean between builds.

gregmagolan: Do you mean we can check the timestamp from lstat against the time of the last sync for all files? There are subtle race conditions with that approach since a sync goes over a timespan from the first copy to the last so its trick to get a timestamp for a sync that works for all files.

js/private/js_run_devserver.js
jbedard: Actually if I'm recalling the `data_files` correctly I don't think we even need recursion and can have one giant `Promise.all`?
```
function sync(...) { return Promise.all(files.map(... syncRecursive(src, dst)) }

function syncRecursive(src, dst) {
   ...if/lstat/check-cache...

   await mkdirp(path.dirname(dst));
   if (symlink) ... link ...
   else if (file) ... copy/chmod ...
   else if (directory) ... mkdir() ...    // <--- no "p"
}
```

Note that the results of `syncRecursive` are put in `Promise.all` and not individual `await`s so it should be 100% parallel?

js/private/js_run_devserver.js
jbedard: > I wonder if the runfiles tree also grows unbounded?

I'd hope it cleans deleted files, otherwise it's not properly sandboxed?

> Might add that before landing for completeness.

It might be true that it grows forever but it's also possible it's not worth doing anything. With what I suggested we wouldn't need it, with the current approach it might not be worth it just to save some tiny map entries in a process that probably won't live long enough for it to matter?

> Do you mean we can check the timestamp from lstat against the time of the last sync for all files?

It would be checking the lastmod time of the src + dst at the time of copy. The lastmod of the dst should be the same as what's in the map... unless a different process touched it.

gregmagolan: Ohhh. I see the source of confusion. The `lstat.mtimeMs` of the dst is the time the copy is made and will never be the same as the `lstat.mtimeMs` of the src. `copy` doesn't preserve the `lstat.mtimeMs`

gregmagolan: Hence the map to keep track of the `lstat.mtimeMs` of the src when we do the copy

js/private/js_devserver.js
gregmagolan: nice catch

js/private/js_run_devserver.js
jbedard: Ohhh I see. We're storing+comparing the time of the src, not comparing the src vs destination 👍

jbedard: So the map is required. I don't think it's wroth trying to clean it atm, at least in this PR

js/private/js_devserver.js
jbedard: Actually if I'm recalling the `data_files` correctly I don't think we even need recursion and can have one giant `Promise.all`?
```
function sync(...) { return Promise.all(files.map(... syncRecursive(src, dst)) }

function syncRecursive(src, dst) {
   ...if/lstat/check-cache...

   await mkdirp(path.dirname(dst));
   if (symlink) ... link ...
   else if (file) ... copy/chmod ...
   else if (directory) ... mkdir() ...    // <--- no "p"
}
```

Note that the results of `syncRecursive` are put in `Promise.all` and not individual `await`s so it should be 100% parallel?

js/private/js_devserver.js
gregmagolan: Fixed up

js/private/js_devserver.js
gregmagolan: Nice catch

js/private/js_devserver.js
gregmagolan: 👍

js/private/js_run_devserver.js
gregmagolan: Yeah. Probably not worth the extra complexity. Will leave cleanup out for now.

gregmagolan:
Another example output of a next.js devserver with more files synced

```
...
Starting js_run_devserver //:dev
Syncing...
21 files synced in 20 ms
Initial sync done
Running './node_modules/.bin/next dev' in /var/folders/0w/2qdthbhd3lqd1p96g4r99h2m0000gn/T/js_run_devserver-QW4kUl/__main__

ready - started server on 0.0.0.0:3000, url: http://localhost:3000
event - compiled client and server successfully in 1635 ms (173 modules)
```

and after a change

```
iBazel [8:29PM]: Changed: "/Users/greg/aspect/rules/bazel-examples/next.js/pages/index.tsx". Rebuilding...
iBazel [8:29PM]: Running //:dev
iBazel [8:29PM]: Notifying of changes
IBAZEL_BUILD_STARTED
...
iBazel [8:29PM]: IBAZEL BUILD SUCCESS
IBAZEL_BUILD_COMPLETED SUCCESS
Syncing...
2 files synced in 13 ms
wait  - compiling...
event - compiled client and server successfully in 106 ms (173 modules)
```

.github/workflows/ci.yaml
```diff
@@ -86,7 +105,7 @@ jobs:
               working-directory: ${{ matrix.folder }}
               # Don't run e2e/npm_translate_lock_auth unless there is an auth token secret; this won't
               # be set on forks
-              if: ${{ matrix.folder != 'e2e/npm_translate_lock_auth' || env.ASPECT_NPM_AUTH_TOKEN != ''}}
+              if: ${{ steps.has_test_sh.outputs.files_exists != 'true' && (matrix.folder != 'e2e/npm_translate_lock_auth' || env.ASPECT_NPM_AUTH_TOKEN != '') }}
```
alexeagle, line 108: is it really useful to be so general and combine that test.sh with the matrix setup?
at least add some comment about when the test.sh outputs is expected to be true, I don't quite follow how this logic works

e2e/js_run_devserver/serve_test.sh
```diff
@@ -0,0 +1,76 @@
...
+_sedi () {
+  case $(uname) in
+    Darwin*) sedi=('-i' '') ;;
+    *) sedi=('-i') ;;
+  esac
+
+  sed "${sedi[@]}" "$@"
+}
+
+./node_modules/.bin/ibazel run "$1" >/dev/null 2>&1 &
```
alexeagle, line 15: reminds me that your ibazel benchmark runner from like 5 years ago can be useful again to measure the new roundtrip

e2e/js_run_devserver/src/BUILD.bazel
```diff
@@ -0,0 +1,42 @@
...
+        "//:node_modules/http-server",
+    ],
+)
+
+http_server_bin.http_server_binary(
+    name = "http_server",
+)
+
+copy_to_directory(
+    name = "http_root",
```
alexeagle, line 37: nit: I'd put this earlier in the file so it's not forward-referenced

e2e/js_run_devserver/src/index.html
```diff
@@ -0,0 +1 @@
+<div>My first website</div>
```
alexeagle, line 1: congratulations! 😛

e2e/js_run_devserver/test.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -o errexit -o nounset -o pipefail
+
+# Integration test for js_run_devserver run with ibazel
+
+bazel run -- @pnpm//:pnpm --dir "$PWD" install
```
alexeagle, line 6: thought for later: this would be a nice thing to syntax sugar in the rules_js plugin for aspect-cli

js/private/js_run_devserver.bzl
```diff
@@ -0,0 +1,95 @@
...
+
+The devserver specified is run in a custom sandbox that is more compatible with
+node tools that come with watch mode such as Webpack devserver or Next.js.
+
+The custom sandbox is populated with the default outputs of all targets in `data`.
+rules_js node_modules links targets such as `//:node_modules/http-server` are supported.
+Since these targets are symlinks in the output tree, they are recreated as symlinks
+in the custom sandbox and do not incur a fully copy of the underlying npm packages.
+
+Supports running with ibazel. Only changes in `data` files are synchronized incrementally
```
alexeagle, line 17: nit: link to bazel-watcher

js/private/js_run_devserver.bzl
```diff
@@ -0,0 +1,95 @@
...
+        doc = """The devserver binary target to run.
+        
+Only one of `command` or `tool` may be specified.""",
+        executable = True,
+        cfg = "exec",
+    ),
+    "command": attr.string(
+        doc = """The devserver command to run.
+
+For example, this could be an executable tool in an npm package that is included
```
alexeagle, line 32: more guidance about how command fixes problems with several copies of react?

alexeagle (approved):

js/private/js_run_devserver.bzl
gregmagolan: sure. will add more to the docstring in the morning and add some examples as well

.github/workflows/ci.yaml
gregmagolan: Borrowed it from our bazel-examples repo; https://github.com/aspect-build/bazel-examples/blob/83d07540a160c28a6ec56c1a383dd9811022049c/.github/workflows/ci.yaml#L61

In short, if there is a test.sh file in a matrix folder then it runs that, otherwise it runs `bazel test ...` and maybe `bazel coverage ...`. Running a test.sh instead of `bazel test ...` is a useful pattern for some e2e tests; any integration test such as this one that doesn't fit nicely into a bazel test case use this pattern.

rules_ts has a slightly different variation of this check https://github.com/aspect-build/rules_ts/blob/1cbe54029c00241a0ea399b97d0b4d4175cee8a8/.github/workflows/ci.yaml#L107

e2e/js_run_devserver/serve_test.sh
gregmagolan: good idea. I should revive that again and make some new ibazel benchmarks and a corresponding blog post to advertise this new rule

e2e/js_run_devserver/src/BUILD.bazel
gregmagolan: 👍

js/private/js_run_devserver.bzl
gregmagolan: 👍

.github/workflows/ci.yaml
gregmagolan: Added better comments and cleaned it up so that folders are dynamically generated as well. The `npm_translate_lock_auth` e2e is special and needs to be filtered on on fork PRs since it requires secrets.

js/private/js_run_devserver.js
```diff
@@ -0,0 +1,209 @@
...
+    return (
+        !!(stats.mode & fs.constants.S_IXUSR) ||
+        !!(stats.mode & fs.constants.S_IXGRP) ||
+        !!(stats.mode & fs.constants.S_IXOTH)
+    )
+}
+
+// Ensure that a directory exists. If it has not been previously created or does not exist then it
+// creates the directory, first recursively ensuring that its parent directory exists.
+// Intentionally synchronous to avoid race conditions between sync promises
```
jbedard, line 26: *async ?

Any examples of such race conditions that could be mentioned?

js/private/js_run_devserver.js
```diff
@@ -0,0 +1,209 @@
...
+                    `Syncing file ${src.slice(RUNFILES_ROOT.length + 1)}`
+                )
+            }
+            if (exists) {
+                await fs.promises.unlink(dst)
+            } else {
+                mkdirpSync(path.dirname(dst))
+            }
+            await fs.promises.copyFile(src, dst)
+            const mode = isExecutable(lstat) ? 0o744 : 0x644
```
jbedard, line 93: One is 0o and one is 0x, can we make them the same?

js/private/js_run_devserver.js
```diff
@@ -0,0 +1,209 @@
...
+            } else {
+                mkdirpSync(path.dirname(dst))
+            }
+            await fs.promises.symlink(src, dst)
+            return 1
+        } else if (lstat.isDirectory()) {
+            const contents = await fs.promises.readdir(src)
+            let totalSynced = 0
+            await Promise.all(
+                contents.map(async (entry) => {
```
jbedard, line 72: If the src directory is empty and there's no recursion then I don't think it will be created at all?

js/private/js_run_devserver.js
```diff
@@ -0,0 +1,209 @@
...
+            if (exists) {
+                await fs.promises.unlink(dst)
+            } else {
+                mkdirpSync(path.dirname(dst))
+            }
+            await fs.promises.symlink(src, dst)
+            return 1
+        } else if (lstat.isDirectory()) {
+            const contents = await fs.promises.readdir(src)
+            let totalSynced = 0
```
jbedard, line 70: Could use `.reduce` instead of the `totalSynced` var and incrementing within the loop 🤷‍♂️

js/private/js_run_devserver.js
jbedard: The returned count will also be 0 in that case. Would that case never actually occur?

js/private/js_run_devserver.js
```diff
@@ -0,0 +1,209 @@
...
+        })
+
+        process.stdin.on('data', async (chunk) => {
+            const chunkString = chunk.toString()
+            if (chunkString.indexOf('IBAZEL_BUILD_COMPLETED SUCCESS') !== -1) {
+                if (process.env.JS_BINARY__LOG_DEBUG) {
+                    console.error('IBAZEL_BUILD_COMPLETED SUCCESS')
+                }
+                // Re-parse the config file to get the latest list of data files to copy
+                await sync(
```
jbedard, line 171: Do we want to do anything if something within this sync call throws?

js/private/js_run_devserver.js
```diff
@@ -0,0 +1,209 @@
...
+
+        proc.on('close', (code) => {
+            console.error(`child tool process exited with code ${code}`)
+            resolve()
+            process.exit(code)
+        })
+
+        process.stdin.on('data', async (chunk) => {
+            const chunkString = chunk.toString()
+            if (chunkString.indexOf('IBAZEL_BUILD_COMPLETED SUCCESS') !== -1) {
```
jbedard, line 166: Could use `.includes` instead of index and -1

js/private/js_run_devserver.js
gregmagolan: When you use `await fs.promises.mkdir(p)` then you could end up calling it twice in two different promises which would error on the 2nd call. This is because `if (!fs.existsSync(p))` followed by `await fs.promises.mkdir(p)` is not atomic so both threads can enter into the condition before either calls mkdir.

js/private/js_run_devserver.js
gregmagolan: Actually we should be fine without calling chmod at all. Removed this line.

js/private/js_run_devserver.js
gregmagolan: Just let the process fail with an error message is sufficient

js/private/js_run_devserver.js
gregmagolan: 👍

js/private/js_run_devserver.js
gregmagolan: Updated comment

js/private/js_run_devserver.js
gregmagolan: Might if there is a tree artifact that contains an empty directory

gregmagolan: Fixed

js/private/js_run_devserver.js
gregmagolan: I don't think that will play nicely with Promises.all

gregmagolan: Oh I see what you mean. A throw here is an unhandled promise exception.

Fixed.

js/private/js_run_devserver.js
gregmagolan: Oh I see what you mean. A throw here is an unhandled promise exception.

Fixed.

e2e/js_run_devserver/src/index.html
gregmagolan: 🎉

js/private/js_run_devserver.js
jbedard: Really just do what you think is most readable though

npm/private/npm_link_package_store.bzl
```diff
@@ -94,12 +94,8 @@ def _impl(ctx):
     files = utils.make_symlink(ctx, root_symlink_path, virtual_store_directory)
 
     for bin_name, bin_path in ctx.attr.bins.items():
-        if ctx.label.package:
-            path_to_root = "/".join([".."] * len(ctx.label.package.split("/")))
-        else:
-            path_to_root = "."
         bin_file = ctx.actions.declare_file(paths.join("node_modules", ".bin", bin_name))
-        bin_path = paths.normalize(paths.join("../..", path_to_root, virtual_store_directory.short_path, bin_path))
+        bin_path = paths.normalize(paths.join("..", package, bin_path))
```
jbedard, line 98: Did this change anything?

js/private/js_run_devserver.js
gregmagolan: ✅

npm/private/npm_link_package_store.bzl
gregmagolan: yea, it changed where the generated `node_modules/.bin/foo` points to. Previously I reached into the virtual store directory but it is cleaner and more portable to run the bin entry point as `node_modules/foo/path/to/bin` that goes through the `node_modules/foo` symlink

jbedard: 👍 

Did that fix anything though or just cleaner (and I think more "correct" as well)?

gregmagolan: It wouldn't have worked in the devserver tmp sandbox without this change since the relative path to the virtual store is not the same as from the runfiles tree.

gregmagolan: Comes into play with the `command` variant for specifying the tool such as
```
js_run_devserver(
    name = "serve_alt",
    args = ["{}/http_root".format(package_name())],
    command = "./node_modules/.bin/http-server",
    data = [
        ":http_root",
        "//:node_modules/http-server",
    ],
)
```

js/private/js_run_devserver.js
```diff
@@ -0,0 +1,212 @@
...
+
+        process.stdin.on('data', async (chunk) => {
+            try {
+                const chunkString = chunk.toString()
+                if (chunkString.includes('IBAZEL_BUILD_COMPLETED SUCCESS')) {
+                    if (process.env.JS_BINARY__LOG_DEBUG) {
+                        console.error('IBAZEL_BUILD_COMPLETED SUCCESS')
+                    }
+                    // Re-parse the config file to get the latest list of data files to copy
+                    await sync(
```
jbedard, line 168: This can get triggered multiple times and then have multiple overlapping `sync`s. Is that ok? Since we are only creating+updating files it might be but if we are deleting files it won't be.

This is where if we want to ensure there are no overlapping `sync`s we'd do something such as:

```
let synching = Promise.resolve();
...
await (synching = synching.then(()( => sync()))
```

jbedard (approved):

js/private/js_run_devserver.js
gregmagolan: 🧙

PR #546, opened 2022-10-22
perf: prevent preparing stack traces in fs *Sync patches

jbedard:
This makes my test of 500 basic files about 2x faster, going from a fairly consistent 7.5-8 to 3.5-4. Those 500 files are just `expect(true).toBeTrue()` so the time invoking tests should be ~0 and it's all jest running, node loading files etc.

Originally I only wrapped some internal functions such as `guardedRealPathSync` but I think it's better to just wrap the full public methods so the `Error` try/catch modifications only happen once. The worry with that is stack traces thrown within the implementations wouldn't be complete but that seems to still work, see the stack trace tests.

See https://github.com/aspect-build/rules_jest/issues/50

### DIFF
--- a/js/private/node-patches/src/fs.ts
+++ b/js/private/node-patches/src/fs.ts
@@ -28,6 +28,68 @@ type Dirent = any
 // es modules
 const _fs = require('fs')
 
+/**
+ * Reduce the information in Errors to both hide the origin and improve performance.
+ * Any nested calls may be effected and also hidden.
+ *
+ * May potentially:
+ * - hide the function from stack traces
+ * - prevent errors thrown within the function from calculating stack traces
+ *
+ * TODO: currently can not hide computing of the stack trace within handleErrorFromBinding:
+ * - https://github.com/nodejs/node/blob/1b971883e0c86f37468e421ffe5f39b585782ad6/lib/internal/fs/utils.js#L340-L356
+ */
+function hideStackFrames<T extends Function>(fn: T): T {
+    const HiddenErrorProps = {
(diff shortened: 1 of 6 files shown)


js/private/node-patches/src/fs.ts
```diff
@@ -125,7 +125,10 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
             // there are no hops so lets report the stats of the real file;
             // we can't use origRealPathSync here since that function calls lstat internally
             // which can result in an infinite loop
-            return origLstatSync(unguardedRealPathSync(args[0]), args.slice(1))
```
jbedard, line 128: This one looks like a bug. I found this when I (temporarily) put proper ts types on the `orig*Sync` methods and this was an error.

matthewjh:
Thanks @jbedard, this looks great. I noticed that webpack-dev-server is 2-3x slower with the fs patches turned on (WDS does a LOT of brute force directory crawling and symlink probing), and my profile showed that a lot of time was spent in node stack trace processing machinery (though I had not raised an issue yet). I will have to try, of course, but I am hopeful this will address that.

jbedard:
I think I've prevented 2/3 of the places I saw major stack trace cpu time, see the comments in the diff (including the TODO) which link to the nodejs source if you're interested. That last one with the TODO is still a major perf hit but my tests seem to have reduced it at least 50% with a super simple jest test.

gregmagolan (approved):
🌮 nice one. if this is good to land then lets get these improvements in and see about the TODO and further improvements in a follow-up.

PR #550, opened 2022-10-24
feat: add additional env vars to js_binary

gregmagolan:
Also add tests for a subset of the envars that are trivial to test (ones that don't change with platform).

Pre-factor for https://github.com/aspect-build/rules_js/pull/540 and the next.js example update in bazel-examples.


js/private/js_binary.bzl
```diff
@@ -258,7 +258,9 @@ def _bash_launcher(ctx, entry_point_path, log_prefix_rule_set, log_prefix_rule,
         "JS_BINARY__BINDIR": "$(BINDIR)",
         "JS_BINARY__BUILD_FILE_PATH": "$(BUILD_FILE_PATH)",
         "JS_BINARY__COMPILATION_MODE": "$(COMPILATION_MODE)",
+        "JS_BINARY__PACKAGE": ctx.label.package,
```
alexeagle, line 261: again, it feels like a bit of duplication since it's pretty easy/reliable to extract from JS_BINARY__TARGET - but OTOH I guess another env variable isn't expensive and it's one fewer thing for the program to be responsible for

js/private/js_binary.sh.tpl
```diff
@@ -118,6 +118,11 @@ trap _exit EXIT
 # ==============================================================================
 {{initialize_runfiles}}
 
+# TODO(2.0): export only JS_BINARY__RUNFILES
+export RUNFILES
+JS_BINARY__RUNFILES="$RUNFILES"
+export JS_BINARY__RUNFILES
```
alexeagle, line 124: hahaha everywhere that already had to special-case JAVA_RUNFILES https://bazel.build/reference/test-encyclopedia

alexeagle (approved):

alexeagle:
no doc update ? is there even a doc for js_binary saying what variables the program can expect to be present?

gregmagolan:
> no doc update ? is there even a doc for js_binary saying what variables the program can expect to be present?

They are all still undocumented

PR #551, opened 2022-10-24
fix: update `expand_template` to return `DefaultInfo`

cgrindel:
Related to aspect-build/silo#550.

- Update to `expand_template`.
- Add test that ensures that the `DefaultInfo` is returned.


gregmagolan (approved):
🌮

PR #556, opened 2022-10-26
fix: support non-https registry protocols

jbedard:
Assigning the default registry protocol is now done when reading them from npmrc instead of when using them. So the `registries` map passed around now has the protocols in the URLs.

I'm not sure how the `//no-protocol.com` cases should be handled though. Normally that means to "use the current protocol", but idk what the "current" is?

### DIFF
--- a/npm/private/utils.bzl
+++ b/npm/private/utils.bzl
@@ -134,11 +134,16 @@ def _parse_package_name(package):
         return (segments[0], segments[1])
     return ("", segments[0])
 
+def _npm_registry_url(package, registries, default_registry):
+    (package_scope, _) = _parse_package_name(package)
+
+    return registries[package_scope] if package_scope in registries else default_registry
+
 def _npm_registry_download_url(package, version, registries, default_registry):
     "Make a registry download URL for a given package and version"
 
-    (package_scope, package_name_no_scope) = _parse_package_name(package)
-    registry = "https://{}".format(registries[package_scope]) if package_scope in registries else default_registry
+    (_, package_name_no_scope) = _parse_package_name(package)
+    registry = _npm_registry_url(package, registries, default_registry)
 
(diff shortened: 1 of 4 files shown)


gregmagolan:
Not clear to me from the description what issue this is solving? Can you elaborate @jbedard

npm/private/utils.bzl
```diff
@@ -134,11 +134,16 @@ def _parse_package_name(package):
...
+    (package_scope, _) = _parse_package_name(package)
+
+    return registries[package_scope] if package_scope in registries else default_registry
+
 def _npm_registry_download_url(package, version, registries, default_registry):
     "Make a registry download URL for a given package and version"
 
-    (package_scope, package_name_no_scope) = _parse_package_name(package)
-    registry = "https://{}".format(registries[package_scope]) if package_scope in registries else default_registry
+    (_, package_name_no_scope) = _parse_package_name(package)
```
gregmagolan, line 145: looks like this like can go; package_name_no_scope not used either

gregmagolan:
I think I see; this preserves the registry protocol in .npmrc if it is explicitly set?

jbedard:
> I think I see; this preserves the registry protocol in .npmrc if it is explicitly set?

Yeah. Previously we were assuming registries defined for scopes had no protocol and we stripped the protocol from the main (non-scoped) registry config. This PR changes it to keep whatever protocol is in the npmrc and default to https instead of forcing it.

gregmagolan (approved):

npm/private/utils.bzl
jbedard: It's a param in the `.format` below

gregmagolan: ooohhh. i see. missed that. 👍

PR #559, opened 2022-10-27
fix: fix windows regression in js_binary launcher bash script

gregmagolan:
This has been broken for a while on Windows. We don't have Windows CI yet for rules_js so it wasn't caught until the BCR pre-submit failed.

### DIFF
--- a/js/private/js_binary.sh.tpl
+++ b/js/private/js_binary.sh.tpl
@@ -183,7 +183,7 @@ if [ ! -f "$JS_BINARY__NODE_BINARY" ]; then
     logf_fatal "node binary '%s' not found in runfiles" "$JS_BINARY__NODE_BINARY"
     exit 1
 fi
-if [ ! -x "$JS_BINARY__NODE_BINARY" ]; then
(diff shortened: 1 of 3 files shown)


js/private/js_binary.sh.tpl
```diff
@@ -183,7 +183,7 @@ if [ ! -f "$JS_BINARY__NODE_BINARY" ]; then
     logf_fatal "node binary '%s' not found in runfiles" "$JS_BINARY__NODE_BINARY"
     exit 1
 fi
-if [ ! -x "$JS_BINARY__NODE_BINARY" ]; then
+if [ "$_IS_WINDOWS" -ne "1" ] && [ ! -x "$JS_BINARY__NODE_BINARY" ]; then
```
alexeagle, line 186: do you mean to leave these changes in??

alexeagle (approved):

js/private/js_binary.sh.tpl
gregmagolan: yes, apparently `-x` doesn't work for checking the "binary" status on Windows

jbedard (approved):

PR #563, opened 2022-10-27
fix: always set JsInfo.declarations on npm_link_package_store targets

jbedard:
This way it will be a valid `ts_project` dep like we were discussing.

This PR currently does the same as what `js_library` does for directories but 100% of the time. Here `js_library` does this only when finding a directory with unknown content: https://github.com/aspect-build/rules_js/blob/2bc5e8622b9509b370fbbd44e9022900d15b5244/js/private/js_library.bzl#L126-L129

### DIFF
--- a/npm/private/npm_link_package_store.bzl
+++ b/npm/private/npm_link_package_store.bzl
@@ -126,6 +126,9 @@ def _impl(ctx):
             runfiles = ctx.runfiles(transitive_files = transitive_files_depset),
         ),
         js_info(
+            # assume a directory contains declarations since we can't know that it doesn't
+            declarations = files_depset,
+            transitive_declarations = transitive_files_depset,
             npm_linked_package_files = files_depset,
             npm_linked_packages = depset([npm_linked_package_info]),
             npm_package_store_deps = depset([store_info]),
--- a/js/private/test/js_library_test.bzl
+++ b/js/private/test/js_library_test.bzl
@@ -37,11 +37,17 @@ def _declarations_test_impl(ctx):
     asserts.true(env, declarations[0].path.find("/importing.d.ts") != -1)
     asserts.true(env, declarations[1].path.find("/data.json") != -1)
 
-    # declarations should only have the source declarations
+    # transitive_declarations should have the source declarations and transitive declarations
     transitive_declarations = target_under_test[JsInfo].transitive_declarations.to_list()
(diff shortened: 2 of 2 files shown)


npm/private/npm_link_package_store.bzl
```diff
@@ -126,6 +126,8 @@ def _impl(ctx):
             runfiles = ctx.runfiles(transitive_files = transitive_files_depset),
         ),
         js_info(
+            # assume a directory contains declarations since we can't know that it doesn't
+            declarations = files_depset,
```
jbedard, line 130: Should this be `virtual_store_directory` instead of `files_depset`?

The `files_depset` is a symlink to the `virtual_store_directory` but possibly also includes both the file and the symlink (based on `experimental_allow_unresolved_symlinks`?).

gregmagolan: `files_depset` is right. You want both the node_modules/foo symlink and the underlying virtual store foo directory

gregmagolan: depending on `experimental_allow_unresolved_symlinks` `ctx.actions.symlink` is either called with `target_path` or `target_file`. If `target_file` is used then the virtual store is included transitively in the symlink. When using `target_path` then underlying virtual store directory needs to be included explicitly along with the symlink File

gregmagolan: technically you should also set `transitive_declarations` to `transitive_files_depset`

jbedard: > technically you should also set transitive_declarations to transitive_files_depset

I guess just like the npm package itself we don't know what's in the transitive files so we should just include it 100% of the time? Otherwise it won't be available to the ts compiler and might fail?

jbedard:
One other command case is `ts_project` depending on a `js_library` but the library having no dts/json/directory so it has empty declarations. Is that ok? That means a `js_library` might change and then any `ts_project` depending on it needs to drop the dep? Or the opposite...

npm/private/npm_link_package_store.bzl
gregmagolan: Yup. I don't know if any rules even use the `transitive_declarations` but they are there incase they are ever needed

gregmagolan (approved):
🌮

PR #564, opened 2022-10-28
fix: make js_run_binary more compatible with tools that are sensitive to being found in multiple node_modules trees

gregmagolan:
Adds run_from_execroot option to js_run_binary and set as default.

Runs the `tool` entry_point out of the execroot output tree instead of the runfiles of `tool`.

Runfiles of `tool` are all hoisted to `srcs` of the underlying `run_binary` so they are included as execroot
inputs to the action.

Running the entry point out of the execroot output tree means that there is no conflicting runfiles
`node_modules` in the node_modules resolution path to confuse npm packages such as next and react which
don't like to be resolved in multiple node_modules trees. This more closely emulates the environment that
tools such as Next.js see when they are run outside of Bazel.

When True, it is recommended that the js_binary tool use the `copy_data_to_bin` default of True
so that all data files needed by the binary are available in the execroot output tree.

This little bit of magic allows us to do away with the custom next.js entry point in the next.js example here: https://github.com/aspect-build/bazel-examples/pull/124


examples/js_binary/BUILD.bazel
```diff
@@ -95,6 +95,28 @@ diff_test(
...
+# Also test with run_from_execroot execution
+js_run_binary(
+    name = "run2_run_from_execroot",
+    srcs = [],
+    outs = ["out2_run_from_execroot"],
+    args = ["out2_run_from_execroot"],
+    chdir = package_name(),
+    # Request that the rules_js launcher prints extra information
+    log_level = "debug",
+    run_from_execroot = True,
```
alexeagle, line 107: what's the downside if this defaults to True?

I barely understand the difference between the two ways of running this, so I'd have a hard time explaining to users. I'm concerned most people won't discover that their problem can be fixed with this flag...

js/private/js_run_binary.bzl
```diff
@@ -275,6 +279,24 @@ See https://github.com/aspect-build/rules_js/tree/main/docs#using-binaries-publi
             name = to_label(name),
         ))
 
+    # Configure run from execroot
+    if run_from_execroot:
```
alexeagle, line 283: there's no complex magic in the implementation, that's nice :)

alexeagle (commented):
let's work out the docstring implications, but it's amazing how this delivers on the compatibility promise of rules_js

examples/js_binary/BUILD.bazel
gregmagolan: yeah; that should work as a better default. flipped it and everything is green. 🌮

js/private/js_run_binary.bzl
gregmagolan: layering is still working out nicely

js/private/js_run_binary.bzl
```diff
@@ -107,6 +109,19 @@ def js_run_binary(
 
             This makes node binaries match the expected bazel paradigm.
 
+        run_from_execroot: Runs the `tool` entry_point out of the execroot output tree instead of the runfiles of `tool`.
```
alexeagle, line 112: nit "run out of" is vernacular, I think you mean setting the working directory?

js/private/js_run_binary.bzl
```diff
@@ -107,6 +109,19 @@ def js_run_binary(
...
+
+            Runfiles of `tool` are all hoisted to `srcs` of the underlying `run_binary` so they are included as execroot
+            inputs to the action.
+
+            Running the entry point out of the execroot output tree means that there is no conflicting runfiles
+            `node_modules` in the node_modules resolution path to confuse npm packages such as next and react which
+            don't like to be resolved in multiple node_modules trees. This more closely emulates the environment that
+            tools such as Next.js see when they are run outside of Bazel.
+
+            When True, it is recommended that the js_binary tool use the `copy_data_to_bin` default of True
```
alexeagle, line 122: nit: instead of "recommended" it would be nice to error if the tool doesn't copy its data to bin

js/private/js_run_binary.bzl
alexeagle: 💎

alexeagle (approved):

js/private/js_run_binary.bzl
gregmagolan: it can be done as long there is an opt-out to disable the validation

js/private/js_run_binary.bzl
gregmagolan: Actually, no. The working directory remains the same. This changes which copy of the entry point we're calling. Node itself is still run out of runfiles (since that is where the platform specific node lives) but the entry point passing to node is now the one in the execroot output tree instead of the one in runfiles.

js/private/js_run_binary.bzl
gregmagolan: ✅

docs/js_run_devserver.md
```diff
@@ -93,7 +93,7 @@ Only `data` files that change on incremental builds are synchronized when runnin
...
+| <a id="js_run_devserver-copy_data_to_bin"></a>copy_data_to_bin |  When True, data files and the entry_point file are copied to the Bazel output tree before being passed         as inputs to runfiles.<br><br>        Defaults to True so that a js_binary with the default value is compatible with js_run_binary with run_from_execroot         set to True, the default there.<br><br>        Setting this to False is more optimal in terms of inputs, but there is a yet unresolved issue of ESM imports         skirting the node fs patches and escaping the sandbox: https://github.com/aspect-build/rules_js/issues/362.         This is hit in some test popular runners such as mocha, which use native <code>import()</code> statements         (https://github.com/aspect-build/rules_js/pull/353). When set to false, a program such as mocha that uses ESM         imports may escape the execroot by following symlinks into the source tree. When set to True, such a program         would escape the sandbox but they will end up in the output tree where node_modules and other inputs required         will be available.   | Boolean | optional | <code>True</code> |
```
jbedard, line 96:
```suggestion
| <a id="js_run_devserver-copy_data_to_bin"></a>copy_data_to_bin |  When True, data files and the entry_point file are copied to the Bazel output tree before being passed         as inputs to runfiles.<br><br>        Defaults to True so that a js_binary with the default value is compatible with js_run_binary with run_from_execroot         set to True, the default there.<br><br>        Setting this to False is more optimal in terms of inputs, but there is a yet unresolved issue of ESM imports         skirting the node fs patches and escaping the sandbox: https://github.com/aspect-build/rules_js/issues/362.         This is hit in some popular test runners such as mocha, which use native <code>import()</code> statements         (https://github.com/aspect-build/rules_js/pull/353). When set to false, a program such as mocha that uses ESM         imports may escape the execroot by following symlinks into the source tree. When set to True, such a program         would escape the sandbox but they will end up in the output tree where node_modules and other inputs required         will be available.   | Boolean | optional | <code>True</code> |
```

js/private/js_run_binary.bzl
```diff
@@ -107,6 +110,21 @@ def js_run_binary(
 
             This makes node binaries match the expected bazel paradigm.
 
+        run_from_execroot: Run the `tool` js_binary entry_point script that is in the execroot of the output tree
```
jbedard, line 113:
```suggestion
        run_from_execroot: Run the `tool` js_binary entry_point script in the execroot of the output tree
```

?

js/private/js_run_binary.bzl
```diff
@@ -107,6 +110,21 @@ def js_run_binary(
...
+            inputs to the action.
+
+            Running the entry point out of the execroot output tree means that there is no conflicting runfiles
+            `node_modules` in the node_modules resolution path to confuse npm packages such as next and react which
+            don't like to be resolved in multiple node_modules trees. This more closely emulates the environment that
+            tools such as Next.js see when they are run outside of Bazel.
+
+            When True, the js_binary tool must have `copy_data_to_bin` set to True (the default) so
+            so that all data files needed by the binary are available in the execroot output tree.
+            This requirement can be turned off with the
```
jbedard, line 126: ?

js/private/js_run_binary.bzl
jbedard: Actually maybe it is correct now that I read it again, but it feels like a very long sentence that I had to read multiple times

js/private/js_run_binary.bzl
```diff
@@ -107,6 +110,21 @@ def js_run_binary(
 
             This makes node binaries match the expected bazel paradigm.
 
+        run_from_execroot: Run the `tool` js_binary entry_point script that is in the execroot of the output tree
+            instead of the copy that is in runfiles.
+
+            Runfiles of `tool` are all hoisted to `srcs` of the underlying `run_binary` so they are included as execroot
```
jbedard, line 116: Is that the `srcs` param? So they are hoisted **as** srcs of the underline `run_binary`?

js/private/js_run_binary.bzl
```diff
@@ -107,6 +110,21 @@ def js_run_binary(
...
+            inputs to the action.
+
+            Running the entry point out of the execroot output tree means that there is no conflicting runfiles
+            `node_modules` in the node_modules resolution path to confuse npm packages such as next and react which
```
jbedard, line 120: "to confuse" => "which confuse"?

js/private/js_run_binary.bzl
```diff
@@ -107,6 +110,21 @@ def js_run_binary(
 
             This makes node binaries match the expected bazel paradigm.
 
+        run_from_execroot: Run the `tool` js_binary entry_point script that is in the execroot of the output tree
+            instead of the copy that is in runfiles.
+
+            Runfiles of `tool` are all hoisted to `srcs` of the underlying `run_binary` so they are included as execroot
+            inputs to the action.
+
+            Running the entry point out of the execroot output tree means that there is no conflicting runfiles
```
jbedard, line 119: "means that there is no conflicting runfiles" => "means there are no conflicting runfiles"?

js/private/js_run_binary.bzl
```diff
@@ -107,6 +110,21 @@ def js_run_binary(
...
+            `node_modules` in the node_modules resolution path to confuse npm packages such as next and react which
+            don't like to be resolved in multiple node_modules trees. This more closely emulates the environment that
+            tools such as Next.js see when they are run outside of Bazel.
+
+            When True, the js_binary tool must have `copy_data_to_bin` set to True (the default) so
```
jbedard, line 124: "so so that all data files" => "so all data files"?

js/private/js_run_binary.bzl
```diff
@@ -178,6 +196,11 @@ def js_run_binary(
             When disabled, node programs can leave the execroot, runfiles and sandbox by following symlinks
             which can lead to non-hermetic behavior.
 
+        allow_run_from_execroot_with_no_copy_data_to_bin: Turn off validation that the js_binary tool
```
jbedard, line 199: js_binary => `js_binary`?

js/private/js_run_binary.bzl
```diff
@@ -275,6 +298,31 @@ See https://github.com/aspect-build/rules_js/tree/main/docs#using-binaries-publi
             name = to_label(name),
         ))
 
+    # Configure run from execroot
+    if run_from_execroot:
+        fixed_env["JS_BINARY__RUN_FROM_EXECROOT"] = "1"
+
+        if allow_run_from_execroot_with_no_copy_data_to_bin:
```
jbedard, line 305: It's probably only applicable when `run_from_execroot` is set, but should we put this `if` outside the `run_from_execroot` `if` just so the `JS_BINARY__` vars directly align with the bzl params?

js/private/js_run_binary.bzl
jbedard, line 113: js_binary => `js_binary`?

js/private/js_binary.bzl
```diff
@@ -182,14 +182,16 @@ _ATTRS = {
         doc = """When True, data files and the entry_point file are copied to the Bazel output tree before being passed
         as inputs to runfiles.
 
-        Ideally, the default for this would be False as it is optimal, but there is a yet unresloved issue of ESM imports
+        Defaults to True so that a js_binary with the default value is compatible with js_run_binary with run_from_execroot
```
jbedard, line 185: backticks around the rule/method/param names?

js/private/js_binary.bzl
```diff
@@ -182,14 +182,16 @@ _ATTRS = {
...
+
+        Setting this to False is more optimal in terms of inputs, but there is a yet unresolved issue of ESM imports
         skirting the node fs patches and escaping the sandbox: https://github.com/aspect-build/rules_js/issues/362.
         This is hit in some test popular runners such as mocha, which use native `import()` statements
-        (https://github.com/aspect-build/rules_js/pull/353). 
-
-        A default of True will prevent program such as mocha from following symlinks into the source tree. They will
-        escape the sandbox but they will end up in the output tree where node_modules and other inputs required will be
-        available. With this in mind, the default will remain true until issue #362 is resolved.
+        (https://github.com/aspect-build/rules_js/pull/353). When set to false, a program such as mocha that uses ESM
```
jbedard, line 191: false => False

js/private/js_binary.bzl
```diff
@@ -182,14 +182,16 @@ _ATTRS = {
...
+        imports may escape the execroot by following symlinks into the source tree. When set to True, such a program
+        would escape the sandbox but they will end up in the output tree where node_modules and other inputs required
```
jbedard, line 193: "but they will end up" => "but will end up"? 🤷

docs/js_run_devserver.md
gregmagolan: 🍔

js/private/js_run_binary.bzl
gregmagolan: 🍔

js/private/js_run_binary.bzl
gregmagolan: Yeah. So they end up in the execroot. Otherwise they won't be present since Bazel only includes the runfiles in the sandbox and not the execroot files that back them.

js/private/js_run_binary.bzl
gregmagolan: 🍔

js/private/js_run_binary.bzl
gregmagolan: 🍔

js/private/js_run_binary.bzl
gregmagolan: 🍔

js/private/js_run_binary.bzl
gregmagolan: 🍔

js/private/js_binary.bzl
gregmagolan: 🍔

js/private/js_binary.bzl
gregmagolan: 🍔

js/private/js_binary.bzl
gregmagolan: 🍔

js/private/js_run_binary.bzl
gregmagolan: Its only ever checked in the launcher if `JS_BINARY__RUN_FROM_EXECROOT` is set but it doesn't hurt to hoist it for alignment

js/private/js_run_binary.bzl
gregmagolan: Whoops. Forgot to finish that sentence.

js/private/js_run_binary.bzl
gregmagolan: Yea. It is tricky to be both precise and to the point with this one.

gregmagolan: Re-worded to this
```
        run_from_execroot: Run the `entry_point` script of the `js_binary` `tool` that is in the execroot output tree
            instead of the copy that is in runfiles.
```
which I'm not sure if it is any better

alexeagle (approved):

PR #569, opened 2022-10-29
perf: cache fs patch fs lookups 

jbedard:
Caches all fs access operations when reading symlinks, specifically: if a file is a symlink, and what it points.

The underlying `readlinkSync` call throws when called on a non-symlink. Previously calculating the exception stack trace used a significant amount of time. This prevents the exception by doing `lstat` first, the combined result of `lstat` + `readlink` are then cached.

### DIFF
--- a/js/private/node-patches/src/fs.ts
+++ b/js/private/node-patches/src/fs.ts
@@ -168,7 +168,7 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
         origLstat(...args)
     }
 
-    fs.lstatSync = hideStackFrames((...args: any[]) => {
+    fs.lstatSync = hideStackFrames(function _lstatSync(...args: any[]) {
         const stats = origLstatSync(...args)
 
         if (!stats.isSymbolicLink()) {
@@ -258,7 +258,7 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
         origRealpathNative(...args)
     }
 
-    fs.realpathSync = hideStackFrames((...args: any[]) => {
+    fs.realpathSync = hideStackFrames(function _realpathSync(...args: any[]) {
         const str = origRealpathSync(...args)
         const escapedRoot: string | false = isEscape(args[0], str)
         if (escapedRoot) {
@@ -267,7 +267,9 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
         return str
     })
 
-    fs.realpathSync.native = hideStackFrames((...args: any[]) => {
+    fs.realpathSync.native = hideStackFrames(function _native_realpathSync(
+        ...args: any[]
+    ) {
         const str = origRealpathSyncNative(...args)
         const escapedRoot: string | false = isEscape(args[0], str)
         if (escapedRoot) {
@@ -331,7 +333,7 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
         origReadlink(...args)
     }
 
-    fs.readlinkSync = hideStackFrames((...args: any[]) => {
+    fs.readlinkSync = hideStackFrames(function _readlinkSync(...args: any[]) {
         const resolved = path.resolve(args[0])
 
         const str = path.resolve(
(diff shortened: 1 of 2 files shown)


js/private/node-patches/src/fs.ts
```diff
@@ -168,7 +189,7 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
         origLstat(...args)
     }
 
-    fs.lstatSync = hideStackFrames((...args: any[]) => {
+    fs.lstatSync = hideStackFrames(function _lstatSync(...args: any[]) {
```
jbedard, line 192: This is just so it's easier to see while profiling

js/private/node-patches/src/fs.ts
```diff
@@ -621,19 +642,16 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
         function _tryParent(): boolean {
             nested.push(path.basename(maybe))
             const dirname = path.dirname(maybe)
-            if (
-                !dirname ||
-                dirname == maybe ||
-                dirname == '.' ||
-                dirname == '/'
-            ) {
+            // TODO: also look for paths that are not within any 'roots'
```
jbedard, line 645: Both of these TODOs were causing test failures that I don't understand atm so I'm just going to do what I have below for now.

js/private/node-patches/src/fs.ts
```diff
@@ -621,19 +642,16 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
...
+            if (dirname == maybe || HOP_ROOTS[dirname] || HOP_ROOTS[maybe]) {
```
jbedard, line 646: This essentially does the same as before, but also stops when hitting `/var/.../bazel-out` or I think there's always (often?) also `/var/.../bazel-out/.../target.runfiles` which will be hit first.

Note that if something the `dirname` or `maybe` skip those paths this does not catch paths *above* those roots which is essentially what the TODO above is.

js/private/node-patches/src/fs.ts
```diff
@@ -621,19 +642,16 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
...
                 // not a link
                 return false
             }
             maybe = dirname
             return true
         }
         for (;;) {
+            // TODO: cache anything within 'roots'
```
jbedard, line 654: Simply caching everything (within 'roots' which we know is readonly) causes test failures as well that I don't understand. Maybe some npm modules write to their own directory? It was causing runtime issues with I think `require` statements failing.

js/private/node-patches/src/fs.ts
```diff
@@ -613,13 +617,55 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
         oneHop(loc, cb)
     }
 
+    const hopLinkCache = new Map<string, string | typeof HOP_NON_LINK>()
+    function readHopLink(p: string) {
```
jbedard, line 621: This caches all hop-related metadata that can be computed from just the path `p` off the filesystem, it does not take things like `nested` or `escapedHop` from the original method into account.

js/private/node-patches/src/fs.ts
```diff
@@ -628,38 +674,11 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
                 dirname == '/'
             ) {
                 // not a link
-                return false
+                return escapedHop
```
gregmagolan, line 677: this line of code stands up? Was returning `false` wrong here?

js/private/node-patches/src/fs.ts
gregmagolan: call it `readHopLinkSync` as per fs convention?

jbedard: Done (30s before I read this)

js/private/node-patches/src/fs.ts
```diff
@@ -628,38 +674,11 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
...
             }
+
+            nested.push(path.basename(maybe))
             maybe = dirname
-            return true
-        }
-        for (;;) {
```
gregmagolan, line 636: nice. I like how is simplified now

js/private/node-patches/src/fs.ts
jbedard: Sorry the diff got a bit ugly there... I inlined the `_tryParent` method now that it is only called once. Previously it was:
```
if (!_tryParent()) {
    return escapedHop
}
```
That false case is now this line...

gregmagolan: Oh I see. 👍

gregmagolan (approved):
🚀

PR #576, opened 2022-11-02
refactor: improve pattern for a derivate rule set with its own npm dependencies

gregmagolan:
This is a nice to have to make mirror versions of npm tools simpler in downstream rules sets.

e2e/rules_foo example updated so show how this will be used in downstream rule sets

In the future the new attributes could be documented and hoisted to the public API but `root_package`  first needs to be made more robust to handle pnpm workspaces.

Downstream usages:
https://github.com/aspect-build/rules_jasmine/pull/17
https://github.com/aspect-build/rules_jest/pull/51
https://github.com/aspect-build/rules_rollup/pull/48
https://github.com/aspect-build/rules_terser/pull/48
https://github.com/aspect-build/rules_webpack/pull/59


npm/private/npm_translate_lock.bzl
```diff
@@ -987,10 +984,27 @@ load("@aspect_rules_js//npm/private:npm_package_store.bzl", _npm_package_store =
...
 
-    rctx.file(_DEFS_BZL_FILENAME, "\n".join(defs_bzl_header + [""] + defs_bzl_body + [""]))
-    rctx.file(_REPOSITORIES_BZL_FILENAME, "\n".join(repositories_bzl))
+    rctx_files[_DEFS_BZL_FILENAME] = ["\n".join(defs_bzl_header + [""] + defs_bzl_body + [""])]
+    rctx_files[_REPOSITORIES_BZL_FILENAME] = ["\n".join(repositories_bzl)]
+
+    for filename, contents in rctx.attr.additional_file_contents.items():
+        if not filename in rctx_files.keys():
+            rctx_files[filename] = contents
+        elif filename[len(filename) - 4:] == ".bzl":
```
jbedard, line 993: There is an [endswith](https://bazel.build/rules/lib/string#endswith) method?

e2e/npm_link_package-esm/WORKSPACE
```diff
@@ -47,6 +47,10 @@ local_repository(
     path = "../rules_foo",
 )
 
-load("@rules_foo//:repositories.bzl", rules_foo_repositories = "repositories")
+load("@rules_foo//foo:repositories.bzl", rules_foo_repositories = "repositories")
 
-rules_foo_repositories()
+rules_foo_repositories(name = "foo")
+
+load("@foo//:repositories.bzl", foo_npm_repositories = "npm_repositories")
```
jbedard, line 54: Do both of these need to be named `repositories.bzl`? Would it be better to have different names?

gregmagolan: It would be but In this case `@foo` is just a customized `npm_translate_lock` and that currently generates a `repositories.bzl`. Suppose the name of the generated file could that customizable.

npm/private/npm_translate_lock.bzl
gregmagolan: 🤦

e2e/npm_link_package-esm/WORKSPACE
gregmagolan: Updated

jbedard: Yeah I like that more. It more clearly shows that one is repos while one is the npm packages of that repo 👍

e2e/rules_foo/foo/repositories.bzl
```diff
@@ -0,0 +1,30 @@
...
+        # Override the Bazel package where pnpm-lock.yaml is located and link to the specified package instead
+        root_package = "",
+        defs_bzl_filename = "npm_link_all_packages.bzl",
+        repositories_bzl_filename = "npm_repositories.bzl",
+        additional_file_contents = {
+            "BUILD.bazel": [
+                """load("//:npm_link_all_packages.bzl", "npm_link_all_packages")""",
+                """npm_link_all_packages(name = "node_modules")""",
+            ],
+            # Test that we can add statements the defs.bzl
```
kormide, line 23: nit: `the` -> `to`

kormide:
Will we want to use this pattern in `rules_jest` as well?

kormide:
I'm curious if this will help with this issue I'm having in rules_jest: https://github.com/aspect-build/rules_jest/pull/52

gregmagolan:
> Will we want to use this pattern in `rules_jest` as well?

Yes. I'm rolling it to rules_jest, rules_rollup, rules_terser & rules_webpack. rules_jasmine was the first I tried it on.

gregmagolan:
> I'm curious if this will help with this issue I'm having in rules_jest: [aspect-build/rules_jest#52](https://github.com/aspect-build/rules_jest/pull/52)

How do you think it would be related?

e2e/rules_foo/foo/repositories.bzl
gregmagolan: 🦖

kormide:
> > I'm curious if this will help with this issue I'm having in rules_jest: [aspect-build/rules_jest#52](https://github.com/aspect-build/rules_jest/pull/52)
> 
> How do you think it would be related?

The jest tool is having trouble resolving another npm dependency it needs. I'm not sure if those new undocumented features might help with resolving it. The reason why it can't resolve the junit output module isn't clear to me.

gregmagolan:
> > > I'm curious if this will help with this issue I'm having in rules_jest: [aspect-build/rules_jest#52](https://github.com/aspect-build/rules_jest/pull/52)
> > 
> > 
> > How do you think it would be related?
> 
> The jest tool is having trouble resolving another npm dependency it needs. I'm not sure if those new undocumented features might help with resolving it. The reason why it can't resolve the junit output module isn't clear to me.

I don't think this will help. The stack trace there looks like `jest-junit` is not a defined dependency of `jest-config`. I'll comment on that PR

npm/private/npm_translate_lock.bzl
```diff
@@ -32,6 +35,10 @@ _ATTRS = {
     "verify_node_modules_ignored": attr.label(),
     "link_workspace": attr.string(),
     "bzlmod": attr.bool(),
+    "root_package": attr.string(default = DEFAULT_ROOT_PACKAGE),
```
jbedard, line 38: Would there be any value prefixing these with `_` or something like that just to further emphasize that they are internal? (Actually... does bazel allow that?).

npm/private/npm_translate_lock.bzl
```diff
@@ -550,24 +556,17 @@ def _impl(rctx):
...
     # check that there isn't conflicting link workspaces if both are set
-    if link_workspace and rctx.attr.link_workspace and link_workspace != rctx.attr.link_workspace:
-        msg = "lock file workspace_name '{}' and link_workspace '{}' are both set but do not match".format(
-            link_workspace,
-            rctx.attr.link_workspace,
-        )
-        fail(msg)
-
-    if not link_workspace:
+    if rctx.attr.link_workspace:
```
jbedard, line 561: Let me know if you end up changing this at all...

e2e/rules_foo/foo/repositories.bzl
```diff
@@ -0,0 +1,30 @@
+"Shows how a custom rule set can export its npm dependencies"
+
+load("@aspect_rules_js//npm:npm_import.bzl", _npm_translate_lock = "npm_translate_lock")
+
+def foo_repositories(name):
```
jbedard, line 5: @thesayyn I think we should replace the rules_ts [npm_dependencies](https://github.com/aspect-build/rules_ts/blob/797610817564c68c56ac53a80a570be02867bf21/ts/private/npm_repositories.bzl#L75-L101) with something like this.

This "repositories function" should (in most rulesets) also accept things like the `npm_translate_lock(npmrc)` so it can be configured to load from other npm repos and other npm config can be set.

@gregmagolan we should try to define a standard way of doing this just to be consistent across repos. Potentially even create a macro (wrapping `npm_translate_lock`) that rulesets like this would use just to force consistency? We should think about that before we start adding params to these methods in each repo...

npm/private/npm_translate_lock.bzl
gregmagolan: Looked at this again. The change here is correct. I was just missing setting link_workspace in the downstream rule sets.

e2e/rules_foo/foo/repositories.bzl
gregmagolan: Potentially; tho the `foo_repositories()` pattern is pretty simple already so a wrapper macro wouldn't add much.

npm/private/npm_translate_lock.bzl
gregmagolan: `_` prefix makes them private so they can't be set from outside

gregmagolan: in could prefix them with something `internal_` but then when they are hoisted to the public API it is breaking for downstream rule sets

gregmagolan: the contract of undocumented means private API is reasonable

jbedard (approved):

e2e/rules_foo/foo/repositories.bzl
jbedard: I just worry it will slowly grow and not be simple, and if we don't think about it each ruleset will grow differently...

npm/private/npm_translate_lock.bzl
jbedard: Yeah I think the `_`/private + breaking for downstream rulesets makes it worth keeping the names as-is, just no telling people that they exist 😅

e2e/rules_foo/foo/repositories.bzl
gregmagolan: Suppose there is some room to standardize on the naming of the npm_repositories.bzl.... but lets leave that for a follow-up if there is merit

PR #579, opened 2022-11-02
Revert "perf: prevent preparing stack traces in fs *Sync patches"

### LINKED ISSUE #577: [Bug]: v1.6.0 breaks popular 3rd party tools like fast-glob
### What happened?

After upgrading rules_js to 1.6.0, some of my Node.js scripts immediately started to fail with the following error and no stack trace:

```
TypeError: Cannot assign to read only property 'stackTraceLimit' of function 'function Error() { [native code] }'
```

This took a number of hours to track down (especially due to no stack trace), and I found it's coming from the use of `fast-glob` (https://www.npmjs.com/package/fast-glob), which is a very popular library (38 million+ weekly downloads)

### Version

Development (host) and target OS/architectures: **MacOS Monterey**

Output of `bazel --version`: **5.3.0**

`WORKSPACE` or `MODULE.bazel` file: **rules_js v1.6.0** (bug was specifically introduced in this version; v1.5.3 is okay)

Language(s) and/or frameworks involved: **Node.js v14.18.1**

### How to reproduce

**BUILD**:

```starlark
js_binary(
    name = "test_fast_glob",
    entry_point = "test-fast-glob.js",
    data = [
        "//:node_modules/fast-glob"
    ]
)
```

**test-fast-glob.js**:

```js
const fastGlob = require('fast-glob');

const files = fastGlob.sync('some-directory/**');
console.log(files);
```

**CLI**: 

```
bazel run //:test_fast_glob
```

### Any other information?

This bug looks to have been specifically introduced by https://github.com/aspect-build/rules_js/pull/546

Couple of things with this:

1. Don't make the global `Error.stackTraceLimit` property unwritable. Can't have 3rd party libraries breaking entire scripts if they want to assign it. We also assign this property on our side when the default Node.js stack trace isn't long enough to track down some of our own errors. 
3. Setting the global `Error.stackTraceLimit` to `0` essentially destroys our ability to debug our Node.js programs. There needs to be a way to opt-out of this functionality (tag on the rule, environment variable, etc.). 
   * Also, any error message that gets raised should tell us how to opt-out of the "set-stack-trace-to-zero" functionality so that developers aren't simply left with a message and no stack trace without knowing that the stack trace has been removed by `rules_js`.

I also question the idea of modifying the global `Error` object at all from within rules_js, but if it increases performance like https://github.com/aspect-build/rules_js/pull/546 says, then at least having the above two points should work around the issues involved with it.

### Fund our work

jbedard:
This reverts commit ca2ba36a719e2f7920fdcb367007a23170020422.

Fixes #577

### DIFF
--- a/js/private/node-patches/fs.js
+++ b/js/private/node-patches/fs.js
@@ -43,61 +43,6 @@ const util = require("util");
 // also even though imports are mutable in typescript the cognitive dissonance is too high because
 // es modules
 const _fs = require('fs');
-/**
- * Reduce the information in Errors to both hide the origin and improve performance.
- * Any nested calls may be effected and also hidden.
- *
- * May potentially:
- * - hide the function from stack traces
- * - prevent errors thrown within the function from calculating stack traces
- *
- * TODO: currently can not hide computing of the stack trace within handleErrorFromBinding:
- * - https://github.com/nodejs/node/blob/1b971883e0c86f37468e421ffe5f39b585782ad6/lib/internal/fs/utils.js#L340-L356
- */
-function hideStackFrames(fn) {
-    const HiddenErrorProps = {
-        stackTraceLimit: {
-            // Minimize stack trace size similar to:
-            // https://github.com/nodejs/node/blob/30187baf2042dee683c5cc759e7ffad7c5a92169/lib/internal/errors.js#L508-L517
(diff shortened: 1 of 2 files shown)


gregmagolan (approved):
🌮

PR #581, opened 2022-11-03
refactor: use is_bazel_6_or_greater() helper to determine if bzlmod is supported

gregmagolan:


kormide:
Is deleting `def_checked.bzl` related to this?

gregmagolan:
> Is deleting `def_checked.bzl` related to this?

Yes. That one has an encoded `@@` or `@` now depending on the Bazel version so would break on CI since we test with both versions.

kormide (approved):

PR #584, opened 2022-11-03
renovate: upgrade aspect deps asap, reduce spam of various dep types/names

jbedard:
Still playing around with this config in rules_js before we put it into each of our repos and enable it everywhere. Main goal is the reduce spam while still automating things. Can hopefully enable automerge as well.

The summary:
* use standard/recommended config, group by monorepo etc, send PRs weekly by default
* group all patches (unless one of the below matches)
* group all bazel minor/patch
* group all in the aspect github org, open PRs instantly
* group a few spamy npm packages (types/rollup/webpack) and do them monthly
* otherwise it will open individual PRs weekly

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -16,38 +16,42 @@
 
     "packageRules": [
         {
-            "matchPackagePatterns": ["*"],
+            "matchPackagePatterns": ["npm"],
             "stabilityDays": 3
         },
         {
-            "matchPackagePatterns": ["*"],
-            "matchUpdateTypes": ["patch"],
             "groupName": "all patch updates",
-            "groupSlug": "all-patch"
+            "matchPackagePatterns": ["*"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -16,38 +16,42 @@
...
         {
-            "groupName": "Webpack",
-            "groupSlug": "webpack",
-            "matchPackagePatterns": ["^webpack-", "^@webpack-cli/", "-webpack-"]
+            "groupName": "Aspect",
+            "matchSourceUrlPrefixes": ["https://github.com/aspect-build/"],
+            "matchUpdateTypes": ["patch", "minor"],
+            "schedule": null
         },
         {
```
jbedard, line 38: From here down is npm specific stuff that probably won't go into non-js related repos.

gregmagolan (approved):
lgtm

PR #588, opened 2022-11-04
fix: allow for absolute path to node specified in target_tool_path in node_toolchain

gregmagolan:
I also ported over the e2e test from rules_nodejs for a vendored node but there still is no test coverage for the absolute path to node & npm as that test would depend on having a node on the local machine. Could be added to a CI only test at some point but not that important atm. I had a need for this change when testing against a locally built node from source which is how this came up. Tested locally that the absolute path to node works:

```
load("@rules_nodejs//nodejs:toolchain.bzl", "node_toolchain")

toolchain(
    name = "node_macos_toolchain",
    exec_compatible_with = [
        "@platforms//os:macos",
        "@platforms//cpu:x86_64",
    ],
    toolchain = ":node_macos",
    toolchain_type = "@rules_nodejs//nodejs:toolchain_type",
)

node_toolchain(
    name = "node_macos",
    target_tool_path = "/Users/greg/aspect/oss/node/out/Release/node",
)
```

```
register_toolchains("//:node_macos_toolchain")
```

### DIFF
--- a/js/private/js_binary.sh.tpl
+++ b/js/private/js_binary.sh.tpl
@@ -195,22 +195,42 @@ if [ ! -f "$entry_point" ]; then
     exit 1
 fi
 
-export JS_BINARY__NODE_BINARY="$JS_BINARY__RUNFILES/{{workspace_name}}/{{node}}"
-if [ ! -f "$JS_BINARY__NODE_BINARY" ]; then
-    logf_fatal "node binary '%s' not found in runfiles" "$JS_BINARY__NODE_BINARY"
-    exit 1
+node="$(_normalize_path "{{node}}")"
+if [ "${node:0:1}" = "/" ]; then
+    # A user may specify an absolute path to node using target_tool_path in node_toolchain
+    export JS_BINARY__NODE_BINARY="$node"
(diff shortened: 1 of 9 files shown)


js/private/js_binary.sh.tpl
```diff
@@ -195,22 +195,42 @@ if [ ! -f "$entry_point" ]; then
...
+        exit 1
+    fi
 fi
 if [ "$_IS_WINDOWS" -ne "1" ] && [ ! -x "$JS_BINARY__NODE_BINARY" ]; then
     logf_fatal "node binary '%s' is not executable" "$JS_BINARY__NODE_BINARY"
     exit 1
 fi
 
-npm={{npm}}
+npm="{{npm}}"
```
jbedard, line 218: Why does node need the `_normalize_path` but not npm?

jbedard (approved):

js/private/js_binary.sh.tpl
gregmagolan: the both have it, I just check that npm is not empty before I call it a few lines down

PR #250, opened 2022-06-27
refactor: expose bzl_library targets for generated bins

alexeagle:
This allows users to document their starlark code which loads from the generated bins.
It also makes bazel query look better when exploring the @npm repository, as there are now targets everywhere.

PR #589, opened 2022-11-04
fix: don't generate bzl_library targets in npm_translate_lock and npm_import by default

gregmagolan:
These targets are only needed in rules_js itself for a test case. Not currently needed in any downstream rule sets AFAIK and definitely not needed in user workspaces.

### DIFF
--- a/npm/npm_import.bzl
+++ b/npm/npm_import.bzl
@@ -325,6 +325,7 @@ def npm_translate_lock(
     additional_file_contents = kwargs.pop("additional_file_contents", {})
     repositories_bzl_filename = kwargs.pop("repositories_bzl_filename", None)
     defs_bzl_filename = kwargs.pop("defs_bzl_filename", None)
+    generate_bzl_library_targets = kwargs.pop("generate_bzl_library_targets", None)
 
     if len(kwargs):
         fail("Invalid npm_translate_lock parameter '{}'".format(kwargs.keys()[0]))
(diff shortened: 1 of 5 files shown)


npm/npm_import.bzl
```diff
@@ -325,6 +325,7 @@ def npm_translate_lock(
     additional_file_contents = kwargs.pop("additional_file_contents", {})
     repositories_bzl_filename = kwargs.pop("repositories_bzl_filename", None)
     defs_bzl_filename = kwargs.pop("defs_bzl_filename", None)
+    generate_bzl_library_targets = kwargs.pop("generate_bzl_library_targets", None)
```
jbedard, line 328: `False` instead of `None` since `generate_bzl_library_targets` is `attr.bool()`?

jbedard (approved):

npm/npm_import.bzl
gregmagolan: `None` implies the default gets used which is better so the default doesn't have to be encoded in two places

jbedard: Isn't the second param of `pop` the value returned if the first param is not in the map?

jbedard: Or you mean passing `None` from here into the actual rule means it will use the rule default?

alexeagle (commented):
We did add these for a reason, explained in the PR that introduced them
https://github.com/aspect-build/rules_js/pull/250
 so that you can `bazel query` to find out what exists, requested by @mattem

gregmagolan:
Hmmm. That's problematic because of the bzl_library graph up to aspect_bazel_lib that depends on the something that doesn't exist in most cases.

gregmagolan:
Guess we can opt-in for silo if we need them. Presumably most users won't?

alexeagle:
I think we made a wrong turn somewhere when removing the old is_windows attribute set by wrapper macros...

alexeagle:
Can we revert this or flip the default back?

Issue #362, opened 2022-08-05
ESM imports escape the sandbox & runfiles


gregmagolan:
This is the underlying reason why we current need `--preserve-symlinks-main` on by default in js_library so the `.mjs` entry points don't escape their runfiles.

It is also the reason mocha was observed to escape the sandbox in the repro https://github.com/aspect-build/rules_js/pull/353#issuecomment-1206659854 and likely related to https://github.com/aspect-build/rules_js/issues/347.

This affects `.mjs` entry points and programs that use esm imports. For example, mocha uses the `import()` built-in that is affected.

The relevent code in node is in resolve.js where there is a realPathSync call

https://github.com/nodejs/node/blob/4c5b96b376aa778cbe651362c29a26e0d4c32ccd/lib/internal/modules/esm/resolve.js#L310

This realPathSycn call does not get the fs monkey paches while the cjs require loader's realPathSync call does. It is unclear why.

gregmagolan:
We suspect the underlying reason why the fs patches apply to the commonjs path and not the esm path in node is the fs import style in https://github.com/nodejs/node/blob/4c5b96b376aa778cbe651362c29a26e0d4c32ccd/lib/internal/modules/esm/resolve.js

```
const {
  realpathSync,
  statSync,
  Stats,
} = require('fs');
```

while the commonjs import is

```
const fs = require('fs');
```

Presumably both modules are evaluated before our fs patches, which are done in a `--require` from the command line. Our patches would apply only to the commonjs path since it has a reference to the fs module which we monkey patch while the esm path has references to the fs functions it needs directly which are not monkey patched.

cgrindel:
Product Planning: still relevant

gregmagolan:
I tested the changed needed in Node.js to resolve this and it works as expected. The change is just https://github.com/aspect-build/node/commit/6616ddd629f9417e0ac2826910ff9756b01cfe34.

gregmagolan:
Minimal repro: https://github.com/gregmagolan/repro_rules_js_362

mkg20001:
https://nodejs.org/api/module.html#modulesyncbuiltinesmexports

Does this fix the issue?

gregmagolan:
> https://nodejs.org/api/module.html#modulesyncbuiltinesmexports
> 
> Does this fix the issue?

[...]

PR #591, opened 2022-11-05
fix: sync fs esm library after patching cjs version

jbedard:
For example as a test... without the `syncBuiltinESMExports` the last log (5) is false:
```
    import('fs').then(importFS => {
        console.log(1, origReaddirSync === fs.readdirSync)          // pre patched
        console.log(2, origReaddirSync === importFS.readdirSync)    // pre patch
        patchfs(fs, roots)
        console.log(3, origReaddirSync !== fs.readdirSync)          // post patched
        console.log(4, origReaddirSync === importFS.readdirSync)    // pre synced
        module.syncBuiltinESMExports()
        console.log(5, origReaddirSync !== importFS.readdirSync)    // post sync
    });
```

### DIFF
--- a/js/private/node-patches/register.js
+++ b/js/private/node-patches/register.js
@@ -35,7 +35,8 @@ if (
     JS_BINARY__PATCH_NODE_FS != '0' &&
     JS_BINARY__FS_PATCH_ROOTS
 ) {
-    const fs = require('fs')
+    const fs = require('node:fs')
+    const module = require('node:module')
     const roots = JS_BINARY__FS_PATCH_ROOTS.split(':')
     if (JS_BINARY__LOG_DEBUG) {
         console.error(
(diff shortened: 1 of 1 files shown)


matthewjh:
@jbedard, will this address https://github.com/aspect-build/rules_js/issues/362?

jbedard:
No it does not.

In #362 the fs methods are destructed using cjs: `{...} = require('fs')`. This PR addresses the case where the fs methods are destructed using esm: `import {...} from 'fs'`, and we haven't actually observed this esm case anywhere so I'm not sure if this will actually land.

matthewjh:
I see. Well, it seems useful to me, just in case someone in 1p code or a library decides to do this.

What are the downsides to merging?

jbedard:
Just the possibility of breaking something. I'd prefer merging and giving it a shot though...

PR #601, opened 2022-11-11
test: explicitly set devserver test ports

jbedard:
I've seen this randomly failing, specifically timing out I think due to a port conflict?

If I'm understanding this correctly the multirun test was assuming the first run would take port 8080 and the second would take 8081 because 8080 was in use? This is a race condition and I'm finding the second one often also tries 8080 and then fails, the "waiting (8081)" loop then loops forever.

I found the logging info useful when trying to debug so leaving it in.

### DIFF
--- a/e2e/js_run_devserver/multirun_test.sh
+++ b/e2e/js_run_devserver/multirun_test.sh
@@ -12,6 +12,8 @@ _sedi () {
   sed "${sedi[@]}" "$@"
 }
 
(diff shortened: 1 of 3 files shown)


jbedard:
The test seems flaky for other reasons as well..? https://github.com/aspect-build/rules_js/actions/runs/3442592504/jobs/5743326071

gregmagolan (approved):
Nice. This should make this test more robust.

PR #604, opened 2022-11-12
chore: update to Aspect CLI Pro 4.2.2

gregmagolan:


.aspect/cli/plugins.yaml
```diff
@@ -0,0 +1,13 @@
+# Plugins fail on CI: https://github.com/aspect-build/rules_js/actions/runs/3451833892/jobs/5761331944
+# ```
+# Error: failed to run test command: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (7366885 vs. 4194304)
+# Error: failed to run test command: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (7366885 vs. 4194304)
+# Error: failed to run test command: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (8031334 vs. 4194304)
+# Error: failed to run test command: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (8031334 vs. 4194304)
+# ```
+# - name: plugin-aspect-pro
```
jbedard, line 8: Are these plugins what would be enabled if the file didn't exist?

.aspect/cli/plugins.yaml
```diff
@@ -0,0 +1,13 @@
+# Plugins fail on CI: https://github.com/aspect-build/rules_js/actions/runs/3451833892/jobs/5761331944
```
jbedard, line 1: Upgrading caused this? So it's a cli issue?

.aspect/cli/plugins.yaml
gregmagolan: I attempted to add the plugins in this PR but there is an issue with the CLI so just left them commented out and created a silo issue for it instead.

.aspect/cli/plugins.yaml
gregmagolan: The issue exists on both 4.2.1 & 4.2.2 of the CLI. Just adding the plugins breaks.

jbedard (approved):

PR #605, opened 2022-11-12
fix: include transitive source in js_run_devserver sandbox

gregmagolan:
Resolves issue reported on Bazel Slack https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1667931292061319?thread_ts=1667857230.697759&cid=CEZUUKQ6P

### DIFF
--- a/e2e/js_run_devserver/src/BUILD.bazel
+++ b/e2e/js_run_devserver/src/BUILD.bazel
@@ -7,15 +7,25 @@ http_server_bin.http_server_binary(
     name = "http_server",
 )
 
+# A trivial graph of sources to include in js_run_devserver data to test
+# that transitive sources are include
+js_library(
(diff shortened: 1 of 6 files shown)


e2e/js_run_devserver/src/BUILD.bazel
```diff
@@ -47,6 +57,7 @@ js_run_devserver(
         ":http_root",
         "//:node_modules/http-server",
     ],
+    log_level = "debug",
```
alexeagle, line 60: Intentional?

gregmagolan: Yes, if these tests ever fail then its nice to have debug logs in the output already on

alexeagle (approved):

PR #610, opened 2022-11-14
build: use Aspect CLI in e2e tests

gregmagolan:
Except for js_run_devserver which seems to be failing on CI in a way that I can't repro locally. 

https://github.com/aspect-build/rules_js/actions/runs/3463927116/jobs/5784823402

### DIFF
--- a/e2e/bzlmod/.bazeliskrc
+++ b/e2e/bzlmod/.bazeliskrc
@@ -1,0 +1,1 @@
+../../.bazeliskrc
--- a/e2e/js_image/.bazeliskrc
+++ b/e2e/js_image/.bazeliskrc
@@ -1,0 +1,1 @@
+../../.bazeliskrc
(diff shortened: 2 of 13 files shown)


kormide (approved):

jbedard (approved):

jbedard:
For the changed test... are you seeing something such as https://github.com/aspect-build/rules_js/pull/601 ?

gregmagolan:
> For the changed test... are you seeing something such as #601 ?

It seems to be timing out on the "added a new file to a target in BUILD file" test case on CI but I can't repro locally 🤷

jbedard:
That's a different error then what that PR was fixing. I had one other error while doing that PR that I didn't fix though, I don't recall the details though.

gregmagolan:
Those js_run_devserver tests just seem flaky in general and too dependent on timing. They could still use some work but having flaky ones is still better than not having any coverage for that rule.

jbedard:
Yeah I don't think we shouldn't disable them, that would be worse imo, but it does seem like they need some work to see why they are so flaky.

I was able to reproduce the port one locally but clearly there's more.

PR #618, opened 2022-11-18
DNL: Testing --remote_download_outputs=toplevel issue with Bazel 6.0.0rc1 & rc2

gregmagolan:
Based on testing it looks like the issue happens with `--remote_download_outputs=toplevel` or `--remote_download_outputs=minimal` but only with sandboxed execution and when there is a `--disk_cache` or a `--remote_cache` configured.

The symptom seems to be that tree artifacts are just not present in the sandbox under these conditions. npm deps fail to resolve since the virtual store tree artifacts are not present.

### DIFF
--- a/.github/workflows/ci.bazelrc
+++ b/.github/workflows/ci.bazelrc
@@ -10,7 +10,7 @@ build:local --disk_cache=~/.cache/bazel
 
 # Generic remote cache
 build --remote_local_fallback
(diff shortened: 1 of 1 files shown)


gregmagolan:
https://github.com/bazelbuild/bazel/issues/16789 is the issue and [fix](https://github.com/bazelbuild/bazel/pull/16812) should get into the 6.0.0 release

PR #624, opened 2022-11-18
fix: fix npm_imports and packages list unaligned bug

gregmagolan:
Tricky one. npm_imports can fall out of line from packages list when one of the packages is a `file:` link. Code was relying on the back practice of indexes lining up in two different lists 🤦


gregmagolan:
Ugh. The test case is really tricky since pnpm generally puts `file:` packages at the end of the list of packages. The only repro I had is in the vscode repo and I'm not sure why pnpm behaved differently there. Going to land this as-is as it is strictly correct and better than using the index.

PR #625, opened 2022-11-18
perf: remove unnecessary additiona_files.to_list() in npm_package

gregmagolan:
The copy_to_direction action accepts a depset https://github.com/aspect-build/bazel-lib/blob/main/lib/private/copy_to_directory.bzl#L744


jbedard (approved):
Nice 👍

PR #626, opened 2022-11-18
feat(npm_package): add attributes to npm_package to configure what JsInfo files & if runfiles are included in list of files to copy

gregmagolan:
This allows users seeing perf hits on the `additional_files.to_list()` in underlying `copy_to_directory_action` to reduce the number of files in the depset.

Long term fix is still to move the replace logic in `copy_to_directory_action` to runtime so no `to_list()` needs to be done there but that is non-trivial. These attributes are still valid when that happens.


jbedard:
Did we have a specific case where this was required?

kormide (approved):

gregmagolan:
> Did we have a specific case where this was required?

Adobe is hitting performance problems in their gigantic repo with 800+ importers and 10,000+ packages so this was a feature request from them so they can tune until we get a runtime implementation of the replacer so we no longer need the to_list

jbedard (approved):

PR #628, opened 2022-11-19
docs: add note to npm_package include_runfiles attribute that it can add performance overhead and will be flipped to False for rules_js 2.0

gregmagolan:


npm/private/npm_package.bzl
```diff
@@ -153,7 +153,13 @@ If unset, a npm_link_package that references this npm_package must define the pa
         default = True,
     ),
     "include_runfiles": attr.bool(
-        doc = """When True, default runfiles from `srcs` targets are included in the list of available files to copy.""",
+        doc = """When True, default runfiles from `srcs` targets are included in the list of available files to copy.
```
alexeagle, line 156: I wish it said more about when I would need this. Do I need it only if I'm trying to ship a "self-contained" npm package?

alexeagle (approved):

npm/private/npm_package.bzl
gregmagolan: I'll add some scenarios when it is needed to the docstring

kormide (approved):

PR #643, opened 2022-11-25
doc: make the release snippet easier to use

alexeagle:
Right now we tell everyone to copy-paste a thing which assumes they have typescript

### DIFF
--- a/e2e/workspace/BUILD.bazel
+++ b/e2e/workspace/BUILD.bazel
@@ -1,0 +1,3 @@
+load("@npm//:defs.bzl", "npm_link_all_packages")
+
+npm_link_all_packages(name = "node_modules")
--- a/.github/workflows/workspace_snippet.sh
+++ b/.github/workflows/workspace_snippet.sh
@@ -41,5 +41,5 @@ http_archive(
 )
 EOF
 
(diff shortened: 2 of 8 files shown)


gregmagolan:
How about just removing the "bins" entirely from the snippet instead?

gregmagolan:
Although, your diff made me think of adding this,
````
# OPTIONAL: reference your .npmrc file if you have one in your repository to pick-up registry & auth configurations
npmrc = "//.npmrc"
```

alexeagle:
I think the `bins` entry here is load-bearing for the e2e to work though

alexeagle:
added .npmrc, good idea

e2e/workspace/BUILD.bazel
```diff
@@ -0,0 +1,3 @@
+load("@npm//:defs.bzl", "npm_link_all_packages")
+
+npm_link_all_packages(name = "node_modules")
```
gregmagolan, line 3: should also add this to `ci.yaml` so it doesn't bit rot & remote the `---SNIP---` from the old location

gregmagolan (approved):

e2e/workspace/BUILD.bazel
gregmagolan: will do in follow-up

PR #646, opened 2022-11-25
Support for NPM basic auth

### LINKED ISSUE #623: [FR]: Basic auth for npm registry

The npm authentication implemented by this repository only supports token-based authentication.

### Describe the feature

My company uses Azure DevOps Artifacts. When configuring your local .npmrc file it requires us to use npm's basic auth when fetching packages. You can see their guide on how to configure your .npmrc file here - https://learn.microsoft.com/en-us/azure/devops/artifacts/npm/scopes?view=azure-devops#set-up-credentials

Without this, our pnpm local installs work, but anything downloaded from the rules here do not.

### Fund our work

LavaToaster:
Resolves #623 

~~@gregmagolan is working on a base64 decoder, this PR will not work until that is complete.~~


npm/private/npm_translate_lock.bzl
```diff
@@ -868,14 +906,27 @@ load("@aspect_rules_js//npm/private:npm_package_store.bzl", _npm_package_store =
...
         npm_token = None
+        npm_auth_username = None
+        npm_auth_password = None
         match_len = 0
         for (auth_registry, auth_token) in npm_tokens.items():
             if _registry.startswith(auth_registry) and len(auth_registry) > match_len:
                 npm_token = auth_token
                 match_len = len(auth_registry)
 
+        if not npm_token:
```
gregmagolan, line 917: I'd drop the `if not npm_token:` here. Let npm_import deal with the case of both being set.

It looks like right now it will give preference to `npm_auth`. Should it fail or warn if both happen to be set in the `.npmrc` for a package or just silently go with one of them>

npm/private/npm_import.bzl
```diff
@@ -382,7 +382,13 @@ def _download_and_extract_archive(rctx):
             "pattern": "Bearer <password>",
             "password": rctx.attr.npm_auth,
         },
-    } if rctx.attr.npm_auth else {}
+    } if rctx.attr.npm_auth else {
```
gregmagolan, line 385: Can you switch this to:
```
if rctx.attr.npm_auth:
   ...
elif rctx.attr.npm_auth_username and rctx.attr.npm_auth_password:
   ...
else:
   ...
```
Easier to read in that case.

Should the rule `fail()` hard if both `npm_auth` and `npm_auth_username` and `npm_auth_password` are set?

Should the rule `fail()` hard if only one of `npm_auth_username` and `npm_auth_password` is set?

gregmagolan:
Looks like you need to run `bazel run //docs:update` as well to green up CI

npm/private/npm_translate_lock.bzl
```diff
@@ -173,8 +178,15 @@ def get_npm_auth(npmrc, npmrc_path, environ):
...
+                "@basic": "somewhere-else.com/basic",
         }
-        auth = (tokens, registries)
+        basic_auth = {
+            "somewhere-else.com/basic": {
+                "username": "someone",
+                "password": "hunter2",
+            },
+        }
+        auth = (tokens, registries, basic_auth)
```
gregmagolan, line 189: last comment, the `Returns` docstring below needs an update:

```
    Returns:
        A tuple (tokens, registries, basic_auth).
```

gregmagolan (approved):

gregmagolan:
🦖

PR #647, opened 2022-11-25
feat: update bzlmod api to be compatible with workspace rules

kormide:
https://github.com/aspect-build/rules_js/issues/644

PR #650, opened 2022-11-27
feat: automatically update pnpm lock file when it is out-of-date when update_pnpm_lock is True

gregmagolan:
`update_pnpm_lock` defaults to False so that `npm_translate_lock` rules from external repositories are never affected. `update_pnpm_lock` default to True if `yarn_lock` or `npm_package_lock` are set so the current `pnpm import` behavior is preserved.

API docstring:
```
        update_pnpm_lock: When True, the pnpm lock file wil be updated automatically when any of its inputs
            have changed since the last update.

            Defaults to False unless `npm_package_lock` or `yarn_lock` are set, in which case it defaults
            to True.

            A `.aspect/external_repository_hashes/npm_translate_lock_HASH` file will be created and used to determine
            when the `pnpm-lock.yaml` file should be updated. This file should be checked into the source control
            along with the `pnpm-lock.yaml` file.

            When the `pnpm-lock.yaml` file needs updating, `npm_translate_lock` will run either
            `pnpm install --lockfile-only` or `pnpm import`. The latter will be used if there
            is a `npm_package_lock` or `yarn_lock` specified.

            To update the `pnpm-lock.yaml` file manually either [install pnpm](https://pnpm.io/installation) and
            run `pnpm install --lockfile-only` or `pnpm import` use the Bazel-managed pnpm by running
            `bazel run -- @pnpm//:pnpm --dir $PWD install --lockfile-only` or `bazel run -- @pnpm//:pnpm --dir $PWD import`

            If the ASPECT_RULES_JS_FROZEN_PNPM_LOCK environment variable is set and `update_pnpm_lock` is True, the build
            will fail if the pnpm lock file needs updating. It is recommended to set this environment variable
            on CI when `update_pnpm_lock` is True.
```

`package_json` attribute (currently used only when `yarn_lock` or `npm_package_lock` are set) is deprecated in favor of adding all `package.json` files to the `data` attribute.

Examples failure on CI when `ASPECT_RULES_JS_FROZEN_PNPM_LOCK` is set and the action cache is out of date.

```
ERROR: `.aspect/rules/external_repository_action_cache/npm_translate_lock_NDg3NzUwNzA2NQ==` is out of date. `pnpm-lock.yaml` may require an update. To update run,

           bazel sync --only=npm
```


npm/private/npm_translate_lock.bzl
```diff
@@ -16,6 +16,7 @@ DEFAULT_DEFS_BZL_FILENAME = "defs.bzl"
 
 _ATTRS = {
     "pnpm_lock": attr.label(),
+    "pnpm_workspace": attr.label(),
```
kormide, line 19: Reminder to myself to add this in https://github.com/aspect-build/rules_js/pull/647.

gregmagolan: You should hold off on #647 as my refactoring here has grown rather large and will conflict with your changes.

gregmagolan: Once this is done & landed then #647 can be finished off

npm/private/npm_translate_lock.bzl
```diff
@@ -456,183 +488,531 @@ def _check_for_conflicting_public_links(npm_imports, public_hoist_packages):
...
+        input_hashes = parse_npmrc(rctx.read(expected_pnpm_lock_label))
+        if _dicts_match(input_hashes, translate_state["input_hashes"]):
+            # Calculated input hashes match saved input hashes; nothing to update
+            return False
+
+    if RULES_JS_FROZEN_PNPM_LOCK_ENV in rctx.os.environ.keys() and rctx.os.environ[RULES_JS_FROZEN_PNPM_LOCK_ENV]:
+        fail("""
+
+ERROR: {pnpm_lock} file's inputs have changed since the last update and an update may be necessary.
+       Run `bazel query @{name}//...` locally to apply updates.""".format(pnpm_lock = translate_state["pnpm_lock"], name = rctx.name))
```
gregmagolan, line 872: What's a better command to run than `bazel query @npm//...`?

gregmagolan: `bazel fetch @npm//...` does the wrong thing. I wish there was a `bazel sync <repo>` command.

e2e/npm_translate_lock/repositories_checked.bzl
```diff
@@ -1,9 +1,9 @@
-"@generated by @aspect_rules_js//npm/private:npm_translate_lock.bzl from //:pnpm-lock.yaml"
+"@generated by npm_translate_lock(name = 'npm')"
 
 load("@aspect_rules_js//npm:npm_import.bzl", "npm_import")
 
 def npm_repositories():
-    "Generated npm_import repository rules corresponding to npm packages in //:pnpm-lock.yaml"
```
jbedard, line 6: Isn't this still true?

e2e/npm_translate_lock_auth/pnpm-workspace copy.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
```
jbedard, line 1: This is a " copy" file...?

WORKSPACE
```diff
@@ -73,6 +73,20 @@ npm_translate_lock(
         "@aspect-test/c": "echo moo > cow.txt",
         "@aspect-test/c@2.0.2": "echo mooo >> cow.txt",
     },
+    data = [
```
jbedard, line 76: Can these just be globs?

npm/extensions.bzl
```diff
@@ -11,12 +11,20 @@ load("//npm/private:transitive_closure.bzl", "translate_to_transitive_closure")
 def _extension_impl(module_ctx):
     for mod in module_ctx.modules:
         for attr in mod.tags.npm_translate_lock:
-            lockfile = utils.parse_pnpm_lock(module_ctx.read(attr.pnpm_lock))
-            trans = translate_to_transitive_closure(lockfile, attr.prod, attr.dev, attr.no_optional)
+            # npm_translate_lock MUST run first incase it updates the pnpm-lock.yaml
+            # file read next when update_pnpm_lock is True
```
jbedard, line 15: "file read next"?

npm/extensions.bzl
```diff
@@ -11,12 +11,20 @@ load("//npm/private:transitive_closure.bzl", "translate_to_transitive_closure")
 def _extension_impl(module_ctx):
     for mod in module_ctx.modules:
         for attr in mod.tags.npm_translate_lock:
-            lockfile = utils.parse_pnpm_lock(module_ctx.read(attr.pnpm_lock))
-            trans = translate_to_transitive_closure(lockfile, attr.prod, attr.dev, attr.no_optional)
+            # npm_translate_lock MUST run first incase it updates the pnpm-lock.yaml
```
jbedard, line 14: Can they be put at the top of the method in their own loop so they all run before any of the other code below?

npm/npm_import.bzl
```diff
@@ -204,35 +206,55 @@ def npm_translate_lock(
...
 
-            Mandatory when using npm_package_lock or yarn_lock, otherwise must be unset.
+        yarn_lock: The `yarn.lock` file written by `yarn install`.
 
-        npm_package_lock: The package-lock.json file written by `npm install`.
+            Only one of [npm_package_lock, yarn_lock] may be set.
 
-            When set, the `package_json` attribute must be set as well.
-            Exactly one of [pnpm_lock, npm_package_lock, yarn_lock] should be set.
+        update_pnpm_lock: When True, the pnpm lock file wil be updated automatically when it is out-of-date.
```
jbedard, line 225: Can we specify what "out of date" means in this context? Maybe "out of sync with yarn/npm" is better terminology?

npm/npm_import.bzl
```diff
@@ -341,14 +363,9 @@ def npm_translate_lock(
 
         pnpm_version: pnpm version to use when generating the @pnpm repository. Set to None to not create this repository.
 
-        data: Data files required by this repository rule.
-
-            When any of these files changes it causes the repository rule to re-run.
+        package_json: Deprecated.
```
jbedard, line 366: Lets create an issue for 2.0 or some type of `TODO(2.0)` comment so we don't forget this?

npm/npm_import.bzl
```diff
@@ -204,35 +206,55 @@ def npm_translate_lock(
...
-            Exactly one of [pnpm_lock, npm_package_lock, yarn_lock] should be set.
+        update_pnpm_lock: When True, the pnpm lock file wil be updated automatically when it is out-of-date.
 
-        yarn_lock: The yarn.lock file written by `yarn install`.
+            Defaults to False unless `npm_package_lock` or `yarn_lock` are set, in which case it defaults
+            to True.
 
-            When set, the `package_json` attribute must be set as well.
-            Exactly one of [pnpm_lock, npm_package_lock, yarn_lock] should be set.
+            A `pnpm-lock.rules_js` file will be created next to `pnpm-lock.yaml` and used to determine when
```
jbedard, line 230: I wonder if we can add metadata to something such as the lockfile itself? If we control the lockfile can we put comments at the top?

npm/private/versions.bzl
```diff
@@ -676,4 +676,30 @@ PNPM_VERSIONS = {
     "6.34.0": "sha512-K79DAT5WtGjrUyLtxAyehtN8uy5cXkzT1v+7RIJ3gygCffoLaLlBoG5gcfjfnX7oZH5rxznv/v8tlMIMfZyzTA==",
     "7.9.0": "sha512-xkIVw73yJm/h5M4VvFIS5Q+gQCRDrp3r92g58PtcCK86aZCa7EQ6q6ivdfTAz0KsAVgloA6Anub28n6wju5v3w==",
     "7.9.1": "sha512-5vyV+FwZj5y31UDDsiq9xcVzF+mvS+IPdgAgkZ9rVxYPNKCCVwVn5LsHh9jcfuApiKYVgpyisAMcuSDbe1/C0Q==",
+    "7.9.2": "sha512-Xsp96ox9FmlEYuiN3+slAS1S9OXPPxa4JxEp4IX8tHIOlQAPRzLcsjkU9ZA5aQa8PVLGEmA9dQenp7sBpIMoCQ==",
```
jbedard, line 679: Add all these in a separate PR?

jbedard (commented):
Looking great 👍

WORKSPACE
kormide: I don't think you can glob in a WORKSPACE.

docs/faq.md
```diff
@@ -19,10 +19,12 @@ See [example](https://github.com/aspect-build/rules_ts/blob/74d54bda208695d7e899
 
 ## Bazel isn't seeing my changes to package.json
 
-rules_js relies on what's in the `pnpm-lock.yaml` file.
-Make sure your changes are reflected there.
+rules_js relies on what's in the `pnpm-lock.yaml` file. Make sure your changes are reflected there.
 
-Want a Bazel test to assert the lockfile isn't stale? See our `examples/assert_lockfile_to_to_date`.
+Set `update_lock_file` to True in your `npm_translate_lock` rule and Bazel will auto-update your `pnpm-lock.yaml` when
```
kormide, line 24: `update_pnpm_lock`?

npm/private/versions.bzl
gregmagolan: Yeah. I'll pre-factor those.

npm/npm_import.bzl
gregmagolan: ✅  left a `TODO(2.0)` which is what I've been using throughout

npm/npm_import.bzl
gregmagolan: Suppose we could but it doesn't seem ideal; for one running `pnpm i --lockfile-only` will remove anything we add.

I'm thinking of putting the "hash" file into an `.aspect/external_repository_hashes/` folder so it doesn't pollute the root of the repository

npm/npm_import.bzl
gregmagolan: I'll update "when any of its inputs have changed since the last update" which is technically most correct

npm/extensions.bzl
gregmagolan: 👍

npm/extensions.bzl
gregmagolan: 🤷

WORKSPACE
gregmagolan: Correct. You can't glob in a repository rule

docs/faq.md
gregmagolan: nice catch

e2e/npm_translate_lock/repositories_checked.bzl
gregmagolan: Yea. The comments were in flux here but I think it can go back to the pnpm-lock.yaml label again now that the code always depends on one being in the source tree even in the `pnpm import` case

e2e/npm_translate_lock_auth/pnpm-workspace copy.yaml
gregmagolan: Whoops. Meant to copy that elsewhere.

npm/private/versions.bzl
gregmagolan: ✅

.aspect/external_repository_hashes/npm_translate_lock_NDg3NzUwNzA2NQ==
```diff
@@ -0,0 +1,16 @@
+# Input hashes for repository rule: npm_translate_lock(name = "npm", pnpm_lock = "@//:pnpm-lock.yaml")
+//:pnpm-lock.yaml=1507139895
```
alexeagle, line 2: the format for the hash looks unusual, it's not a sha256 or a resource integrity hash or anything else I'm familiar with

.aspect/external_repository_hashes/npm_translate_lock_NDg3NzUwNzA2NQ==
```diff
@@ -0,0 +1,16 @@
+# Input hashes for repository rule: npm_translate_lock(name = "npm", pnpm_lock = "@//:pnpm-lock.yaml")
```
alexeagle, line 1: I still need help to understand why this file is needed. What goes wrong if we just use those files as inputs to the repository rule directly? Does it sometimes invalidate when it shouldn't? Fail to invalidate when it should?

docs/faq.md
```diff
@@ -19,10 +19,12 @@ See [example](https://github.com/aspect-build/rules_ts/blob/74d54bda208695d7e899
 
 ## Bazel isn't seeing my changes to package.json
 
-rules_js relies on what's in the `pnpm-lock.yaml` file.
-Make sure your changes are reflected there.
+rules_js relies on what's in the `pnpm-lock.yaml` file. Make sure your changes are reflected there.
 
-Want a Bazel test to assert the lockfile isn't stale? See our `examples/assert_lockfile_to_to_date`.
+Set `update_pnpm_lock` to True in your `npm_translate_lock` rule and Bazel will auto-update your `pnpm-lock.yaml` when
+any of its inputs change. When you do this, you'll have to specify all `package.json` files in your pnpm workspace and
```
alexeagle, line 25: something is wrong with this sentence at "workspace and required"

.aspect/external_repository_hashes/npm_translate_lock_NDg3NzUwNzA2NQ==
gregmagolan: yup; hash function which is the only reasonable choice in starlark returns an `int32` as the hash

.aspect/external_repository_hashes/npm_translate_lock_NDg3NzUwNzA2NQ==
gregmagolan: This allows the repo rule to update the `pnpm-lock.yaml` file in the source tree when its inputs change but only for the developer who makes the changes to the inputs. Another developer who pulls the changes will invalidate the repo rule but it will determine that the `pnpm-lock.yaml` file is up-to-date with its inputs so it won't run `pnpm install --lockfile-only` or `pnpm import` (depending on your SoT)

gregmagolan: This is also better than build action since developer machines typically don't share a remote cache with CI or with other developers; especially for developers on macos. Making the lock file update a build action would mean that every developer in a monorepo would have to have pnpm fetch 100% of the deps in the monorepo even if they don't build any js targets. `pnpm install` and `pnpm import` will cache the npm packages but that cost will always have to be paid to bootstrap. With the external repository hash, only javascript developers in a monorepo will have to ever fetch npm packages to their machine with pnpm.

gregmagolan: Perhaps `.aspect/external_repository_action_cache/` is clearer naming

alexeagle: Yeah I get it. You're basically creating something like an action graph inside repository rules. The first one is expected to invalidate frequently but is quick to run, and as long as it produces a stable output it prevents the next "action" from running.

If you're going to check in a file anyway, could it also contain the translated pnpm lock? Just make it a .bzl file that does both? That would make this more similar to how Java, Go and Python work when you vendor the "pinned" file as starlark. The translated file doesn't have to be portable to work as an external repo, like we ran into with rules_ts.

Does that make sense?

gregmagolan: Yeah. I think I know what you're getting at.

gregmagolan: The pnpm lock file is translated to a number of starlark files. There is the `npm_link_all_packages()` function that is the primary one but there are also the `package_json.bzl` pass-through stubs & the root BUILD file. Wouldn't make any sense to vendor just part of this generation since you'd still have to run the same generation code for the rest. I was thinking of vendoring in some intermediate format that is post calculating the transitive closure but that would require a lot more refactoring than already done here. So just checking the hashes of the inputs as is already done here is probably best for this step function forward IMO.

alexeagle: yup, that makes sense

WORKSPACE
```diff
@@ -94,14 +110,14 @@ npm_translate_lock(
         "@gregmagolan/test-a@0.0.1": ["//examples/npm_deps:patches/test-a@0.0.1.patch"],
     },
     pnpm_lock = "//:pnpm-lock.yaml",
-    pnpm_version = "6.32.19",
```
alexeagle, line 97: curious, why remove that?

e2e/js_run_devserver/.aspect/rules/external_repository_action_cache/npm_translate_lock_NDg3NzUwNzA2NQ==
```diff
@@ -0,0 +1,4 @@
+# Input hashes for repository rule: npm_translate_lock(name = "npm", pnpm_lock = "@//:pnpm-lock.yaml")
+//:pnpm-lock.yaml=1962368226
+//:.npmrc=664934919
```
alexeagle, line 3: should these be sorted? will the order be stable?

WORKSPACE
```diff
@@ -73,6 +73,21 @@ npm_translate_lock(
         "@aspect-test/c": "echo moo > cow.txt",
         "@aspect-test/c@2.0.2": "echo mooo >> cow.txt",
     },
+    data = [
```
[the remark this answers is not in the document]
gregmagolan: The code does this already

WORKSPACE
```diff
@@ -94,14 +110,14 @@ npm_translate_lock(
...
     pnpm_lock = "//:pnpm-lock.yaml",
-    pnpm_version = "6.32.19",
     public_hoist_packages = {
         # Instructs the linker to hoist the ms@2.1.3 npm package to `node_modules/ms` in the `examples/npm_deps` package.
         # Similar to adding `public-hoist-pattern[]=ms` in .npmrc but with control over which version to hoist and where
         # to hoist it. This hoisted package can be referenced by the label `//examples/npm_deps:node_modules/ms` same as
         # other direct dependencies in the `examples/npm_deps/package.json`.
         "ms@2.1.3": ["examples/npm_deps"],
     },
+    update_pnpm_lock = True,
```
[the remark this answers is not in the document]
gregmagolan: we wouldn't have to with this on; mainly for dog-fooding the feature

e2e/js_run_devserver/WORKSPACE
```diff
@@ -24,8 +24,10 @@ npm_translate_lock(
             "http-server": "./bin/http-server",
         },
     },
+    data = ["//:package.json"],
     npmrc = "//:.npmrc",
     pnpm_lock = "//:pnpm-lock.yaml",
+    update_pnpm_lock = True,
```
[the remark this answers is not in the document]
gregmagolan: primarily for dog-fooding the feature and for e2e test coverage; but I can it from most places in this PR to reduce the size and still get good e2e test coverage

npm/private/utils.bzl
```diff
@@ -49,16 +56,35 @@ def _parse_pnpm_name(pnpmName):
         fail(msg)
     return (segments[0], segments[1])
 
-def _parse_pnpm_lock(lockfile_content):
-    """Parse a pnpm lock file.
+def _parse_pnpm_lock(content):
+    """Parse the content of a pnpm-lock.yaml file.s
```
kormide, line 60: Typo, trailing `s`.

npm/private/utils.bzl
```diff
@@ -49,16 +56,35 @@ def _parse_pnpm_name(pnpmName):
...
 
     Returns:
-        dict containing parsed lockfile
+        A tuple of (importers dict, packages dict)
     """
-    return _parse_yaml(lockfile_content)
+    parsed = _parse_yaml(content)
+
+    if not types.is_dict(parsed):
+        fail("lockfile should be a starlark dict")
```
kormide, line 71: When does this happen? Is there an unhandled parser case I could fix?

npm/private/utils.bzl
```diff
@@ -156,6 +182,113 @@ def _npm_registry_download_url(package, version, registries, default_registry):
 def _is_git_repository_url(url):
     return url.startswith("git+ssh://")
 
+def _is_url(url):
+    return url.find("://") != -1
```
kormide, line 186: How about `is_probably_url`?

npm/private/utils.bzl
gregmagolan: I believe this is just a defence in depth; I haven't seen this hit

npm/private/utils.bzl
gregmagolan: 👍

npm/private/utils.bzl
gregmagolan: Right. That function is not very robust. I'll just move it back inline to the one file that needs it.

npm/private/utils.bzl
gregmagolan: (this line was just moved around in this PR)

WORKSPACE
jbedard: did we want to do this?

jbedard:
What are the new `docs/*.md` files.

jbedard:
Can we do all the file permission changes in a separate PR?

WORKSPACE
alexeagle: I think this is here to provide a bit of test coverage for that attribute so yeah, let's keep it and just update the version

e2e/npm_translate_lock/package.json
```diff
@@ -1,12 +1,15 @@
 {
     "private": true,
     "dependencies": {
-        "debug": "^4.3.4",
-        "is-odd": "^3.0.1",
-        "semver": "^7.3.8"
+        "debug": "4.3.4",
```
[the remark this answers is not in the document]
gregmagolan: its a preference thing IMO; I prefer pinned deps because I know exactly what the inputs to the lock file generator are; the renovator bots for GH will update the deps one way or the other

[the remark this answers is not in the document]
jbedard: renovator bots will update it either way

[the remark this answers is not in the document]
alexeagle: okay I'll stop picking on you... it just gives developers a false sense of confidence that they think they've pinned the version, and are losing data about whether a given upgrade is actually safe. So it makes me crazy that "most of us are doing it wrong" and I get on my soapbox. It's nice on here!

npm/private/utils.bzl
```diff
@@ -156,6 +182,110 @@ def _npm_registry_download_url(package, version, registries, default_registry):
...
+    result = str(hash(s))
+    if result.startswith("-"):
+        # Bazel's hash() resolves to a signed 32bit number [-2,147,483,648 to 2,147,483,647].
+        # Convert manually to a unique unsigned number here.
+        result = result[1:]
+        if len(result) <= 9:
+            padding = "0" * (9 - len(result))
+            result = "3" + padding + result
+        elif result[0] == "1":
+            result = "4" + result[1:]
```
jbedard, line 201: Why replace "1" with "4"? That makes this non-unique...?

npm/private/utils.bzl
```diff
@@ -156,6 +182,110 @@ def _npm_registry_download_url(package, version, registries, default_registry):
...
+def _consistent_label_str(label):
+    return "//{}:{}".format(
+        # Starting in Bazel 6, the workspace name is empty for the local workspace and there's no other way to determine it.
+        # This behavior differs from Bazel 5 where the local workspace name was fully qualified in str(label).
+        label.package,
+        label.name,
+    )
+
+# Copies a file from the external repository to the same relative location in the source tree
+# TODO: test on Windows
```
jbedard, line 228: Since we have windows specific logic we should probably do this?

gregmagolan: yea, I'm going to dust off my windows box before landing

npm/private/utils.bzl
```diff
@@ -156,6 +182,110 @@ def _npm_registry_download_url(package, version, registries, default_registry):
...
+set -o errexit -o nounset -o pipefail
+if [ ! -f $1 ]; then exit 42; fi
+""", executable = True)
+        result = rctx.execute(["./_exists.sh", str(p)])
+    if result.return_code == 0:  # file exists
+        return True
+    elif result.return_code == 42:  # file does not exist
+        return False
+    else:
+        fail(INTERNAL_ERROR_MSG)
```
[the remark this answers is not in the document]
gregmagolan: should never happen but if it does then the fail() callstack will have the line number for the bug report

npm/private/utils.bzl
gregmagolan: bash hash `function` returns a signed int `[-2,147,483,648 to 2,147,483,647]` which bothers me to put in to the hashes file.

So I did some string manipulation to make it a unique unsigned int string:
`-xxx,xxx,xxx` to `3,xxx,xxx,xxx`
`-1,xxx,xxx,xxx` to `4,xxx,xxx,xxx`
`-2,xxx,xxx,xxx` to `5,xxx,xxx,xxx`

jbedard: But now both `-1,xxx,xxx,xxx` and `4,xxx,xxx,xxx` result in the same hash (I think).

Shouldn't we prepend to make it unique instead of replacing the first char? Just replace any `-` with a `2` and put a `1` in front of non-negatives?

npm/private/utils.bzl
```diff
@@ -156,6 +182,110 @@ def _npm_registry_download_url(package, version, registries, default_registry):
 def _is_git_repository_url(url):
     return url.startswith("git+ssh://")
 
+def _to_registry_url(url):
+    return "{}://{}".format(DEFAULT_REGISTRY_PROTOCOL, url) if url.find("//") == -1 else url
+
+def _default_registry():
+    return "{}://registry.npmjs.org/".format(DEFAULT_REGISTRY_PROTOCOL)
```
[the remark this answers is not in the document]
gregmagolan: 👍

npm/private/utils.bzl
gregmagolan: `4,xxx,xxx,xxx` is out of the positive range so will never happen

gregmagolan: max positive int32 is 2,147,483,647

jbedard: Oh I see, that is quite easy to miss reading this.

If we're already making it too high to be an int I'd still vote to not trim it. Just put a `0` in front of a positive number and replace `-` with `1` in front of negative? 🤷

e2e/npm_translate_lock/package.json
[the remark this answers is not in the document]
alexeagle: I've been fixing the renovate to stop doing that wrong thing. there's a "preserve semver"  option that we want on. IMO if you have a lockfile, then renovate default should be to leave your dependency ranges alone.

https://docs.renovatebot.com/dependency-pinning/

npm/private/utils.bzl
gregmagolan: sure. we can tweak in a follow-up.

e2e/npm_translate_lock/package.json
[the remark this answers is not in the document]
gregmagolan: sounds like a longer discussion; I personally I think `^` and `~` are left haphazardly and don't have much meaning anyway since you can't be confident that a bump in version will work or not until renovate makes the PR either way. and in general I think renovate should always bump patch & minor and never major on PRs unless explicitly configured to ignore a certain package in its config. this keeps the logic for what to bump in the renovate config and not in the package.json.

WORKSPACE
gregmagolan: roger roger

npm/private/utils.bzl
jbedard: Normally that means it never happens :P

Also won't tweaking it in the future break caches?

e2e/npm_translate_lock/package.json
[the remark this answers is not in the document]
alexeagle: This isn't our area of expertise or influence so I think we can just leave it how it is

docs/faq.md
```diff
@@ -19,10 +19,12 @@ See [example](https://github.com/aspect-build/rules_ts/blob/74d54bda208695d7e899
 
 ## Bazel isn't seeing my changes to package.json
 
-rules_js relies on what's in the `pnpm-lock.yaml` file.
-Make sure your changes are reflected there.
+rules_js relies on what's in the `pnpm-lock.yaml` file. Make sure your changes are reflected there.
 
-Want a Bazel test to assert the lockfile isn't stale? See our `examples/assert_lockfile_to_to_date`.
+Set `update_pnpm_lock` to True in your `npm_translate_lock` rule and Bazel will auto-update your
+`pnpm-lock.yaml` when any of its inputs change. When you do this, add all files required to
```
jbedard, line 25: "add all files required to for pnpm to generate"

npm/private/test/repositories_checked.bzl
```diff
@@ -1,4 +1,4 @@
-"@generated by @aspect_rules_js//npm/private:npm_translate_lock.bzl from //:pnpm-lock.yaml"
```
jbedard, line 1: What caused this to change? Don't want "aspect_rules_js" in there?

npm/private/utils.bzl
jbedard: can we drop the comment then?

npm/private/test/repositories_checked.bzl
gregmagolan: just a tweak so the repository rule type & name is identified instead of the source file

gregmagolan: this is all behind the scenes anyway will never seen by the average user

kormide (approved):

alexeagle:
We'll need an update to https://github.com/aspect-build/rules_js/blob/main/docs/faq.md#can-i-use-bazel-managed-pnpm in this PR or right after.

npm/private/repository_label_store.bzl
```diff
@@ -0,0 +1,115 @@
+"""In order to avoid late restarts, repository rules should pre-compute dynamic labels and
```
alexeagle, line 1: TODO to upstream this to bazel-lib? Nothing JS/pnpm specific about it. Also looks nicely unit-testable

gregmagolan: Yea, I was thinking it would end up there. Have to learn how to unit-test repository rule code tho since this one depends on `rctx.path`

npm/private/npm_translate_lock_state.bzl
```diff
@@ -0,0 +1,398 @@
...
+        if utils.dicts_match(input_hashes, priv["input_hashes"]):
+            # Calculated input hashes match saved input hashes; nothing to update
+            return False
+    return True
+
+########################################################################################################################
+def _write_action_cache(priv, rctx, label_store):
+    contents = [
+        "# Input hashes for repository rule: npm_translate_lock(name = \"{}\", pnpm_lock = \"{}\")".format(rctx.name, label_store.label("pnpm_lock")),
+    ]
```
alexeagle, line 278:
```suggestion
        "# This file should be checked into version control along with the pnpm-lock.yaml file.",
    ]
```

npm/private/npm_translate_lock_state.bzl
```diff
@@ -0,0 +1,398 @@
...
+    priv["importers"] = importers
+    priv["packages"] = packages
+
+########################################################################################################################
+def _has_workspaces(priv):
+    importer_paths = priv["importers"].keys()
+    return importer_paths and (len(importer_paths) > 1 or importer_paths[0] != ".")
+
+########################################################################################################################
+def _update_pnpm_lock(priv):
```
alexeagle, line 334: "update" seems like an action would be performed. I wonder if these getter functions ought to be prefixed with get_

npm/private/npm_translate_lock_state.bzl
gregmagolan: Good call

npm/private/repository_label_store.bzl
gregmagolan: https://github.com/aspect-build/rules_js/issues/693

gregmagolan:
> We'll need an update to https://github.com/aspect-build/rules_js/blob/main/docs/faq.md#can-i-use-bazel-managed-pnpm in this PR or right after.

Yes. Although it will only bootstrap if you have a yarn or npm lock file. It won't bootstrap without any lock file specified since the repository rule has no idea which folder you want to run pnpm in.

Given a single `package.json` or a set of `package.json` files and a `pnpm-workspace.yaml` file in `data` it could. The tricky bit is the sequencing for users since you can't pass a label to a repository rule to a non-existant file.

npm/private/utils.bzl
```diff
@@ -156,6 +182,115 @@ def _npm_registry_download_url(package, version, registries, default_registry):
...
 
+def _to_registry_url(url):
+    return "{}://{}".format(DEFAULT_REGISTRY_PROTOCOL, url) if url.find("//") == -1 else url
+
+def _default_registry():
+    return _to_registry_url("registry.npmjs.org/")
+
+def _hash(s):
+    result = str(hash(s))
+    if result.startswith("-"):
```
alexeagle, line 193: meh, this seems like more code for us to maintain that's not really worth it just because you don't particularly enjoy the negative numbers. Is there any benefit to them all being positive?

npm/private/npm_translate_lock_rule.bzl
```diff
@@ -0,0 +1,220 @@
...
+        packages,
+        state.root_package(),
+        state.default_registry(),
+        state.npm_registries(),
+        state.npm_tokens(),
+        state.npm_basic_auth(),
+        state.link_workspace(),
+    )
+
+npm_translate_lock = repository_rule(
```
alexeagle, line 57: nit: I'd wish for this file to be `npm_translate_lock.bzl` since it declares the rule. What could the other one be named?
oh I see you already have a TODO for that after landing.

alexeagle (approved):
I made it through a lot of the code...

npm/private/utils.bzl
gregmagolan: just to keep the hashes to `[a-zA-Z0-9]`. negative numbers are surprising hash values but then so are integers.

I spend a whole 10 minutes on this function so I could go either way

npm/private/npm_translate_lock_state.bzl
gregmagolan: I like the golang guidance of not prefixing getters with `Get`, just setters with `Set`

gregmagolan: better to rename the value if it is confusing to something like `should_update_pnpm_lock`

npm/private/utils.bzl
alexeagle: I think we should just allow negative. Maybe someone will have a reason to produce this file themselves. Simpler to use the value starlark provides

gregmagolan: Ahh there it is: https://github.com/bazelbuild/starlark/issues/36#issuecomment-1115352085

npm/private/npm_translate_lock_state.bzl
gregmagolan: https://go.dev/doc/effective_go#Getters for refere

npm/private/utils.bzl
gregmagolan: that's a good point; it may be a dumb looking hash but you can generate the same value with other languages

PR #651, opened 2022-11-27
doc: say how to bootstrap a pnpm lockfile with managed pnpm

### LINKED ISSUE #649: [FR]: Document how to bootstrap a pnpm lock file

The docs currently have [this](https://github.com/aspect-build/rules_js/blob/main/docs/faq.md#can-i-use-bazel-managed-pnpm) to say about invoking `pnpm`:
> just run `bazel run -- @pnpm//:pnpm --dir $PWD` followed by the usual arguments to pnpm.

I wanted to use that invocation to create a lock file from our `package.json` file.

However, as far as I can tell, that only works once you have an `npm_translate_lock()` call in your WORKSPACE file. That call, in turn, requires a pnpm lock file.

After some trial and error, here are the steps I went through to create a pnpm lock file.

1. Create a `npm_translate_lock()` call in the workspace pointing at an empty `pnpm-lock.yaml`.
2. Put the following contents in the `pnpm-lock.yaml` file:

        lockfileVersion: 5.4
        packages: {}
3. Patch `rules_js` with the following:

        --- a/npm/private/npm_translate_lock.bzl
        +++ b/npm/private/npm_translate_lock.bzl
        @@ -236,8 +236,8 @@ def _gen_npm_imports(lockfile, root_package, attr, registries, default_registry)
                 fail("prod and dev attributes cannot both be set to true")

             packages = lockfile.get("packages")
        -    if not packages:
        -        fail("expected packages in processed lockfile")
        +    #if not packages:
        +    #    fail("expected packages in processed lockfile")

             importers = lockfile.get("importers")
             if not importers:
        @@ -620,8 +620,8 @@ def _impl(rctx):
             ]

             packages = lockfile.get("packages")
        -    if not packages:
        -        fail("expected packages in processed lockfile")
        +    #if not packages:
        +    #    fail("expected packages in processed lockfile")

             importers = lockfile.get("importers")
             if not importers:
4. Then run `bazel run -- @pnpm//:pnpm --dir $PWD install --lockfile-only`.

That gives me a functional lock file as far as I can tell.

### Describe the feature

I would like to offer to document this bootstrapping process. Alternatively I'll document a better one if folks have one.

### Fund our work

alexeagle:
Fixes #649

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -44,7 +44,7 @@ This defines the `@pnpm` repository so that you can create the lockfile with
 `bazel run -- @pnpm//:pnpm --dir $PWD install --lockfile-only`, and then once the file exists you'll
 be able to add the `pnpm_translate_lock` to the `WORKSPACE` which requires the lockfile.
 
-Consider documenting this as a good practice for your team, so that all developers run the exact same pnpm and node versions that Bazel does.
+Consider documenting running pnpm through bazel as a good practice for your team, so that all developers run the exact same pnpm and node versions that Bazel does.
 
 ## Why can't Bazel fetch an npm package?
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -32,7 +32,19 @@ Try running Bazel with `--experimental_check_output_files=false` so that your ed
 
 Yes, just run `bazel run -- @pnpm//:pnpm --dir $PWD` followed by the usual arguments to pnpm.
 
-Document this as a good practice so that all developers run the exact same pnpm and node versions that Bazel does.
+If you're bootstrapping a new project, you'll need to add this to your WORKSPACE:
+
+```starlark
+load("@aspect_rules_js//npm:npm_import.bzl", "pnpm_repository")
+
+pnpm_repository(name = "pnpm")
+```
+
+This defines the `@pnpm` repository so that you can create the lockfile with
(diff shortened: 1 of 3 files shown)


docs/faq.md
```diff
@@ -32,7 +32,19 @@ Try running Bazel with `--experimental_check_output_files=false` so that your ed
...
+load("@aspect_rules_js//npm:npm_import.bzl", "pnpm_repository")
+
+pnpm_repository(name = "pnpm")
+```
+
+This defines the `@pnpm` repository so that you can create the lockfile with
+`bazel run -- @pnpm//:pnpm --dir $PWD install --lockfile-only`, and then once the file exists you'll
+be able to add the `pnpm_translate_lock` to the `WORKSPACE` which requires the lockfile.
+
+Consider documenting this as a good practice for your team, so that all developers run the exact same pnpm and node versions that Bazel does.
```
philsc, line 47: The `this` is a bit ambiguous here; it's not 100% clear what it refers to. I.e. it could refer to the bootstrapping, but it really doesn't. Perhaps something like this?
```suggestion
Consider documenting running pnpm through bazel as a good practice for your team, so that all developers run the exact same pnpm and node versions that Bazel does.
```

philsc (approved):
Thanks a ton! Really appreciate it :heart:

gregmagolan (approved):
Very nice

PR #659, opened 2022-11-29
chore: update to Bazel last_green

gregmagolan:
Testing that fix in https://github.com/bazelbuild/bazel/pull/16812 resolves issue https://github.com/bazelbuild/bazel/issues/16789 seen in Bazel 6.0.0rc1 & rc2


alexeagle (approved):
SWEET fixing bazel

PR #661, opened 2022-11-29
Blank outdir or output_dir examples

rc-glean:


rc-glean:
Related link for my personal convenience https://github.com/aspect-build/rules_js/issues/662

alexeagle:
I'm confused - is this a repro of a bug, or is it encoding expected behavior?

rc-glean:
@alexeagle sorry for the confusion. This is a minimal bug repro. Basically in rules_nodejs, output_dir = True would emit all webpack bundles.

However, in rules_js, no assets are created.

It's probably possible to split what was previously one target in rules_nodejs into several in rules_js, but raising the issue in case this is a problem for other common generated rules

alexeagle:
It's because webpack was told to write to a certain location
```
    output: {
        filename: 'main.js',
        path: out_path,
    },
```
and that's not where Bazel expected it to write.

By the way, we have a https://github.com/aspect-build/rules_webpack for rules_js which is probably easier to use than the webpack CLI directly.

gregmagolan:
More specifically, `out_path` is set to `const out_path = path.resolve(__dirname, 'dist')` so your example would work if you set `out_dirs = ["dist"],` or changed the webpack config to output to `raytest` as you have set.

Issue #660, opened 2022-11-29
[Bug]: Running with ESM breaks js_dev_server


Aghassi:
### What happened?

When trying to run js_dev_server, node throws an error because the aspect runner is a .js extension that uses a commonjs module syntax

```
file:///private/var/tmp/_bazel_davidaghassi/30b12e6784b9304c95faa3f4506c5839/execroot/rh/bazel-out/darwin_arm64-fastbuild/bin/web/web-app/app/dev.sh.runfiles/aspect_rules_js/js/private/js_run_devserver.js:1
const path = require('path')
             ^

ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and '/private/var/tmp/_bazel_davidaghassi/30b12e6784b9304c95faa3f4506c5839/execroot/rh/bazel-out/darwin_arm64-fastbuild/bin/web/web-app/app/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at file:///private/var/tmp/_bazel_davidaghassi/30b12e6784b9304c95faa3f4506c5839/execroot/rh/bazel-out/darwin_arm64-fastbuild/bin/web/web-app/app/dev.sh.runfiles/aspect_rules_js/js/private/js_run_devserver.js:1:14
    at ModuleJob.run (node:internal/modules/esm/module_job:183:25)
    at async Loader.import (node:internal/modules/esm/loader:178:24)
    at async Object.loadESM (node:internal/process/esm_loader:68:5)
    at async handleMainPromise (node:internal/modules/run_main:63:12)
^C
```

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:
bazel 6.0.0-pre.20221020.1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: 1.6.9

Language(s) and/or frameworks involved: Typescript/SWC

### How to reproduce

```shell
Set `package.json` to `type: "module"` and then run the dev server. Here is how mine looks


js_run_devserver(
    name = "dev",
    args = [
        "serve",
        "--config",
        "webpack.config.mjs",
    ],
    chdir = package_name(),
    # See WORKSPACE file for the linking setup for this
    command = "./node_modules/.bin/webpack-cli",
    data = [
        "server/template.html",
        "webpack.config.mjs",
        ":node_modules",
        "//web/web-app/app/kernel",
    ],
[...]

PR #670, opened 2022-11-30
fix: use .mjs entry point for js_run_devserver

gregmagolan:
Resolves https://github.com/aspect-build/rules_js/issues/660


Aghassi (approved):

jbedard:
I assume existing tests verify cjs still works? Worth adding an mjs test?

gregmagolan:
> I assume existing tests verify cjs still works? Worth adding an mjs test?

Good call. I added the additional test

jbedard (approved):

PR #676, opened 2022-12-02
chore: add more details to docstring of js_library_helpers gather_runfiles helper function

gregmagolan:
The subtle difference in this helper between data & deps could use some more detail. Also added links to relevant spots in the Bazel docs that apply to handling runfiles in build rules.


alexeagle (approved):

jbedard (commented):
👍

PR #684, opened 2022-12-05
feat: move @bazel/worker

thesayyn:
still not finalized.

TODO
- docs
- generating .d.ts output of the API.

### DIFF
--- a/.aspect/rules/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU=
+++ b/.aspect/rules/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU=
@@ -1,9 +1,9 @@
-# Input hashes for repository rule npm_translate_lock(name = "npm", pnpm_lock = "@//:pnpm-lock.yaml").
+# Input hashes for repository rule npm_translate_lock(name = "npm", pnpm_lock = "//:pnpm-lock.yaml").
 # This file should be checked into version control along with the pnpm-lock.yaml file.
 .npmrc=-2065072158
-pnpm-lock.yaml=1507139895
+pnpm-lock.yaml=-1744054231
 package.json=-1140838692
-pnpm-workspace.yaml=-1869942826
+pnpm-workspace.yaml=1306476322
 examples/js_binary/package.json=-41174383
 examples/macro/package.json=-696073599
 examples/npm_deps/package.json=-462506805
@@ -15,3 +15,4 @@ js/private/coverage/bundle/package.json=-975746706
 npm/private/test/package.json=1632753039
 npm/private/test/vendored/is-odd/package.json=1041695223
 npm/private/test/vendored/semver-max/package.json=578664053
+js/private/worker/src/package.json=837209084


js/private/worker/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+        abortionMap.set(request.request_id, abortController)
+
+        const response = new blaze.worker.WorkResponse({
+            request_id: request.request_id,
+        })
+
+        const outputBuffer = Buffer.alloc(0)
+        const outputStream = new Writable({ write: outputBuffer.write })
+
+        implementation({
```
thesayyn, line 49: no use of await here to achieve a non-blocking loop.  the loop has to be blocked on stdin only.

js/private/worker/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+                    responseBytes.byteLength
+                )
+                process.stdout.write(
+                    Buffer.concat([responseSizeBytes, responseBytes])
+                )
+            })
+    }
+}
+
+export function shouldRunAsWorker(args: string[]): boolean {
```
thesayyn, line 80: i feel like there is a better name for this function.

js/private/worker/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+
+        const request = blaze.worker.WorkRequest.deserialize(current!)
+
+        if (request.cancel) {
+            abortionMap.get(request.request_id)?.abort()
+            continue
+        }
+
+        const abortController = new AbortController()
+        abortionMap.set(request.request_id, abortController)
```
thesayyn, line 40: not to be confused with child abortion...

js/private/worker/src/index.ts
thesayyn: necessary for multiplexing and cancellation

js/private/worker/src/index.ts
jbedard: When would the "bazel worker" not be run as a worker?

jbedard:
Is this editing it or only moving it? Can we see a diff of any changes you made?

js/private/worker/src/index.ts
thesayyn: when you put `--spawn_strategy=TsProject=local|sandboxed` into your .bazelrc

thesayyn:
> Is this editing it or only moving it? Can we see a diff of any changes you made?

we got a completely different API now.

gregmagolan (approved):

gregmagolan:
> > Is this editing it or only moving it? Can we see a diff of any changes you made?
> 
> we got a completely different API now.

Looks like `size.ts` hasn't changed and `worker_protoco.ts` is generated so the new impl is in `index.ts`

js/private/worker/src/index.ts
alexeagle: may as well name it after the flag you read so you don't have to invent a name.

`isPersistentWorker` ?

gregmagolan:
Is the gc.ts --expose_gc special handling no longer needed: https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/worker/src/gc.ts

I vaguely recall that being added at some point

js/private/worker/src/index.ts
thesayyn: that's better. Thanks

js/private/worker/src/index.ts
gregmagolan: 👍

PR #691, opened 2022-12-07
fix: fix .npmrc default registry logic

gregmagolan:
Not sure when this regressed since we only had the bzl snapshot test for coverage. I added a `test.sh` script now to make appropriate assertions on the bzl snapshot.

### DIFF
--- a/npm/private/npm_translate_lock.bzl
+++ b/npm/private/npm_translate_lock.bzl
@@ -416,14 +416,20 @@ def _gen_npm_imports(lockfile, root_package, attr, registries, default_registry)
             url = repo
         elif tarball:
             if _is_url(tarball):
-                if registry and tarball.startswith(default_registry):
-                    url = registry + tarball[len(default_registry):]
+                # pnpm sometimes prefixes the `tarball` url with the default npm registry `https://registry.npmjs.org/`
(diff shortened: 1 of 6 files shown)


npm/private/npm_translate_lock.bzl
```diff
@@ -416,14 +416,19 @@ def _gen_npm_imports(lockfile, root_package, attr, registries, default_registry)
             url = repo
         elif tarball:
             if _is_url(tarball):
-                if registry and tarball.startswith(default_registry):
-                    url = registry + tarball[len(default_registry):]
+                # tarball url may start with the default registry in pnpm-lock.yaml which we must replace:
```
jbedard, line 419: I don't follow... why use the global default instead of local one?

Why does the comment use .org and .com domains?

gregmagolan: pnpm does a funny thing and sometimes adds `https://registry.npmjs.org/` (`DEFAULT_REGISTRY`) on the tarball URL and then puts the registry you asked for in the `registry` field. No idea why but we have to strip

gregmagolan: added some more color to the comment

jbedard (approved):

PR #696, opened 2022-12-08
feat: use settings in ~/.npmrc when use_home_npmrc is set to True in npm_translate_lock

gregmagolan:


jbedard:
I assume we want to discourage this? Maybe put more info in the param doc saying it is not recommended and then specifying some examples where it might be required?

gregmagolan:
> I assume we want to discourage this? Maybe put more info in the param doc saying it is not recommended and then specifying some examples where it might be required?

Unfortunately we need it for sourcegraph as they put their auth in their `~/.npmrc`. There is some precedent from rules_jvm_external https://github.com/bazelbuild/rules_jvm_external/commit/2d492e48cd9a85c0e8cfa92d776b25bbf28aa9b1

jbedard:
Yeah I assumed it was for auth, and that could be an example use case put into the param docs? While also discouraging uses it for other purposes that would change behaviour depending on the config?

gregmagolan:
Expanded docstring to say:

```
            WARNING: The repository rule will not be invalidated by changes to the home `.npmrc` file since there
            is no way to specify this file as an input to the repository rule. If changes are made to the home
            `.npmrc` you can force the repository rule to re-run and pick up the changes by running:
            `bazel sync --only={name}` where `name` is the name of the `npm_translate_lock` you want to re-run.

            Because of the repository rule invalidation issue, using the home `.npmrc` is not recommended.
            `.npmrc` settings should generally go in the `npmrc` in your repository so they are shared by all
            developers. The home `.npmrc` should be reserved for authentication settings for private npm repositories.
```

alexeagle (approved):

PR #698, opened 2022-12-08
feat: add support for '_auth' basic authentication in .npmrc

gregmagolan:
plus support for global `_auth`, `_authtoken`, `username` and `_password` keys in `.npmrc`


npm/private/npm_import.bzl
```diff
@@ -394,14 +392,27 @@ def _download_and_extract_archive(rctx):
...
+    if rctx.attr.npm_auth_basic:
+        auth = {
+            download_url: {
+                "type": "pattern",
+                "pattern": "Basic <password>",
+                "password": rctx.attr.npm_auth_basic,
+            },
+        }
+        auth_count += 1
+    if rctx.attr.npm_auth_username and rctx.attr.npm_auth_password:
```
alexeagle, line 405: nit: I think you're missing a check for e.g. npm_auth_basic + npm_apth_username (but no _password)

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -195,39 +196,55 @@ WARNING: Issue while reading "{npmrc}". Failed to replace env in config: ${{{tok
...
+        if k.endswith(_NPM_PKG_SCOPE_KEY):
             # @myorg:registry=https://somewhere-else.com/myorg
             # scope: @myorg
             # registry: somewhere-else.com/myorg
             scope = k.removesuffix(_NPM_PKG_SCOPE_KEY)
             registry = utils.to_registry_url(v)
             registries[scope] = registry
 
-        if k.find(_NPM_USERNAME) != -1:
+        if k == _NPM_AUTH or k.endswith(":" + _NPM_AUTH):
```
alexeagle, line 214: nit: seems like the npmrc handling could be in a separate starlark source file?

alexeagle (approved):
thanks!

npm/private/npm_import.bzl
gregmagolan: that isn't valid. npm_auth_basic is the base64 encoded `username:password` so it stands on its own

npm/private/npm_translate_lock_generate.bzl
gregmagolan: yeah; I've been meaning to move it out of here. still here from the original .bzl before the big refactor

npm/private/npm_import.bzl
gregmagolan: there is also a check above that username & password always travel together:
```
    if rctx.attr.npm_auth_username or rctx.attr.npm_auth_password:
        if not rctx.attr.npm_auth_username:
            fail("'npm_auth_password' was provided without 'npm_auth_username'")
        if not rctx.attr.npm_auth_password:
            fail("'npm_auth_username' was provided without 'npm_auth_password'")
```

PR #707, opened 2022-12-11
fix: js_image_layer build on RBE

bazaglia:
The `js_image_layer`  rule currently relies on a hacky approach to get the symlinks targets working in the built Docker images by stripping the `execroot` from the path. The problem is that there is no `execroot` on the RBE, so I tried to get the paths corrected on remote builds too.

This has been tested on the open-source [Bazel Buildfarm](https://github.com/bazelbuild/bazel-buildfarm) RBE.

### DIFF
--- a/e2e/js_image/BUILD.bazel
+++ b/e2e/js_image/BUILD.bazel
@@ -60,8 +60,6 @@ container_test(
     name = "image_test",
     configs = [":smoketest.yaml"],
     image = ":image",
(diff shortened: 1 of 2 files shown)


thesayyn:
It would be great if you could turn on RBE for our github actions. relevant line https://github.com/aspect-build/rules_js/blob/f02a68c2e9d4e38ee15ff1535bfafc3d2d1b9a16/e2e/js_image/BUILD.bazel#L64

thesayyn (approved):

PR #712, opened 2022-12-14
feat: run prepare lifecycle hooks

gregmagolan:
oddly, the `runPostinstallHooks` function we were using from `@pnpm/lifecycle` was only running `preinstall`, `install`, `postinstall` while pnpm itself runs `preinstall`, `install`, `postinstall` and `prepare` using `runLifecycleHooksConcurrently` from `@pnpm/lifecycle`: https://github.com/pnpm/pnpm/blob/bc18d33fe00d9ed43f1562d4cc6d37f49d9c2c38/pkg-manager/core/src/install/index.ts#L1108


package.json
```diff
@@ -16,6 +16,13 @@
                 "peerDependencies": {
                     "mocha-junit-reporter": "*"
                 }
+            },
+            "@kubernetes/client-node": {
+                "dependencies": {
+                    "@types/node": "^10.12.0",
```
jbedard, line 22: Where are these versions coming from?

npm/private/lifecycle/lifecycle-hooks.js
```diff
@@ -169,17 +210,23 @@ async function copyRecursive(src, dest) {
     } catch (e) {
         // Note: use .log rather than .error. The former is deferred and the latter is immediate.
         // The error is harder to spot and parse when it appears in the middle of the other logs.
-        if (e.code === "ELIFECYCLE" && !!e.pkgid && !!e.stage && !!e.script) {
```
jbedard, line 172: Is this block all just formatting? What's causing this?

npm/private/lifecycle/lifecycle-hooks.js
```diff
@@ -68,6 +69,46 @@ async function makeBins(nodeModulesPath, scope, segmentsUp) {
     }
 }
 
+// Helper from not exported by @pnpm/lifecycle
```
jbedard, line 72: Add a link to what we're cutting?

npm/private/lifecycle/lifecycle-hooks.js
```diff
@@ -68,6 +69,46 @@ async function makeBins(nodeModulesPath, scope, segmentsUp) {
...
+async function checkBindingGyp(root, scripts) {
+    if (await exists(path.join(root, 'binding.gyp'))) {
+        scripts['install'] = 'node-gyp rebuild'
+    }
+}
+
+// Like runPostinstallHooks from @pnpm/lifecycle but also runs prepare hook which is notably missing
+// from that function and does not match what pnpm does on its install:
+// https://github.com/pnpm/pnpm/blob/bc18d33fe00d9ed43f1562d4cc6d37f49d9c2c38/pkg-manager/core/src/install/index.ts#L1108
```
jbedard, line 81: This seems to run them concurrently? The method also seems to be public from `@pnpm/lifecycle`, do/can we depend on that?

alexeagle (approved):

package.json
gregmagolan: I so wish it was easier to leave comments in package.json files. I'll see if I can inject a comment field.

TL;DR is that these come from the `devDependencies` of `@kubernetes/client-node` and the are needed for the `prepare` lifecycle hook. Surprisingly the `prepare` step of that npm package runs tsc and makes the `dist` folder and for it to pass the `@types/node` must be an older version.

npm/private/lifecycle/lifecycle-hooks.js
gregmagolan: It's the formatter. Guessing that who ever changed this file last didn't have the precommit hooks in place & we don't check formatting on CI presently.

npm/private/lifecycle/lifecycle-hooks.js
gregmagolan: I spoke good english

npm/private/lifecycle/lifecycle-hooks.js
gregmagolan: The concurrently method is a lot more than I'm comfortable changing in this PR. Want to keep the scope limited to just adding the `prepare` hook on top of what we currently have. Could be explored in a follow-up at some point.

npm/private/lifecycle/lifecycle-hooks.js
```diff
@@ -68,6 +69,46 @@ async function makeBins(nodeModulesPath, scope, segmentsUp) {
...
 
+// Helper from not exported by @pnpm/lifecycle
+async function checkBindingGyp(root, scripts) {
+    if (await exists(path.join(root, 'binding.gyp'))) {
+        scripts['install'] = 'node-gyp rebuild'
+    }
+}
+
+// Like runPostinstallHooks from @pnpm/lifecycle but also runs prepare hook which is notably missing
+// from that function and does not match what pnpm does on its install:
```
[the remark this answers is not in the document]
gregmagolan: It looks like the `runPostinstallHooks` method is only used by the `@pnpm/plugin-commands-rebuild` npm package in `pnpm` which is described as "Commands for rebuilding dependencies". Maybe it is intentionally ommitted since the function is named "PostinstallHooks" so it only runs "preinstall", "install" and "postinstall"

npm/private/lifecycle/lifecycle-hooks.js
```diff
@@ -1,7 +1,8 @@
 const fs = require('fs')
+const exists = require('path-exists')
 const path = require('path')
-const { runPostinstallHooks } = require('@pnpm/lifecycle')
-const runLifecycleHook = require('@pnpm/lifecycle').default
+const { safeReadPackageJsonFromDir } = require('@pnpm/read-package-json')
+const { runLifecycleHook } = require('@pnpm/lifecycle')
```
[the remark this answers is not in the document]
gregmagolan: the exports changed with the version bump of `@pnpm/lifecycle` included in this PR

npm/private/lifecycle/lifecycle-hooks.js
gregmagolan: ✅

package.json
gregmagolan: added
```
                "//1": "these deps are from devDependencies of @kubernetes/client-node and needed for 'prepare' lifecycle hook",
                "//2": "https://github.com/kubernetes-client/javascript/blob/763decf6e09d0e246b80cefdb2c5fd570373801a/package.json#L72",
```

PR #713, opened 2022-12-14
feat: add lifecycle_hooks_sandboxed attribute to npm_translate_lock and npm_import

gregmagolan:
Deprecates lifecycle_hooks_no_sandbox attribute.

The original attribute was badly formed. This style makes it simpler to opt-in to lifecycle hooks sandboxing for a subset of npm packages instead of enabling it for all of them when only a subset may require it.


WORKSPACE
```diff
@@ -107,6 +107,18 @@ npm_translate_lock(
...
+        # If @kubernetes/client-node is not sandboxed, will fail with
+        # ```
+        # src/azure_auth.ts(97,43): error TS2575: No overload expects 2 arguments, but overloads do exist that expect either 1 or 4 arguments.
+        # src/azure_auth.ts(98,34): error TS2575: No overload expects 2 arguments, but overloads do exist that expect either 1 or 4 arguments.
+        # src/gcp_auth.ts(93,43): error TS2575: No overload expects 2 arguments, but overloads do exist that expect either 1 or 4 arguments.
+        # src/gcp_auth.ts(94,34): error TS2575: No overload expects 2 arguments, but overloads do exist that expect either 1 or 4 arguments.
+        # ```
+        # since a `jsonpath-plus@7.2.0` that is newer then the transitive dep `jsonpath-plus@0.19.0` is found outside of the sandbox that
+        # includes typings that don't match the 0.19.0 "any" usage.
+        "@kubernetes/client-node",
```
alexeagle, line 120: it feels like it's going to be a forever project for our community to discover all the brokenness and add them to their WORKSPACE files. I'm not sure what to do about it.

npm/npm_import.bzl
```diff
@@ -323,10 +327,18 @@ def npm_translate_lock(
             ```
             lifecycle_hooks_execution_requirements: {
                 "*": ["requires-network"],
-                "@foo/bar": ["no-sandbox"],
+                "@foo/bar": ["no-remote-exec"],
             }
             ```
 
+        lifecycle_hooks_sandboxed: A list of npm packages for which lifecycle hooks, if they exist,
```
alexeagle, line 334: I liked how the old way was just syntax sugar for execution requirements. I'm not clear why reversing the boolean is better.

gregmagolan: The better default for lifecycle hooks is `no-sandbox` but the old way there was no way to opt-in a package or two. Hence the flip. But I'm open to suggestions if you have a better idea for selective sandbox opt-in

WORKSPACE
gregmagolan: 3p deps are one of the harder parts and lifecycle hooks will be one of those sharp edges. The `@kubernetes/client-node` package is the first one I've seen the requires `prepare` and actually runs typescript in the `prepare` step to build a `dist` folder. This is an example of one that could use a BUILD file.

alexeagle: I keep wishing for us to have some registry of known ones that ppl need, maybe even fetch that as a default input to the rule so it gets easier over time. Like that repo that yarn and pnpm co-maintain that has common hoisting.

But it would have to be versioned (some later `@kubernetes/client-node` might fix it). Seems intractable.

Maybe the answer is just to keep improving error messages so that users discover the right WORKSPACE settings.

Another thought: instead of these having to go in the root workspace it would be nice to put them in package.json with the pnpm hoist settings.

Thoughts for another day

npm/npm_import.bzl
alexeagle: I think lifecycle_hooks_execution_requirements should express it all, the others are syntax sugar

Let's see if I can work this through...

`lifecycle_hooks_execution_requirements` default value is `{"*": ["no-sandbox"]}` for performance.

```
lifecycle_hooks_execution_requirements: {
    "*": ["requires-network", "no-sandbox"],
    "@foo/bar": ["no-remote-exec"],
}
```

Is how you'd allow all of them to use network and disable remote exec for one of them.

Can we deprecate `lifecycle_hooks_no_sandbox` and not replace it with anything?

gregmagolan: The trouble is there is no `sandbox` flag. If I set `"*": ["no-sandbox"],` I have no way to express _not_ setting `no-sandbox` on just one package that needs the sandbox.

gregmagolan:
Replaced by https://github.com/aspect-build/rules_js/pull/718

PR #717, opened 2022-12-14
docs: mention how to use pnpm ls to find package.json data

alexeagle:


gregmagolan (approved):

npm/npm_import.bzl
```diff
@@ -259,11 +259,14 @@ def npm_translate_lock(
 
         data: Data files required by this repository rule when auto-updating the pnpm lock file.
 
-            Only needed with `update_pnpm_lock` is True.
+            Only needed when `update_pnpm_lock` is True.
 
             This should include all `package.json` files in the pnpm workspace or required to run
```
gregmagolan, line 264: Can you also mention that the `pnpm-workspace.yaml` file is also needed. Looks like I forgot that.

PR #718, opened 2022-12-14
feat: add lifecycle_hooks attribute to npm_translate_lock and npm_import

gregmagolan:
The not-yet-released `prepare` lifecycle hook which landed in https://github.com/aspect-build/rules_js/commit/70f46edc780ea5c449174a324f756a3e9feefc25 will now not run unless explicitly set to run for a package via the `lifecycle_hooks` attribute.

This hook should only be run for npm packages fetched from source from github, for example, so a better default for it is to not be run unless explicitly specified. In the future we could turn it on automatically for packages that we detect are fetched from source.

MINOR BREAKING CHANGES:

`patch_args` in `npm_translate_lock` are no longer additive if there are multiple matches; this allows you to set a global `"*": ["-p1"]` for example and then override that arg for specific packages.

`lifecycle_hooks_execution_requirements` in `npm_translate_lock` are no longer additive if there are multiple matches; this allows you to set a global `"*": ["no-sandbox"]` for example and then override that for specific packages.

`bins` in `npm_translate_lock` are no longer additive if there are multiple matches; adding them up doesn't have any valid use case and may have led to unexpected bins


alexeagle (commented):
I think we should add a new markdown file in docs/ that explains all of this in prose, we're pushing how much you can read in the stardoc-generated attribute table

npm/npm_import.bzl
```diff
@@ -313,24 +373,40 @@ def npm_translate_lock(
...
-            For example:
+            The execution requirements can be defined per package by package name or globally using "*". For example,
 
             ```
-            lifecycle_hooks_execution_requirements: {
-                "*": ["requires-network"],
-                "@foo/bar": ["no-sandbox"],
+            lifecycle_hooks_execution_requirements = {
+                "*": ["no-sandbox"],
+                "@foo/bar": [],
```
alexeagle, line 395: would be good to add a comment here "this will be run in a sandbox"

npm/npm_import.bzl
```diff
@@ -313,24 +373,40 @@ def npm_translate_lock(
...
+                "fum@0.0.1": ["no-sandbox", "no-remote-exec"],
             }
             ```
 
+            Execution requirements are not additive. The move specific name match takes precedence.
+
+        lifecycle_hooks_no_sandbox: If True, a "no-sandbox" execution requirement is added to all lifecycle hooks
```
alexeagle, line 402: want to deprecate this one?

npm/npm_import.bzl
```diff
@@ -347,16 +423,11 @@ def npm_translate_lock(
             }
             ```
 
+            Dicts of bins not additive. The move specific name match takes precedence.
```
alexeagle, line 426: do we have a mix of these where some are additive and some are most-specific-wins?

npm/npm_import.bzl
gregmagolan: maybe in a follow-up

npm/npm_import.bzl
gregmagolan: Yes. Depending on the attribute one makes more sense than the other

npm/npm_import.bzl
gregmagolan: ✅

gregmagolan:
> I think we should add a new markdown file in docs/ that explains all of this in prose, we're pushing how much you can read in the stardoc-generated attribute table

follow-up?

alexeagle:
We can add `docs/pnpm_settings.md` in a follow-up I guess, but may as well do it todayish while this is all fresh. I'm happy to help

npm/npm_import.bzl
alexeagle: can you make all of them indicate Additive or Most-Specific in a consistent way? seems like this will be confusing

WORKSPACE
```diff
@@ -90,18 +90,30 @@ npm_translate_lock(
...
+            "postinstall",
+        ],
         # 'install' hook fails as it assumes the following path to `node-pre-gyp`: ./node_modules/.bin/node-pre-gyp
         # https://github.com/stultuss/protoc-gen-grpc-ts/blob/53d52a9d0e1fe3cbe930dec5581eca89b3dde807/package.json#L28
-        "protoc-gen-grpc@2.0.3",
-    ],
+        "protoc-gen-grpc@2.0.3": [],
+    },
     lifecycle_hooks_execution_requirements = {
+        "*": [
```
alexeagle, line 106: wdyt about calling this `default` rather than `*`? does that make it more clear that it won't be added to packages below? `*` matches everything but `default` applies if not overridden

alexeagle: and obvs there had better not be https://npmjs.com/default

gregmagolan: that would be more of a breaking change since we already support "*" here

gregmagolan: Good APIs are hard build; especially when you'd trying not to make breaking changes on top of existing APIs

npm/npm_import.bzl
```diff
@@ -271,22 +273,54 @@ def npm_translate_lock(
...
+            custom_postinstalls = {
+                "@foo/bar": "echo something > somewhere.txt",
+                "fum@0.0.1": "echo something_else > somewhere_else.txt",
+            },
+            ```
+
+            Custom postinstalls are additive and joined with ` && ` when there are multiple matches for a package.
+            More specific matches are appended to previous matches.
+
+        prod: If True, only install dependencies.
```
alexeagle, line 323: add "but not devDependencies" otherwise it's not very clear. And putting in code quotes will also help explain this doesn't mean dependencies generally, it means that field in package.json

npm/npm_import.bzl
```diff
@@ -295,16 +329,43 @@ def npm_translate_lock(
             wild-cards are not yet supported and npm_translate_lock will fail if there are multiple versions of a package that
             are to be hoisted.
 
-        dev: If true, only install devDependencies
+            ```
+            public_hoist_packages = {
+                "@foo/bar": [""]
```
alexeagle, line 334: add a comment about what empty string means (the root)

npm/npm_import.bzl
```diff
@@ -295,16 +329,43 @@ def npm_translate_lock(
...
+                "fum@0.0.1": ["some/sub/package"]
+            },
+            ```
+
+            List of public hoist packages are additive when there are multiple matches for a package.
+
+        dev: If True, only install devDependencies
+
+        no_optional: If True, optionalDependencies are not installed
```
alexeagle, line 343: opt: link to that issue about platform-specific optional deps?

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -98,15 +97,20 @@ def _is_url(url):
...
-        if name:
-            v = keyed_lists.get(name, [])
-            if type(v) == "list":
-                result.extend(v)
+        if name and name in keyed_lists:
+            v = keyed_lists[name]
+            if additive:
+                if type(v) == "list":
+                    result.extend(v)
+                else:
```
alexeagle, line 108: should assert that it's a string here?

alexeagle (approved):

npm/npm_import.bzl
gregmagolan: I have,

```
Foo are not additive. The move specific name match takes precedence.
```

and

```
Bar are additive when there are multiple matches for a package. More specific matches are
appended to previous matches.
```

You have something better in mind?

npm/npm_import.bzl
```diff
@@ -425,6 +497,24 @@ WARNING: `package_json` attribute in `npm_translate_lock(name = "{name}")` is de
     if update_pnpm_lock == None and (npm_package_lock or yarn_lock):
         update_pnpm_lock = True
 
+    # lifecycle_hooks_exclude is a convenience attribute to set `<value>: []` in `lifecycle_hooks`
+    for p in lifecycle_hooks_exclude:
+        if p in lifecycle_hooks:
+            fail("expected '{}' to be in only one of lifecycle_hooks or lifecycle_hooks_exclude")
```
jbedard, line 505: The `{}` needs a value?

npm/npm_import.bzl
jbedard: Note there's already a TODO to remove it in 2.0.. 🤷

npm/npm_import.bzl
alexeagle: ooh good buildifier lint check

npm/npm_import.bzl
gregmagolan: ✅

npm/npm_import.bzl
gregmagolan: ✅

npm/private/npm_translate_lock_generate.bzl
gregmagolan: ✅

npm/npm_import.bzl
gregmagolan: nice catch

gregmagolan: ✅

npm/npm_import.bzl
gregmagolan: The TODO is for the one in `npm_import`. That one is rarely used and is not an API that is used by most users.

The one in `npm_translate_lock` I'm not sure what to do about yet so I'll punt the decision to a follow-up.

npm/npm_import.bzl
gregmagolan: ✅

PR #719, opened 2022-12-14
fix: empty lifecycle_hooks_execuction_requirements should be set on generated npm_import targets in npm_translate_lock

gregmagolan:
Follow-up to https://github.com/aspect-build/rules_js/pull/718.


npm/private/npm_import.bzl
```diff
@@ -758,14 +756,13 @@ _ATTRS_LINKS = dicts.add(_COMMON_ATTRS, {
...
     "lifecycle_hooks_execution_requirements": attr.string_list(),
-    "lifecycle_hooks_no_sandbox": attr.bool(default = True),
     "npm_translate_lock_repo": attr.string(),
 })
 
 _ATTRS = dicts.add(_COMMON_ATTRS, {
     "extra_build_content": attr.string(),
     "integrity": attr.string(),
-    "patch_args": attr.string_list(default = ["-p0"]),
+    "patch_args": attr.string_list(),
```
alexeagle, line 765: why remove this? seems like npm_import users might still rely on it?

alexeagle (approved):

npm/private/npm_import.bzl
gregmagolan: the public API macro which is documented sets the default; this one is sneaky and undocumented

PR #723, opened 2022-12-15
feat: add optional 'scripts' to run in  before updating the pnpm lock file

gregmagolan:


npm/npm_import.bzl
```diff
@@ -128,6 +129,14 @@ def npm_translate_lock(
 
             Read more: [using update_pnpm_lock](/docs/pnpm.md#update_pnpm_lock)
 
+        preupdate: Node.js scripts to run in this repository rule before auto-updating the pnpm lock file.
+
+            Scripts are run sequentially in the order they are listed. The working directory is set to the root of the
+            external repository. Make sure all files required by preupdate scripts are added to the `data` attribute.
+
+            Only needed when `update_pnpm_lock` is True.
```
alexeagle, line 137: can you `fail` if there's preupdate but not update_pnpm_lock?

alexeagle (approved):

npm/npm_import.bzl
gregmagolan: ✅

PR #725, opened 2022-12-17
chore: add test case for a node-gyp install step on an npm package

gregmagolan:


jbedard:
The overall build succeeding is the indicator the test passed? I really wish there was a better way of doing that or at least better way of declaring the test case, even just a `build_test` somewhere?

gregmagolan:
> The overall build succeeding is the indicator the test passed? I really wish there was a better way of doing that or at least better way of declaring the test case, even just a `build_test` somewhere?

There is a filegroup target that captures all of the npm packages in `npm/private/test/package.json` that should be built. I'll switch it over to a `build_test` so it is move obvious that it is testing that the targets build.

gregmagolan:
This is currently blocked on the Workflows runners not having `g++` available on the PATH

PR #728, opened 2022-12-19
fix: don't mutate the frozen values in lifecycle_hooks dict

mattem:
https://github.com/aspect-build/rules_js/pull/718 introduced a bug when using lifecycle hook exclusions where the values in the frozen dict `lifecycle_hooks` are mutated, which results in a hard error.

Just before the check, create a copy of the original frozen value. This is likely very cheap as these dicts are small if not empty.

Unblocks https://github.com/aspect-build/silo/pull/1005


alexeagle:
Oops, another "we should deploy to silo before releasing" process issue

gregmagolan:
> Oops, another "we should deploy to silo before releasing" process issue

😞

gregmagolan (approved):

PR #737, opened 2022-12-28
prefactor: sort bzlmod attributes and fix npm_translate_lock name

kormide:
Prefactor for https://github.com/aspect-build/rules_js/pull/647. Step towards https://github.com/aspect-build/rules_js/issues/644.

Sorting the attributes alphabetically makes it easier to keep track of what's missing from the WORKSPACE api.

### DIFF
--- a/npm/extensions.bzl
+++ b/npm/extensions.bzl
@@ -10,13 +10,12 @@ load("//npm:npm_import.bzl", "npm_import", "npm_translate_lock")
 load("//npm/private:transitive_closure.bzl", "translate_to_transitive_closure")
 
 def _extension_impl(module_ctx):
-    npm_translate_lock_name = "npm"
     for mod in module_ctx.modules:
         for attr in mod.tags.npm_translate_lock:
             # npm_translate_lock MUST run before parse_pnpm_lock below since it may update
             # the pnpm-lock.yaml file when update_pnpm_lock is True.
             npm_translate_lock(
-                name = npm_translate_lock_name,
+                name = attr.name,
                 pnpm_lock = attr.pnpm_lock,
                 # TODO: get this working with bzlmod
                 # update_pnpm_lock = attr.update_pnpm_lock,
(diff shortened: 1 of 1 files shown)


npm/extensions.bzl
```diff
@@ -10,13 +10,12 @@ load("//npm:npm_import.bzl", "npm_import", "npm_translate_lock")
 load("//npm/private:transitive_closure.bzl", "translate_to_transitive_closure")
 
 def _extension_impl(module_ctx):
-    npm_translate_lock_name = "npm"
```
kormide, line 13: The repository name should be parameterized like it is in WORKSPACE.

npm/extensions.bzl
```diff
@@ -28,45 +27,61 @@ def _extension_impl(module_ctx):
...
+                url = i.url,
+                version = i.version,
             )
 
+def _npm_translate_lock_attrs():
+    attrs = dict(**npm_translate_lock_lib.attrs)
+
+    # Add macro attrs that aren't in the rule attrs.
+    attrs["name"] = attr.string()
+
```
kormide, line 70: Not much here in the prefactor, but there are several attributes I will need to add here and below from the other PR.

kormide: Since we invoke the macro, we won't get all of the attributes we need from `**npm_translate_lock_lib.attrs`, which just includes the rule attributes. There are a few macro parameters that we need to represent as extension tag class attributes for bzlmod.

gregmagolan: 👍

npm/extensions.bzl
kormide:
```
npm.npm_translate_lock(
    name = "npm",
)

use_repo(npm, "npm")
```
This way users can replace `"npm"` with whatever they want.

kormide: The bug is that the `name` attribute for `npm_translate_lock` isn't currently being used.

gregmagolan: 👍

gregmagolan (approved):

PR #749, opened 2023-01-03
chore: add bzlmod test matrix

kormide:
In support of: https://github.com/aspect-build/rules_js/issues/644

This adds a test matrix vector that exercises bzlmod for all unit and e2e tests. This way we can test that all of our e2es work under bzlmod and can keep the WORKSPACE/bzlmod apis in sync.

Currently this will fail because we need to add `MODULE.bazel` files to each e2e and actually fix the bzlmod api. The module files should be ignored during non-bzlmod test runs.

I can either ignore failing tests for now, or we can keep this PR open until all are fixed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -127,6 +127,7 @@ jobs:
                 config: ${{ fromJSON(needs.matrix-prep-config.outputs.configs) }}
                 bazelversion: ${{ fromJSON(needs.matrix-prep-bazelversion.outputs.bazelversions) }}
                 folder: ${{ fromJSON(needs.matrix-prep-folder.outputs.folders) }}
+                bzlmodEnabled: [true, false]
                 exclude:
                     # Don't test RBE with Bazel 5 (not supported)
                     - config: rbe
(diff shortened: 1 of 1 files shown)


kormide:
It might be best for me to manually ignore the failing tests for now so that as I update the bzlmod api I can un-ignore them, rather than leaving this testing until the end.

alexeagle (commented):
I guess we would remove e2e/bzlmod and add WORKSPACE.bzlmod files to existing e2e/examples so they are testable under both? In this PR?

kormide:
> I guess we would remove e2e/bzlmod and add WORKSPACE.bzlmod files to existing e2e/examples so they are testable under both? In this PR?

My assumption was that they would keep the `WORKSPACE` file and just ignore `MODULE.bazel` when `--enable_bzlmod` is set. Yeah, with this we wouldn't need `e2e/bzlmod`.

kormide:
Excluded the individual tests so that I can include them individually as I update the bzlmod api.

gregmagolan (approved):

PR #751, opened 2023-01-04
feat: improved npm_package performance & globbing support using copy_to_directory_bin_action

gregmagolan:
New toolchain registration is hidden in `npm_import` and `npm_translate_lock` macros. In 2.0 we can introduce a new required `rules_js_repositories()` WORKSPACE function.

e2e/bzlmod won't be green until https://github.com/aspect-build/bazel-lib/pull/311 lands and is released


npm/private/npm_package.bzl
```diff
@@ -224,11 +221,13 @@ def _impl(ctx):
         targets = ctx.attr.data,
     ))
 
-    copy_to_directory_action(
+    copy_to_directory_bin_action(
```
alexeagle, line 224: nit: would it read better for bazel-lib to expose a factory function for creating actions?

npm/private/npm_package.bzl
```diff
@@ -65,64 +72,57 @@ If unset, a npm_link_package that references this npm_package must define the pa
...
 
-        Defaults to ["node_modules/**", "**/node_modules/**"] which excludes all node_modules folders
+        Defaults to ["**/node_modules/**"] which excludes all node_modules folders
         from the output directory.
 
         Files and directories that do not have matching output directory paths are subject to subsequent
         filters and transformations to determine if they are copied and what their path in the output
         directory will be.
 
+        {glob_support_docstring}
```
alexeagle, line 119: nit: it's a lot of text for attributes, maybe this goes in separate markdown?

npm/npm_import.bzl
```diff
@@ -30,6 +30,7 @@ load("//npm/private:npm_import.bzl", _npm_import_lib = "npm_import", _npm_import
 load("//npm/private:versions.bzl", "PNPM_VERSIONS")
 load("//npm/private:utils.bzl", _utils = "utils")
 load("//npm/private:npm_translate_lock.bzl", _npm_translate_lock = "npm_translate_lock")
+load("@aspect_bazel_lib//lib:repositories.bzl", _register_copy_to_directory_toolchains = "register_copy_to_directory_toolchains")
```
alexeagle, line 33: nit: IIUC something changed, and you no longer have to rename these to underscore on import - imported symbols aren't automatically treated as exports by stardoc

npm/npm_import.bzl
```diff
@@ -675,6 +688,11 @@ def npm_import(
 
         **kwargs: Internal use only
     """
+
+    # TODO(2.0): move this to a new required rules_js_repositories() WORKSPACE function
```
alexeagle, line 692: I dunno if we should ever bother, since we'll put effort into bzlmod migrations.

alexeagle (approved):

npm/private/npm_package.bzl
gregmagolan: hmmm. not sure what you mean. this function is a wrapper around ctx.actions.run: https://github.com/aspect-build/bazel-lib/blob/fe867981ee5606415a0b1656ee9e034fb89924c7/lib/private/copy_to_directory.bzl#L618

npm/private/npm_package.bzl
gregmagolan: yeah; it should get cleaned up soon starting with upstream. The rendering of these is completely broken on the docsite right now

npm/npm_import.bzl
gregmagolan: in this case `register_copy_to_directory_toolchains` conflicts with the attribute name so needs the rename but that is good to know

npm/npm_import.bzl
gregmagolan: it would be a small change one way or the other but yes I agree bzlmod is the future for all rulesets anyway

PR #758, opened 2023-01-06
chore: add npmrc to bzlmod example

alexeagle:
we recommend all users set this


gregmagolan:
This is now done. https://github.com/aspect-build/rules_js/blob/cefe1395149ff4745dc13063f680744bb6c07923/e2e/npm_translate_lock/MODULE.bazel#L22.

Issue #381, opened 2022-08-11
js_image_layer.bzl should not depend on rules_python directly.


thesayyn:
Currently, js_image_layer is using rules_python to create a proxy py_binary that passed down to `pkg_tar#build_tar` to fix up manifest for symlinks. We should migrate that to js_binary instead.

thesayyn:
closing this as we no longer depend on either rules_pkg or rules_python.

PR #771, opened 2023-01-13
feat: promote js_image_layer to public api

### LINKED ISSUE #304: Promote js_image_layer to Public API
js_image_layer should be promoted to public API before 1.0

thesayyn:
closes #304 and #381


thesayyn:
probably will the ci fail. something in private/test/package.json causes pnpm install to fail. will investigate

js/private/js_image_layer.bzl
```diff
@@ -0,0 +1,140 @@
+"creates tar layers from js_binary targets"
```
alexeagle, line 1: this docstring appears at the top of the markdown, so it needs more text and capitalization, etc

e2e/js_image/BUILD.bazel
```diff
@@ -1,19 +1,37 @@
 load("@npm//:defs.bzl", "npm_link_all_packages")
-load("@aspect_bazel_lib//lib:docs.bzl", "stardoc_with_diff_test", "update_docs")
-load("@aspect_rules_js//js:defs.bzl", "js_binary")
-load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("@aspect_rules_js//js:defs.bzl", "js_binary", "js_image_layer")
+load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "copy_to_bin")
 load("@io_bazel_rules_docker//container:container.bzl", "container_image")
 load("@io_bazel_rules_docker//contrib:test.bzl", "container_test")
 load("@io_bazel_rules_docker//container:layer.bzl", "container_layer")
```
alexeagle, line 6: hm, at what point would we use rules_oci here? Maybe it will be another package under this same e2e that shows the usage with those rules? I suppose ppl will want to see usage with rules_docker for at least a year while the migrations are happening

js/private/BUILD.bazel
```diff
@@ -156,3 +170,16 @@ copy_to_bin(
     srcs = ["js_run_devserver.mjs"],
     visibility = ["//visibility:public"],
 )
+
+write_source_files(
+    name = "js_image_layer_checked",
+    files = {
+        "js_image_layer.mjs": "//js/private/image:image.mjs",
```
alexeagle, line 177: could you add some docs somewhere explaining why we need to check in a generated file rather than just execute the sources directly?

js/private/image/index.ts
```diff
@@ -0,0 +1,120 @@
+import { Archiver, create } from 'archiver'
```
alexeagle, line 1: is this still a workaround for some problem in rules_pkg with the pkg_tar rule? Or do we just not like their python dependency?

js/private/image/index.ts
```diff
@@ -0,0 +1,120 @@
...
+                path.join(dir, dirent.name),
+                path.join(accumulate, dirent.name)
+            )
+        } else {
+            yield path.join(accumulate, dirent.name)
+        }
+    }
+}
+
+for (const key of Object.keys(entries).sort()) {
```
alexeagle, line 69: this whole file needs some commenting and maybe some tests

js/private/js_image_layer.bzl
```diff
@@ -0,0 +1,140 @@
...
+
+load("@aspect_bazel_lib//lib:paths.bzl", "to_manifest_path")
+load("@bazel_skylib//lib:paths.bzl", "paths")
+
+_doc = """Create container image layers from js_binary targets.
+
+```starlark
```
alexeagle, line 8: missing some text here like "for example, you could..."

and/or link to the e2e where it's easier for us to show long starlark code listings

e2e/js_image/BUILD.bazel
thesayyn: there's already a js example in rules_oci. https://github.com/bazel-contrib/rules_oci/tree/main/example/js. perhaps i should move it here?

js/private/image/index.ts
thesayyn: python dependency is a big factor here. I don't feel comfortable adding another dependency to rules_js.

js/private/image/index.ts
thesayyn: agreed. i'll get on that

e2e/js_image/BUILD.bazel
alexeagle: The example is also here to provide test coverage right? If rules_js changes could break it, then having an example here is good. If rules_oci changes might break it, then we might *also* want to have an example over there?

js/private/image/index.ts
alexeagle: okay that sounds right to me, could you just add some commenting or even a short README.md saying why it's designed this way?

e2e/js_image/BUILD.bazel
thesayyn: Yes, examples here are for rules_js test coverage. It's less likely that rules_oci will break just rules_js given it doesn't care about what creates the image layers.  Nevertheless, I think we should have both.

I'll add a rules_oci example here as well.

js/private/js_image_layer.bzl
```diff
@@ -0,0 +1,182 @@
...
+
+js_image_layer(
+    name = "layers",
+    binary = ":binary",
+    root = "/app"
+)
+
+oci_image(
+    name = "image",
+    tars = [
```
gregmagolan, line 40: does the entry_point & cmd get automatically set when using rules_oci?

js/private/js_image_layer.bzl
```diff
@@ -0,0 +1,182 @@
...
+    output_group = "node_modules"
+)
+container_layer(
+    name = "node_modules_layer",
+    tars = [":node_modules_tar"],
+)
+
+container_image(
+    name = "image",
+    cmd = ["/app/main.sh"],
```
gregmagolan, line 92: nit: perhaps make an `/app/main` entry point that redirects to `main.sh` since the `.sh` extension is an implementation detail of `js_binary` and not otherwise visible in a BUILD file?

js/private/js_image_layer.bzl
thesayyn: oh no. it doesn't. good catch.

js/private/js_image_layer.bzl
thesayyn: I like the idea. though I am a little worried that it might make people neglect the fact we need bash to start the program.

gregmagolan: technically the extension should be `.bash` if it depends on bash 😉 

the bash dependency will be encoded in the generated executable file anyway:

```
#!/usr/bin/env bash
exec main.sh
```

thesayyn: say no more 👍 done

.github/workflows/ci.bazelrc
```diff
@@ -18,10 +18,14 @@ build --grpc_keepalive_time=30s
 
 # Generic remote build execution
 build:rbe --extra_execution_platforms=@aspect_rules_js//platforms:x86_64_linux_remote
-build:rbe --genrule_strategy=remote
+build:rbe --genrule_strategy=remote,local
```
thesayyn, line 21: allows genrule to be executed locally if requested explicitly.

WORKSPACE
```diff
@@ -35,6 +37,11 @@ nodejs_register_toolchains(
     node_version = "16.13.1",
 )
 
+nodejs_register_toolchains(
+    name = "node18",
+    node_version = "18.12.0",
```
thesayyn, line 42: is this fine?

e2e/js_image_oci/pull.bzl
```diff
@@ -0,0 +1,30 @@
+"a repository rule fetching container images using crane"
```
alexeagle, line 1: Should users implement their own oci_pull rule like this? 
I'd expect at least a TODO about making a supported upstream API to do that.

e2e/js_image_oci/README.md
```diff
@@ -0,0 +1,50 @@
+# An example/e2e for rules_js + rules_oci
+
+The rule returns `tar` artifacts, suitable to include in the `tars` attribute of the `oci_image` rule from rules_oci.
```
alexeagle, line 3: "the rule" isn't clear, can you say which one

e2e/js_image_oci/README.md
```diff
@@ -0,0 +1,50 @@
...
+On the other hand, if the application code is changed, then only `app.tar` will be updated and pushed.
+
+## Selecting the right NodeJS interpreter
+
+By default `js_binary` gets the nodejs interpreter for the host platform. However, this is not the case when including the js_binary in js_image_layer thanks to transitions. See [#3373](https://github.com/bazelbuild/rules_nodejs/pull/3373) for how this works.
+
+Toolchain selection is controlled by `platform` attribute on `js_image_layer`.
+NodeJS interpreter for a different platform can be obtained by changing `platform`.
+
+Here is what the final image looks like when the platform is `linux/arm64`
```
alexeagle, line 21: It would be easier to understand a snippet of starlark where `platform = "linux/arm64"` so that you can see what you type in order to get the final image below

js/private/image/README.md
```diff
@@ -0,0 +1,16 @@
+# js_image_layer faq
+
+# why js_image_layer isn't like nodejs_image from rules_docker
```
alexeagle, line 3: nit: make these headings H2 so they are beneath the main one

js/private/image/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+That would require rules_js to take rules_docker as a dependency which would be not ideal given that rules_docker is in maintenance mode and [other](https://github.com/bazel-contrib/rules_oci) rulesets exists.
+So instead of being full-fledged js_image rule that is specific to one container building ruleset, it tries to be not opinionated and work with any ruleset out there.
+
+# why use custom builder script instead of using pkg_tar
+
+There are two reasons why js_image_layer doesn't use pkg_tar. First, it's python dependency and secondly, it doesn't work very well with `ctx.actions.symlink`. We needed to maintain a [patch](https://github.com/aspect-build/rules_js/blob/28dd54f6bdeb8d7d916271666782d327b258741b/e2e/js_image/js_image_layer.bzl#L135-L191) to make it work with rules_js which is specific to rules_js and can't be pushed upstream.
```
alexeagle, line 10: could you link to https://github.com/bazelbuild/rules_pkg/issues/115 instead? That's the reason we can't use pkg_tar yet, the patch is our workaround. I think users should start from the top if they read this FAQ and want to know "why".

js/private/image/index.ts
```diff
@@ -0,0 +1,258 @@
+import { createReadStream, createWriteStream } from 'node:fs'
```
alexeagle, line 1: add some file-level doc please

js/private/image/index.ts
```diff
@@ -0,0 +1,258 @@
...
+type Compression = 'gzip' | 'none'
+
+function findKeyByValue(entries: Entries, value: string): string {
+    for (const [key, { dest: val }] of Object.entries(entries)) {
+        if (val == value) {
+            return key
+        }
+    }
+    throw new Error(
+        `couldn't map ${value} to a path. please file a bug at https://github.com/aspect-build/rules_js/issues/new/choose`
```
alexeagle, line 35: nit: try to print at least the beginning of `Entries` here as well, or else such a filed bug will be hard for us to do anything with?

js/private/js_image_layer.bzl
```diff
@@ -0,0 +1,226 @@
...
+        "//command_line_option:platforms": str(attr.platform),
+    }
+
+_js_image_layer_transition = transition(
+    implementation = _js_image_layer_transition_impl,
+    inputs = [],
+    outputs = ["//command_line_option:platforms"],
+)
+
+js_image_layer = rule(
```
alexeagle, line 213: nit: to guarantee users can write custom rules that re-use our logic, you should also export a struct so that the attrs list and implementation function is accessible. See all the examples in bazel-lib

js/private/js_image_layer.bzl
```diff
@@ -0,0 +1,226 @@
...
+    implementation = _js_image_layer_impl,
+    doc = _doc,
+    attrs = {
+        "_allowlist_function_transition": attr.label(
+            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+        ),
+        "binary": attr.label(mandatory = True, cfg = _js_image_layer_transition, doc = "Label to an js_binary target"),
+        "root": attr.string(doc = "Path where the files from js_binary will reside in. eg: /apps/app1 or /app"),
+        "compression": attr.string(doc = "Compression algorithm. Can be one of `gzip`, `none`.", values = ["gzip", "none"], default = "gzip"),
```
alexeagle, line 222: nit: is this buildifier-formatted? make sure you've got the `pre-commit install`ed

I expected to see more newlines/indents in these attribute lists

.prettierignore
```diff
@@ -7,4 +7,5 @@ js/private/node-patches/fs.js
 min/
 npm/private/test/vendored/
 js/private/worker/worker.js
-js/private/worker/src/worker_protocol.ts
\ No newline at end of file
+js/private/worker/src/worker_protocol.ts
+js/private/js_image_layer.mjs
```
alexeagle, line 11: could you add it to .gitattributes as well so later changes to it don't appear in the code review?
same with coverage.js which appears edited in this PR

WORKSPACE
alexeagle: I don't see what it's used for, removing for now

e2e/js_image_oci/pull.bzl
alexeagle: filed https://github.com/bazel-contrib/rules_oci/issues/63

e2e/js_image_docker/BUILD.bazel
```diff
@@ -0,0 +1,91 @@
...
+)
+
+container_image(
+    name = "image",
+    architecture = "amd64",  # or arm64
+
+    # Since js_binary depends on bash we have to bring in a base image that has bash
+    base = "@debian_amd64//image",  # or "@debian_arm64//image", if you are on an arm machine
+    # This is going to be /{root of js_image_layer}/{package_name()}/{name of js_binary}
+    cmd = ["/app/main"],
```
gregmagolan, line 77: FYI. we chatted about also dropping the `.sh` extension from js_binary in the future. It seems we only added it to rules_js so js_binary was more backward compatible with rules_nodejs js_binary.

e2e/js_image_oci/README.md
gregmagolan: +1 one to this nit of being explicit with `platform = "linux/arm64"` instead of "platform is `linux/arm64`"

e2e/js_image_oci/WORKSPACE
```diff
@@ -0,0 +1,66 @@
...
+load("@contrib_rules_oci//oci:repositories.bzl", "LATEST_CRANE_VERSION", "LATEST_ZOT_VERSION", "oci_register_toolchains")
+
+oci_register_toolchains(
+    name = "oci",
+    crane_version = LATEST_CRANE_VERSION,
+    zot_version = LATEST_ZOT_VERSION,
+)
+
+## Pull base images
+load(":pull.bzl", "oci_pull")
```
gregmagolan, line 60: Is `oci_pull` going to end up in rules_oci still?

alexeagle: I filed https://github.com/bazel-contrib/rules_oci/issues/63

e2e/js_image_docker/BUILD.bazel
alexeagle: okay but nothing to comment here, right?

WORKSPACE
```diff
@@ -11,12 +11,14 @@ load("//js:repositories.bzl", "rules_js_dependencies")
...
 rules_js_dependencies()
 
-load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies", "register_jq_toolchains")
+load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies", "register_coreutils_toolchains", "register_jq_toolchains")
 
 aspect_bazel_lib_dependencies(override_local_config_platform = True)
 
 register_jq_toolchains()
 
+register_coreutils_toolchains()
```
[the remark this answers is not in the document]
alexeagle: It's not even required right now because the one assertion that used it is unstable and failing on CI, I'll remove

js/private/js_image_layer.bzl
```diff
@@ -0,0 +1,247 @@
...
+        destination = _runfile_path(ctx, file, runfiles_dir)
+        entry = {"dest": file.path, "root": file.root.path, "is_source": file.is_source, "is_directory": file.is_directory}
+        if destination == real_executable_path:
+            entry["remove_non_hermetic_lines"] = True
+        files[destination] = entry
+
+    entries = ctx.actions.declare_file("{}_entries.json".format(ctx.label.name))
+    ctx.actions.write(entries, content = json.encode(files))
+
+    extension = ".tar"
```
gregmagolan, line 168: nit: could shorten to
```
extension = ".tar.gz" if ctx.attr.compression == "gzip" else ".tar"
```

e2e/js_image_docker/BUILD.bazel
gregmagolan: yup

gregmagolan (approved):
🧙🧙🧙🧙🧙🧙

js/private/js_image_layer.bzl
alexeagle: done

e2e/js_image_docker/WORKSPACE
```diff
@@ -31,46 +31,10 @@ load("@npm//:repositories.bzl", "npm_repositories")
...
-    urls = ["https://github.com/bazelbuild/rules_go/releases/download/v0.35.0/rules_go-v0.35.0.zip"],
-)
-
-load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
-
-go_rules_dependencies()
-
-go_register_toolchains(version = "1.19.3")
-# WORKAROUND END
-
```
davidmcnamee, line 73: Not sure if it was intentional to remove this workaround, but I found the build wouldn't work on my M1 mac without it. Just wanted to make sure you knew; let me know if there's a better workaround that I'm missing

Issue #769, opened 2023-01-13
[Bug]: Using binaries from npm not working as expected


kreempuff:
### What happened?

I'm trying to use the [remotion cli](https://www.remotion.dev/) as [documented](https://github.com/aspect-build/rules_js/tree/d0e08695e50b868346835d8023f5557acd0c77bb/docs#using-binaries-published-to-npm) with this load statement:  `load("@video_production_playground_npm//:@remotion/cli/package_json.bzl", remotion_bin = "bin")` but I get an error:
```
ERROR: error loading package 'video-production/playground': cannot load '@video_production_playground_npm//:@remotion/cli/package_json.bzl': no such file
INFO: Elapsed time: 0.531s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (5 packages loaded)
FAILED: Build did NOT complete successfully (5 packages loaded)
currently loading: video-production/playground ... (2 packages)
```

It also fails when I try to replicate the typescript example. It feels like I'm missing something small but critical. 

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:
`bazel 5.4.0`

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
```
## Nodejs dependencies
http_archive(
    name = "aspect_rules_js",
    sha256 = "686d5b345592c1958b4aea24049d935ada11b83ae5538658d22b84b353cfbb1e",
    strip_prefix = "rules_js-1.13.1",
    url = "https://github.com/aspect-build/rules_js/archive/refs/tags/v1.13.1.tar.gz",
)


load("@aspect_rules_js//js:repositories.bzl", "rules_js_dependencies")
rules_js_dependencies()

load("@rules_nodejs//nodejs:repositories.bzl", "DEFAULT_NODE_VERSION", "nodejs_register_toolchains")
nodejs_register_toolchains(
    name = "nodejs",
    node_version = DEFAULT_NODE_VERSION,
)

load("@aspect_rules_js//npm:npm_import.bzl", "npm_translate_lock")

npm_translate_lock(
    name = "video_production_playground_npm",
    pnpm_lock = "//video-production/playground:pnpm-lock.yaml",
    verify_node_modules_ignored = "//:.bazelignore",
)

load("@video_production_playground_npm//:repositories.bzl", "npm_repositories")
npm_repositories()

## END Nodejs dependencies
```
Language(s) and/or frameworks involved:

### Fund our work

[...]

PR #773, opened 2023-01-13
docs: add example of how to use npm binary in subpackage

kreempuff:
More explicitly documents how to use an npm binary. Addresses #769

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -177,10 +177,49 @@ with a rule like [`copy_to_bin`](https://docs.aspect.build/aspect-build/bazel-li
 rules_js automatically mirrors the `bin` field from the `package.json` file of your npm dependencies
 to a Starlark API you can load from in your BUILD file or macro.
 
-For example, if you depend on the `typescript` npm package, you write this in `BUILD`:
+The format of the generated starlark file is `@<npm external repo name>//<package name where npm binary is loaded>:<npm package name>/package_json.bzl`.
+
+For example, let's imagine you have this folder structure:
+
+```
+app/
+├─ BUILD
+BUILD
+package.json
+WORKSPACE
+
+```
+and you have imported npm packages in your `WORKSPACE` file with the external repo name `npm`:
+
+```starlark=
+npm_translate_lock(
+    name = "npm",
+    ...
+)
+```
+
+To use the `typescript` npm package in the root bazel package of your project, you write this in the `BUILD` file:
 
 ```starlark=
 load("@npm//:typescript/package_json.bzl", typescript_bin = "bin")
+
+typescript_bin.tsc(
+    name = "compile",
+    srcs = [
+        "fs.ts",
+        "tsconfig.json",
+        "//:node_modules/@types/node",
+    ],
+    outs = ["fs.js"],
+    chdir = package_name(),
+    args = ["-p", "tsconfig.json"],
+)
+```
+
+To use the `typescript` npm package in the `//app` bazel package of your project, you write this in the `app/BUILD` file:
+
+```starlark=
+load("@npm//app:typescript/package_json.bzl", typescript_bin = "bin")
 
 typescript_bin.tsc(
     name = "compile",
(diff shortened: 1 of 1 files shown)


kreempuff:
Friendly ping @alexeagle

docs/README.md
```diff
@@ -177,7 +177,28 @@ with a rule like [`copy_to_bin`](https://docs.aspect.build/aspect-build/bazel-li
 rules_js automatically mirrors the `bin` field from the `package.json` file of your npm dependencies
 to a Starlark API you can load from in your BUILD file or macro.
 
-For example, if you depend on the `typescript` npm package, you write this in `BUILD`:
+The format of the generated starlark file is `@<npm external repo name>//<package name where npm binary is loaded>:<npm package name>/package_json.bzl`.
```
gregmagolan, line 180: Not quite right. The package segment `<package name where npm binary is loaded>` should be `<package name where npm package is linked>`. This is the same package of the package.json that defines the dependency on the npm package you're loading the bin from.

In the example below, the load statement would be

```
load("@npm//:typescript/package_json.bzl", typescript_bin = "bin")
```

since the `typescript` dep is coming from the root `package.json` and it is linked to the root package at  `//:node_modules/typescript`

kreempuff: Interesting. I'll update the PR! 

Can you point to some code and/or docs so I can solidify my understanding of the way this works?

gregmagolan: There are scattered examples throughout the rules & examples repos. Here is a simple one that shows the load package being different than the use package:

https://github.com/aspect-build/rules_ts/blob/5ea2d011369bf371d2431871bd178d6893766392/examples/typecheck_only/BUILD.bazel#L9

corresponding to the typescript version from this package.json:

https://github.com/aspect-build/rules_ts/blob/5ea2d011369bf371d2431871bd178d6893766392/examples/package.json#L17

kreempuff: Updated! Thanks for the pointers!

docs/README.md
```diff
@@ -177,24 +177,74 @@ with a rule like [`copy_to_bin`](https://docs.aspect.build/aspect-build/bazel-li
 rules_js automatically mirrors the `bin` field from the `package.json` file of your npm dependencies
 to a Starlark API you can load from in your BUILD file or macro.
 
-For example, if you depend on the `typescript` npm package, you write this in `BUILD`:
+The format of the generated starlark file is `@<npm external repo name>//<package name where npm package is linked>:<npm package name>/package_json.bzl`.
+
+`<package name where npm package is linked>` is the name of the *bazel* package where the npm package is linked. Linking npm packages happens with `npm_link_package` or
+`npm_link_all_packages`. When [`npm_link_package`](./npm_link_package.md) is used, the `root_package` parameter specifies the bazel package. When `npm_link_all_packages` is used, the bazel package is the location where `npm_link_all_packages` was called.
```
gregmagolan, line 183: Ahh. The `root_package` is something different altogether. It is the package that the `node_modules/.aspect_rules_js` virtual store is located at. This is typically the bazel package where the pnpm lock file is located. So the correct description here is that for both `npm_link_package` and `npm_link_all_packages`, the "package name where npm package is linked" is the package that those targets are located in.

docs/README.md
```diff
@@ -177,24 +177,74 @@ with a rule like [`copy_to_bin`](https://docs.aspect.build/aspect-build/bazel-li
...
-        "//:node_modules/@types/node",
-    ],
-    outs = ["fs.js"],
-    chdir = package_name(),
-    args = ["-p", "tsconfig.json"],
+    ...
 )
 ```
 
+this is because the `typescript` npm package is linked in the root bazel package of your project.
```
gregmagolan, line 246: perhaps the 2nd example should loading a bin from a npm package from an `app/package.json` such as
```
load("@npm//app:http-server/package_json.bzl", http_server_bin = "bin")
```
(or some other package)

jbedard:
@kreempuff would you still be able to finish this one off?

kreempuff:
> @kreempuff would you still be able to finish this one off?

Yeah, I just realized I don't fully understand how this repo is implemented. If you need me to, I'll close this out and reopen it again in when I have a better understanding.

jbedard:
Closing for now. Feel free to open a new one if you get back to this.

PR #780, opened 2023-01-14
chore: update devDeps of @kubernetes/client-node needed for prepare lifecycle hook to match the commit hash we're pulling

gregmagolan:
Last one

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,11 +25,10 @@
             },
             "@kubernetes/client-node": {
                 "//1": "these deps are from devDependencies of @kubernetes/client-node and needed for 'prepare' lifecycle hook",
-                "//2": "https://github.com/kubernetes-client/javascript/blob/763decf6e09d0e246b80cefdb2c5fd570373801a/package.json#L72",
+                "//2": "https://github.com/kubernetes-client/javascript/blob/fc681991e61c6808dd26012a2331f83671a11218/package.json",
                 "dependencies": {
-                    "@types/node": "^10.12.0",
-                    "@types/underscore": "^1.8.9",
-                    "typescript": "~4.7.4"
+                    "@types/underscore": "^1.11.4",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -25,11 +25,10 @@
...
             "@kubernetes/client-node": {
                 "//1": "these deps are from devDependencies of @kubernetes/client-node and needed for 'prepare' lifecycle hook",
-                "//2": "https://github.com/kubernetes-client/javascript/blob/763decf6e09d0e246b80cefdb2c5fd570373801a/package.json#L72",
+                "//2": "https://github.com/kubernetes-client/javascript/blob/fc681991e61c6808dd26012a2331f83671a11218/package.json",
                 "dependencies": {
-                    "@types/node": "^10.12.0",
-                    "@types/underscore": "^1.8.9",
-                    "typescript": "~4.7.4"
+                    "@types/underscore": "^1.11.4",
+                    "typescript": "~4.9.4"
```
jbedard, line 31: Why these ones?

jbedard (approved):

package.json
gregmagolan: comment explains why,
```
            "@kubernetes/client-node": {
                "//1": "these deps are from devDependencies of @kubernetes/client-node and needed for 'prepare' lifecycle hook",
                "//2": "https://github.com/kubernetes-client/javascript/blob/fc681991e61c6808dd26012a2331f83671a11218/package.json",
                "dependencies": {
                    "@types/underscore": "^1.11.4",
                    "typescript": "~4.9.4"
                }
            }
```

TL;DR is they are needed for the prepare lifecycle hook but they are specified only as devDependencies in the package.json of the package

jbedard: Oh right, I was looking at that 10 minutes ago 😅

(and on mobile right now the diff looks ugly and it's hard to see 🫣)

PR #793, opened 2023-01-15
Set fs._unpatched when patching the 'node:fs' module

gonzojive:
This allows inspecting the unsandboxed view of the file system if it's needed, such as for writing other sandbox guarding logic.


gregmagolan (approved):
LGTM. Thanks for the PR

PR #803, opened 2023-01-17
feat: add option to verify that all patches are declared in npm_translate_lock

kormide:
Here's an example of the output:
```
ERROR: in verify_patches_folder:

The following patches were found in /Users/derek/Code/rules_js/examples/npm_deps/patches but were not listed
in the `patches` attribute of `npm_translate_lock`.

- test-a@0.0.1.patch
- test-b.patch

To disable this check, remove the `verify_patches_folder` attribute from `npm_translate_lock`.
```

Do you think it's worth adding an e2e?


WORKSPACE
```diff
@@ -151,6 +151,7 @@ npm_translate_lock(
     },
     update_pnpm_lock = True,
     verify_node_modules_ignored = "//:.bazelignore",
+    verify_patches_folder = "//examples/npm_deps:patches",
```
kormide, line 154: Both this label and `"//:examples/npm_deps/patches"`. Are either of those more correct than the other?

gregmagolan: Because repository rules come before the analysis phase they are not away where the BUILD files are and the package boundaries so labels are ambiguous where `//examples/npm_deps:patches` is the same as `//:examples/npm_deps/patches`. Both work in this context tho I think sticking to where the Bazel package is and using `//examples/npm_deps:patches` is _more_ correct and also better to read

gregmagolan:
Hmm. How about just `verify_patches`? The folder seems redundant and that will be explained in the docstring anyway

kormide:
Changed name to `verify_patches`.

npm/npm_import.bzl
```diff
@@ -302,6 +303,9 @@ def npm_translate_lock(
 
             See https://github.com/bazelbuild/bazel/issues/8106
 
+        verify_patches: Verify that all patches in this folder appear in the `patches` attribute. Assumes
```
gregmagolan, line 306: nit: "Verify that all patches in the folder appear in the `patches` attribute. Does not recurse into subfolders."

kormide: 👍

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -566,6 +567,57 @@ or disable this check by setting `verify_node_modules_ignored = None` in `npm_tr
             )
             fail(msg)
 
+################################################################################
+# https://github.com/bazelbuild/bazel/issues/3901#issuecomment-421318536
+def _directory_path(repository_ctx, label):
```
gregmagolan, line 572: mind moving these into `npm_translate_lock.bzl` or perhaps a new `npm_translate_lock_helpers.bzl`. The `helpers` struct in this one is an artifact of the big refactor since I wanted to minimize the diff. I plan to move the functions that are not starlark generation specific out of this file.

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -566,6 +567,57 @@ or disable this check by setting `verify_node_modules_ignored = None` in `npm_tr
...
+    split = split[1].split(":", 1)
+    dir_path = split[0]
+    dir_name = split[1]
+    if dir_path == "/":
+        path = "/" + dir_name
+    else:
+        path = dir_path + "/" + dir_name
+
+    return repository_ctx.path(
+        str(repository_ctx.path(Label(prefix + "//:WORKSPACE"))).rsplit("/", 1)[0] +
```
gregmagolan, line 590: hmm. this will fail if the user has a `WORKPACE.bazel` file

kormide:
Added an e2e and added support for `patches` and `patch_args`.

kormide:
Hm, it looks like Bazel `5.1.1` doesn't have `repository_ctx.workspace_root`. Was there a reason we needed to support `5.1.1`?

kormide:
> Hm, it looks like Bazel `5.1.1` doesn't have `repository_ctx.workspace_root`. Was there a reason we needed to support `5.1.1`?

Same with `5.3.2` :(

kormide:
> > Hm, it looks like Bazel `5.1.1` doesn't have `repository_ctx.workspace_root`. Was there a reason we needed to support `5.1.1`?
> 
> Same with `5.3.2` :(

An alternative solution was just to pass the helper a reference label from the user workspace. Seems that we can rely on `pnpm_lock` always being there. The bootstrap logic should occur before this.

kormide:
Ready for another review pass.

npm/npm_import.bzl
```diff
@@ -302,6 +303,9 @@ def npm_translate_lock(
 
             See https://github.com/bazelbuild/bazel/issues/8106
 
+        verify_patches: Verify that all patches in the folder appear in the patches attribute. Does not recurse
```
gregmagolan, line 306: Oh. Should specify that this takes a path to a folder in the WORKSPACE. I think it would be confusing to take a label actually since labels to folders that are also packages are not valid.

WORKSPACE
```diff
@@ -151,6 +151,7 @@ npm_translate_lock(
     },
     update_pnpm_lock = True,
     verify_node_modules_ignored = "//:.bazelignore",
+    verify_patches = "//:examples/npm_deps/patches",
```
gregmagolan, line 154: This might read better as just a straight path and not a label
```
verify_patches = "examples/npm_deps/patches"
```
since labels to folders that are also package are not valid... and it would prevent someone from trying to put a label to an external WORKSPACE.

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -0,0 +1,52 @@
...
+    return rctx.path(
+        str(rctx.path(user_workspace_label)).rsplit("/", 1)[0] +
+        path,
+    )
+
+def _verify_patches(rctx):
+    if rctx.attr.verify_patches and rctx.attr.patches != None:
+        patches_folder_path = _directory_path(rctx, rctx.attr.verify_patches, rctx.attr.pnpm_lock)
+
+        all_patch_files = sets.make(patches_folder_path.readdir())
```
gregmagolan, line 28: lets filter on `.patch` and `.diff` files? perhaps make this configurable as well will with a `verify_patches_extensions` attribute that takes a list of extensions and defaults to `[".patch", "diff"]`?

kormide:
Addressed your comments and ready for another pass.

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -0,0 +1,49 @@
...
+def _workspace_path(rctx, short_path, user_workspace_label):
+    """Get the absolute path in the user repository of a short path using a reference label from the user repository."""
+    label_path = rctx.path(user_workspace_label)
+    label_short_path = user_workspace_label.package + ("/" if user_workspace_label.package else "") + user_workspace_label.name
+    workspace_path = str(label_path).removesuffix(label_short_path)
+    return rctx.path(workspace_path + "/" + short_path)
+
+def _verify_patches(rctx):
+    if rctx.attr.verify_patches and rctx.attr.patches != None:
+        patches_folder_path = _workspace_path(rctx, rctx.attr.verify_patches, rctx.attr.pnpm_lock)
```
gregmagolan, line 14: why not just
```
rctx.workspace_root + "/" + rctx.attr.verify_patches
```
?

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -0,0 +1,49 @@
...
+
+        files_in_patch_folder = patches_folder_path.readdir()
+        all_patch_files = sets.make()
+        for file in files_in_patch_folder:
+            if file.basename.find(".") >= 0:
+                if file.basename.rsplit(".")[-1] in rctx.attr.verify_patches_extensions:
```
gregmagolan, line 20: incase the extension the user gave has a dot like "foo.bar" this check is more robust as
```
for e in rctx.attr.verify_patches_extensions:
    if file.basename.endsiwth("."+e):
```

npm/private/npm_translate_lock_helpers.bzl
gregmagolan: ahh. its a bazel version thing 👍 ... yup, you can rely on pnpm_lock always being there

gregmagolan (approved):

PR #807, opened 2023-01-18
fix: fix a bug where patch extensions weren't correctly verified

kormide:
🤦

### DIFF
--- a/e2e/verify_patches/test.sh
+++ b/e2e/verify_patches/test.sh
@@ -20,6 +20,23 @@ if ! bazel build $BZLMOD_FLAG //...; then
   exit 1
 fi
 
+# Should succeed after adding an ignored patch extension
+touch patches/foo.diff
+if ! bazel build $BZLMOD_FLAG //...; then
+  echo "ERROR: expected 'bazel build $BZLMOD_FLAG //...' to pass"
+  exit 1
+fi
+rm patches/foo.diff
+
+# Should fail after adding a patch that isn't in `patches`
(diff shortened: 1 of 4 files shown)


e2e/verify_patches/test.sh
```diff
@@ -20,6 +20,23 @@ if ! bazel build $BZLMOD_FLAG //...; then
...
+# Should succeed after adding an ignored patch extension
+touch patches/foo.diff
+if ! bazel build $BZLMOD_FLAG //...; then
+  echo "ERROR: expected 'bazel build $BZLMOD_FLAG //...' to pass"
+  exit 1
+fi
+rm patches/foo.diff
+
+# Should fail after adding a patch that isn't in `patches`
+bazel clean --expunge # Need to invalidate the repository cache to see the missing file
```
kormide, line 32: I noticed this when writing the e2e. We won't detect new patch files that are missing until the repo rule invalidates.

gregmagolan: you can `bazel sync --only npm` instead of `bazel clean --expunge`; faster

gregmagolan (approved):

e2e/verify_patches/test.sh
kormide: Bazel sync doesn't work under bzlmod.

gregmagolan: ohh. i see i see.

you could touch any file that is an input to the npm_translate_lock in that case

PR #808, opened 2023-01-18
feat: verify patches via a generated source file

kormide:
This solves the problem where npm_translate_lock wouldn't invalidate when a patch was added to the patch folder.

### DIFF
--- a/npm/npm_import.bzl
+++ b/npm/npm_import.bzl
@@ -30,7 +30,9 @@ load("//npm/private:npm_import.bzl", _npm_import_lib = "npm_import", _npm_import
 load("//npm/private:versions.bzl", "PNPM_VERSIONS")
 load("//npm/private:utils.bzl", _utils = "utils")
 load("//npm/private:npm_translate_lock.bzl", _npm_translate_lock = "npm_translate_lock")
+load("//npm/private:list_sources.bzl", "list_sources")
 load("@aspect_bazel_lib//lib:repositories.bzl", _register_copy_directory_toolchains = "register_copy_directory_toolchains", _register_copy_to_directory_toolchains = "register_copy_to_directory_toolchains")
+load("@aspect_bazel_lib//lib:write_source_files.bzl", "write_source_file")
 
 LATEST_PNPM_VERSION = PNPM_VERSIONS.keys()[-1]
 
@@ -86,7 +88,6 @@ def npm_translate_lock(
         bins = {},
         verify_node_modules_ignored = None,
         verify_patches = None,
-        verify_patches_extensions = [".diff", ".patch"],
         quiet = True,
         link_workspace = None,
         pnpm_version = LATEST_PNPM_VERSION,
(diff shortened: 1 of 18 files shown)


npm/private/npm_translate_lock_state.bzl
```diff
@@ -104,6 +106,10 @@ def _init_common_labels(rctx, label_store):
     # pnpm-workspace.yaml file
     label_store.add_sibling("lock", "pnpm_workspace", PNPM_WORKSPACE_FILENAME)
 
+    # verify patches list
+    if attr.verify_patches:
```
gregmagolan, line 110: nit: could move this one to within `_init_patch_labels` since it is related to patches

npm/private/list_sources.bzl
```diff
@@ -0,0 +1,19 @@
+"Output a list of source files to a file"
+
+def _impl(ctx):
+    output = ctx.actions.declare_file(ctx.label.name + ".txt")
+    content = "\n".join([file.short_path for file in ctx.files.srcs]) + "\n"
```
gregmagolan, line 5: lets use file.path here so its obvious in the output if a generated files sneaks in here. for source files in the user workspace short_path and path are equivalent.

npm/npm_import.bzl
```diff
@@ -787,3 +795,33 @@ def npm_import(
         bins = bins,
         npm_translate_lock_repo = npm_translate_lock_repo,
     )
+
+def list_patches(name, include_patterns = ["*.diff", "*.patch"], exclude_patterns = []):
+    """Write a file containing all patches in the current folder to the source tree.
```
gregmagolan, line 800: nit: Write a file containing a list of all patches in the current folder to the source tree.

gregmagolan:
Looks great overall. I like the new pattern.

Lets drop the implicit `.txt` extension tho. Its not obvious without reading the rule code or the docs that the extension is added implicitly. I think I'm also allergy to `.txt` as it reminds me of Windows.

Sending you a .patch file over Slack that has my suggestion for that.

npm/npm_import.bzl
```diff
@@ -304,17 +305,28 @@ def npm_translate_lock(
...
+        verify_patches: Label to a patch list file.
+
+            Use this in together with the `list_patches` macro to guarantee that all patches in a patch folder
+            are included in the `patches` attribute.
 
             For example:
 
             ```
-            verify_patches = "path/to/patches",
+            verify_patches = "//patches:patches",
```
gregmagolan, line 316: in this case it would be `"//patches:patches.list"  to match the `out` of the `list_patches` in the example below

npm/npm_import.bzl
```diff
@@ -787,3 +798,33 @@ def npm_import(
...
+
+def list_patches(name, out = None, include_patterns = ["*.diff", "*.patch"], exclude_patterns = []):
+    """Write a file containing a list of all patches in the current folder to the source tree.
+
+    Use this together with the `verify_patches` attribute of `npm_translate_lock` to verify
+    that all patches in a patch folder are included. This macro stamps a test to ensure the
+    file stays up to date.
+
+    Args:
+        name: Name of the patch file (.txt is automatically appended)
```
gregmagolan, line 810: `.txt` is no longer automatically appended :)

gregmagolan (approved):

PR #809, opened 2023-01-18
chore: run e2e/vendored_node under bzlmod

kormide:


e2e/vendored_node/MODULE.bazel
```diff
@@ -0,0 +1,57 @@
...
+    module_name = "aspect_bazel_lib",
+    path = "../../../bazel-lib",
+)
+
+node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
+
+http = use_extension("@aspect_bazel_lib//lib:extensions.bzl", "http")
+
+http.http_archive(
+    name = "vendored_node_linux_amd64",
```
alexeagle, line 28: meh, I think leaving these in WORKSPACE.bzlmod is just as good for our current purposes and avoids the bazel-lib change being a blocker

kormide: Good point. Updated.

alexeagle (approved):

PR #810, opened 2023-01-18
chore: do not run workspace e2e under bzlmod

kormide:


.github/workflows/ci.yaml
```diff
@@ -143,6 +143,9 @@ jobs:
                       folder: e2e/bzlmod
                     - bzlmodEnabled: false
                       folder: e2e/bzlmod
+                    # Don't run workspace smoke test under bzlmod
```
alexeagle, line 146: just moving these lines higher?

kormide: Yep just clarifying that it should not be run, rather than to be converted later.

alexeagle (approved):

PR #811, opened 2023-01-19
fix: print the error message to log if verbose

thesayyn:


js/private/worker/src/rollup.config.js
```diff
@@ -6,5 +6,8 @@ import terser from '@rollup/plugin-terser'
 
 /** @type {import("rollup").RollupOptions} */
 export default {
-    plugins: [resolve(), commonjs(), json(), ts(), terser()],
+    plugins: [resolve(), commonjs(), json(), ts({ sourceMap: true }), terser()],
+    output: {
+        sourcemap: 'inline',
```
jbedard, line 11: Does this effect the stack traces outputted?

jbedard (approved):

js/private/worker/src/rollup.config.js
thesayyn: makes stack traces a bit more humanly.

PR #812, opened 2023-01-19
chore: run e2e/update_pnpm_lock_with_import under bzlmod

kormide:
* Required some changes to the extension and the label store
* Workflow for bootstrapping under bzlmod still works but is slightly different from WORKSPACE (see e2e test script).

Related to https://github.com/aspect-build/rules_js/issues/644


alexeagle (commented):
There should be a documentation update somewhere corresponding to the test.sh script changes, so that bzlmod users know how to update the lockfile too

e2e/update_pnpm_lock_with_import/test.sh
```diff
@@ -27,18 +38,30 @@ _sedi 's#"@types/node": "18.11.18"#"@types/node": "16"#' package.json
...
-  exit 1
+if [ ! $BZLMOD_FLAG ]; then
+  print_step "It should update the lockfile after a bazel sync with ASPECT_RULES_JS_FROZEN_PNPM_LOCK unset"
+  if ! bazel sync --only=npm; then
+    echo "ERROR: expected 'bazel sync --only=npm' to pass"
+    exit 1
+  fi
+else
+  print_step "It should update the lockfile after a failed bazel build with ASPECT_RULES_JS_FROZEN_PNPM_LOCK unset"
+  # bazel sync isn't load bearing under bzlmod.
```
gregmagolan, line 56: hmmm. the instructions in
```
def _fail_if_frozen_pnpm_lock(rctx, state):
    if RULES_JS_FROZEN_PNPM_LOCK_ENV in rctx.os.environ.keys() and rctx.os.environ[RULES_JS_FROZEN_PNPM_LOCK_ENV]:
        fail("""

ERROR: `{action_cache}` is out of date. `{pnpm_lock}` may require an update. To update run,

           bazel sync --only={rctx_name}

""".format(
``` 
are wrong then under bzlmod

gregmagolan: I also left a `use: bazel sync --only npm` reference in the docstring of `use_home_npmrc` that needs updating to handle the bzlmod case
```
        use_home_npmrc: Use the `$HOME/.npmrc` file (or `$USERPROFILE/.npmrc` when on Windows) if it exists.
            ...
            WARNING: The repository rule will not be invalidated by changes to the home `.npmrc` file since there
            is no way to specify this file as an input to the repository rule. If changes are made to the home
            `.npmrc` you can force the repository rule to re-run and pick up the changes by running:
            `bazel sync --only={name}` where `name` is the name of the `npm_translate_lock` you want to re-run.
```

kormide:
Going through the workflow manually again under bzlmod, it seems a bit off even if we were to change the messaging. I'm looking for a command other than `bazel build --enable_bzlmod //...` that can re-evaluate the `npm` repo because while it shows a message it also gives an unrelated build error.

kormide:
Okay I think I have a solution here which works for bzlmod and workspace. There's a new target in the npm_translate_lock repo that you can call:
```
bazel run @npm//:invalidate
```

Updated the docs. Ready for another review pass.

kormide:
Renamed to `bazel run @npm//:sync` after some bikeshedding with @gregmagolan .

gregmagolan:
> 

The old way will still work for WORKSPACE users so I wouldn't consider it breaking. We're just adding support for bzlmod too and changing the instructions for WORKSPACE users.

gregmagolan (approved):
🏋️

PR #818, opened 2023-01-24
feat: add dev_deps attribute to js_library

gregmagolan:
Inspired by https://github.com/aspect-build/rules_ts/pull/280.

Addresses https://github.com/aspect-build/rules_ts/issues/268.

Will require a **MINOR** release after this lands as this changes the default behaviour of `deps`. Still needs some tests.


js/private/js_library.bzl
```diff
@@ -82,6 +82,19 @@ _ATTRS = {
...
+
+        Npm dependencies in `deps` are propagated to the transitive npm dependencies of downstream `npm_package` targets when
+        they are linked.
+        """,
+        providers = [JsInfo],
+    ),
+    "dev_deps": attr.label_list(
+        doc = """Development dependencies of this target.
+
+        This may include other js_library targets or other targets that provide JsInfo.
```
jbedard, line 94: backticks around `js_library` and `JsInfo`?

js/private/js_library.bzl
```diff
@@ -82,6 +82,19 @@ _ATTRS = {
...
         runfiles of this target. They should appear in the '*.runfiles' area of any executable which is output by or has a
         runtime dependency on this target.
+
+        Npm dependencies in `deps` are propagated to the transitive npm dependencies of downstream `npm_package` targets when
+        they are linked.
+        """,
+        providers = [JsInfo],
+    ),
+    "dev_deps": attr.label_list(
+        doc = """Development dependencies of this target.
```
jbedard, line 92: "development" means compile-time but not runtime? Is there a better term for that? :/

gregmagolan: npm docs `"devDependencies": Packages that are only needed for local development and testing."` tho that doesn't translate over perfectly to rules_js. That is the idea, however, when it comes propagating deps through the linker.

https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file

js/private/js_library.bzl
gregmagolan: 👍

js/private/js_library_helpers.bzl
```diff
@@ -12,13 +12,8 @@ JS_LIBRARY_DATA_ATTR = attr.label_list(
...
     a runtime dependency on this target.
 
-    If this list contains linked npm packages, npm package store targets or other targets that provide `JsInfo`,
-    `NpmPackageStoreInfo` providers are gathered from `JsInfo`. This is done directly from the
-    `npm_package_store_deps` field of these. For linked npm package targets, the underlying npm_package_store
-    target(s) that back the links is used.
-
-    Gathered `NpmPackageStoreInfo` providers are used downstream as direct dependencies when linking a downstream
-    `npm_package` target with `npm_link_package`.
+    Npm dependencies in `data` are propagated to the transitive npm dependencies of downstream `npm_package` targets when
```
jbedard, line 15: and those in `deps`?

jbedard: The difference being the ones in `deps` are also part of the ts compilation, otherwise both get passed along to the `npm_package` the same way for use at runtime. Correct?

gregmagolan: `data` is actually very similar to deps except that the `DefaultInfo` files of `data` are also included as inputs & passed along to runfiles. `data` allows files and doesn't require the target have a `JsInfo` provider.

jbedard: and only `deps` are included in the ts compilation, while both are passed along to runfiles...?

gregmagolan: correct; data is not available to the tsc action itself.

jbedard: Great, that all aligns with how I was thinking about it 👍 

Should `deps` also be mentioned here though? Also maybe the fact that `dev_deps` are not propagated?

gregmagolan: In `data`? I don't think so. The docstrings for `deps` and `dev_deps` will be right beside the docstring for `data`. This `JS_LIBRARY_DATA_ATTR` is just here because I re-use this docstring in rules_webpack and rules_esbuild

gregmagolan: the js_library `data` docstring uses this const as well:
```
"data": JS_LIBRARY_DATA_ATTR,
```

jbedard: > In data? 

Ah I see, I forgot what this comment thread was actually on 👍

gregmagolan:
We may add this in the future but for now we're going to go with https://github.com/aspect-build/rules_js/pull/860

PR #825, opened 2023-01-27
fix: delete .git folder when downloading npm package from git

kormide:
Found that the `.git` folder was non-hermitic in some client work.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -178,16 +178,16 @@ jobs:
               with:
                   persist-credentials: false
 
-            # - name: Reconfigure git to use HTTP authentication
-            #   run: >
-            #       git config --global url."https://github.com/".insteadOf
-            #       ssh://git@github.com/
+            - name: Reconfigure git to use HTTP authentication
+              run: >
+                  git config --global url."https://github.com/".insteadOf
+                  ssh://git@github.com/
 
             # - uses: webfactory/ssh-agent@v0.7.0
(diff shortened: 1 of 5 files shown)


gregmagolan-codaio:
Nice one. Can you add a test to verify that the .git folder is not present in the one package we pull down with git+ssh:
```
"protoc-gen-grpc": "git+ssh://git@github.com:gregmagolan-codaio/protoc-gen-grpc-ts#be5580b06348d3eb9b4610a4a94065154a0df41f",
```

kormide:
Hm, I'm writing a test, but strangely I can't reproduce the `.git` folder existing on my local box. `.gitignore` is present 🤔

kormide:
Oh interesting, generated `repositories.bzl` file strips the transforms the `git+ssh` url into `https://codeload.github.com/...`. So by the time the git+ssh check occurs in npm_import, it doesn't go through this logic.

gregmagolan:
Hmm. You might need to switch the repo it pulls. I swear that one took the right code path when I landed it 🤨

kormide:
So from what I can tell, `pnpm` will take a `git+ssh` dependency and automatically convert it into a tarball url resolution in `pnpm-lock.yaml`. I think I can only replicate this case with a private repository.

It should be easier to replicate this case for a test using an `npm_import`.

gregmagolan:
Testing with npm import sgtm

kormide:
Updated

PR #828, opened 2023-01-29
[DO NOT MERGE] chore: Add the macOS runner to the matrix strategy

realtimetodie:
Adds the macOS runner to the matrix strategy of the test workflow job.

Notes 

- All test cases that depend on the Docker daemon are excluded for the macos-latest runner.

- The devserver test fails on the macos-latest runner.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -118,7 +118,7 @@ jobs:
             folders: ${{ toJSON(steps.*.outputs.folder) }}
 
     test:
-        runs-on: ubuntu-latest
+        runs-on: ${{ matrix.os }}
 
         needs:
             - matrix-prep-config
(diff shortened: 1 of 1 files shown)


realtimetodie:
Warning: This rules repository uses an extremely large matrix strategy. This change almost doubles the amount of matrix tests.

realtimetodie:
Related

- https://github.com/aspect-build/rules_esbuild/pull/118
- https://github.com/aspect-build/rules_terser/pull/64
- https://github.com/aspect-build/rules_ts/pull/298
- https://github.com/aspect-build/rules_webpack/pull/89
- https://github.com/aspect-build/rules_rollup/pull/65
- https://github.com/aspect-build/rules_jest/pull/94
- https://github.com/aspect-build/rules_jasmine/pull/35
- https://github.com/aspect-build/rules_cypress/pull/27
- https://github.com/aspect-build/rules_deno/pull/15

realtimetodie:
I can not confirm whether this works since I'm running into a limit of my GitHub pro account.

alexeagle:
MacOS also consumes GH Actions minutes at 10x the rate of Linux, so aspect-build has burned through our whole February quota already. We need to revert all of these.

realtimetodie:
@alexeagle Wait, can we not keep it at least for those rules that use binaries, eg. like rules_swc and rules_cypress? The latter seems to be broken on macOS https://github.com/aspect-build/rules_cypress/pull/27.

realtimetodie:
I caused this mess so let me create the appropriate PR's to revert the changes. For rules_swc, I come up with some minimal idea. I'm sorry for the trouble. I did not know about the tenfold burn rate for the macOS runner.

realtimetodie:
https://github.com/aspect-build/rules_deno/pull/17

realtimetodie:
@alexeagle I would integrate @gregmagolan's main branch change for rules that use either Node.js NAPI bindings (eg. rules_esbuild) or are independent from Node.js and binary based (eg. rules_swc) and remove it from all pure Node.js based rules. This is a minimal solution. What do you think?

alexeagle:
I should add, Google pays for the Bazel central registry precommit so we find out about mac/windows breakages at that time, though it means cutting a patch release if we do break it

PR #831, opened 2023-01-30
ci: make folder part of cache key

thesayyn:
Currently, all matrix members use the same cache key for storing caches, preventing all matrix folders to upload their own caches and use the from the root folder all the time. this slows down all matrices except the root.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -117,6 +117,16 @@ jobs:
             # Will look like [".", "e2e/bzlmod", ...]
             folders: ${{ toJSON(steps.*.outputs.folder) }}
 
+    matrix-prep-root-hash:
+        # Prepares the 'folder' axis of the test matrix
+        runs-on: ubuntu-latest
+        steps:
+            # need to checkout repository to hash files.
+            - uses: actions/checkout@v3
+            - id: root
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -182,8 +193,14 @@ jobs:
                   path: |
                       ~/.cache/bazel
                       ~/.cache/bazel-repo
-                  key: bazel-cache-${{ hashFiles('**/BUILD.bazel', '**/*.bzl', 'WORKSPACE', '**/*.js') }}
-                  restore-keys: bazel-cache-
+                  key: >-
+                      bazel-cache-
+                      ${{ matrix.config }}-${{ matrix.bazelversion }}-
```
thesayyn, line 198: leave out bazelversion and bzlmod from the key and make bazelisk have it's own cache.

gregmagolan: Nice one. Can you please put up a PR with the same change in the bazel-control rules-template? https://github.com/bazel-contrib/rules-template

gregmagolan:
Nice one. Can you please put up a PR with the same change in the bazel-control rules-template? https://github.com/bazel-contrib/rules-template

.github/workflows/ci.yaml
gregmagolan: 👍

jbedard:
@thesayyn was this ready to land? I thought we wanted to make changes after our last discussion?

PR #832, opened 2023-02-01
feat: support pnpm.patchedDependencies

kormide:
Automatically applies `pnpm.patchedDependencies`. Any patches in `patches` are applied after.


npm/private/npm_translate_lock_state.bzl
```diff
@@ -137,20 +137,26 @@ def _init_update_labels(rctx, label_store):
...
         label_store.add("verify_patches", rctx.attr.verify_patches)
 
     patches = []
     for pkg_patches in rctx.attr.patches.values():
         patches.extend(pkg_patches)
 
+    # Add patches from `pnpm.patchedDependencies`
+    for patch_info in priv["patched_dependencies"].values():
+        patches.append("//:%s" % patch_info.get("path"))
+
```
kormide, line 151: There is a bug here that I haven't solved. If you rename a patch file and try to update the lockfile with `update_pnpm_lock = True`, then this will fail because the lockfile is referencing a patch file that no longer exists, and the lockfile won't get updated.

npm/private/npm_translate_lock_state.bzl
```diff
@@ -53,6 +51,8 @@ WARNING: `update_pnpm_lock` attribute in `npm_translate_lock(name = "{rctx_name}
     if _should_update_pnpm_lock(priv):
         _init_importer_labels(priv, label_store)
 
+    _init_patch_labels(priv, rctx, label_store)
```
kormide, line 54: The patch labels now have to be processed after parsing the lockfile since some of the labels now come from the lockfile.

pnpm-lock.yaml
```diff
@@ -7,6 +7,11 @@ overrides:
 
 packageExtensionsChecksum: 60d08949101fc004c7564f98037b7764
 
+patchedDependencies:
+  meaning-of-life@1.0.0:
+    hash: o3deharooos255qt5xdujc3cuq
+    path: examples/npm_deps/patches/meaning-of-life@1.0.0-pnpm.patch
+
```
kormide, line 14: When you use `pnpm.patchedDependencies` you get this entry in the lockfile. It's an index of all of the patches.

pnpm-lock.yaml
```diff
@@ -3782,6 +3789,11 @@ packages:
       is-buffer: 1.1.6
     dev: true
 
+  /meaning-of-life/1.0.0_o3deharooos255qt5xdujc3cuq:
+    resolution: {integrity: sha512-fVA4xSydqtK9owabGcYw1r4EKEsMOVVeYQLeCXPu77Z+8Y2j2B2I16UqZlKIOHnYkJ4RSvpJ00ywy9IWjmuxYw==}
+    dev: true
+    patched: true
+
```
kormide, line 3796: Package entries in `pnpm-lock.yaml` can have the `patched` arg, and not how the alias changes, similar to peer deps.

npm/private/npm_translate_lock_state.bzl
kormide: Is one solution to parse these from `package.json` instead?

kormide: This is needed in order to precompute labels for the `verify_patches` feature.

kormide:
I'm planning to add an e2e so we can exercise it under bzlmod.

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -344,13 +345,34 @@ def _gen_npm_imports(importers, packages, root_package, rctx_name, attr, registr
         # gather patches & patch args
         patches, patches_keys = _gather_values_from_matching_names(True, attr.patches, name, friendly_name, unfriendly_name)
 
+        # Apply patch from `pnpm.patchedDependencies` first
+        if pnpm_patched:
+            patch_path = "//:%s" % patched_dependencies.get(friendly_name).get("path")
```
jbedard, line 350: Can we assume the patch is in `//:`? What if it's in a `filegroup` in `//my-patches:patches`?

kormide: Because it's a repository rule and runs before the analysis phase, the form of the label is flexible, so long as it has the correct path to the file.

WORKSPACE
```diff
@@ -136,11 +137,13 @@ npm_translate_lock(
     patch_args = {
         "*": ["-p1"],
         "@gregmagolan/test-a": ["-p4"],
+        "meaning-of-life@1.0.0": ["-p1"],
```
gregmagolan, line 140: `"*": ["-p1"],` should cover this case so this isn't needed right?

WORKSPACE
```diff
@@ -136,11 +137,13 @@ npm_translate_lock(
...
     },
     patches = {
         "@gregmagolan/test-a": ["//examples/npm_deps:patches/test-a.patch"],
         "@gregmagolan/test-a@0.0.1": ["//examples/npm_deps:patches/test-a@0.0.1.patch"],
         "@gregmagolan/test-b": ["//examples/npm_deps:patches/test-b.patch"],
+        "meaning-of-life@1.0.0": ["//examples/npm_deps:patches/meaning-of-life@1.0.0-after_pnpm.patch"],
```
gregmagolan, line 146: is this entry here now redundant since the patch file is in data and listed in the lock file?

kormide: This is to test that a patch can be applied via `patches` after a patch is applied from `pnpm.patchedDependencies`.

gregmagolan: ohh I see. this is a 2nd patch to show you can patch twice. 👍

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -344,13 +345,34 @@ def _gen_npm_imports(importers, packages, root_package, rctx_name, attr, registr
...
 
         patch_args, _ = _gather_values_from_matching_names(False, attr.patch_args, "*", name, friendly_name, unfriendly_name)
+
+        # Patches in `pnpm.patchedDependencies` must have the -p1 format. Therefore any
+        # patches applied via `patches` must also use -p1 since we don't support different
+        # patch args for different patches.
+        if pnpm_patched and not _has_strip_prefix_arg(patch_args, 1):
+            if _has_strip_prefix_arg(patch_args):
+                msg = """\
+ERROR: patch_args for package {package} contains a strip prefix that is incompatible with a patch applied via `pnpm.patchedDependencies`.
```
gregmagolan, line 367: great guard rails here

gregmagolan:
This looks great. Just one question of what happens with a patch is listed in pnpm. patchedDependencies but the user forgets to add it to data? Is it a hard fail or a warning that it is missing from data?

kormide:
> This looks great. Just one question of what happens with a patch is listed in pnpm. patchedDependencies but the user forgets to add it to data? Is it a hard fail or a warning that it is missing from data?

It not operates the same as the other data files: you'll get a warning but it will implicitly copy the patches over.

kormide:
Updated and ready for another pass.

npm/private/npm_translate_lock_state.bzl
kormide: Fixed this by reading patches from package.json.

npm/private/npm_translate_lock_state.bzl
```diff
@@ -151,8 +161,10 @@ def _init_patch_labels(rctx, label_store):
     for i, d in enumerate(patches):
         label_store.add("patches_{}".format(i), d)
 
+    priv["num_patches"] = len(patches)
+
 ################################################################################
-def _init_importer_labels(priv, label_store):
+def _init_importer_labels(priv, rctx, label_store):
```
gregmagolan, line 167: `rctx` added by mistake?

npm/private/npm_translate_lock_state.bzl
```diff
@@ -281,6 +293,31 @@ WARNING: Implicitly using package.json file `{package_json}` since the `{pnpm_lo
                 fail(msg)
             _copy_input_file(priv, rctx, label_store, package_json_key)
 
+    # pnpm.patchedDependencies patch files
+    root_package_json = _read_root_package_json(rctx, label_store)
```
gregmagolan, line 297: it looks like you're calling _read_root_package_json in two spots. You could cache the read to eliminate the duplicate file read & json parsing

kormide: Added caching inside `_read_root_package_json`.

gregmagolan (approved):
🏄

PR #833, opened 2023-02-01
release: publish release artifact

alexeagle:
GitHub's stability guarantee for the archive is iffy, and we want metrics on downloads. See bazel-contrib/SIG-rules-authors#11 (comment)


kormide (approved):
:surfing_man:

PR #834, opened 2023-02-01
fix: delete .git folder when downloading npm package from git

kormide:
An ssh keypair is required to clone via ssh. Secrets aren't accessible in GitHub actions on forks, so I had to move the test to an e2e that is exclusively run inside this repo.


gregmagolan (approved):
🧙

PR #837, opened 2023-02-02
fix: provide a better experience for patch_args when pnpm patches are used

kormide:
Revisiting this change as it was brought up recently in the Bazel slack group: https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1691179585032929

### DIFF
--- a/e2e/npm_translate_lock_subdir_patch/MODULE.bazel
+++ b/e2e/npm_translate_lock_subdir_patch/MODULE.bazel
@@ -11,9 +11,6 @@ npm.npm_translate_lock(
         "//subdir:package.json",
         "//subdir:patches/debug@4.3.4.patch",
     ],
-    patch_args = {
-        "*": ["-p1"],
-    },
     pnpm_lock = "//subdir:pnpm-lock.yaml",
     verify_node_modules_ignored = "//:.bazelignore",
 )
--- a/npm/private/npm_translate_lock.bzl
+++ b/npm/private/npm_translate_lock.bzl
@@ -521,6 +521,15 @@ WARNING: `package_json` attribute in `npm_translate_lock(name = "{name}")` is de
         lifecycle_hooks_execution_requirements,
     )
 
+    # The default `patch_args` of {"*": ["-p0"]} conflicts with the pnpm.patchedDependencies
+    # requirement of "-p1" for patches. When pnpm patches are used, it's highly likely that
+    # `patches` is unset, so when it is empty remove any strip prefixes from the default or
+    # any user-supplied argument as they won't have an effect anyway.
+    if len(patches) == 0 and patch_args:
+        patch_args = dict(**patch_args)
+        for matcher, args in patch_args.items():
+            patch_args[matcher] = _remove_strip_prefix_arg(args)
+
     npm_translate_lock_rule(
         name = name,
         pnpm_lock = pnpm_lock,
@@ -787,3 +796,14 @@ ERROR: `{action_cache}` is out of date. `{pnpm_lock}` may require an update. To 
             pnpm_lock = state.label_store.relative_path("pnpm_lock"),
             rctx_name = rctx.name,
         ))
+
+def _remove_strip_prefix_arg(patch_args):
+    patch_args = patch_args[:]
+    remove_arg = None
+    for arg in patch_args:
+        if arg.startswith("-p") or arg.startswith("--strip="):
+            remove_arg = arg
+            break
+    if remove_arg:
+        patch_args.remove(remove_arg)
+    return patch_args
--- a/e2e/npm_translate_lock_subdir_patch/WORKSPACE
+++ b/e2e/npm_translate_lock_subdir_patch/WORKSPACE
@@ -22,9 +22,6 @@ npm_translate_lock(
         "//subdir:package.json",
         "//subdir:patches/debug@4.3.4.patch",
     ],
-    patch_args = {
-        "*": ["-p1"],
-    },
     pnpm_lock = "//subdir:pnpm-lock.yaml",
     verify_node_modules_ignored = "//:.bazelignore",
 )
--- a/npm/private/npm_translate_lock_generate.bzl
+++ b/npm/private/npm_translate_lock_generate.bzl
@@ -373,7 +373,9 @@ def _gen_npm_imports(importers, packages, patched_dependencies, root_package, rc
                 msg = """\
 ERROR: patch_args for package {package} contains a strip prefix that is incompatible with a patch applied via `pnpm.patchedDependencies`.
 
-`pnpm.patchedDependencies` requires a strip prefix of `-p1`. All applied patches must use the same strip prefix.
+`pnpm.patchedDependencies` requires a strip prefix of `-p1`. All applied patches to a package must use the same strip prefix.
+
+Double check that the `patch_args` in `npm_translate_lock` does not declare a different strip prefix.
 
 """.format(package = friendly_name)
                 fail(msg)


npm/npm_import.bzl
```diff
@@ -430,6 +430,12 @@ WARNING: `package_json` attribute in `npm_translate_lock(name = "{name}")` is de
         if "no-sandbox" not in lifecycle_hooks_execution_requirements["*"]:
             lifecycle_hooks_execution_requirements["*"].append("no-sandbox")
 
+    # TODO (2.0): Stop defaulting patch_args with {"*": ["-p0"]} when no patches are supplied
+    # as this conflicts with the default pnpm.patchedDependencies format of "-p1" and
+    # needs to be explicitly overridden by the user.
+    if len(patches) == 0:
+        patch_args = {}
+
```
kormide, line 438: I figure this isn't a breaking change because it won't affect anyone who doesn't have `patches`.

jbedard:
Did an issue come up with this or is it just confusing / a bad API now that the pnpm one has a different default?

kormide:
> Did an issue come up with this or is it just confusing / a bad API now that the pnpm one has a different default?

I just noticed that I have to override `patch_args` when I don't have any `patches1 in some client work.

kormide:
Alternatively, what I could do is have patches ignore the `patch_args` in `npm_import` if there are no non-pnpm patches. Wdyt @gregmagolan ?

jbedard:
> Patch args only make sense when there are patches since pnpm patches are always p1

Aren't there other patch args that the pnpm patches might need?

kormide:
> > Patch args only make sense when there are patches since pnpm patches are always p1
> 
> Aren't there other patch args that the pnpm patches might need?

pnpm patches have to be `-p1`.

jbedard:
Aren't there other arguments though? I thought it was a generic array of arguments and not only the `p*`?

gregmagolan (approved):

kormide:
> Aren't there other arguments though? I thought it was a generic array of arguments and not only the `p*`?

None that are defaulted in the macro.

kormide:
Oh I see what you mean @jbedard . I shouldn't remove any non-strip args here. I'll do another pass.

kormide:
> Oh I see what you mean @jbedard . I shouldn't remove any non-strip args here. I'll do another pass.

Actually no, if there are no `patches` then there should be no need to apply any other args to pnpm patches...

jbedard:
But now if patches are added then the default args for pnpm suddenly change. I don't think adding an entry to `patches` should suddenly change the args to pnpm patches. Or am I reading that wrong?

kormide:
> But now if patches are added then the default args for pnpm suddenly change. I don't think adding an entry to `patches` should suddenly change the args to pnpm patches. Or am I reading that wrong?

Yeah, that's true. I think there might be another way to solve this. Maybe "*" patch_args just shouldn't apply to pnpm patches (at least any strip args in there).

jbedard:
Is the pnpm format a default of "-p1" but it can be changed? Maybe aligning our default with pnpm is best then?

kormide:
> Is the pnpm format a default of "-p1" but it can be changed? Maybe aligning our default with pnpm is best then?

Cannot change the format for pnpm patches, and changing our default would be a breaking change.

jbedard:
I guess not applying `*` to pnpm patches is one option like you said, but that's an odd exception :/

Do we need to apply the patch args to pnpm at all? Normally with pnpm how are patch args applied? Why do we have to add our own patch args?

kormide:
> I guess not applying `*` to pnpm patches is one option like you said, but that's an odd exception :/
> 
> Do we need to apply the patch args to pnpm at all? Normally with pnpm how are patch args applied? Why do we have to add our own patch args?

It's just because pnpm patches and `patches` follow the same implementation path, and `patches` allows you to be flexible with what arguments are passed.

jbedard:
> It's just because pnpm patches and patches follow the same implementation path

So it would be hard to treat pnpm different and ignore the patch args?

kormide:
> > It's just because pnpm patches and patches follow the same implementation path
> 
> So it would be hard to treat pnpm different and ignore the patch args?

I think it's doable. I'll do a pass when I get back to this.

kormide:
Finally got back to this. I improved the error message when a strip prefix is set for `patches` that won't work with `pnpm.patchedDependencies`. I also removed the strip prefix when there are no patches supplied so that the default of `-p1` will be used for pnpm patches.

e2e/npm_translate_lock_subdir_patch/MODULE.bazel
```diff
@@ -11,9 +11,6 @@ npm.npm_translate_lock(
         "//subdir:package.json",
         "//subdir:patches/debug@4.3.4.patch",
     ],
-    patch_args = {
-        "*": ["-p1"],
-    },
```
kormide, line 16: The patch arg has to be `-p1` for this e2e where `pnpm.patchedDependencies` is used, so this exclusion exercises the logic.

npm/private/npm_translate_lock.bzl
```diff
@@ -521,6 +521,15 @@ WARNING: `package_json` attribute in `npm_translate_lock(name = "{name}")` is de
         lifecycle_hooks_execution_requirements,
     )
 
+    # The default `patch_args` of {"*": ["-p0"]} conflicts with the pnpm.patchedDependencies
+    # requirement of "-p1" for patches. When pnpm patches are used, it's highly likely that
+    # `patches` is unset, so when it is empty remove any strip prefixes from the default or
+    # any user-supplied argument as they won't have an effect anyway.
+    if len(patches) == 0 and patch_args:
```
jbedard, line 528: Would it be possible and any nicer changing the default to `patch_args = None` and adding the default when there are patches instead of clearing it when no patches? Mainly to avoid the extra strip method...

```
def npm_translate_lock(
        name,
        ....
        patches = {},
        patch_args = None,

...

    if len(patches) > 0 and patch_args == None:
       patch_args = {"*": ["-p0"]}
```

kormide: My intention to perform the substitution whether `patch_args` is defaulted or the user supplies a value. As long as `patches` is empty, the strip prefix should be irrelevant because we always override the pnpm patches with `-p1`.

jbedard: Then they'd be overriding the `None` default and we wouldn't have to do anything though?

kormide: I didn't want to default to `None` because then the stardocs would be inaccurate about what the default it.

jbedard: Isn't it even more inaccurate having the magic defaults depending on a combo of attributes and values?

kormide: In general I would say so, however these changes to the defaults cannot have any effect if there are no manual patches supplied, and only helps to avoid a common error case. When there are only pnpm patches, any other strip prefix doesn't make sense, so it's harmless to remove it as we override it anyway.

kormide: In the current state, someone who doesn't set `patches` or `patch_args` and just uses pnpm patches gets the error.

kormide: @gregmagolan Did you have any opinions on this?

gregmagolan: Catching up on my `patch`-fu. It looks like by default (if you don't specify `-p0`) then `patch` will strip off all leading directories and just leave the base filename. That explains why we want `-p0` as the reasonable default.

How about calling `_remove_strip_prefix_arg` right before appending the `-p1` to `patch_args` if `pnpm_patched` is True in `npm/private/npm_translate_lock_generate.bzl` instead of doing it here?

jbedard: Can we avoid `_remove_strip_prefix_arg` completely though? I feel like it's odd (and dangerous) to modify a user-provided config in certain cases, it just leads to too many edge cases and hard to understand scenarios?

gregmagolan: That's what we do currently and it causes issues with the default `-p0` when using pnpm patches which prefer the `-p1`

jbedard: I mean change how our defaults work. Different defaults depending if we are using `patches` or assuming we're using pnpm (because `patches` is empty). Different defaults in these 2 cases instead of modifying/converting the args in 1/2 cases.

If that makes sense?

gregmagolan: It's hard to document in that case since stardoc wouldn't show the default.

I think it is cleaner to document that `-p1` is forced for any patches that are listed in the pnpm package.json patches and both the default and any user supplied `patch_args` for `-p` or `--strip` are ignored in that case.

jbedard: Can you think of a way of doing that without the current `_remove_strip_prefix_arg()` implementation though?

Scanning args for various `-p*` or `--strip=*` names and modifying the args is just something I'd prefer avoiding, especially with almost no test coverage :/

gregmagolan: I don't have any better ideas besides that. Maybe just appending `-p1` is sufficient if patch lets the last strip argument win but we'd have to test that.

jbedard:
Done with https://github.com/aspect-build/rules_js/pull/1694

PR #838, opened 2023-02-03
feat: add external_repository_action_cache attribute to npm_translate_lock to make action cache location configurable

### LINKED ISSUE #824: Configurable .aspect dir location?
Is it possible to configure the path of the `.aspect` directory? We would like to avoid adding additional dot folders to the root of our repository.

gregmagolan:
We've got this request a few times now on slack and now via an issue

Fixes #824


npm/npm_import.bzl
```diff
@@ -332,8 +333,9 @@ def npm_translate_lock(
 
         quiet: Set to False to print info logs and output stdout & stderr of pnpm lock update actions to the console.
 
-        link_workspace: The workspace name where links will be created for the packages in this lock file.
+        external_repository_action_cache: The location of the external repository action cache to write to when `update_pnpm_lock` = True.
```
jbedard, line 336: This is just one thing within the `.aspect` directory, will we potentially ever want to customize more? Should we just make the `.aspect` configurable instead of everything within it (if we add more...)?

gregmagolan: I don't see this repository rule every using any thing else but the `external_repository_action_cache` directory which currently defaults to `.aspect/rules/external_repository_action_cache`. Maybe the customization could be at the `.aspect/rules` level but then I bet users would complain that they can't control the `external_repository_action_cache` subdirectory path. This way is gives users maximum flexibility. It should be a rare case anyway once the Aspect CLI is in wide use since that uses the `.aspect` folder and it will not be configurable.

kormide (approved):

npm/npm_import.bzl
jbedard: It just means if we put more files in there then each and every one will need its own argument like this one. So if all you want to do is change `.aspect` to `.foo` you'll have to do it for every one of these options. I think just replacing `.aspect` with something already git-ignored was the request in https://github.com/aspect-build/rules_js/issues/824?

gregmagolan: Having it gitignored would defeat the purpose of the action cache tho

jbedard: Sorry it was to reduce "addition dot folders". For some reason that thought made me think of reducing folders in a .gitignore 🤦. Same idea though, that request is just looking to rename the top level `.aspect` to reduce the number of dot folders.

My thought was if we add additional things in `.aspect` then we'd need a separate param for all of them. Although I guess we'd need a separate param on each rule anyway so maybe it makes no difference :/

gregmagolan: Yeah. I see it more of an opt-out for some users and then they would want maximum flexibility. If a user is also using the CLI then they won't want to change this anyway.

jbedard (approved):

npm/npm_import.bzl
jbedard: 👍

PR #839, opened 2023-02-03
fix: don't read root package.json if there is no root importer

gregmagolan:
Fixes issue introduced in https://github.com/aspect-build/rules_js/pull/832. It hasn't been reported yet as not having a root package.json is not a common use case but someone will hit it eventually.


npm/private/npm_translate_lock_generate.bzl
```diff
@@ -522,12 +522,7 @@ def _find_missing_bazel_ignores(root_package, importer_paths, bazelignore):
 
     # The pnpm-lock.yaml file package needs to be prefixed on paths
     for i in importer_paths:
-        if i == ".":
-            expected = root_package
-        else:
-            expected = paths.normalize(paths.join(root_package, i))
-
-        expected = paths.join(expected, "node_modules")
+        expected = paths.normalize(paths.join(root_package, i, "node_modules"))
```
gregmagolan, line 525: the logic here was broken; if the importer path was `..` you'd end up with `./node_modules` as the expected instead of just `node_modules`

kormide (approved):
🦈

PR #840, opened 2023-02-03
chore: update workflows config to latest format

gregmagolan:


gregmagolan:
Dup of https://github.com/aspect-build/rules_js/pull/841

PR #844, opened 2023-02-04
fix: support single quotes in npmrc properties

### LINKED ISSUE #822: [Bug]: .npmrc registry is not sanitized
### What happened?

If your `.npmrc` has the following

```
registry = 'https://url.com'
```

this rule will take `'` and not remove the strings. This will cause installs to fail.

### Version

Development (host) and target OS/architectures: Ventura 13.2

Output of `bazel --version`: 6.0.0

### How to reproduce

```shell
add a registry to .npmrc that is equal to a string
```

### Any other information?

Maybe surface a validation message. Took me a while to find this. Had to `bazel fetch` the target to find the failure

### Fund our work

jbedard:
Fixes #822

### DIFF
--- a/npm/private/npmrc.bzl
+++ b/npm/private/npmrc.bzl
@@ -47,7 +47,7 @@ def parse_npmrc(npmrc_content):
         name = p[0].strip()
         value = p[2].strip()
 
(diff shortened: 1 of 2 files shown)


gregmagolan (approved):
🌮

PR #855, opened 2023-02-08
ci: build and run unit tests with bzlmod

kormide:
Fixes https://github.com/aspect-build/rules_js/issues/644

This should close off the bzlmod support for rules_js ticket.

Depends on: https://github.com/aspect-build/bazel-lib/pull/375

### DIFF
--- a/.bazelrc.common
+++ b/.bazelrc.common
@@ -70,3 +70,7 @@ build --incompatible_strict_action_env
 
 # Enable with --config=release
 build:release --stamp --workspace_status_command=$(pwd)/workspace_status.sh
+
+# Config to enable building and testing with bzlmod
+common:bzlmod --enable_bzlmod
+common:bzlmod --//:flag_bzlmod
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -174,8 +174,6 @@ jobs:
                     - bazelversion: 5.1.1
                       bzlmodEnabled: true
                     # TODO: un-exclude the following bzlmod tests once they work
-                    - folder: .
-                      bzlmodEnabled: true
                     - folder: e2e/js_image_docker
                       bzlmodEnabled: true
                     - folder: e2e/js_image_oci
@@ -236,17 +234,24 @@ jobs:
                   ENGFLOW_CLIENT_CRT: ${{ secrets.ENGFLOW_CLIENT_CRT }}
                   ENGFLOW_PRIVATE_KEY: ${{ secrets.ENGFLOW_PRIVATE_KEY }}
 
-            - name: Set bzlmod flag
+            - name: Set bzlmod flag (for integration tests)
               # Store the --enable_bzlmod flag that we add to the test command below
               # only when we're running bzlmod in our test matrix.
               id: set_bzlmod_flag
-              if: matrix.bzlmodEnabled
+              if: ${{ matrix.bzlmodEnabled && matrix.folder != '.'}}
               run: echo "bzlmod_flag=--enable_bzlmod" >> $GITHUB_OUTPUT
+
+            - name: Set bzlmod configuration (for test //...)
+              # Store the --config=bzlmod flag that we add to the test command below
+              # only when we're running bzlmod in our test matrix.
+              id: set_bzlmod_config
+              if: ${{ matrix.bzlmodEnabled && matrix.folder == '.'}}
+              run: echo "bzlmod_config=--config=bzlmod" >> $GITHUB_OUTPUT
 
             - name: bazel test //...
               working-directory: ${{ matrix.folder }}
               run: |
-                  bazel --bazelrc=$GITHUB_WORKSPACE/.github/workflows/ci.bazelrc --bazelrc=.bazelrc test --config=${{ matrix.config }} ${{ steps.set_bzlmod_flag.outputs.bzlmod_flag }} //...
+                  bazel --bazelrc=$GITHUB_WORKSPACE/.github/workflows/ci.bazelrc --bazelrc=.bazelrc test --config=${{ matrix.config }} ${{ steps.set_bzlmod_flag.outputs.bzlmod_flag }} ${{ steps.set_bzlmod_config.outputs.bzlmod_config }} //...
                   ls $(bazel info output_base)/external | grep -v __links | grep -vz unused
               env:
                   # Bazelisk will download bazel to here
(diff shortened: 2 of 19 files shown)


.bazelrc.common
```diff
@@ -70,3 +70,7 @@ build --incompatible_strict_action_env
 
 # Enable with --config=release
 build:release --stamp --workspace_status_command=$(pwd)/workspace_status.sh
+
+# Config to enable building and testing with bzlmod
+build:bzlmod --enable_bzlmod
+build:bzlmod --//:flag_bzlmod
```
kormide, line 76: Unfortunately we can't declare a `config_setting` on `--enable_bzlmod`, so we have to detect it for `select`s this way.

.github/workflows/ci.yaml
```diff
@@ -236,17 +234,24 @@ jobs:
...
+              if: ${{ matrix.bzlmodEnabled && matrix.folder != '.'}}
               run: echo "bzlmod_flag=--enable_bzlmod" >> $GITHUB_OUTPUT
 
+            - name: Set bzlmod configuration (for test //...)
+              # Store the --enable_bzlmod flag that we add to the test command below
+              # only when we're running bzlmod in our test matrix.
+              id: set_bzlmod_config
+              if: ${{ matrix.bzlmodEnabled && matrix.folder == '.'}}
+              run: echo "bzlmod_config=--config=bzlmod" >> $GITHUB_OUTPUT
+
```
kormide, line 250: In the case of `test //...` we need to set `--config=bzlmod` rather than just an `--enable_bzlmod` flag.

docs/BUILD.bazel
```diff
@@ -3,46 +3,97 @@ load("@aspect_bazel_lib//lib:docs.bzl", "stardoc_with_diff_test", "update_docs")
 stardoc_with_diff_test(
     name = "js_binary",
     bzl_library_target = "//js/private:js_binary",
+    # https://github.com/bazelbuild/stardoc/pull/141
+    target_compatible_with = select({
+        "//:bzlmod": ["@platforms//:incompatible"],
+        "//conditions:default": [],
+    }),
```
kormide, line 10: Stardoc doesn't work with bzlmod enabled.

js/dev_repositories.bzl
```diff
@@ -21,14 +21,16 @@ def rules_js_dev_dependencies():
...
-        strip_prefix = "bazel-skylib-1.3.0",
-        urls = ["https://github.com/bazelbuild/bazel-skylib/archive/refs/tags/1.3.0.tar.gz"],
+        sha256 = "b8a1527901774180afc798aeb28c4634bdccf19c4d98e7bdd1ce79d1fe9aaad7",
+        urls = ["https://github.com/bazelbuild/bazel-skylib/releases/download/1.4.1/bazel-skylib-1.4.1.tar.gz"],
+    )
+
+    http_archive(
+        name = "bazel_skylib_gazelle_plugin",
+        sha256 = "0a466b61f331585f06ecdbbf2480b9edf70e067a53f261e0596acd573a7d2dc3",
+        urls = ["https://github.com/bazelbuild/bazel-skylib/releases/download/1.4.1/bazel-skylib-gazelle-plugin-1.4.1.tar.gz"],
```
kormide, line 33: We no longer have to download the source repo :tada:

kormide: I can move this out into a prefactor.

kormide:
Ready for review

e2e/pnpm_workspace_deps/MODULE.bazel
```diff
@@ -16,6 +16,7 @@ node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
 
 node.toolchain(
     name = "nodejs",
+    node_version = "16.9.0",
```
kormide, line 19: I had to make the version the same as in MODULE.bazel in rules_js due to this bug which needs to be fixed upstream.

https://github.com/bazelbuild/rules_nodejs/pull/3626

alexeagle: that bug fix is released, want to update this PR?

kormide: Yup just publishing the new rnj version to bcr.

kormide:
Node toolchain fixes are in. This one is ready for another review @gregmagolan @alexeagle.

BUILD.bazel
```diff
@@ -61,3 +62,15 @@ buildifier(
     lint_mode = "warn",
     mode = "diff",
 )
+
+bool_flag(
+    name = "flag_bzlmod",
```
alexeagle, line 67: comment about why we need this strange thing

MODULE.bazel
```diff
@@ -10,6 +10,15 @@ bazel_dep(name = "aspect_bazel_lib", version = "1.27.2")
 bazel_dep(name = "bazel_skylib", version = "1.4.1")
 bazel_dep(name = "rules_nodejs", version = "5.8.2")
 bazel_dep(name = "platforms", version = "0.0.4")
+bazel_dep(name = "gazelle", version = "0.28.0", repo_name = "bazel_gazelle")
```
alexeagle, line 13: shouldn't this be a devdep?

docs/BUILD.bazel
alexeagle: fix is coming soon I hear. this is an interesting workaround

docs/BUILD.bazel
```diff
@@ -3,46 +3,97 @@ load("@aspect_bazel_lib//lib:docs.bzl", "stardoc_with_diff_test", "update_docs")
...
 )
 
 stardoc_with_diff_test(
     name = "npm_import",
     bzl_library_target = "//npm:npm_import",
+    # https://github.com/bazelbuild/stardoc/pull/141
```
alexeagle, line 85: this is a lot of boilerplate, maybe it could go in a wrapper macro or even in bazel-lib where this symbol is loaded?

BUILD.bazel
alexeagle: and, we should file the issue if someone hasn't already, that it should be possible to sense this like we do for `--stamp` https://github.com/aspect-build/bazel-lib/blob/main/lib/private/stamping.bzl#L29

alexeagle (commented):
cool

MODULE.bazel
```diff
@@ -10,9 +10,22 @@ bazel_dep(name = "aspect_bazel_lib", version = "1.27.2")
...
+
+ext.host()
+
+use_repo(ext, "aspect_bazel_lib_host")
 
 node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
 
+node.toolchain(
+    node_version = "16.9.0"
+)
```
kormide, line 27: WORKSPACE registers this specific version. If we don't do it here, one of our tests fails which checks the npm version.

BUILD.bazel
kormide: @Wyverald didn't think that could be easily done.

docs/BUILD.bazel
kormide: Added a macro

BUILD.bazel
kormide: Added some clarifying comments.

.bazelrc.common
alexeagle: That seems like a bug (I assume you tried it)

kormide: Yep, I tried.

alexeagle (approved):

PR #860, opened 2023-02-10
feat: propagate non-dev npm dependencies in js_library deps to direct dependencies of downstream linked npm_package targets

gregmagolan:
This also fixes the bug where "dev" dependencies in the `data` attribute are incorrectly propagated to the direct dependencies of downstream linked `npm_package` targets.

Once landed, downstream build rules such as `ts_project` will be updated to follow the same pattern.


kormide:
Needs rebase.

jbedard (approved):

PR #865, opened 2023-02-11
chore: update to latest version of Aspect CLI and turn on additional workflows features

gregmagolan:
Can't turn on gazelle quite yet as the bzl_library targets are out-of-date and not sure if gazelle is generating the right things there yet

### DIFF
--- a/.aspect/workflows/config.yaml
+++ b/.aspect/workflows/config.yaml
@@ -5,49 +5,79 @@ workspaces:
         label: rules_js
     e2e/bzlmod:
         icon: bazel
+        tasks:
+            buildifier:
+                without: true
     e2e/js_image_docker:
         icon: docker
+        tasks:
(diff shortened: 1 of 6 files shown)


.aspect/workflows/config.yaml
```diff
@@ -48,6 +48,8 @@ workspaces:
 tasks:
     warming:
     branch_freshness:
+        update_strategy: rebase
```
mattem, line 51: These can be removed really as they are all the defaults, you get the `warming` and `branch_freshness` tasks for free.

gregmagolan: default for branch_freshness is `rebase`?

mattem: Yup

mattem (approved):

PR #876, opened 2023-02-13
fix(performance): cache patch labels early to avoid unnecessary reloading of lockfile

kormide:


npm/private/npm_translate_lock_state.bzl
```diff
@@ -108,8 +118,9 @@ def _init_common_labels(rctx, label_store):
     # pnpm-workspace.yaml file
     label_store.add_sibling("lock", "pnpm_workspace", PNPM_WORKSPACE_FILENAME)
 
-    # root package.json file
-    label_store.add_sibling("lock", "package_json_root", PACKAGE_JSON_FILENAME)
+    # root package.json file (if one exists)
+    if "//:package.json" in rctx.attr.data:
```
gregmagolan, line 122: not quite right; the root package.json will be a sibling to the pnpm-lock file so its Bazel package needs to be derived from where the lock file is located

kormide: Updated

npm/private/npm_translate_lock_state.bzl
```diff
@@ -93,6 +103,12 @@ def _init_common_labels(rctx, label_store):
     if attr.pnpm_lock:
         label_store.add("pnpm_lock", attr.pnpm_lock, seed_root = True)
         label_store.add("lock", attr.pnpm_lock)
+
+        # root package.json file
+        root_package_json_label = str(attr.pnpm_lock).replace(PNPM_LOCK_FILENAME, PACKAGE_JSON_FILENAME)
+        if root_package_json_label in rctx.attr.data:
```
gregmagolan, line 109: hmmm. it is possible that this won't work if someone doesn't get their labels straight:
```
pnpm_lock = "//path/to:pnpm-lock.yaml",
data = ["//:path/to/package.json"],
```

gregmagolan (changes requested):

npm/private/npm_translate_lock_state.bzl
kormide: Good catch. Fixed.

gregmagolan (approved):
🚀

Issue #899, opened 2023-02-22
[Bug]: Local vendored Node.js binary doesn't resolve


mattrobmattrob:
### What happened?

Custom `http_archive` provided Node.js binary works but locally vendored one does not.

```
FATAL: aspect_rules_js[js_binary]: node binary '/private/var/tmp/_bazel_matt.robinson/4eb5286194401ed417482737bd8dd3cc/execroot/my-repo/bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/external/aspect_rules_js/npm/private/lifecycle/lifecycle-hooks.sh.runfiles/my-repo/my-repo/tools/snoozel/nodejs/node-v16.19.0-darwin-arm64/bin/node' not found in runfiles
```

### Version

Development (host) and target OS/architectures: macOS, arm64

Output of `bazel --version`: 6.0.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
```Starlark
http_archive(
    name = "aspect_rules_js",
    sha256 = "9fadde0ae6e0101755b8aedabf7d80b166491a8de297c60f6a5179cd0d0fea58",
    strip_prefix = "rules_js-1.20.0",
    url = "https://github.com/aspect-build/rules_js/releases/download/v1.20.0/rules_js-v1.20.0.tar.gz",
)

http_archive(
    name = "aspect_rules_ts",
    sha256 = "db77d904284d21121ae63dbaaadfd8c75ff6d21ad229f92038b415c1ad5019cc",
    strip_prefix = "rules_ts-1.3.0",
    url = "https://github.com/aspect-build/rules_ts/releases/download/v1.3.0/rules_ts-v1.3.0.tar.gz",
)
```

Language(s) and/or frameworks involved:

### How to reproduce

`WORKSPACE`:
```
http_archive(
    name = "aspect_rules_js",
    sha256 = "9fadde0ae6e0101755b8aedabf7d80b166491a8de297c60f6a5179cd0d0fea58",
    strip_prefix = "rules_js-1.20.0",
    url = "https://github.com/aspect-build/rules_js/releases/download/v1.20.0/rules_js-v1.20.0.tar.gz",
)

http_archive(
    name = "aspect_rules_ts",
    sha256 = "db77d904284d21121ae63dbaaadfd8c75ff6d21ad229f92038b415c1ad5019cc",
    strip_prefix = "rules_ts-1.3.0",
[...]

PR #900, opened 2023-02-22
fix: add support for local `node_toolchain`

mattrobmattrob:
# Summary
This closes https://github.com/aspect-build/rules_js/issues/899 and it has all the context. Make the `_target_tool_short_path` path resolve when it’s in an external repo or repo-local.

### DIFF
--- a/js/private/coverage/coverage.sh.tpl
+++ b/js/private/coverage/coverage.sh.tpl
@@ -30,7 +30,7 @@ if [ ! -f "$entry_point" ]; then
     exit 1
 fi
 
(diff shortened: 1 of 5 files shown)


gregmagolan (approved):
Looks great. Thanks for the fix!

gregmagolan:
Pushed some more commits to fix the node path in coverage.sh.tpl. The failure there also exposed an unset var there that I removed so actual errors can surface. Should go green now.

mattrobmattrob:
Thanks, @gregmagolan!

PR #901, opened 2023-02-22
fix: workaround ambiguous symlinks on osx

kormide:
This bug only surfaces on OSX. Can't reproduce on Linux.

### DIFF
--- a/js/private/image/index.ts
+++ b/js/private/image/index.ts
@@ -39,7 +39,27 @@ function findKeyByValue(entries: Entries, value: string): string {
 async function* walk(dir: string, accumulate = '') {
     const dirents = await readdir(dir, { withFileTypes: true })
     for (const dirent of dirents) {
-        if (dirent.isDirectory()) {
+        let isDirectory = dirent.isDirectory();
+
+        if (dirent.isSymbolicLink() && !dirent.isDirectory() && !dirent.isFile()) {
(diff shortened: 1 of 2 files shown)


js/private/image/index.ts
```diff
@@ -39,7 +39,27 @@ function findKeyByValue(entries: Entries, value: string): string {
...
     const dirents = await readdir(dir, { withFileTypes: true })
     for (const dirent of dirents) {
-        if (dirent.isDirectory()) {
+        let isDirectory = dirent.isDirectory();
+
+        if (dirent.isSymbolicLink() && !dirent.isDirectory() && !dirent.isFile()) {
+            // On OSX we sometimes encounter this bug: https://github.com/nodejs/node/issues/30646
+            // The entry is apparently a symlink, but it's ambiguous whether it's a symlink to a
+            // file or to a directory, and lstat doesn't tell us either. Determine the type by
+            // attempting to read it as a directory.
```
gregmagolan, line 48: woah. that is a sneaky one

thesayyn (approved):

thesayyn:
it would be great to have a test for this.

kormide:
> it would be great to have a test for this.

I agree, but since we don't run ci on macox I'm not sure the best way to do that :thinking:

thesayyn:
right...

Issue #902, opened 2023-02-22
[Bug]: js_image_layer doesn't work with `--nobuild_runfile_links`


kormide:
### What happened?

The binary's runfiles aren't included in the app layer tar when the `--nobuild_runfile_links` option is set.
### Version

Development (host) and target OS/architectures: osx, arm64

Output of `bazel --version`: 6.0.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: v1.20.0

### How to reproduce

```shell
cd e2e/js_image_docker
bazel test //... --nobuild_runfile_links
```

kormide:
fyi @thesayyn

gregmagolan:
This intersects with what we do with Workflows so we should figure out what the principled fix is here.

cc @mattem @alexeagle

alexeagle:
I see that the e2e/js_image_oci also fails with that flag.

PR #905, opened 2023-02-24
fix: include runfiles tree in js_image_layer when --nobuild_runfile_links is set

kormide:
Fixes https://github.com/aspect-build/rules_js/issues/902.

The runfiles tree doesn't need to exist, we just need the correct path to it.

### DIFF
--- a/js/private/js_image_layer.bzl
+++ b/js/private/js_image_layer.bzl
@@ -136,7 +136,15 @@ def _runfile_path(ctx, file, runfiles_dir):
 def _runfiles_dir(root, default_info):
     manifest = default_info.files_to_run.runfiles_manifest
 
-    runfiles = manifest.short_path.replace(manifest.basename, "")[:-1]
+    nobuild_runfile_links_is_set = manifest.short_path.endswith("_manifest")
+
+    if nobuild_runfile_links_is_set:
+        # When `--nobuild_runfile_links` is set, runfiles_manifest points to the manifest
+        # file sitting adjacent to the runfiles tree rather than within it.
+        runfiles = default_info.files_to_run.runfiles_manifest.short_path.replace("_manifest", "")
+    else:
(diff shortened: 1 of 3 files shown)


js/private/js_image_layer.bzl
```diff
@@ -136,7 +136,15 @@ def _runfile_path(ctx, file, runfiles_dir):
 def _runfiles_dir(root, default_info):
     manifest = default_info.files_to_run.runfiles_manifest
 
-    runfiles = manifest.short_path.replace(manifest.basename, "")[:-1]
+    nobuild_runfile_links_is_set = manifest.short_path.endswith("_manifest")
+
+    if nobuild_runfile_links_is_set:
+        # When `--nobuild_runfile_links` is set, runfiles_manifest points to the manifest
+        # file sitting adjacent to the runfiles tree rather than within it.
+        runfiles = default_info.files_to_run.runfiles_manifest.short_path.replace("_manifest", "")
```
kormide, line 144: I wonder if this is a safe assumption to make.

kormide:
I should add the same test to `js_image_oci`.

js/private/js_image_layer.bzl
kormide: I asked a question on the Bazel slack here: https://bazelbuild.slack.com/archives/CA31HN1T3/p1677196663633109

gregmagolan:
Looks good to me. I'll defer to @thesayyn for approval since he knows this code best.

js/private/js_image_layer.bzl
thesayyn: I might need to check the Runfiles.java to make this is the case. is this blocking you currently?

kormide: Yeah it's blocking some client work.

thesayyn (approved):
looks good to me. I'll have detailed look at the behavior of Runfiles.java.

Issue #864, opened 2023-02-11
[Bug]: Multiple conflicting toolchains declared for name nodejs


condemil:
### What happened?

When bzlmod is used and node toolchain is registered with specified `node_version` the following error appears:

```
ERROR: Traceback (most recent call last):
        File "/home/me/.cache/bazel/_bazel_me/25283051ada197906a83de4165c71df9/external/rules_nodejs~5.8.0/nodejs/extensions.bzl", line 13, column 21, in _toolchain_extension
                fail("Multiple conflicting toolchains declared for name {} ({} and {})".format(
Error in fail: Multiple conflicting toolchains declared for name nodejs (16.19.0 and 18.12.1)
```

There is no error if WORKSPACE is used instead of bzlmod.

### Version

Development (host) and target OS/architectures: Ubuntu 22.04 x86_64

Output of `bazel --version`: bazel 6.0.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: 1.19.0

Language(s) and/or frameworks involved: nodejs

### How to reproduce

Minimal reproducible example:

.bazelrc

```
common --enable_bzlmod
```

BUILD.bazel

```
load("@aspect_rules_js//js:defs.bzl", "js_binary")

js_binary(
    name = "version",
    entry_point = "main.js",
)
```

main.js

```js
console.log(process.version);
```

MODULE.bazel

```starlark
bazel_dep(name = "aspect_rules_js", version = "1.19.0")

bazel_dep(name = "rules_nodejs", version = "5.8.0")

node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")

node.toolchain(
    name = "nodejs",
    node_version = "18.12.1",
)

use_repo(node, "nodejs_toolchains")
use_repo(node, "nodejs_linux_amd64")

register_toolchains(
    "@nodejs_toolchains//:all"
)
```

WORKSPACE

```
# empty
```

### Fund our work

- [ ] Sponsor our open source work by donating a [bug bounty](https://opencollective.com/aspect-build/)

gregmagolan:
Thanks for the bug report. Looks like something is mis-wired. @kormide or I will look into it soon.

kormide:
@condemil I'm investigating the right way to handle toolchain registration in bzlmod, but if you want a workaround in the meantime you can just rename `nodejs` to something like `node18`, as well as in the subsequent `use_repo`s and toolchain registration.

[...]

PR #906, opened 2023-02-24
chore: update to rules_nodejs 5.8.2

kormide:
In support of https://github.com/aspect-build/rules_js/issues/644.

Fixes https://github.com/aspect-build/rules_js/issues/864.

Node toolchain no longer needs to be explicitly registered unless using a different default version or registering a second copy.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -8,14 +8,10 @@ module(
 
 bazel_dep(name = "aspect_bazel_lib", version = "1.27.2")
 bazel_dep(name = "bazel_skylib", version = "1.4.1")
-bazel_dep(name = "rules_nodejs", version = "5.8.0")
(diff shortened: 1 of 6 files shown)


MODULE.bazel
```diff
@@ -8,15 +8,11 @@ module(
 
 bazel_dep(name = "aspect_bazel_lib", version = "1.27.2")
 bazel_dep(name = "bazel_skylib", version = "1.4.1")
-bazel_dep(name = "rules_nodejs", version = "5.8.0")
+bazel_dep(name = "rules_nodejs", version = "5.8.2")
```
kormide, line 11: Including this dep is enough to register the node toolchain at the default version.

alexeagle: yeah looks much better

alexeagle (approved):

PR #919, opened 2023-03-02
docs: update bzlmod snippet to demonstrate pinning node version

kormide:


.github/workflows/release_prep.sh
```diff
@@ -18,6 +18,14 @@ cat << EOF
...
 \`\`\`starlark
 bazel_dep(name = "aspect_rules_js", version = "${TAG:1}")
+bazel_dep(name = "rules_nodejs", version = "5.8.2")
+
+# Pin a node version (optional)
+node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
+
+node.toolchain(
+    node_version = "16.9.0",
+)
```
alexeagle, line 28: nit: it's hard to tell which lines here are optional, maybe

```
###
# By default you get the node version from [XX default location]
# Optionally you can pin a node version:
bazel_dep(name = "rules_nodejs", version = "5.8.2")
node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
node.toolchain(node_version = "16.9.0")
# End pinning node version
################
```

alexeagle: nit: also ideally this would come from a SNIP in the actual MODULE.bazel so we know it works, but don't spend too much time on it :)

kormide: Updated to what you have above, but we could leave this open until I figure out SNIP semantics.

PR #929, opened 2023-03-07
feat: add use_execroot_entry_point option to js_run_devserver (when running tool)

gregmagolan:
Similar to https://github.com/aspect-build/rules_webpack/pull/98 but for js_run_devserver. js_run_binary also has this capability.

This resolves issues of for tools like webpack devserver & next.js (which uses webpack devserver) where webpack & react can be found in multiple node_modules trees leading to build failures.


js/private/js_run_devserver.bzl
```diff
@@ -4,103 +4,20 @@ load(":js_binary.bzl", "js_binary_lib")
...
-`//:node_modules/next` are handled efficiently. Since these targets are symlinks in the output
-tree, they are recreated as symlinks in the custom sandbox and do not incur a fully copy of the
-underlying npm packages.
-
-Supports running with [ibazel](https://github.com/bazelbuild/bazel-watcher).
-Only `data` files that change on incremental builds are synchronized when running with ibazel.
-"""
-
 _attrs = dicts.add(js_binary_lib.attrs, {
-    "tool": attr.label(
```
jbedard, line 85: Is this a breaking change?

gregmagolan: No. The macro the replaces the rule in the public API still has `tool`

docs/js_run_devserver.md
```diff
@@ -90,28 +88,16 @@ Supports running with [ibazel](https://github.com/bazelbuild/bazel-watcher).
 Only `data` files that change on incremental builds are synchronized when running with ibazel.
 
 
-**ATTRIBUTES**
```
jbedard, line 93: Were most of these params incorrect?

js/private/js_run_devserver.bzl
```diff
@@ -4,103 +4,20 @@ load(":js_binary.bzl", "js_binary_lib")
 load(":js_binary_helpers.bzl", _gather_files_from_js_providers = "gather_files_from_js_providers")
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 
-_DOC = """Runs a devserver via binary target or command.
```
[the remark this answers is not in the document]
gregmagolan: Sadly not possible since the docs for a macro have to be an inline string and not a variable

docs/js_run_devserver.md
gregmagolan: They were all inherited from js_binary... now most are not re-documented and instead `**kwargs` just links to the js_binary docs:

```
All other args from <code>js_binary</code> except for <code>entry_point</code> which is set implicitly.<br><br><code>entry_point</code> is set implicitly by <code>js_run_devserver</code> and cannot be overridden.<br><br>See https://docs.aspect.build/rules/aspect_rules_js/docs/js_binary
```

We do the same thing in the `jest_test` macro.

js/private/js_run_devserver.bzl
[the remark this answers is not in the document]
jbedard: Could still do the macro in a separate commit but I guess that's its own mess then :(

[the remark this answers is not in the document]
gregmagolan: Not with the `tool` attribute change since I need to macro to pipe `tool` from the public API to two different attributes in the rule now that allow you to switch between `cfg = "exec"` and `cfg = "target"`

js/private/js_run_devserver.bzl
jbedard: Yeah I see it now 😅

dgp1130:
Just FYI, I suspect there might be an unintentional breaking change here. Upgrading to version `1.21.0` and running `live-server` via `js_run_devserver()` but executed [as a subprocess](https://github.com/dgp1130/rules_prerender/blob/c737aea5e0ee2143bbf44a54536e1f6b30399fc4/common/testing/devserver.mts#L128) via a [`data` dependency on the server](https://github.com/dgp1130/rules_prerender/blob/c737aea5e0ee2143bbf44a54536e1f6b30399fc4/examples/minimal/BUILD.bazel#L56) led to:

```
FATAL: aspect_rules_js[js_binary]: the entry_point '/home/doug/.cache/bazel/_bazel_doug/0000b3b3fa770637f16a71f2a8299a7e/sandbox/linux-sandbox/5100/execroot/rules_prerender/bazel-out/k8-fastbuild/bin/node_modules/.aspect_rules_js/live-server@1.2.2/node_modules/live-server/./live-server.js' not found
```

I was able to fix it with `use_execroot_entry_point = False`, so not that big a deal, but seems like this wasn't intended as a breaking change, so figured I'd mention it. I am using the devserver in an uncommon way by leveraging it as a test server though, so it might not be that broadly impacting of an issue.

gregmagolan:
Yeah, that is possible if you're doing something unusual. 

Is the file `bazel-out/k8-fastbuild/bin/node_modules/.aspect_rules_js/live-server@1.2.2/node_modules/live-server/./live-server.js` found under the execution platform transition `k8-opt-exec`?

dgp1130:
Not quite, in `--sandbox_debug` and digging around, there is only `k8-fastbuild`. The package is actually under runfiles for my test since it's a data dependency executed at test time.

```
.../execroot/rules_prerender/bazel-out/k8-fastbuild/bin/examples/minimal/test_chromium-local.sh.runfiles/rules_prerender/node_modules/.aspect_rules_js/live-server@1.2.2/node_modules/live-server/live-server.js
```

gregmagolan:
I see. The `use_execroot_entry_point` intentionally uses the node_modules in the execroot instead of the runfiles so that npm packages are not inadvertently found in both locations which causes build failures with tools such are react & next.

In your case opting out with `use_execroot_entry_point = False` is correct unless include the deps you need as direct inputs instead of transitive runfiles. The underlying mechanism is that an `JS_BINARY__USE_EXECROOT_ENTRY_POINT` env var [is set ](https://github.com/aspect-build/rules_js/blob/f940df4a10fae79c6133c1bca08840691ab6d55b/js/private/js_run_devserver.mjs#L172)that affects how a js_binary tool calculates the location of the entry point script.

gregmagolan:
js_run_binary also uses the same entry point by default and how a special block to hoist runfiles to inputs https://github.com/aspect-build/rules_js/blob/f940df4a10fae79c6133c1bca08840691ab6d55b/js/private/js_run_binary.bzl#L309

dgp1130:
Yeah, in this case the devserver and all its deps should be in runfiles of the test, so I think using that is appropriate for my situation.

Issue #935, opened 2023-03-09
[FR]: Allow tools to write to files in the js_run_devserver sandbox


gregmagolan:
### What is the current behavior?

Files copied to the js_run_devserver sandbox from bazel-out are write protected since bazel makes all files in bazel-out write protected.

### Describe the feature

chmod files after copying to allow them to be written too since some tools such as Next.js want to write to certain files such as the tsconfig.json file when running the devserver

### Fund our work

- [ ] Sponsor our open source work by donating a [feature bounty](https://opencollective.com/aspect-build/)

PR #936, opened 2023-03-10
feat: add grant_sandbox_write_permissions attribute to js_run_devserver

gregmagolan:
Resolves https://github.com/aspect-build/rules_js/issues/935


js/private/js_run_devserver.mjs
```diff
@@ -88,6 +89,14 @@ async function syncRecursive(src, dst) {
                 mkdirpSync(path.dirname(dst))
             }
             await fs.promises.copyFile(src, dst)
+            if (writePerm) {
+                // add write permissions to file (preserving executable permissions if file has them)
+                if (await fs.promises.access(dst, fs.constants.X_OK)) {
```
jbedard, line 94: Can we get the permissions and add read instead of hardcoding the `755` and `644`?

Something such as (I haven't tested this 😅)...:
```
const p = await fs.promises.access(dst);
await fs.promises.chmod(dst, p | fs.constants.S_IXUSR)
```

gregmagolan: Its a temporary sandbox of files so I think the hard-coded values would be fine but I'll try your suggestion out. Thanks

kormide (approved):

PR #960, opened 2023-03-24
chore: update dependencies

gregmagolan:


kormide (approved):

js/repositories.bzl
```diff
@@ -7,14 +7,10 @@ See https://docs.bazel.build/versions/main/skylark/deploying.html#dependencies
 load("//js/private:maybe.bzl", http_archive = "maybe_http_archive")
 
 def rules_js_dependencies():
-    "Dependencies for users of aspect_rules_js"
```
jbedard, line 10: Any specific reason for dropping this? It's not used/visible somewhere?

jbedard (approved):

js/repositories.bzl
gregmagolan: Just cleanup as the equivalent macros in all other rule sets don't have the docstring comment. Its not used/visible anywhere on the docsite.

jbedard: 👍

PR #961, opened 2023-03-24
chore: fix spacing of js_library docstrings

gregmagolan:
These docstrings indents are off


alexeagle:
it's fixed in our stardoc actually...
https://github.com/bazelbuild/bazel/pull/13403
https://github.com/aspect-build/stardoc/commit/34d1bdbf6aa58f676ff4b69d77ff434672ba49ac
😁

gregmagolan:
The docsite I keep having to manually fix things after slurping & generating 😢

js/private/js_library.bzl
```diff
@@ -46,45 +46,45 @@ _ATTRS = {
...
-        ```
-        js_library(
-            name = "js_lib",
-            declarations = ["index.js"],
-        )
-        ```
-        """,
+```
+js_library(
+    name = "js_lib",
```
jbedard, line 70: I think these lines will still be broken if I'm reading this right... :(

jbedard (approved):

js/private/js_library.bzl
gregmagolan: They should be good. The code block is left aligned.

jbedard: I see no new lines or `<br>` in the generated html though...?

gregmagolan:
Ohhh.. I see what the root cause is on the docsite. Some of these docstrings end up with mixed indentations since they are build up from string partials with format 😵‍💫

js/private/js_library.bzl
```diff
@@ -46,45 +46,45 @@ _ATTRS = {
     "srcs": attr.label_list(
         doc = """Source files that are included in this library.
```
alexeagle, line 47: nit: when I'm paranoid and de-dent docs to the file level, I start with backslash-newline so the first line is also at column 0
```
          """\
First line goes here
```

PR #970, opened 2023-03-29
chore: remove Bazel 5.1.1 from CI matrix since to reduce CI queue times

gregmagolan:
Less concerned with Bazel 5.1 regressions at this point with Bazel 6.1 out.


jbedard (approved):

kormide (approved):
This is jogging a memory about not being able to use some feature because we supported 5.1, but I can't recall what that was..

Issue #971, opened 2023-03-29
[Bug]: Support for pnpm npm_translate_lock lock_version 6.0


flolu:
### What happened?

After running `pnpm i`, my `pnpm-lock.yaml` updated to `lockfileVersion: '6.0'`, which isn't supported by `rules_js` yet:

```
ERROR: Error computing the main repository mapping: no such package '@npm//': npm_translate_lock currently supports a maximum lock_version of 5.4, but found 6.0. Please file an issue on rules_js
```

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: bazel 6.0.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: WORKSPACE

flolu:
Also the syntax seems to have changed in the `pnpm-lock.yaml`:
from
```
lockfileVersion: 5.4
```
to
```
lockfileVersion: '6.0'
```

gregmagolan:
Looks like there are load bearing breaking changes to the lock file that rules_js will have to account for. Should have a PR up later today or tomorrow to support the both v5 and v6 of the pnpm lock file format.

gregmagolan:
#973 should do the trick. Planning to add some tests there before landing.

gregmagolan:
Support for v6 lockfile landed in https://github.com/aspect-build/rules_js/commit/002397780d51735f1a98b362fca7018df31e85ac

I'm going to do some testing downstream over the next week before cutting a release.

flolu:
@gregmagolan thank you so much!

PR #973, opened 2023-03-29
feat: support pnpm lockfile version 6.0 which is generated by pnpm 8

gregmagolan:
pnpm 8 was just released and comes with some breaking changes (from rules_js's perspective) in its lockfile format version 6.0. This PR adds support for the new lock file format.

Resolves https://github.com/aspect-build/rules_js/issues/971


kormide (approved):

jbedard (changes requested):
I think we need new tests? Can we create a test for each lockfile version we support?

Probably create a v5 one as the base, then upgrade to v6 and check that in as a new test and leave the v5 as is etc...

gregmagolan:
Yes, planning to add better test coverage before this lands.

gregmagolan:
Added unit test for v5 and v6 formats and updated a subset of e2e tests to v6. I'm happy with that coverage.

kormide (approved):

Issue #977, opened 2023-03-31
[Bug]: Windows host detection fails, when a custom non-windows platform is used


nosamad:
### What happened?

When using a custom target platform for a different os than windows, on a windows host, the linux launcher (shell script) is selected.
This leads to an error, while the bash script is not directly executable by CreateProcess.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:
aspect 5.2.1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
809ce7f1c52bc6eb55afcec94bcf97acc1f8cdb1

Language(s) and/or frameworks involved:
C/C++ and JS, TypeScript

### How to reproduce

```shell
Add a custom target platform for cross compile to a different OS than windows and try to build under windows.
```

### Any other information?

https://github.com/nosamad/rules_js/tree/topic/nosamad/fix_windows_platform_detection

gregmagolan:
I believe Bazel is doing the right thing here and building the js_binary for the target platform. What is your use case?

nosamad:
> I believe Bazel is doing the right thing here and building the js_binary for the target platform. For build actions that use a js_binary, those should still be built for the execution platform via the `cfg=exec` transition, which would be the host if you're not using RBE. What is your use case?

Already fixed by clean in main. Thanks

gregmagolan:
Can you provide more details as to how the js_binary is being executed? Is it a custom rule? Are you calling `bazel build` or `bazel run`? Code snippets may be helpful.

PR #978, opened 2023-03-31
fix: windows host platform detection

nosamad:
#977

### DIFF
--- a/js/defs.bzl
+++ b/js/defs.bzl
@@ -32,6 +32,10 @@ def js_binary(**kwargs):
             "@aspect_rules_js//js/private:enable_runfiles": True,
             "//conditions:default": False,
         }),
+        is_windows_host = select({
+            str(Label("@bazel_tools//src/conditions:host_windows")): True,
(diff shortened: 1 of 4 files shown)


gregmagolan:
Thanks for the PR. The js_library && js_binary locations where `is_windows` is present are no longer load bearing. https://github.com/aspect-build/rules_js/pull/986 cleans those up.

Using the windows host platform to choose where to use bash or cmd for the coverage_merger rule doesn't seem right since what is actually load bearing is the execution platform. What is currently there is not right either since it is using the target platform is checked which can be wrong if the execution platform is different than the target platform.

nosamad:
> Thanks for the PR. The js_library && a subset of the js_binary locations where `is_windows` is used are no longer load bearing. #986 cleans those up.
> 
> Using the windows host platform to choose where to use bash or cmd for the coverage_merger and js_binary rules doesn't seem right since what is actually load bearing is the execution platform which may be different from the host.

Thanks for the explanation. Works after cleanup.

PR #984, opened 2023-04-01
refactor: split node-patches to main (bazel 6 with unresolved symlinks) and legacy (bazel 5 or no unresolved symlinks)

gregmagolan:
# Description

Optimizations to node-patches are going to be made to the main patches since we can make changes that are possible with fixes in unresolved symlinks in Bazel 6 that would not work with Bazel 5 or with unresolved symlinks disabled. These will come in follow-ups.

# Type of change

- [x] Refactor (a code change that neither fixes a bug or adds a new feature)

# Test plan

How has this been tested?

- [x] Covered by existing tests cases
- [x] New test cases added for node-patches_legacy


jbedard (approved):
😄 👍

PR #985, opened 2023-04-01
perf: optimize fs patches when using Bazel 6 with unresolved symlinks

gregmagolan:
# Description

With Bazel 6 & unresolved symlinks enabled (default for Bazel 6) relative symlinks are maintained in the sandbox and in runfiles. This means we can remove the additional logic to check if following hops outside of the sandbox has a corresponding path within the sandbox. This should have a noticeable impact on the performance of some actions that are slowed down by the fs patches such as some jest tests.

# Type of change

- [x] Performance (a code change that improves performance)

# Test plan

How has this been tested?

- [x] Covered by existing tests cases


js/private/node-patches/src/fs.ts
```diff
@@ -634,23 +634,7 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
             }
             if (isEscape(loc, next)) {
                 // this hop takes us out of the guard
-                return nextHop(next, (next2: string | false) => {
```
jbedard, line 637: If I recall correct this was one of the major spots wasting time.

gregmagolan: Sounds about right

kormide (approved):

PR #990, opened 2023-04-03
docs: add note of js_binary to npm binaries section

jbedard:
This seems to be a point of confusion sometimes 🤷

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -195,17 +195,19 @@ typescript_bin.tsc(
 )
 ```
 
+Each bin exposes three rules, one for each Bazel command ("verb"): build, test and run - each aligning with the corresponding [js_run_binary](./js_run_binary.md), [js_test](#js_test) and [js_binary](./js_binary.md) rule APIs.
+
+For example:
+
+| Rule         | Underlying Rule | Invoked with  | To              |
+| ------------ | --------------- | ------------- | --------------- |
+| `foo`        | `js_run_binary` | `bazel build` | produce outputs |
+| `foo_binary` | `js_binary`     | `bazel run`   | side-effects    |
+| `foo_test`   | `js_test`       | `bazel test`  | assert exit `0` |
+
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -175,7 +175,7 @@ with a rule like [`copy_to_bin`](https://docs.aspect.build/aspect-build/bazel-li
 ### Using binaries published to npm
 
 rules_js automatically mirrors the `bin` field from the `package.json` file of your npm dependencies
-to a Starlark API you can load from in your BUILD file or macro.
+to a Starlark API you can load from in your BUILD file or macro. Each mirrored `bin` entry is a [js_binary](./js_binary.md) supporting the same attributes such as `data`, `node_options` etc.
```
gregmagolan, line 178: not quite right. the `<bin>` is a `js_run_binary`, `<bin>_binary` a `js_binary` and `<bin>_test` a `js_test`

gregmagolan: you could put those in the table below instead of up here?

docs/README.md
```diff
@@ -195,18 +195,20 @@ typescript_bin.tsc(
...
 
-> Note: this doesn't cause an eager fetch!
-> Bazel doesn't download the typescript package when loading this file, so you can safely write this
-> even in a BUILD.bazel file that includes unrelated rules.
+Each bin exposes three rules, one for each Bazel command ("verb"): build, test and run - each aligning with the corresponding [js_run_binary](./js_run_binary.md), [js_test](#js_test) and [js_binary](./js_binary.md) rule APIs.
 
-Each bin exposes three rules, one for each Bazel command ("verb"):
+For example:
 
 | Use          | With          | To              |
```
gregmagolan, line 202: Looking better. Thanks for updating. Can you add a column to the table as to which underlying rule is used?

gregmagolan (approved):

gregmagolan (approved):
🌮

PR #1005, opened 2023-04-17
feat: add selective opt-out for copying files to bin to js_binary & js_test with new no_copy_to_bin attribute

gregmagolan:
This case is particularly useful for passing files from external repositories to js_binary & js_test. For example,

```
js_test(
    name = "test",
    data = [
        test_files,
        "@chromium961656_toolchains//:resolved_toolchain",
    ],
    no_copy_to_bin = [
        "@chromium961656_toolchains//:resolved_toolchain",
    ],
    args = ["--chromium=\"$(rootpaths %s)\"" % "@chromium961656_toolchains//:resolved_toolchain"],
)
```

**Type of change**

- [x] New feature or functionality (change which adds functionality)
- [x] Documentation (updates to documentation or READMEs)

- [x] Relevant documentation has been updated

**Test plan**

- [x] New test cases added


kormide (approved):

gregmagolan:
> What the motivation for this? Just to avoid copying large files like a chromium distribution?

Not copying large files is a plus yes. Also, without this you'd get the following error because you can't automatically copy_to_bin a file from a different package and being in an external repository you can't easily modify the targets.

```
ERROR: .../BUILD.bazel:15:11: in _jest_test rule //foo:test: 
	File "/private/var/tmp/_bazel_greg/0095ff7128bae05979f6d697fa82e126/external/aspect_rules_jest/jest/private/jest_test.bzl", line 102, column 45, in _impl
		launcher = js_binary_lib.create_launcher(
	File "/private/var/tmp/_bazel_greg/0095ff7128bae05979f6d697fa82e126/external/aspect_rules_js/js/private/js_binary.bzl", line 448, column 53, in _create_launcher
		files.append(copy_file_to_bin_action(ctx, d))
	File "/private/var/tmp/_bazel_greg/0095ff7128bae05979f6d697fa82e126/external/aspect_bazel_lib/lib/private/copy_to_bin.bzl", line 42, column 13, in copy_file_to_bin_action
		fail(_file_in_external_repo_error_msg(file))
Error in fail: 
Cannot use copy_to_bin to copy BUILD.bazel from the external repository @chromium961656_Mac.
Files can only be copied from the source tree to their short path equivalent in the output tree.
```

jbedard:
> you can't automatically copy_to_bin a file from a different package

I assume detecting that isn't easy?

jbedard:
Does this basically introduce 2x as many possible fs layouts? Should we try hiding this and avoid putting it into the public API docs?

gregmagolan:
> > you can't automatically copy_to_bin a file from a different package
> 
> I assume detecting that isn't easy?

Yes, the failure is at analysis time. Easy to check the package name of an input file and whether it is a source file or an output tree (generated) file.

gregmagolan:
> Does this basically introduce 2x as many possible fs layouts? Should we try hiding this and avoid putting it into the public API docs?

I can mention that it is advanced option and you should know what you are doing if you use this to the docstring to discourage general use but I don't think we should hide advanced options.

PR #1007, opened 2023-04-18
perf: fs performance improvements

jbedard:


js/private/node-patches/src/fs.ts
```diff
@@ -760,16 +804,23 @@ export const patcher = (fs: any = _fs, roots: string[]) => {
 // =========================================================================
 
 export function isSubPath(parent: string, child: string): boolean {
-    return !path.relative(parent, child).startsWith('..')
+    return (
+        parent === child ||
+        (child[parent.length] === path.sep && child.startsWith(parent))
```
gregmagolan, line 809: hmmm. This one may be a problem on Windows vs Linux/Macos for the path.sep since the params are not normalized. Should probably check for either `/` or `\` to be safe.

gregmagolan: Unit test this little function?

jbedard: I think the params are always normalized, it is only ever invoked after other fs operations such as `path.resolve` or often in the callback of the operation being wrapped.

js/private/node-patches/src/fs.ts
```diff
@@ -782,26 +833,40 @@ export const escapeFunction = (_roots: string[]) => {
...
+    }
+
+    function fs_canEscape(
+        maybeLinkPath: string,
+        roots = defaultRoots
+    ): boolean {
+        // maybeLinkPath is the path which may be a symlink
+
+        if (!path.isAbsolute(maybeLinkPath)) {
+            maybeLinkPath = path.resolve(maybeLinkPath)
```
jbedard, line 853: @gregmagolan I think if we do this line 100% of the time then we don't have to worry about non-resolved/normalized paths being passed into the helpers, so the `path.sep` won't be an issue? For me that made my test go from a consistent 88s to 90s. But we might have to?

gregmagolan: Sounds like its all normalized. We just need to run CI on Windows to be sure.

js/private/node-patches/src/fs.ts
jbedard: FWIW I think the use of `startsWith` means the path must be normalized, so I think we should stick with `path.sep` and ensure the path is normalized.

Note that I think ensuring the path is normalized will always be faster the nthe original `path.relative`.

js/private/node-patches/src/fs.ts
jbedard: Moving the conversation to the other thread about the `/`s...

js/private/node-patches/src/fs.ts
jbedard: Ok I updated this a bit to ensure anything passed to `canEscape` is already normalized and absolute, and the second argument to `isEscape` is already normalized and absolute.  This way this `startsWith` + `path.sep` should work, and it avoids calling `path.resolve|normalize` except on the first argument to `isEscape` which always calls it now.

See https://github.com/aspect-build/rules_js/compare/8b7b14dbea95395c08ca2c9e2ebd84b93b3b5e73..3e4b9fb6137b25f79016ff17cfa93208d6884db6

gregmagolan (approved):
🚀

PR #1010, opened 2023-04-19
fix: fix a bug where patches from another repo failed to apply

### LINKED ISSUE #956: [Bug]: Error when trying to use a patch from a local repository
### What happened?

I'm trying to use the `patches` attribute on the `npm_translate_lock` rule with the `patch` coming from a local repository and hitting this error:
```
/Users/amyzhao/bazel_patches_repro/WORKSPACE:44:17: //external:npm__at_types_google-apps-script__1.0.55: invalid label '@@nested_workspace//:patches/@types__google-apps-script@1.0.55.patch' in element 0 of attribute 'patches' in '_npm_import' rule: invalid repository name '@@nested_workspace': workspace names may contain only A-Z, a-z, 0-9, '-', '_' and '.'
```

I think this [line](https://github.com/aspect-build/rules_js/blob/f940df4a10fae79c6133c1bca08840691ab6d55b/npm/private/npm_translate_lock_generate.bzl)  in npm_translate_lock_general.bzl is what's prepending the extra `@`.

### Version

Ubuntu 18.04

Output of `bazel --version`:
5.4.0 (note: we don't hit this error on bazel 6.0 but are having trouble upgrading b/c of mysterious cache misses) 

- aspect_bazel_lib - 1.27.2
- rule_nodejs-core-5.7.2
- rules_js-v1.21.0
- rules_ts-v1.3.2

- pnpm

### How to reproduce

```shell
I set up a minimal repro in https://github.com/amybzhao/bazel_patches_repro

Commands that we expect to succeed:
- bzl build //:node_modules
- bzl build @nested_workspace//example
- cd nested_workspace && bzl build //example
```

### Any other information?

Greg suggested copying the external patches to the main workspace. This does work but it seems like we need to copy over the patches, the .npmrc, the package.json, and the pnpm-lock.yaml which is a little unfortunate.

This change to using a package.json from the main workspace also seems to be creating runtime issues for `js_binary` targets in the main workspace that depend on node modules from the local repository. They seem to be using the node_modules that live at `bazel-out/host/bin/external/tools/postcss.sh.runfiles/__main__/node_modules/.aspect_rules_js/` rather than `bazel-out/host/bin/external/tools/postcss.sh.runfiles/__main__/external/nested_workspace/node_modules/.aspect_rules_js/`

kormide:
For older versions of Bazel where labels referencing the root repository could optionally begin with `@`, we were incorrectly prepending a `@` when the label belonged to a different repository.

Fixes #956 

**Type of change**

- [x] Bug fix (change which fixes an issue)
- [ ] New feature or functionality (change which adds functionality)
- [ ] Style (white-space, formatting, etc...)
- [ ] Refactor (a code change that neither fixes a bug or adds a new feature)
- [ ] Performance (a code change that improves performance)
- [ ] Documentation (updates to documentation or READMEs)
- [ ] Chore (any other change that doesn't affect source or test files)

- [ ] Breaking change (this change will force users to change their own code or config)
- [ ] Relevant documentation has been updated
- [ ] Suggested release notes are provided below:

**Test plan**

How has this been tested?

- [ ] Covered by existing test cases
- [x] New test cases added
- [ ] Manual testing


e2e/patch_from_repo/MODULE.bazel
```diff
@@ -0,0 +1,37 @@
...
+    patch_args = {
+        "meaning-of-life": ["-p1"],
+    },
+    patches = {
+        # TODO: Because `patches` is a string_list_dict that doesn't take labels, we can't
+        # pass in a label like `@local_repo//:....` because only the root repository knows
+        # how to map this repo to its local override. This causes rules_js to fail when trying
+        # to convert the label into a path.
+        "meaning-of-life@1.0.0": ["@@local_repo~override//:meaning-of-life@1.0.0.patch"],
+    },
```
kormide, line 32: @gregmagolan This was a fun one... Because `patches` doesn't take labels, we lose the repo mapping to the local override and have to use an absolute label like this.

js/private/test/image/structure/digests.sum
```diff
@@ -1,2 +1,2 @@
-099acc2677cf8fbcd5c49d79169a21a8f541234d312e8e71b260af484093e061
+db4b47b2410911b723bed458f4be3ac6eae5859c232e924f7401856224233196
```
gregmagolan, line 1: ??? how'd this get updated?

e2e/patch_from_repo/MODULE.bazel
gregmagolan: 🏋️

gregmagolan (approved):

js/private/test/image/structure/digests.sum
kormide: I'm not sure. Still trying to track this down..

kormide: My best guess is that somehow rules_js files are being included in the shasum, perhaps defined by the files and dirs here? https://github.com/aspect-build/rules_js/blob/main/js/private/test/image/structure/app_structure.mf

kormide: @thesayyn Do you know what's causing this?

kormide: Oh actually I'm getting the same error on main, so I think it's unrelated.

kormide: Removed the changes.

kormide: Fixed in https://github.com/aspect-build/rules_js/pull/1012.

PR #1022, opened 2023-04-25
docs: add note documenting js_binary(args)

### LINKED ISSUE #1019: [Bug]: Seems that the 'args' attribute of js_binary and js_test is not documented
Didn't realize I could provide `args` to `js_binary` and `js_test` until I tried it, but seems undocumented :)

jbedard:
Fix #1019

All binary+test targets have an `args` but I don't see a way to document it. I think this is a stardoc issue and an issue should be created?

### DIFF
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -553,8 +553,7 @@ js_binary = rule(
 js_test = rule(
     doc = """Identical to js_binary, but usable under `bazel test`.
 
-All [common binary attributes](https://bazel.build/reference/be/common-definitions#common-attributes-binaries)
-and [common test attributes](https://bazel.build/reference/be/common-definitions#common-attributes-tests) are
+All [common test attributes](https://bazel.build/reference/be/common-definitions#common-attributes-tests) are
 supported including `args` as the list of arguments passed node.js.
 
 Bazel will set environment variables when a test target is run under `bazel test` and `bazel run`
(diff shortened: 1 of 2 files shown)


js/private/js_binary.bzl
```diff
@@ -32,6 +32,8 @@ based on the requested target platform. Use the
 [`--toolchain_resolution_debug`](https://docs.bazel.build/versions/main/command-line-reference.html#flag--toolchain_resolution_debug)
 Bazel option to see more detail about the selection.
 
+The `js_binary` `args` attribute is a list of strings that will be passed as arguments to node.js.
```
gregmagolan, line 35: Should probably list all of the "common" attributes that aren't documented and link to
https://bazel.build/reference/be/common-definitions#common-attributes-binaries
https://bazel.build/reference/be/common-definitions#common-attributes-tests

gregmagolan:
Yes. Stardoc doesn't add the common built-in attributes to the generated docs. In other rules, we can work-around this since we document the args in a wrapper macro docstring but js_binary doesn't have one of those.

js/private/js_binary.bzl
```diff
@@ -550,6 +553,10 @@ js_binary = rule(
 js_test = rule(
     doc = """Identical to js_binary, but usable under `bazel test`.
 
+All [common binary attributes](https://bazel.build/reference/be/common-definitions#common-attributes-binaries)
```
gregmagolan, line 556: Lets just link to common test attributes for js_test.

gregmagolan (approved):
🌮

PR #1026, opened 2023-04-26
fix: cleanup flaky devserver tests

jbedard:
I'm not sure if or what could have change but these devserver tests are failing frequently on CI atm, often >50% of the time, and we run them 4x (bzlmod x bazel 5/6).

### DIFF
--- a/e2e/js_run_devserver/multirun_test.sh
+++ b/e2e/js_run_devserver/multirun_test.sh
@@ -14,13 +14,13 @@ _sedi () {
   sed "${sedi[@]}" "$@"
 }
 
-echo "TEST - $0: $1"
+echo "$$: TEST - $0: $1"
 
 ./node_modules/.bin/ibazel run "$1" "$BZLMOD_FLAG" >/dev/null 2>&1 &
 ibazel_pid="$!"
 
 function _exit {
-  echo "Cleanup..."
+  echo "$$: Cleanup..."
   kill "$ibazel_pid"
   wait "$ibazel_pid"
   git checkout src/index.html src/BUILD.bazel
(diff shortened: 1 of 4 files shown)


e2e/js_run_devserver/multirun_test.sh
```diff
@@ -14,107 +14,119 @@ _sedi () {
   sed "${sedi[@]}" "$@"
 }
 
-echo "TEST - $0: $1"
+echo "$$: TEST - $0: $1"
```
jbedard, line 17: I found putting the pid as a prefix helpful to distinguish between the stdout of the shell script, bazel and the http server.

e2e/js_run_devserver/multirun_test.sh
```diff
@@ -14,107 +14,119 @@ _sedi () {
...
 
-./node_modules/.bin/ibazel run "$1" "$BZLMOD_FLAG" >/dev/null 2>&1 &
+./node_modules/.bin/ibazel run "$1" "$BZLMOD_FLAG" 2>&1 &
```
jbedard, line 19: This spam seemed worth it while debugging 🤷

e2e/js_run_devserver/multirun_test.sh
```diff
@@ -14,107 +14,119 @@ _sedi () {
...
 ibazel_pid="$!"
 
 function _exit {
-  echo "Cleanup..."
+  echo "$$: Cleanup..."
   kill "$ibazel_pid"
+  wait "$ibazel_pid"
```
jbedard, line 25: One error was port 8080 not being available, I assume because the previous process was not dead yet and still using the port. This should wait for it to fully die if I'm understanding correctly.

e2e/js_run_devserver/multirun_test.sh
```diff
@@ -14,107 +14,119 @@ _sedi () {
...
   git checkout src/index.html src/BUILD.bazel
   rm -f src/new.html
 }
 trap _exit EXIT
 
-echo "Waiting for $1 devserver to launch on 8080..."
+echo "$$: Waiting for $1 devserver to launch on 8080..."
 
+n=0
 while ! nc -z localhost 8080; do
```
jbedard, line 34: This was possibly an infinite loop before, and I did see CI go for 3+h stuck in it a few times.

e2e/js_run_devserver/multirun_test.sh
```diff
@@ -14,107 +14,119 @@ _sedi () {
...
-echo "Waiting 10 seconds for ibazel rebuild after change to src/index.html..."
-sleep 10
+echo "$$: <div>A second line</div>" >> src/index.html
 
-if ! curl http://localhost:8080/index.html --fail 2>/dev/null | grep "A second line"; then
-  echo "ERROR: Expected http://localhost:8080/index.html to contain 'A second line'"
-  exit 1
-fi
+n=0
+while ! curl http://localhost:8080/index.html --fail 2>/dev/null | grep "A second line"; do
```
jbedard, line 100: The ugly loop avoids waiting 10s every single time. With the amount of times this script (and the other) are run it seemed worth copy+pasting the loop a few times 🤷

kormide (approved):
Nice, thanks for fixing this. These failures have been bugging me as well.

gregmagolan (approved):
🏋️

Issue #910, opened 2023-02-27
[FR]: Read pnpm.patches from lockfile when update_pnpm_lock = False


Toxaris:
### What is the current behavior?

When `update_pnpm_lock = True`, `npm_translate_lock` reads the direct dependencies as well as the pnpm patches from the `package.json` files in the repository.

When `update_pnpm_lock = False`, `npm_translate_lock` reads the direct dependencies from the lockfile, and prior to #832, there was no need to even have `package.json` files in the repository. But since #832, the pnpm patches are taken from the `package.json` file in the same folder as the lockfile and the build fails if there is no such `package.json` file.

### Describe the feature

When `update_pnpm_lock = False`, `npm_translate_lock` should not read any `package.json` files in the repository for the following reasons:

1. changes to the `package.json` files should only influence the build after an explicit call to `pnpm install` to regenerate the lockfile. it is inconsistent to apply the pnpm patches but ignore the direct dependencies in the same file.
2. the build should not fail if `package.json` files are missing (and the lockfile was created in some other way).

To achieve this, `npm_translate_lock` should read the pnpm patches from the lockfile if `update_pnpm_lock = False`.

### Fund our work

- [ ] Sponsor our open source work by donating a [feature bounty](https://opencollective.com/aspect-build/)

Toxaris:
As [discussed on Slack](https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1677445364979689), this is not blocking me, I just created an empty package.json to make the build succeed. I do believe this would be more consistent though, so I wanted to create this feature request to conclude the Slack discussion.

gregmagolan:
As part of this, the docs for `data` and `update_pnpm_lock` should be updated

PR #1029, opened 2023-04-27
fix: read list of pnpm patches from lock file when update_pnpm_lock is False

gregmagolan:
When `update_pnpm_lock` is `False` then we can read the list of packages from the lock file since that won't break the update pnpm lock workflow when it is not enabled. This means that there is no phantom read of an undeclared package.json file for users in this case.

Fixes https://github.com/aspect-build/rules_js/issues/910.

Also related to Bazel Slack conversation: https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1682614797831909

### Type of change

### Test plan

### DIFF
--- a/npm/private/npm_translate_lock_state.bzl
+++ b/npm/private/npm_translate_lock_state.bzl
@@ -37,8 +37,12 @@ WARNING: `update_pnpm_lock` attribute in `npm_translate_lock(name = "{rctx_name}
     _init_patches_labels(priv, rctx, label_store)
 
     root_package_json_declared = label_store.has("package_json_root")
-    if root_package_json_declared:
+    if root_package_json_declared and _should_update_pnpm_lock(priv):
+        # If we need to read the patches list from the package.json since
+        # update_pnpm_lock is enabled and the user has declared the root package.json
+        # as an input file then we can read it right away.
         _init_patched_dependencies_labels(priv, rctx, label_store)
+        priv["patched_dependencies_labels_initialized"] = True
 
     if _should_update_pnpm_lock(priv) or not rctx.attr.pnpm_lock:
         # labels only needed when updating or bootstrapping the pnpm lock file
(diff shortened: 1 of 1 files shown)


kormide (approved):
🆒

npm/private/npm_translate_lock_state.bzl
```diff
@@ -56,11 +59,11 @@ WARNING: `update_pnpm_lock` attribute in `npm_translate_lock(name = "{rctx_name}
     if pnpm_lock_exists:
         _load_lockfile(priv, rctx, label_store)
 
-    if not root_package_json_declared:
-        # If a root package.json label wasn't declared then we didn't load pnpm.patchedDependencies
-        # above. Try again now that the lockfile has been read so that we can derive the root
-        # package.json from the `importers` field.
-        _init_patched_dependencies_labels(priv, rctx, label_store)
+    # If we didn't initialize the patched dependency labels above then try again now. We'll either
```
jbedard, line 62: "if we didn't ... then try again" - don't we try again every time now?

jbedard (approved):

npm/private/npm_translate_lock_state.bzl
gregmagolan: I simplified the logic so the function itself marks itself as called already and exits early

jbedard: Should the comment change then?

gregmagolan: Sure. I'll re-phrase so it is more obvious

gregmagolan: Actually, I'll move the "already intialized" state to this function so it reads clearly

jbedard: > Actually, I'll move the "already intialized" state to this function so it reads clearly

sgtm 👍

PR #1041, opened 2023-05-03
feat: align js_binary+library copy_to_bin APIs

jbedard:


js/private/js_library.bzl
```diff
@@ -219,6 +245,12 @@ def _js_library_impl(ctx):
         sources = transitive_sources,
         data = ctx.attr.data,
         deps = ctx.attr.srcs + ctx.attr.declarations + ctx.attr.deps,
+        files = ctx.files.data,
```
jbedard, line 248: These files being copied is what fixes the jest use case.

js/private/js_helpers.bzl
```diff
@@ -159,6 +160,16 @@ def gather_runfiles(ctx, sources, data, deps):
 
         deps: list of dependency targets; only transitive runfiles are gather from these targets
 
+        files: a list of files which should be included in runfiles
+
+        copy_files_to_bin: whether to copy files to the bin directory
```
jbedard, line 165: Note that this is "files" instead of "data" because it is copying the `gather_runfiles` `files` attribute  (which is normally `ctx.files.data`, but possibly more such as `entry_point`). Where the `gather_runfiles` `data` attribute is `ctx.attr.data` and used to collect providers instead of files.

Any suggestions for better naming so we can potentially keep the `copy_data_to_bin` naming consistent? Maybe `copy_data_files_to_bin` and rename `files` to `data_files`?

js/private/js_binary.bzl
```diff
@@ -479,44 +478,31 @@ def _create_launcher(ctx, log_prefix_rule_set, log_prefix_rule, fixed_args = [],
...
+        sources = [],
+        data = ctx.attr.data,
+        files = [entry_point] + ctx.files.data,
+        deps = [],
+        copy_files_to_bin = ctx.attr.copy_data_to_bin,
+        no_copy_to_bin = ctx.files.no_copy_to_bin,
+        include_transitive_sources = ctx.attr.include_transitive_sources,
+        include_declarations = ctx.attr.include_declarations,
+        include_npm_linked_packages = ctx.attr.include_npm_linked_packages,
+    )
```
gregmagolan, line 501: nit: you could do
```
runfiles = gather_runfiles(...).merge(ctx.runfiles(files = launcher_files))
```
instead so that
```
runfiles = runfiles,
```
can stay clean below

js/private/js_library.bzl
gregmagolan: Hmm. Lets call this `data_files` to be very clear that you have to pass both `ctx.attr.data` and `ctx.files.data`

js/private/js_library.bzl
```diff
@@ -88,6 +88,32 @@ runtime dependency on this target.
...
+        This is useful for exceptional cases where a `copy_to_bin` is not possible or not suitable for an input
+        file such as a file in an external repository. In most cases, this option is not needed.
+        See `copy_data_to_bin` docstring for more info.
+        """,
+    ),
+    "copy_data_to_bin": attr.bool(
+        doc = """When True, `data` files and the `entry_point` file are copied to the Bazel output tree before being passed
+        as inputs to runfiles.
+
+        Defaults to True so that a `js_library` with the default value is compatible with `js_run_binary` with
```
gregmagolan, line 104: 2nd and 3rd paragraphs here are js_binary specific and can be dropped. Just the 1st paragraph will suffice

gregmagolan (approved):

PR #1042, opened 2023-05-03
fix: include data sources in js_library transitive sources

jbedard:
Fixes https://github.com/aspect-build/rules_jest/issues/128


js/private/js_library.bzl
```diff
@@ -183,7 +183,7 @@ def _js_library_impl(ctx):
     sources, declarations = _gather_sources_and_declarations(
         ctx = ctx,
         targets = ctx.attr.srcs,
-        files = ctx.files.srcs,
+        files = ctx.files.srcs + ctx.files.data,
```
jbedard, line 186: This is the fix. The other 2 lines are to align with this but I'm not 100% sure of those, maybe those 2 should be dropped for now?

jbedard: Actually this makes the resulting jest config have multiple source files so the `ctx.file.config` in `jest_test` fails. Instead I think we have to treat "data sources" different "srcs", see the diff [changing this](https://github.com/aspect-build/rules_js/compare/b0ea90c9aed691cee0b138dc046403e3a31ea9af..c506f36cb52c3e0f0fabf4664aa7fb0fd81312f2)

jbedard: See the `extract-*` tests.

jbedard:
closing in favour of https://github.com/aspect-build/rules_js/pull/1041

PR #1043, opened 2023-05-03
fix: when symlinking node_modules symlinks in js_run_devserer, prefer the bazel-out copy instead of the runfiles copy if it exists

gregmagolan:
This resolves an issue in a complex Next.js devserver run downstream.

This fix is in the same spirit as running `js_binary` and `js_run_binary` targets from the output tree instead of the runfiles to prevent node being able to find libraries such as react in multiple node_modules trees. This applies the same principle to `js_run_devserver` by symlinks node_modules packages from the devserver custom sandbox to the output tree instead of the runfiles.

---

### Type of change

### Test plan

- Manual testing:

Ran the devserver in `e2e/webpack_devserver` manually.

Tested downstream with existing `js_run_devserver` targets that use Next.js and react-CRA in bazel-examples, https://github.com/aspect-build/bazel-examples/pull/212 as well as a very complex Next.js target at a client where this fix was necessary to prevent React from being found in two locations at runtime.

### DIFF
--- a/js/private/js_run_devserver.mjs
+++ b/js/private/js_run_devserver.mjs
@@ -48,10 +48,26 @@ async function syncRecursive(src, dst, writePerm) {
         const exists = synced.has(src) || fs.existsSync(dst)
         synced.set(src, lstat.mtimeMs)
         if (lstat.isSymbolicLink()) {
+            const srcWorkspacePath = src.slice(RUNFILES_ROOT.length + 1)
             if (process.env.JS_BINARY__LOG_DEBUG) {
-                console.error(
-                    `Syncing symlink ${src.slice(RUNFILES_ROOT.length + 1)}`
+                console.error(`Syncing symlink ${srcWorkspacePath}`)
+            }
+            if (path.basename(path.dirname(src)) == 'node_modules') {
+                // Special case for node_modules symlinks where we should _not_ symlink to the runfiles but rather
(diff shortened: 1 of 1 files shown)


js/private/js_run_devserver.mjs
```diff
@@ -48,10 +48,26 @@ async function syncRecursive(src, dst, writePerm) {
...
+            }
+            if (path.basename(path.dirname(src)) == 'node_modules') {
+                // Special case for node_modules symlinks where we should _not_ symlink to the runfiles but rather
+                // the bin copy of the symlink to avoid finding npm packages in multiple node_modules trees
+                const maybeBinSrc = path.join(
+                    process.env.JS_BINARY__EXECROOT,
+                    process.env.JS_BINARY__BINDIR,
+                    srcWorkspacePath
                 )
+                if (fs.existsSync(maybeBinSrc)) {
```
jbedard, line 63: Most of the fs ops in this file are the `async` versions  which I think would be np here as well?

gregmagolan: Would be a very minor optimization. Could do in a follow-up if we feel its worth it.

gregmagolan: Switching to use fs.access() to check existence here since that is what is recommended in the Node docs as exists is deprecated. Tho I'll keep existsSync for now for mkdirSync that is intentionally non-yielding and atomic.

jbedard (approved):

PR #1052, opened 2023-05-08
Revert "chore(deps): update dependency aspect_bazel_lib to v1.31.0"

alexeagle:
Reverts aspect-build/rules_js#1039

That version of bazel-lib had a breaking change in yq and was yanked.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,7 +6,7 @@ module(
     compatibility_level = 1,
 )
 
-bazel_dep(name = "aspect_bazel_lib", version = "1.31.0")
+bazel_dep(name = "aspect_bazel_lib", version = "1.29.2")
 bazel_dep(name = "bazel_skylib", version = "1.4.1")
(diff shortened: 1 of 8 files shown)


MODULE.bazel
```diff
@@ -6,7 +6,7 @@ module(
     compatibility_level = 1,
 )
 
-bazel_dep(name = "aspect_bazel_lib", version = "1.31.0")
+bazel_dep(name = "aspect_bazel_lib", version = "1.29.2")
```
kormide, line 9: Can this be updated to `1.30.2`?

kormide: Oh this is a revert PR. Nevermind.

kormide (approved):

MODULE.bazel
alexeagle: it shouldn't be. We need to take a different policy with updating our MODULE.bazel deps - we ought to be giving the lower bound here so we don't churn users. I'll post something on slack.

kormide (approved):

PR #1054, opened 2023-05-09
fix: improve error message for invalid js_library(data) files

jbedard:
### Type of change

### Test plan

- New test cases added


js/private/js_helpers.bzl
```diff
@@ -138,6 +138,38 @@ def gather_npm_package_store_deps(targets):
 
     return depset([], transitive = npm_package_store_deps)
 
+def copy_js_file_to_bin_action(ctx, file):
+    if ctx.label.workspace_name != file.owner.workspace_name or ctx.label.package != file.owner.package:
+        msg = """
+
+Expected to find source file {file_basename} in {this_package}, but instead it is in {file_package}.
+
+All source files in rules_js rules must be in the same package as the target.
```
gregmagolan, line 147: This sentence may be more confusing than helpful for users since "source file" is ambiguous and overloaded and means something else outside of Bazel.

@alexeagle WDYT the best error message is here? You're usually good with these.

The other thing to mention in this error message is the option to add the file to `no_copy_to_bin` to prevent copying to bin... and perhaps link to our docsite for the context: https://docs.aspect.build/rules/aspect_rules_js/docs/#javascript

gregmagolan: Guess you need to update the title of this PR as well :)

gregmagolan: "All source and data files that are not in the Bazel output tree must be in the same package as the target so that they can be copied to the output tree in an action" ?

jbedard: The docs link is in the next sentence.

I've updated this sentence and added the one at the end mentioning `no_copy_to_bin`.

js/private/js_helpers.bzl
```diff
@@ -138,6 +138,43 @@ def gather_npm_package_store_deps(targets):
 
     return depset([], transitive = npm_package_store_deps)
 
+def copy_js_file_to_bin_action(ctx, file):
+    if ctx.label.workspace_name != file.owner.workspace_name or ctx.label.package != file.owner.package:
+        msg = """
+
+Expected to find source file {file_basename} in {this_package}, but instead it is in {file_package}.
```
gregmagolan, line 145: Hmm. Should probably quote the packages since the root package is the empty string `""`

js/private/js_helpers.bzl
```diff
@@ -138,6 +138,43 @@ def gather_npm_package_store_deps(targets):
...
+
+All source and data files that are not in the Bazel output tree must be in the same package as the
+target so that they can be copied to the output tree in an action.
+
+See https://github.com/aspect-build/rules_js/tree/dbb5af0d2a9a2bb50e4cf4a96dbc582b27567155/docs#javascript
```
gregmagolan, line 150: nit: link to docsite instead of github: https://docs.aspect.build/rules/aspect_rules_js/docs/#javascript

js/private/js_helpers.bzl
gregmagolan: A few spots below as well

js/private/js_helpers.bzl
```diff
@@ -138,6 +138,43 @@ def gather_npm_package_store_deps(targets):
...
+
+buildozer 'new_load @aspect_rules_js//js:defs.bzl js_library' {file_package}:__pkg__
+buildozer 'new js_library {new_target_name}' {file_package}:__pkg__
+buildozer 'add srcs {file_basename}' {file_package}:{new_target_name}
+buildozer 'add visibility {this_package}:__pkg__' {file_package}:{new_target_name}
+buildozer 'remove srcs {file_package}:{file_basename}' {this_target}
+buildozer 'add srcs {file_package}:{new_target_name}' {this_target}
+
+For exceptional cases where copying is not possible or not suitable for an input file such as
+a file in an external repository exceptions can be added to 'no_copy_to_bin'. In most cases,
```
gregmagolan, line 164: nit: missing comma:
```
a file in an external repository, exceptions can be added to 'no_copy_to_bin'
```

PR #1074, opened 2023-05-18
feat: introduce .publish target for npm_package

### LINKED ISSUE #155: Demonstrate how to publish to npm

thesayyn:
### Type of change

- New feature or functionality (change which adds functionality) fixes #155 

### Test plan

- New test cases added


MODULE.bazel
```diff
@@ -6,7 +6,7 @@ module(
     compatibility_level = 1,
 )
 
-bazel_dep(name = "aspect_bazel_lib", version = "1.31.2")
+bazel_dep(name = "aspect_bazel_lib", version = "1.32.0")
```
thesayyn, line 9: this wasn't synced with workspace one somehow 🤷‍♂️

gregmagolan: I believe this is intentional since the MODULE.bazel can be minimal version (since bzlmod will resolve the latest from all requested versions from all dependencies) but the WORKSPACE should be latest because of how confusing transitives are in WORKSPACES

gregmagolan: now that you're using expand_template from bazel-lib would be the right time to bump the minimum tho :)

thesayyn: right. I was about to say that. :P

npm/private/npm_package.sh
```diff
@@ -0,0 +1,18 @@
...
+source "$(grep -sm1 "^$f " "$0.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+{ echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
+# --- end runfiles.bash initialization v3 ---
+
+readonly PNPM_PATH="$(rlocation pnpm/pnpm.sh)"
+readonly PACKAGE_DIR="$(rlocation {{PACKAGE_DIR}})"
+
+
+"${PNPM_PATH}" publish $PACKAGE_DIR --no-git-checks $@
```
gregmagolan, line 18: how does `pnpm publish` compare to `npm publish`?

thesayyn: AFAIK, it performs some prechecks and delegates the work to `npm publish`.

thesayyn: users transitively depend on npm.

gregmagolan: hmmm. how does it get npm? might not be hermetic here

thesayyn: it most probably relies on PATH.

gregmagolan: you can get npm itself from the node toolchain. check what js_binary does what `include_npm` is set to True. https://docs.aspect.build/rules/aspect_rules_js/docs/js_binary#include_npm

gregmagolan: if `pnpm publish` doesn't add much value over `npm publish` then seems simpler to just call the latter here

gregmagolan: one thing I know `pnpm publish` will do is resolve `workspace:*` deps in the package.json of the package being published but I think that complexity can be dealt with in a future PR

thesayyn: in npm_package case looks like it doesn't add much value.

thesayyn: it's just a fancy wrapper that knows how to do pnpm-workspace multi-package releases etc.

thesayyn: done. now uses js_binary instead.

npm/private/npm_package.bzl
```diff
@@ -172,6 +175,13 @@ def npm_package(
 
     This target can be used as the `src` attribute to `npm_link_package`.
 
+    Every npm_package target has a sub target named after its name, which is `<name>.publish`, that can be run
+    to publish to an npm registry.
+
+    You can pass arguments to npm by escaping them from Bazel using a double-hyphen, for example:
```
gregmagolan, line 181: "Under the hood, this target run `npm publish. You can pass arguments to npm by escaping..."

npm/private/npm_package.bzl
```diff
@@ -459,6 +480,35 @@ def npm_package(
         **kwargs
     )
 
+def _publish_target(name, package, args = [], visibility = None, tags = [], testonly = False):
+    # Always tag the target manual since we should only build it when the final target is built.
+    tags = tags + ["manual"]
+
+    expand_template(
+        name = "{}_mjs".format(name),
+        template = "@aspect_rules_js//npm/private:npm_package.mjs",
```
gregmagolan, line 489: nit: name this script `npm_publish.mjs` for clarity

npm/private/npm_package.bzl
```diff
@@ -459,6 +480,35 @@ def npm_package(
...
+        out = "{}.mjs".format(name),
+        substitutions = {
+            "{{PACKAGE_DIR}}": "$(rlocationpaths :{})".format(package),
```
gregmagolan, line 492: Do you know what version of bazel `rlocationpaths` was added it. Did it exist in Bazel 5?

npm/private/test/npm_package_publish/test.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+readonly PUBLISH_A="$1"
```
gregmagolan, line 3: can you comment on what the two test cases are asserting?

npm/private/npm_package.mjs
```diff
@@ -0,0 +1,14 @@
+import { spawnSync } from 'node:child_process'
+import fs from 'node:fs'
+
+const PACKAGE_DIR = `${process.env.JS_BINARY__RUNFILES}/{{PACKAGE_DIR}}`
+const restArgs = process.argv.slice(2)
+
+const spawn = spawnSync('npm', ['publish', PACKAGE_DIR, ...restArgs], {
```
gregmagolan, line 7: so simple and clean 👍

npm/private/npm_package.mjs
```diff
@@ -0,0 +1,14 @@
+import { spawnSync } from 'node:child_process'
+import fs from 'node:fs'
+
+const PACKAGE_DIR = `${process.env.JS_BINARY__RUNFILES}/{{PACKAGE_DIR}}`
```
gregmagolan, line 4: you could just pass PACKAGE_DIR as the first arg instead and avoid expand_template all together.. tho I suppose in that case the js_binary has to be run with `bazel run` and can't be run with the .sh script directly.

there is Bazel Slack thread going atm to add `fixed_args` to the js_binary public API so if that lands then we can use that here to bake the package dir arg into the js_binary

gregmagolan: https://github.com/aspect-build/rules_js/pull/1077

npm/private/npm_package.bzl
gregmagolan: Thanks for checking. Technically this would be a breaking change for anymore not yet updated to 5.4.0 and using npm_package so I think we should implement without relying `rlocations`. We could leave a TODO for rules_js 2.0 and switch to rlocations at that point.

gregmagolan:
@thesayyn Can you also add an opt-out attribute so that users can disable the `.publish` target for cases where they don't want it to show up in the graph at all?

thesayyn:
> @thesayyn Can you also add an opt-out attribute so that users can disable the `.publish` target for cases where they don't want it to show up in the graph at all?

Does `no_publish` seem reasonable? or `publish_binary = True | False`?

gregmagolan:
> 

Chatted with Alex and he thinks it's not needed. We can add it in the future if someone asks.

npm/private/npm_package.mjs
thesayyn: Prerequisite of this pr; https://github.com/aspect-build/rules_js/pull/1079

npm/private/npm_package.bzl
thesayyn: rootpath is also okay.

PR #1077, opened 2023-05-19
feat: add fixed_args to js_binary & js_test

gregmagolan:
TODO: add test case before landing

Fixed command line arguments to pass to the Node.js when this binary target is executed.

Unlike the built-in `args`, which are only passed to the target when it is
executed either by the `bazel run` command or as a test, `fixed_args` are baked
into the generated launcher script so are always passed even when the binary
target is run outside of Bazel directly from the launcher script.

`fixed_args` are passed before the ones specified in `args` and before ones
that are specified on the `bazel run` or `bazel test` command line.

See https://bazel.build/reference/be/common-definitions#common-attributes-binaries
for more info on the built-in `args` attribute.

---

### Type of change

### Test plan

- New test cases added


thesayyn (approved):

thesayyn (approved):

jbedard:
Are there any restrictions that should be noted in the docs? Can all the regular things like `$(location)` still be used?

gregmagolan:
> Are there any restrictions that should be noted in the docs? Can all the regular things like `$(location)` still be used?

No restrictions. The fixed args are passed through `expand_variables` so the makevars should all work. Good call tho to mention that in the docstring.

```
fixed_args_expanded = [expand_variables(ctx, fixed_arg, attribute_name = "fixed_args") for fixed_arg in fixed_args]
```

PR #1103, opened 2023-06-06
fix: non-experimental allow_unresolved_symlinks

### LINKED ISSUE #1102: rules_js is failing with error "while parsing configuration settings: unknown option: 'experimental_allow_unresolved_symlinks' " with Bazel@HEAD in CI
rules_js is breaking with Bazel@HEAD and some projects depending on it.

https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3068#018879de-8112-4f1d-a1bd-c19508b89002

[https://github.com/aspect-build/rules_js/blob/2f64180ba84693d0bfb975b44b520e95ecb80adb/js/private/BUILD.bazel#L2](https://github.com/aspect-build/rules_js/blob/2f64180ba84693d0bfb975b44b520e95ecb80adb/js/private/BUILD.bazel#L29)

Error logs: 
```
ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/6cd8061c7967fd286ed2de8f0692b4eb/external/aspect_rules_js/js/private/BUILD.bazel:26:15: in values attribute of config_setting rule @aspect_rules_js//js/private:experimental_allow_unresolved_symlinks: error while parsing configuration settings: unknown option: 'experimental_allow_unresolved_symlinks'

```

Reference  : https://github.com/bazelbuild/bazel/commit/215f6982d3caec22e67d2fe345a61d63351c7937#commitcomment-116145231

CC @alexeagle @mai93

jbedard:
Fix #1102

I don't think we can simply use multiple `config_setting` values because the config value must exist when that `config_setting` is accessed. @thesayyn and I were looking for other ways of doing it but I think if we are using `config_setting` we need to determine the config value name ahead of time?

Any other ideas?

### Type of change

### Test plan


js/private/BUILD.bazel
```diff
@@ -24,9 +24,9 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
-config_setting(
+alias(
     name = "experimental_allow_unresolved_symlinks",
-    values = {"experimental_allow_unresolved_symlinks": "true"},
+    actual = "@aspect_config_settings//:allow_unresolved_symlinks",
```
alexeagle, line 29: Actually it sounds like this belongs in https://github.com/bazel-contrib/bazel_features - the whole point there is to make it possible to write rules that detect bazel features are present/not present. Then we'd only have to do these tricks in one place and have other Bazel experts like @fmeum who can help review.

thesayyn: problem with bazel_features is that it requires two macro calls in the WORKSPACE, which we don't have in rules_js. so we can only call repository rules that belong to rules_js.

thesayyn: one http_archive and then features_deps call to load deps. but in rules_js we have register_node_toolchains and rules_js_dependencies.

js/repositories.bzl
```diff
@@ -25,3 +26,7 @@ def rules_js_dependencies():
         strip_prefix = "bazel-lib-1.32.1",
         url = "https://github.com/aspect-build/bazel-lib/releases/download/v1.32.1/bazel-lib-v1.32.1.tar.gz",
     )
+
+    _config_settings(
+        name = "aspect_config_settings",
```
alexeagle, line 31: this is creating a repo for rules_js in a global namespace of workspaces, so I think the name needs to be fixed like aspect_rules_js_config_settings? What about aspect_bazel_lib as a home for this? I assume we'll have a problem with this flag rename in some of our other rulesets

js/private/repository_utils.bzl
```diff
@@ -0,0 +1,24 @@
...
+        prop = "allow_unresolved_symlinks"
+
+    rctx.file("BUILD.bazel", """# @generated by @aspect_rules_js//js/private:repository_utils.bzl
+
+config_setting(
+    name = "allow_unresolved_symlinks",
+    values = {"experimental_allow_unresolved_symlinks": "true"},
+    visibility = ["//visibility:public"],
+)
+""".replace("experimental_allow_unresolved_symlinks", prop))
```
thesayyn, line 19: nit: {{}} could be used to escape the {} and pass the key as format arg.

thesayyn (approved):

js/repositories.bzl
jbedard: Yeah I don't think this location is where this should live long term. I'm not sure if we want to release a bazel-lib change or get this fixed within rules_js asap?

jbedard: Or it might actually have to be here due to what @thesayyn was pointing out.

I've done the name change though.

js/private/repository_utils.bzl
```diff
@@ -0,0 +1,25 @@
+"""
+Repository rule utilities.
+"""
+
+def _config_settings_impl(rctx):
+    v = native.bazel_version.split(".")
+    is_6_3_or_greater = len(v) > 1 and v[0].isdigit() and (int(v[0]) < 6 or (int(v[0]) == 6 and v[1].isdigit() and int(v[1]) < 3))
```
jbedard, line 7: This was originally using bazel_skylib [versions.bzl](https://github.com/bazelbuild/bazel-skylib/blob/288731ef9f7f688932bd50e704a91a45ec185f9b/lib/versions.bzl) however we can't even use that dep because it is loaded within the `rules_js_dependencies()` which is too late because this file is also loaded there.

The unit tests only passed because bazel_skylib is in the dev repos for this repository, but e2e tests were failing when using bazel_skylib.

js/private/repository_utils.bzl
```diff
@@ -0,0 +1,25 @@
+"""
+Repository rule utilities.
+"""
+
+def _config_settings_impl(rctx):
+    v = native.bazel_version.split(".")
+    is_less_than_6_3 = len(v) > 1 and v[0].isdigit() and (int(v[0]) < 6 or (int(v[0]) == 6 and v[1].isdigit() and int(v[1]) < 3))
```
jbedard, line 7: This was originally using bazel_skylib [versions.bzl](https://github.com/bazelbuild/bazel-skylib/blob/288731ef9f7f688932bd50e704a91a45ec185f9b/lib/versions.bzl) however we can't even use that dep because it is loaded within the `rules_js_dependencies()` which is too late because this file is also loaded there.

The unit tests only passed because bazel_skylib is in the dev repos for this repository, but e2e tests were failing when using bazel_skylib.

jbedard:
I'm converting to a draft for now. I think we need at least a pre-release tagged to test this before merging.

alexeagle:
Did we get that tagged pre-release yet?

jbedard:
Doesn't look like it: https://github.com/bazelbuild/bazel/commit/215f6982d3caec22e67d2fe345a61d63351c7937

Also looks like this might *not* land in 6.3 like originally suggested: https://github.com/bazelbuild/bazel/issues/18553

PR #1104, opened 2023-06-06
ci: enable testing on bazel 7

jbedard:
Excluding e2e and rbe for now to reduce the number of jobs to just some essentials.

---

### Type of change

### Test plan

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,10 +46,12 @@ jobs:
             - uses: actions/checkout@v3
             - id: bazel_6
               run: echo "bazelversion=$(head -n 1 .bazelversion)" >> $GITHUB_OUTPUT
-            - id: bazel_4_1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -177,6 +181,15 @@ jobs:
                       bzlmodEnabled: true
                     - folder: e2e/rules_foo
                       bzlmodEnabled: true
+                include:
```
gregmagolan, line 184: oh. You can do an exclude & then include pattern with GHA 😎

gregmagolan (approved):

.github/workflows/ci.yaml
jbedard: That was the best way I could find to only run a few core tests on v7, while also trying to reduce the number of times we repeat that version string.

PR #1107, opened 2023-06-08
ci: enable buildifier check on ci

jbedard:
Are we in favour of doing this to all the repos?

---

### Type of change

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ default_stages: [commit]
 repos:
     # Check formatting and lint for starlark code
     - repo: https://github.com/keith/pre-commit-buildifier
-      rev: 5.0.1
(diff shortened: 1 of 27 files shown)


.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@ default_stages: [commit]
 repos:
     # Check formatting and lint for starlark code
     - repo: https://github.com/keith/pre-commit-buildifier
-      rev: 5.0.1
+      rev: 6.1.0.1
```
gregmagolan, line 12: super semver

gregmagolan (approved):

.pre-commit-config.yaml
jbedard: Alex made the same change when adding this buildifier ci step in the rules_template 🤷

PR #1108, opened 2023-06-08
fix: multiple versions of a package using aliases in pnpm workspaces crashes

jbedard:
No fix yet but the e2e test reproduces it, note that the unit test does not reproduce it so it might have something to do with references between workspaces?

---

### Type of change

### Test plan

- New test cases added


gregmagolan (approved):

e2e/pnpm_workspace_deps/lib-dupes/BUILD.bazel
```diff
@@ -0,0 +1,26 @@
...
+load("@npm//:defs.bzl", "npm_link_all_packages")
+
+npm_link_all_packages(name = "node_modules")
+
+# A library with 2 aliases to one simple dependency (package with no dependencies)
+js_library(
+    name = "js_lib_dupes",
+    srcs = ["index.js"],
+    data = [
+        ":node_modules/@aspect-test/c",
```
gregmagolan, line 12: add a comment that the `data` here is intentional and necessary to check for regressions on #1110

PR #1117, opened 2023-06-24
fix: allow for a package.json file referenced in npm_translate_lock data to not exist if update_pnpm_lock is False

gregmagolan:
Follow-up to https://github.com/aspect-build/rules_js/pull/1116

---

### Type of change

### Test plan

- New test cases added


npm/private/npm_translate_lock_state.bzl
```diff
@@ -102,12 +102,14 @@ def _validate_attrs(attr, is_windows):
...
 
 ################################################################################
-def _init_common_labels(rctx, label_store):
+def _init_common_labels(priv, rctx, label_store):
     attr = rctx.attr
 
     # data files
-    for i, d in enumerate(attr.data):
-        label_store.add("data_{}".format(i), d)
+    # only initialize if update_pnpm_lock is set since data files are unused otherwise
```
jbedard, line 109: Is there any guarantee or at least docs somewhere that ensure the data files are only used if `update_pnpm_lock`?

Can this `if` be moved and merge into the other one (within `if attr.pnpm_lock`)? I guess that just makes the "data is for lockfile" stricter and clearer...

e2e/npm_translate_lock_partial_clone/root/pnpm-workspace.yaml
```diff
@@ -0,0 +1,4 @@
+packages:
+    - '.'
+    - '../project-a'
```
jbedard, line 3: Was the lockfile generated with a `project-a/package.json` existing? Without that file if someone runs `pnpm install` will the lockfile change?

npm/private/npm_translate_lock_state.bzl
jfirebaugh: Yeah the [docs for `data`](https://docs.aspect.build/rules/aspect_rules_js/docs/npm_translate_lock/#data) say "Only needed when `update_pnpm_lock` is True".

e2e/npm_translate_lock_partial_clone/root/pnpm-workspace.yaml
gregmagolan: Yes. If someone runs `pnpm install` in this project the lockfile would change and it would break the test case.

npm/private/npm_translate_lock_state.bzl
gregmagolan: I think this branch is intentionally not under `if attr.pnpm_lock` for the pnpm lock bootstrap case.

e2e/npm_translate_lock_partial_clone/root/pnpm-workspace.yaml
jbedard: I just wonder if we'll break/remove the original intention of the test when that happens, such as when renovate tries to update something

jbedard (approved):

PR #1125, opened 2023-06-28
feat: add src_directory attribute to NpmPackageStoreInfo provider

gregmagolan:
This is useful when you want to know where a linked package was originally copied from before the virtual store.

---

### Type of change


jbedard (approved):

jbedard:
Did you have any ideas for tests? Or I can add one...

gregmagolan:
> Did you have any ideas for tests? Or I can add one...

We don't have a lot of tests for the providers in rules_js. I'm fine with a YOLO on this one as the code path is trivial and singular so it is not likely to regress.

PR #1139, opened 2023-07-03
chore: test a large npm package tgz.bz2 such as pyoxide

gregmagolan:
Test that rules_js can handle a large npm package tgz bz2 archive.

Bazel Slack thread for more context: https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1688396780749809

---

### Type of change

### Test plan

- New test cases added


jbedard:
Is it worth testing just because it's a large download? Or did you also want to test the postinstall stuff discussed in that thread?

gregmagolan:
> Is it worth testing just because it's a large download? Or did you also want to test the postinstall stuff discussed in that thread?

I'm on the fence. I'll leave it as a DRAFT for now and think about it after I'm back from vacation.

jbedard:
@gregmagolan this is old and untouched so I'm closing for now...

PR #1140, opened 2023-07-04
chore: bump rules_go dep to 0.40.1 since 0.35.0 is yanked from BCR

gregmagolan:
CI is broken at HEAD due to rules_go 0.35.0 dep since it is yanked; 0.38.1 exists but it depends on a yanked gazelle version so bumping all the way to the latest 0.40.1.


jbedard (commented):
👍

PR #1143, opened 2023-07-05
feat: add npm_translate_lock(update_pnpm_lock_node_toolchain_prefix) to specify node toolchain

jbedard:


alexeagle (commented):
This new attribute is only meant to be used when we are doing `update_pnpm_lock` so 
- add a validation so we `fail` if this is set but `update_pnpm_lock` isn't
- (maybe) name the attribute so the pairing is clear, if we expect we'll never want the node toolchain in `npm_translate_lock` in any other case

npm/private/npm_translate_lock_state.bzl
```diff
@@ -150,7 +150,7 @@ def _init_pnpm_labels(label_store, rctx):
     #
     # TODO: Try to understand this better and see if we can go back to using
```
alexeagle, line 151: since we're here and studying this, could we remove the TODO as well?

jbedard:
I've added the `update_pnpm_lock_` prefix to the attribute name. I'm not sure if it's worth adding validation though which would require more logic then just `= "nodejs"` as a default param for it. Do you think it's worth it?

npm/private/npm_translate_lock_state.bzl
jbedard: I think that's out of scope here and I'd rather not put the time in atm.

alexeagle (approved):
sure, looks good enough to me, thanks

PR #1145, opened 2023-07-05
chore: add Windows and MacOS CI

gregmagolan:
Windows & MacOS CI run on main branch only and only for `e2e/bzlmod` job for now. They will run on PRs if the branch name contains `windows` or `macos` respectively.

This PR also removes RBE CI from PRs to reduce the matrix size for PRs.

---

### Type of change

### Test plan

This PR will run the new CI matrix jobs

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,6 +23,25 @@ jobs:
     # inspired from
     # https://stackoverflow.com/questions/65384420/how-to-make-a-github-action-matrix-element-conditional
 
+    matrix-prep-os:
+        # Prepares the 'os' axis of the test matrix, to reduce costs since GitHub hosted runners cost more on some platforms.
+        # https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes
+        runs-on: ubuntu-latest
+        steps:
+            - id: linux
+              run: echo "os=ubuntu-latest" >> $GITHUB_OUTPUT
+            - id: windows
+              run: echo "os=windows-latest" >> $GITHUB_OUTPUT
+              # Only run on main branch (or PR branches that contain 'windows') to minimize Windows minutes (billed at 2X)
+              if: github.ref == 'refs/heads/main' || contains(github.head_ref, 'windows')
+            - id: macos
+              run: echo "os=macos-latest" >> $GITHUB_OUTPUT
+              # Only run on main branch (or PR branches that contain 'macos') to minimize macOS minutes (billed at 10X)
+              if: github.ref == 'refs/heads/main' || contains(github.head_ref, 'macos')
+        outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -234,8 +283,18 @@ jobs:
...
-              # Writes EngFlow credential files for RBE configurations
+            - name: Configure Bazel version (Windows)
+              if: matrix.os == 'windows-latest'
+              working-directory: ${{ matrix.folder }}
+              # Overwrite the .bazelversion instead of using USE_BAZEL_VERSION so that Bazelisk
+              # still bootstraps Aspect CLI from configuration in .bazeliskrc. Aspect CLI will
+              # then use .bazelversion to determine which Bazel version to use.
+              run: |
+                  echo "${{ matrix.bazelversion }}" > .bazelversion
+                  $BAZEL_MAJOR_VERSION = "${{ matrix.bazelversion }}".substring(0, 1)
```
jbedard, line 294: Where is `BAZEL_MAJOR_VERSION` used? That's not used with non-windows?

gregmagolan: Nice catch. That is a copy-paste left-over from https://github.com/aspect-build/bazel-lib/blob/794df714d7efbf5f2b986470428bea311f4fd772/.github/workflows/ci.yaml#L168 where it is used.

.github/workflows/ci.yaml
```diff
@@ -234,8 +273,16 @@ jobs:
...
 
-            - name: Write EngFlow credentials
-              # Writes EngFlow credential files for RBE configurations
+            - name: Configure Bazel version (Windows)
+              if: matrix.os == 'windows-latest'
+              working-directory: ${{ matrix.folder }}
+              # Overwrite the .bazelversion instead of using USE_BAZEL_VERSION so that Bazelisk
+              # still bootstraps Aspect CLI from configuration in .bazeliskrc. Aspect CLI will
+              # then use .bazelversion to determine which Bazel version to use.
+              run: echo "${{ matrix.bazelversion }}" > .bazelversion
```
jbedard, line 282: Why do we have a windows specific "Configure Bazel version"? It seems like we only drop the bazel5 setup in the windows one?

.github/workflows/ci.yaml
```diff
@@ -273,6 +321,18 @@ jobs:
                   ASPECT_NPM_AUTH_TOKEN: ${{ secrets.ASPECT_NPM_AUTH_TOKEN }}
                   ASPECT_GH_PACKAGES_AUTH_TOKEN: ${{ secrets.ASPECT_GH_PACKAGES_AUTH_TOKEN }}
 
+            - name: bazel test //... (Windows)
```
jbedard, line 324: This just drops the `ls | ... | grep`? Was that causing issues worth copy+pasting this for windows?

jbedard (approved):

Issue #1131, opened 2023-06-30
[Bug]: js_run_devserver should produce a .bat file on Windows


alexeagle:
### What happened?

Running the devserver target produces an error like `entry_point not found`, see screenshot from Windows:

![Screenshot 2023-06-28 at 9 30 48 AM](https://github.com/aspect-build/rules_js/assets/47395/5c2a8b0d-1202-4e9f-b305-fba97c810fff)

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: 6.2.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: aspect_rules_js 1.26.1

### How to reproduce

```shell
A simple example with an http-server is enough to cause this.
```

gregmagolan:
With https://github.com/aspect-build/rules_js/pull/1156 this should work using the `command` entry point for `js_run_devserver` with the `bins` feature of `npm_translate_lock`.

The trouble with the `http_server_bin.http_server_binary(name = "http_server")` approach on Windows is that this method relies on the `bin` of the npm package to be executable, which it typically is a linux/darwin since it is typically an executable file with `#!/usr/bin/env node`. This is true for `http-server`. On Windows, the `bin` can't be used directly but must instead be used via the generated `node_modules/.bin/http-server.cmd` file that yarn, npm & pnpm all create.

#1156 adds the functionality to create `node_modules/.bin/http-server.cmd` and `js_run_devserver` already has the ability to run an arbitrary command with `command`. So the wiring would look like so:

```
npm_translate_lock(
    name = "npm",
    bins = {
        # derived from "bin" attribute in node_modules/http-server/package.json
        "http-server": {
            "http-server": "./bin/http-server",
        },
    },
    npmrc = "//:.npmrc",
    pnpm_lock = "//:pnpm-lock.yaml",
    verify_node_modules_ignored = "//:.bazelignore",
)
```

```
js_run_devserver(
    name = "devserver_alt",
    chdir = package_name(),
    command = "../node_modules/.bin/http-server.cmd",
    data = [
        ":http_root",
        ":web_files",
        "//:node_modules/http-server",
    ],
    env = {"BAZEL_BINDIR": "."},
    log_level = "debug",
)
```

[...]

PR #1156, opened 2023-07-11
feat: add support for .cmd bin entry points for npm_translate_lock bin attribute

gregmagolan:
---

### Type of change

### Test plan

- New test cases added


js/private/js_run_devserver.mjs
```diff
@@ -139,6 +160,21 @@ async function syncRecursive(src, dst, writePerm) {
 async function sync(files, sandbox, writePerm) {
     console.error('Syncing...')
     const startTime = perf_hooks.performance.now()
+    const isWindows = process.platform === 'win32'
+    if (isWindows) {
+        // special handling on Windows for node_modules
+        files = files.map((file) => {
+            const nodeModulesIndex = file.indexOf('/node_modules/')
```
jbedard, line 167: Are nested `node_modules` not possible? I would have assumed the `if/else` would be the opposite order...

js/private/js_run_devserver.mjs
```diff
@@ -48,15 +48,36 @@ function mkdirpSync(p) {
 // pointing to the source symlink.
 async function syncRecursive(src, dst, writePerm) {
     try {
-        const lstat = await fs.promises.lstat(src)
+        const isWindows = process.platform === 'win32'
+        if (isWindows && path.basename(src) == 'node_modules') {
+            // special handling on Windows for node_modules
```
jbedard, line 53: Can we explain what that handling is and why? Why it's different then the non-windows special handling?

jbedard (approved):

gregmagolan:
Hmm. Putting this back to DRAFT for now since I discovered that https://github.com/aspect-build/rules_js/pull/1157 solves https://github.com/aspect-build/rules_js/issues/1131 cleaner and with less changes. This will still be needed to get the `command` in `js_run_devserver` running on Windows but the `tool` method can now be used on Windows with the fix #1157 which is simpler and principled.

alexeagle:
#1157 fixed it for the one client that asked for this, FWIW

alexeagle:
Should we close this one?

gregmagolan:
> Should we close this one?

It can stay around as a DRAFT so it doesn't get lost. Worth doing one day but its a low priority.

PR #1169, opened 2023-07-21
Fix starlark patching example syntax

pqn:
### Type of change

Documentation correctness.

### Test plan

N/A


jbedard (approved):
Thanks! 👍

PR #1172, opened 2023-07-24
docs: add package path to bin example

jbedard:


docs/README.md
```diff
@@ -180,7 +180,7 @@ to a Starlark API you can load from in your BUILD file or macro.
 For example, if you depend on the `typescript` npm package, you write this in `BUILD`:
 
 ```starlark=
-load("@npm//:typescript/package_json.bzl", typescript_bin = "bin")
+load("@npm//[path/to/package]:typescript/package_json.bzl", typescript_bin = "bin")
```
gregmagolan, line 183: I'd say update some of the text instead and give two example,

```
For example, if you depend on the `typescript` npm package in your root `package.json`, you write this in `BUILD`:
...
```

```
If you depend on the `typescript` npm package from a nested package.json such as `myapp/package.json`, the load statement would look like this instead:

load("@npm//myapp:typescript/package_json.bzl", typescript_bin = "bin")
```

kourtni: I agree with @gregmagolan here. Having some extra sentences that point out the difference would be beneficial.

jbedard: Updated, a bit different then suggested so let me know what you think.

alexeagle (approved):

PR #1175, opened 2023-07-25
`js_run_devserver` should use `existsSync` instead of a custom `fileExists`

ghorbanzade:
We recently upgraded `rules_js` from `v1.25.4` to `v1.30.0` and discovered a regression introduced by #1043 in `v1.26.0` that impacts `js_run_devserver`. That PR introduces a custom function `fileExists` to leverage `fs.promises.access` as an alternative to the deprecated `fs.exists` (see [comment](https://github.com/aspect-build/rules_js/blob/d565889912e97a4eba9d0d059263f1c22ba2f452/js/private/js_run_devserver.mjs#L16-L17) and [review thread](https://github.com/aspect-build/rules_js/pull/1043#discussion_r1187955118)).

But the implementation of `fileExists` is missing a return statement, resulting in the two callsites to interpret `undefined` as a signal that the given file does not exist.

While it is tempting to add a return statement, this PR proposes the removal of `fileExists` in favor of [`fs.existsSync`](https://nodejs.org/api/fs.html#fsexistssyncpath) which is not deprecated. Our reasoning is:
- [`fs.access`](https://nodejs.org/api/fs.html#fsaccesspath-mode-callback) with `fs.constants.F_OK` does not have the exact same behavior as `fs.existsSync`. It checks the file permissions with respect to the calling process which, in our case, resulted in the file to be perceived as missing while `fs.existsSync` returned `true`.
- The two call-sites of `fileExists` immediately `await` its return value so there is no performance gain in using an async function. For this use-case, `existsSync` feels simpler, clearer, and more appropriate.

### Type of change

### Test plan

Tested manually on linux by running `bazel run //src:devserver` in `e2e/js_run_devserver`

### DIFF
--- a/js/private/js_run_devserver.mjs
+++ b/js/private/js_run_devserver.mjs
@@ -12,15 +12,6 @@ const RUNFILES_ROOT = path.join(
 const synced = new Map()
 const mkdirs = new Set()
 
-async function fileExists(p) {
-    // Node recommends using `access` to check for file existence; fs.exists is deprecated.
-    // https://nodejs.org/api/fs.html#fsexistspath-callback
-    await fs.promises
-        .access(p, fs.constants.F_OK)
-        .then(() => true)
-        .catch((_) => false)
-}
-
 // Ensure that a directory exists. If it has not been previously created or does not exist then it
 // creates the directory, first recursively ensuring that its parent directory exists. Intentionally
 // synchronous to avoid race conditions between async promises. If we use `await fs.promises.mkdir(p)`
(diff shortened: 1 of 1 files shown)


jbedard:
Using the async version does still have the performance gain AFAIU, other async code can run while the promise is being `await`ed just like callbacks but unlike the `*Sync` methods.

Can you describe your case more where `access` doesn't work? In what case does the calling process not have file permissions? I guess something such as a `js_binary` could run with any permissions... 🤔

jbedard (approved):

PR #1178, opened 2023-07-27
fix: allow for Bazel 7 flag rename

### LINKED ISSUE #1102: rules_js is failing with error "while parsing configuration settings: unknown option: 'experimental_allow_unresolved_symlinks' " with Bazel@HEAD in CI

alexeagle:
blocked on https://github.com/bazel-contrib/bazel_features/pull/15
fixes #1102

### Type of change

### Test plan

`USE_BAZEL_VERSION=last_green bazel build --nobuild js/...  --override_repository=bazel_features=$HOME/Projects/bazel_features`

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -7,6 +7,7 @@ module(
 )
 
 bazel_dep(name = "aspect_bazel_lib", version = "1.31.2")
+bazel_dep(name = "bazel_features", version = "0.0.1")
 bazel_dep(name = "bazel_skylib", version = "1.4.1")
 bazel_dep(name = "rules_nodejs", version = "5.8.2")
 bazel_dep(name = "platforms", version = "0.0.4")
--- a/npm/repositories.bzl
+++ b/npm/repositories.bzl
@@ -1,11 +1,16 @@
 """Repository rules to fetch third-party npm packages"""
 
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
 load("//npm/private:npm_import.bzl", _npm_import = "npm_import")
 load("//npm/private:npm_translate_lock.bzl", _list_patches = "list_patches", _npm_translate_lock = "npm_translate_lock")
 load("//npm/private:pnpm_repository.bzl", _LATEST_PNPM_VERSION = "LATEST_PNPM_VERSION", _pnpm_repository = "pnpm_repository")
 
 npm_import = _npm_import
-npm_translate_lock = _npm_translate_lock
+
+def npm_translate_lock(**kwargs):
+    bazel_features_deps()
+    _npm_translate_lock(**kwargs)
+
 pnpm_repository = _pnpm_repository
 LATEST_PNPM_VERSION = _LATEST_PNPM_VERSION
 list_patches = _list_patches
--- a/js/defs.bzl
+++ b/js/defs.bzl
@@ -33,7 +33,7 @@ def js_binary(**kwargs):
             "//conditions:default": False,
         }),
         unresolved_symlinks_enabled = select({
-            "@aspect_rules_js//js/private:experimental_allow_unresolved_symlinks": True,
+            "@aspect_rules_js//js/private:allow_unresolved_symlinks": True,
             "//conditions:default": False,
         }),
         **kwargs
@@ -46,7 +46,7 @@ def js_test(**kwargs):
             "//conditions:default": False,
         }),
         unresolved_symlinks_enabled = select({
-            "@aspect_rules_js//js/private:experimental_allow_unresolved_symlinks": True,
+            "@aspect_rules_js//js/private:allow_unresolved_symlinks": True,
             "//conditions:default": False,
         }),
         **kwargs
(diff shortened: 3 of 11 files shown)


npm/repositories.bzl
```diff
@@ -1,11 +1,16 @@
...
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
 load("//npm/private:npm_import.bzl", _npm_import = "npm_import")
 load("//npm/private:npm_translate_lock.bzl", _list_patches = "list_patches", _npm_translate_lock = "npm_translate_lock")
 load("//npm/private:pnpm_repository.bzl", _LATEST_PNPM_VERSION = "LATEST_PNPM_VERSION", _pnpm_repository = "pnpm_repository")
 
 npm_import = _npm_import
-npm_translate_lock = _npm_translate_lock
+
+def npm_translate_lock(**kwargs):
+    bazel_features_deps()
```
alexeagle, line 11: this is a bit awkward, but allows us a place to create the `bazel_features_versions` codegen repo without a breaking change for WORKSPACE users.
We assume everyone calls this at least once. If they don't, they'll have to call `bazel_features_deps()` themselves.
Sadly I don't know of a way to warn them about that when they get the error from the `load` statement.

alexeagle: I added to the "workspace snippet" we publish on releases, so at least it's documented somewhere.

e2e/workspace/WORKSPACE
```diff
@@ -16,6 +16,11 @@ nodejs_register_toolchains(
     node_version = DEFAULT_NODE_VERSION,
 )
 
+# NB: you can skip these "bazel_features" lines, so long as you call `npm_translate_lock`.
```
jbedard, line 19: Do we want to keep it in `npm_translate_lock` long term? I think we'd want a `rules_js_deps()` instead...

alexeagle: I think we do, yes. WORKSPACE users shouldn't have an extra thing leaking out, there should only need to be two calls, "fetch deps" and "initialize". Plus, WORKSPACE is the "legacy way" to be removed by Bazel 9 so we don't care that much about changing this setup.

jbedard: Can we not wrap this in a `rules_js_deps()` though?

Then leave the call in `npm_translate_lock` to prevent a breaking change for most people, but I'd say drop that in 2.0 if we add `rules_js_deps`...

alexeagle: Well, I just would rather not add yet more surface area to the WORKSPACE setup, ever. Under bzlmod this works great, and if we add a `rules_js_deps()` call then it's just another thing to document, and would never (?) be used for another purpose.
@gregmagolan WDYT?

jbedard: Do we need to make our dependency on bazel_features public though? Can we wrap it in our own `_deps` method even if we call it here?

gregmagolan: Given the legacy WORKSPACE way is going to be removed from Bazel v9 in approx 1 year I'd just go with Alex's hack inside npm_translate_lock and not add a thing that is going to be obsolete in 1 year. 99.9% of users will have that. I'm suddenly really looking forward to Bazel 9.

gregmagolan (approved):

e2e/workspace/WORKSPACE
alexeagle: cool, thanks.

MODULE.bazel
```diff
@@ -6,7 +6,10 @@ module(
     compatibility_level = 1,
 )
 
+# Lower-bounds for runtime dependencies.
+# Do not bump these unless rules_js requires a newer version to function.
```
gregmagolan, line 10: As an aside, I'd also advocate for minimum version bumps if there are critical bug fixes or notable performance improvements to would improve function.

PR #763, opened 2023-01-09
feat: automatically update pnpm lockfile via bzlmod api

kormide:
In support of https://github.com/aspect-build/rules_js/issues/644

The goal of this PR was to get `e2e/npm_translate_lock` running with/without bzlmod. The attributes I added to the bzlmod api might unlock more e2es, but I'll go through those one at a time in different PRs.

PR #1186, opened 2023-08-02
feat: accept (version, integrity) for pnpm_version

### LINKED ISSUE #1173: [FR]: Support arbitrary pnpm versions if metadata is provided manually

`pnpm_repository` only supports pnpm versions that have their metadata hard-coded in `versions.bzl`.

### What do we want to achieve?

We have a policy of using the most recent version of dependencies, so we would like to use the most recent release of pnpm before its metadata makes it into a `rules_js` release.

### Describe the feature

Can you add an optional argument to `pnpm_repository` where we can override the hard-coded metadata dictionary?

alexeagle:
This reverts commit 6474e4870061e8ff06d2b91948a0d5e8332fbbd7.

Roll-forward #1184 with fixes.

Fixes #1173 

---

### Type of change

Will add missing docs for `pnpm_repository.bzl` in a followup.

Minor breaking change!

### Test plan

I changed our root module to use the API.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -26,9 +26,10 @@ use_repo(node, "nodejs_windows_amd64")
 pnpm = use_extension("@aspect_rules_js//npm:extensions.bzl", "pnpm")
 pnpm.pnpm(
     name = "pnpm",
+    pnpm_version = "8.6.7",
+    pnpm_version_integrity = "sha512-vRIWpD/L4phf9Bk2o/O2TDR8fFoJnpYrp2TKqTIZF/qZ2/rgL3qKXzHofHgbXsinwMoSEigz28sqk3pQ+yMEQQ==",
 )
-use_repo(pnpm, "pnpm")
-use_repo(pnpm, "pnpm__links")
+use_repo(pnpm, "pnpm", "pnpm__links")
 
 ####### Dev dependencies ########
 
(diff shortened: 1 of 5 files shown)


MODULE.bazel
```diff
@@ -23,14 +23,19 @@ use_repo(node, "nodejs_linux_s390x")
 use_repo(node, "nodejs_linux_ppc64le")
 use_repo(node, "nodejs_windows_amd64")
 
-pnpm = use_extension("@aspect_rules_js//npm:extensions.bzl", "pnpm")
+####### Dev dependencies ########
+
+pnpm = use_extension(
+    "@aspect_rules_js//npm:extensions.bzl",
+    "pnpm",
+    dev_dependency = True,
```
alexeagle, line 31: FYI @kormide I think it was a mistake that we installed a fixed pnpm version for everyone, as this was not a devdep, e2e/js_run_devserver was unable to select a different pnpm version.
Sound right?

alexeagle: It was introduced in https://github.com/aspect-build/rules_js/pull/763

kormide: I agree that it was a mistake. I think at the time I was trying to make the module extension do what the workspace rules does, but those are different idioms.

kormide (approved):

MODULE.bazel
alexeagle: oh, it's not this simple. rules_js references pnpm in load statements, so it does have to declare the dependency.

alexeagle:
Got this green by making it a pure refactor for now.

PR #1187, opened 2023-08-02
fix(public api): expose two targets that downstream rulesets need

alexeagle:
rules_jest, rules_jasmine and others use the js_binary API which requires a select() that needs these config_settings.
So, they are broken by the (unreleased) #1178

See https://github.com/aspect-build/rules_jasmine/blob/049a30281525c0fd4819d2989515d47f6728f288/jasmine/defs.bzl#L71-L78

So, we should just make these public API so we know we cannot change them. Add two `alias` targets to make this change non-breaking.

---

### Type of change

### Test plan

### DIFF
--- a/js/BUILD.bazel
+++ b/js/BUILD.bazel
@@ -1,10 +1,23 @@
 "Public API"
 
+load("@bazel_features//:features.bzl", "bazel_features")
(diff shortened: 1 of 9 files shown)


gregmagolan (approved):

Aghassi (approved):
This works for us internally, thanks for the fix!

PR #1203, opened 2023-08-08
created failing bazel run for js_image_docker e2e test

TarunGunampalli:
This PR creates a failing test case in order to reproduce an issue where bazel/docker can't resolve symlinks in pnpm node_modules.

Type of change
Red PR

Test plan
Covered by existing test cases


thesayyn:
this turned out to be a mistake on the app code instead of a js_image_layer bug.

PR #1205, opened 2023-08-08
feat: introduce option to override npm_package#include_declarations

alexeagle:
Tested in bazel-examples repo that by passing this flag, I can build an app that is downstream of a typecheck error, even though an npm_package is used to link the package.

This helps to solve https://github.com/aspect-build/rules_ts/issues/352 and makes rules_ts users less reliant on the Persistent Worker mode for performant dev round-trip.

### Type of change

### Test plan

- Manual testing; instructions above

### DIFF
--- a/npm/private/npm_package.bzl
+++ b/npm/private/npm_package.bzl
@@ -170,19 +170,16 @@ def npm_package(
         hardlink = "auto",
         verbose = False,
         **kwargs):
-    """A rule that packages sources into a directory (a tree artifact) and provides an `NpmPackageInfo`.
+    """A macro that packages sources into a directory (a tree artifact) and provides an `NpmPackageInfo`.
 
     This target can be used as the `src` attribute to `npm_link_package`.
 
-    Every npm_package target has a sub target named after its name, which is `<name>.publish`, that can be run
-    to publish to an npm registry.
-
+    The macro also produces a target `[name].publish`, that can be run to publish to an npm registry.
     Under the hood, this target runs `npm publish`. You can pass arguments to npm by escaping them from Bazel using a double-hyphen,
-    for example: `bazel run my_package.publish -- --tag=next`
-
+    for example: `bazel run //path/to:my_package.publish -- --tag=next`
 
     Files and directories can be arranged as needed in the output directory using
-    the `root_paths`, `include_srcs_patters`, `exclude_srcs_patters` and `replace_prefixes` attributes.
+    the `root_paths`, `include_srcs_patterns`, `exclude_srcs_patterns` and `replace_prefixes` attributes.
 
     Filters and transformations are applied in the following order:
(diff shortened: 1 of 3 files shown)


npm/private/npm_package.bzl
```diff
@@ -216,6 +213,11 @@ def npm_package(
     using the `include_sources`, `include_transitive_sources`, `include_declarations`, `include_transitive_declarations`
     attributes.
 
+    The two `include*_declarations` options may cause type-check actions to run, which slows down your
```
alexeagle, line 216: It's going to be a challenge for users to discover this feature... we could use `--compilation_mode fastbuild` to signal that we want to skip these typechecks, but it's less explicitly named and still not discoverable.

jbedard:
So this is an override, not a default? Do we need a way to ensure this override is not being used when a package is being built+published?

Do we still want to allow type-checking as a test? I don't think this would work for that case.

alexeagle:
It's up to the user to pass this flag only when they want to bypass type-checking, and not to include it in their release config. It has a long, explicit name which I would hope makes it clear that you don't want to do this all the time.

I don't think we need to have another path to allow type-checking as a test, again, you just don't include this flag in the build when you want to verify.

npm/private/npm_package.bzl
alexeagle: We think we can instrument both `js_run_devserver` and `js_test` such that we can detect the value of the flag, and warn "it looks like you're doing a development loop and still have extra type-checks turned on"

alexeagle: will try this in a followup. #1206

npm/private/npm_package.bzl
```diff
@@ -440,8 +442,14 @@ def npm_package(
             srcs = srcs,
             include_sources = include_sources,
             include_transitive_sources = include_transitive_sources,
-            include_declarations = include_declarations,
-            include_transitive_declarations = include_transitive_declarations,
+            include_declarations = select({
+                "@aspect_rules_js//npm:exclude_declarations_from_npm_packages_flag": False,
+                "//conditions:default": include_declarations,
+            }),
+            include_transitive_declarations = select({
```
jbedard, line 449: Would only setting this one to `False` ever be an improvement on its own? 🤔

alexeagle: I don't think so, because producing the direct .d.ts inputs will run the typechecker and it will use the transitive .d.ts as inputs, so they'll be requested from Bazel as well.

jbedard: Yeah, that's the standard case, I'm trying to think if there are any others 🤔 🤷

jbedard (approved):

gregmagolan (approved):

PR #1214, opened 2023-08-14
Add module not found troubleshooting guide

alexeagle:
### Type of change

### Test plan
None

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -87,12 +87,10 @@ https://github.com/aspect-build/rules_js/blob/a8c192eed0e553acb7000beee00c60d60a
 #### It's a plugin
 
 Sometimes the package intentionally doesn't list dependencies, because it discovers them at runtime.
-`eslint` and `prettier` are common typical examples.
+This is used for tools that locate their "plugins"; `eslint` and `prettier` are common typical examples.
 
 The solution is based on pnpm's [public-hoist-pattern](https://pnpm.io/npmrc#public-hoist-pattern).
-Use the [`public_hoist_packages` attribute of `npm_translate_lock`](./npm_translate_lock.md#npm_translate_lock-public_hoist_packages). This makes the `require` statement appear in the
-"public" root of the `node_modules` tree, so the resolution algorithm will search sibling packages.
-
+Use the [`public_hoist_packages` attribute of `npm_translate_lock`](./npm_translate_lock.md#npm_translate_lock-public_hoist_packages).
 The documentation says the value provided to each element in the map is:
 
 >  a list of Bazel packages in which to hoist the package to the top-level of the node_modules tree
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -51,6 +51,14 @@ config file to allow the `require` syntax.
 
 Once you've done this, it's handled like the "require appears in your code" case above.
 
+For example, the
+[documentation for the postcss-loader for Webpack](https://webpack.js.org/loaders/postcss-loader/#sugarss)
+suggests that you `npm install --save-dev sugarss`
+and then pass the string "sugarss" to the `options.postcssOptions.parser` property of the loader.
+However this violates symmetry and would require workarounds listed below.
+You can simply pass `require("sugarss")` instead of the bare string, then include the `sugarss`
+package in the `data` (runtime dependencies) of your `webpack.config.js`.
+
 #### It's a bug
 
 This is the case when a package has a `require` statement in its runtime code for some package, but
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -6,8 +6,9 @@ This is the most common error rules_js users encounter.
 These problems generally stem from a runtime `require` call of some library which was not declared as a dependency.
 
 Fortunately, these problems are not unique to Bazel.
-As described in https://github.com/aspect-build/rules_js/blob/main/docs/pnpm.md#hoisting, rules_js
-should behave the same way `pnpm` does with `hoist=false`.
+As described in [our documentation](./pnpm.md#hoisting),
+rules_js should behave the same way `pnpm` does with [`hoist=false`](https://pnpm.io/npmrc#hoist).
+
 These problems are also reproducible under [Yarn PnP](https://yarnpkg.com/features/pnp) because it
 also relies on correct dependencies.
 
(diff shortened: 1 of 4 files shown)


docs/troubleshooting.md
```diff
@@ -0,0 +1,90 @@
...
+https://nodejs.org/api/modules.html#loading-from-node_modules-folders
+
+Since the resolution starts from the callsite, the remedy depends on where the `require` statement appears.
+
+### require appears in your code
+
+This is the case when you write a `config.js` file and pass it to a tool.
+
+> This is the "ideal" way for JavaScript libraries to be configured, because it allows an easy
+> "symmetry" where you `require` a library and declare your dependency on it in the same package.
```
jbedard, line 24: "package" is a bit of an overloaded word, in this case we mean bazel target? The `srcs` and `deps of the same target...

docs/troubleshooting.md
```diff
@@ -0,0 +1,90 @@
...
+
+In this case you should add the runtime dependency to your BUILD file where the `config.js` is a source.
+
+For example,
+
+```starlark
+js_library(
+    name = "requires_foo",
+    srcs = ["config.js"],                  # contains "require('foo')"
+    data = ["//my/pkg:node_modules/foo"],  # satisfies that require
```
jbedard, line 34: Where is `my/pkg` relative to `config.js`?

alexeagle: It's unspecified. Here I'm trying to suggest you need to figure out where the foo package is linked. Do you have a suggestion how to change this?

jbedard: I think `config.js` needs to be within `my/pkg` for this to work. Is the "satisfies that require" comment enough? Or could change it to just `:node_modules/foo`.

What you have might be fine, I'm not sure what the clearest way of saying it is.

alexeagle: got it, I agree just removing `//my/pkg` makes it more typical of what a user will have to do.

Would love to have more examples you've run across included or linked in this doc!

docs/troubleshooting.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+### require appears in third-party code
+
+This case itself breaks down into three possible remedies, depending on whether you can move the
+require to your own code, the missing dependency can be considered a "bug",
+or the third-party package uses the "plugin pattern" to discover its
+plugins dynamically at runtime based on finding them based on a string you provided.
+
+#### The `require` can move to first-party
```
jbedard, line 45: This "require" has backticks while the others in headers don't 🤷

jbedard (approved):

docs/troubleshooting.md
```diff
@@ -0,0 +1,99 @@
+# Common troubleshooting tips
+
+## Module not found errors
+
+This is the most common error rules_js users encounter.
+These problems generally stem from a runtime `require` call of some library which was not declared as a dependency.
+
+Fortunately, these problems are not unique to Bazel.
+As described in https://github.com/aspect-build/rules_js/blob/main/docs/pnpm.md#hoisting, rules_js
```
gregmagolan, line 9: Link right to `docs.aspect.build` docsite equivalent from here?

docs/troubleshooting.md
```diff
@@ -0,0 +1,99 @@
...
+should behave the same way `pnpm` does with `hoist=false`.
```
gregmagolan, line 10: Could link `hoist=false` to https://pnpm.io/npmrc#hoist

docs/troubleshooting.md
```diff
@@ -0,0 +1,99 @@
...
+
+The recommended fix for both pnpm and rules_js is to use
+[pnpm.packageExtensions](https://pnpm.io/package_json#pnpmpackageextensions)
+in your `package.json` to add the missing `dependencies` or `peerDependencies`.
+
+Example,
+
+https://github.com/aspect-build/rules_js/blob/a8c192eed0e553acb7000beee00c60d60a32ed82/package.json#L12
+
+Make sure you run `pnpm install` after changing `package.json`, as rules_js only reads the
```
gregmagolan, line 82: Mention `--lockfile-only` here since it is faster? Also mention how to run the bazel provisioned pnpm?

docs/troubleshooting.md
```diff
@@ -0,0 +1,99 @@
...
+`pnpm-lock.yaml` file to gather dependency information.
+
+#### It's a plugin
+
+Sometimes the package intentionally doesn't list dependencies, because it discovers them at runtime.
+`eslint` and `prettier` are common typical examples.
+
+The solution is based on pnpm's [public-hoist-pattern](https://pnpm.io/npmrc#public-hoist-pattern).
```
gregmagolan, line 90: There is a devilish detail here that the `require` of the plugin from 3rd party code happens in the virtual store which is in the "root" `node_modules` tree of the pnpm workspace (the one next to the pnpm lock file). Any publicly hoisted plugins need to be hoisted to that node_modules tree in order to be found from the requires within the `node_modules/.aspect_rules_js` tree.

gregmagolan: Worth mentioning that public hoisting on its own isn't sufficient. You also have to add the hoisted `//:node_modules/foo` package to the deps of the targets that require it.

docs/troubleshooting.md
alexeagle: this should just be a relative link so you stay on whichever site you're on, fixed

docs/troubleshooting.md
alexeagle: pointed to https://docs.aspect.build/rules/aspect_rules_js/docs/#fetch-third-party-packages-from-npm which now has the `--lockfile-only` flag, to avoid this repeating in many places

docs/troubleshooting.md
gregmagolan: 🤔 It is a tricky one. I'll give it a go.

docs/troubleshooting.md
```diff
@@ -0,0 +1,134 @@
...
+> `pnpm-lock.yaml` file to gather dependency information.
+> See [Fetch third-party packages](./README.md#fetch-third-party-packages-from-npm)
+
+#### It's a plugin
+
+Sometimes the package intentionally doesn't list dependencies, because it discovers them at runtime.
+`eslint` and `prettier` are common typical examples.
+
+The solution is based on pnpm's [public-hoist-pattern](https://pnpm.io/npmrc#public-hoist-pattern).
+Use the [`public_hoist_packages` attribute of `npm_translate_lock`](./npm_translate_lock.md#npm_translate_lock-public_hoist_packages). This makes the `require` statement appear in the
```
gregmagolan, line 93: "This makes the `require` statement appear in the "public" root of the `node_modules` tree" 

not quite right...

alexeagle: removed that sentence

PR #1217, opened 2023-08-17
fix(npm_translate_lock): hash is stable when version changes

alexeagle:
This is a requested feature from a large client who rely on `update_pnpm_lock=True`. They have frequent merge conflicts on the npm_translate_lock_[hash] file in our external_repository_action_cache because they frequently edit the version number in their packages. We should never need to perform a 'pnpm import' just because a first-party package.json version field changed.

---

### Type of change

### Test plan

I changed the `version` in e2e/pnpm_workspace/app/b/package.json and ran the build, it no longer updates the `e2e/pnpm_workspace/.aspect/rules/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU=` file.

### DIFF
--- a/npm/private/npm_translate_lock_state.bzl
+++ b/npm/private/npm_translate_lock_state.bzl
@@ -426,13 +426,26 @@ def _copy_input_file(priv, rctx, label_store, key):
         fail("key not found '{}'".format(key))
 
     if _should_update_pnpm_lock(priv):
-        # NB: rctx.read will convert binary files to text but that is acceptable for
-        # the purposes of calculating a hash of the file
-        _set_input_hash(
-            priv,
-            label_store.relative_path(key),
-            utils.hash(rctx.read(label_store.path(key))),
-        )
+        path = label_store.path(key)
+        written = False
+        if path.endswith("package.json"):
+            content = json.decode(rctx.read(path))
+            if "version" in content.keys():
+                content.pop("version")
+                _set_input_hash(
+                    priv,
+                    label_store.relative_path(key),
+                    utils.hash(json.encode(content)),
+                )
+                written = True
+        if not written:
+            # NB: rctx.read will convert binary files to text but that is acceptable for
+            # the purposes of calculating a hash of the file
+            _set_input_hash(
+                priv,
+                label_store.relative_path(key),
+                utils.hash(rctx.read(path)),
+            )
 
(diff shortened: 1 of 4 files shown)


alexeagle:
Note, I've taken care to not churn files which didn't previously have a `version`.

npm/private/npm_translate_lock_state.bzl
```diff
@@ -426,13 +426,26 @@ def _copy_input_file(priv, rctx, label_store, key):
...
-        # the purposes of calculating a hash of the file
-        _set_input_hash(
-            priv,
-            label_store.relative_path(key),
-            utils.hash(rctx.read(label_store.path(key))),
-        )
+        path = label_store.path(key)
+        written = False
+        if path.endswith("package.json"):
+            content = json.decode(rctx.read(path))
```
jbedard, line 432: Do we care about handling failures here? I guess we parse package.json in the exact same way elsewhere so it doesn't matter?

npm/private/npm_translate_lock_state.bzl
```diff
@@ -426,13 +426,26 @@ def _copy_input_file(priv, rctx, label_store, key):
...
+            if "version" in content.keys():
+                content.pop("version")
```
jbedard, line 434: Can't we pop a lot more here? Or only read a couple key properties such as name, *dependencies? That might be a big change that should be saved for 2.0 though?

alexeagle: For each thing we pop, we have to be able to reason that it shouldn't affect pnpm import. So I'd rather be conservative and think about the use cases that matter. Have you seen any other fields that are updated frequently? I'm think it's the only one on https://docs.npmjs.com/cli/v9/configuring-npm/package-json

jbedard: In my case the ones updated frequently are the package version and dependency versions that are overridden with `pnpm.overrides`. You've solved the first one, the `pnpm.overrides` one is too complicated to do here.

There are other common fields we could drop such as `scripts` it's probably not worth it since they don't change to *that* frequently.

alexeagle: I think there's a good case for dropping scripts.

alexeagle: Discussed over VC - the problem is that users typically also update the deps of some other package.json to point to the new version, so they're still going to have a merge conflict even with this change.

Issue #1225, opened 2023-08-22
[Bug]: Support for cross-compilation of pnpm packages with lifecycle


sfc-gh-ptabor:
(Propesed fix contributed in: https://github.com/aspect-build/rules_js/pull/1226)

### What happened?

I was prototyping installing `sharp` (https://sharp.pixelplumbing.com/) using bazel. 
The created artifact (regardless of using `--platform=...` flag and constraints for `linux`) was being built for host's arch (darwin_arm64).  I would like the artifact to be build for the requested architecture (driven by `--platforms` flag or field or transition). 
 
In particular it led to a situation where the artifact couldn't be packed into docker image (using rules_oci), as the built module was for `darwin`. The `node` runtime embedded into image was proper, but the dependency was wrong: 

```
.../examples/bazel-out/darwin_arm64-fastbuild/bin/node_modules/.aspect_rules_js/sharp@0.32.5/node_modules/sharp/build/Release/sharp-darwin-arm64r8.node
```

I will send for review a PR with an idea for a fix: https://github.com/aspect-build/rules_js/pull/1226

### Version

```
bazel --version
aspect 5.5.2
```

```
uname -a
Darwin.... 22.6.0 Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:05 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T6000 arm64
```

```
http_archive(
    name = "aspect_rules_js",
    sha256 = "7b2a4d1d264e105eae49a27e2e78065b23e2e45724df2251eacdd317e95bfdfd",
    strip_prefix = "rules_js-1.31.0",
    url = "https://github.com/aspect-build/rules_js/releases/download/v1.31.0/rules_js-v1.31.0.tar.gz",
)
```

### How to reproduce

```shell
load("@aspect_rules_js//js:defs.bzl", "js_binary", "js_image_layer", "js_run_binary", "js_test")
load("@rules_oci//oci:defs.bzl", "oci_image", "oci_tarball")
load("@npm//:defs.bzl", "npm_link_all_packages")
load("@sfc_container_images_gcbi//:defs.bzl", "sfc_images_from_js_binary")
load("@container_structure_test//:defs.bzl", "container_structure_test")
load("//:blueprint.bzl", "BLUEPRINT")

npm_link_all_packages()

js_binary(
    name = "sharp",
    data = [
        "//node:node_modules/sharp",
    ],
    entry_point = "sharp.js",
)

platform(
[...]

PR #1226, opened 2023-08-22
Pass --arch and --platform flag to the 'lifecycle' hook to support cross-compilation (or proper download).

sfc-gh-ptabor:
Fixes: #1225

pnpm 'install' gets proper platform flags (--arch & --platform) such that right dependencies are being built/downloaded. 

Extensive description in: #1225.

### Type of change

- (with a small test of:) New feature or functionality (change which adds functionality)

### Test plan

Will update `e2e/js_image_oci` if the change is considered OK.


sfc-gh-ptabor:
@alexeagle Hi. Could you, please take a look ?

sfc-gh-ptabor:
@gregmagolan @alexeagle: Could you, please take, a look.

npm/private/npm_import.bzl
```diff
@@ -155,8 +155,16 @@ def npm_imported_package_store(
             args = [
                 "{package}",
                 "../../../$(execpath {npm_package_target_lc})",
-                "../../../$(@D)",
-            ],
+                "../../../$(@D)"] +
+                select({{
+                  "@platforms//os:osx": ["--platform=darwin"],
```
gregmagolan, line 160: Clever way to wire it up 👍

npm/private/lifecycle/lifecycle-hooks.js
```diff
@@ -173,6 +187,8 @@ async function main(args) {
         pkgRoot: path.resolve(outputDir),
         rawConfig: {
             stdio: 'inherit',
+            platform: platform,
```
gregmagolan, line 190: Nice. I wasn't aware you could set these here. I tracked this down to ending up in `config` on the `lifecycle` call to `@pnpm/npm-lifecycle` and ending up the `env`. Could leave a brief comment here as to how it works?

gregmagolan: Links to docs or directly to code would be helpful for the next person who looks at this

npm/private/npm_import.bzl
sfc-gh-ptabor: Thank you.

npm/private/lifecycle/lifecycle-hooks.js
sfc-gh-ptabor: Added.

sfc-gh-ptabor:
@gregmagolan Please Take Another Look.

sfc-gh-ptabor:
@gregmagolan - are the new descriptions sufficient ?

gregmagolan (approved):
🚀

gregmagolan:
> @gregmagolan - are the new descriptions sufficient ?

Those are great. Thanks for updating. Running CI and will merge when green.

gregmagolan (changes requested):
Hmmm. Looks like this is going to need some more work to be compatible with bzlmod. Looks issues accessing `@platform` and with the select statements. You should be able to reproduce locally with you run with `--enable_bzlmod`.

sfc-gh-ptabor:
I'm not sury why, but bumping the version of `platforms` to 0.0.5 resolved the problem for me locally.

gregmagolan:
> I'm not sury why, but bumping the version of `platforms` to 0.0.5 resolved the problem for me locally.

That does ring a bell with platforms 0.0.4 having issues with bzlmod tho I can't recall the details. I've seen a problem of a different shape with platforms and bzlmod before.

I pushed a commit that added the dep on "platforms" to all of the failing e2e tests. Want to see if that is sufficient to green them up. It is an issue, however, that this change adds that dependency on users since that would make it a BREAKING CHANGE. I think we'll have to figure out how to wire it up so that "platforms" dep is internal to rules_js and the generated starlark that ends up in the user's workspace doesn't have any references to `@platforms`.

gregmagolan:
Yeah, looks all green with the added `@platform` deps on the e2e tests. So we just have to figure out how to wire it up so that those are not needed.

I think aliasing the `@platform` targets in use in `@aspect_rules_js` and using those in the generated code instead will work.

sfc-gh-ptabor:
Thank you @gregmagolan for the advice and review.

 I added the platforms. I did it in the pre-existing file: //platforms.
Alternative is to have it as part of `/js/private/(platforms)?/` package.

Please approve the tests to make sure it works.

gregmagolan:
That looks great. I'm fine with them going under `//platforms`. I pushed a small tweak so the targets are under `//platforms/os/` and `//platforms/cpu/` so they match closer with the `@platforms` repository.

Looks good to me. Will let @alexeagle have a look before merging.

gregmagolan (approved):

alexeagle (commented):
I think this works for your use case, but it's a bit naive.

This unconditionally passes platform,arch to the lifecycle rawConfig which means any package that has compilation will be affected. If a platform-dependent npm binary is run in a build action, it be built for the target platform and not work. In theory, we need three different actions:
- run the lifecycle hooks for the host platform so the tool can be used in a local action
- run the lifecycle hooks for the exec platform so the tool can be used in a remote action
- run the lifecycle hooks for the target platform so the library can be packaged for runtime

Maybe one example is the `esbuild` package which has platform-dependent `optionalDependencies`, that seems like something we should test before landing this. Maybe we'll end up with a linux esbuild binary when trying to bundle on a mac.

gregmagolan:
Varying host/exec platform with RBE doesn't work in general with Bazel even at the toolchain layer so I don't think it is a use case we need to solve for right now. If you're bazel running a target on the host then target platform will be set to the host and it should work as expected.

I think this is an improvement over what we have now which is that setting or transitioning the target platform, for example if you're building for a docker container on a mac, then this change will allow for cross-compilation.

platforms/os/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
+alias(
```
thesayyn, line 1: i don't see the value here for these aliases. why can't we just use them in the select statement?

gregmagolan: bzlmod isn't happy about that because the select statements ends up in the user's workspace which forces the user to dep on `platforms` in their MODULE.bazel

thesayyn: oh, i see that makes sense. in that case can we move these into private?

gregmagolan: We could but they would then go in `npm/private` since they are being by `npm`; we may want to use them in `js/` the future so I figure `platforms` is a better home. Perhaps `platforms/private` tho we don't advertise `//platforms` as public API anywhere.

alexeagle:
Even without remote execution taken into consideration, won't this always run lifecycle hooks for the target platform even for a package that you intend to use with `js_run_binary` in a build step? I think that works today but will fail with this change.

gregmagolan:
> Even without remote execution taken into consideration, won't this always run lifecycle hooks for the target platform even for a package that you intend to use with `js_run_binary` in a build step? I think that works today but will fail with this change.

`js_run_binary` will transition the `js_binary` it is running to the execution platform along with all of its deps. It should work as expected with this change.

sfc-gh-ptabor:
1. @alexeagle Thank you for taking a look. I'm likely missing some, piece, but thought that executionTransitions
(https://github.com/bazelbuild/proposals/blob/main/designs/2019-02-12-execution-transitions.md) is the answer to the problem you are signaling, i.e. target should always be build for the 'target' platform, but it's responsibility of infrastructure to transition target platform to proper execution platform if a target is going to be used as 'tool'. 

2. How about `//private/platforms/{cpu,os}` for both: 
    - signaling it's not a public API
    - mimicking @platforms as closely as possible ?

sfc-gh-ptabor:
@alexeagle @gregmagolan @thesayyn: Could, you, please comment whether `//private/platforms/{cpu,os}` is the desired path ?

platforms/os/BUILD.bazel
alexeagle: I don't think this is a meaningful "public API surface" to bother about the private distinction.

alexeagle (approved):

alexeagle:
A user reports errors in some of their builds, likely related to this change:

```
configurable attribute "args" in //fullscreen:.aspect_rules_js/node_modules/source-map@0.1.43/lc doesn't match this configuration. Would a default condition help?
2023-09-27 17:39:57 UTC	
2023-09-27 17:39:57 UTC	Conditions checked:
2023-09-27 17:39:57 UTC	 @aspect_rules_js//platforms/os:osx
2023-09-27 17:39:57 UTC	 @aspect_rules_js//platforms/os:linux
2023-09-27 17:39:57 UTC	 @aspect_rules_js//platforms/os:windows
```

jbedard:
@sfc-gh-ptabor would you be interested in fixing the error that user is reporting? I think we just need to add a default value when the os/architecture is unknown, basically falling back to the previous behaviour in that case?

sfc-gh-ptabor:
@jbedard @alexeagle the fix is in: https://github.com/aspect-build/rules_js/pull/1297

Alternative is to learn which specific arch/os user needs and add explicit values for that mapping.

PR #1229, opened 2023-08-23
Revert "js_run_devserver symlinks scoped node_modules to bazel-out in…

alexeagle:
…stead of runfiles (unscoped node_modules are already linked to bazel-out) (#1210)"

This reverts commit da7e731c6a64c50570af88ab6193b75cc3a9c462.

CI has been red on RBE since this landed, so let's see if reverting makes it green.


alexeagle:
replace with #1230

PR #1230, opened 2023-08-23
Green up CI on main

alexeagle:
Somehow #1210 broke the RBE test on the main branch. Sadly we were not running RBE from PRs, unless "rbe" appears as a substring of the branch name the PR originates from.

We need to revert this to push a working release again. FYI @gregjacobs let's work together to roll-forward again from a branch with "rbe" in the name (or, perhaps we should always test PRs with at least one RBE job, #1231 for example)


jbedard (approved):

gregjacobs:
Oof! Thanks for letting me know @alexeagle. Let's chat on #1233

PR #1241, opened 2023-08-28
chore: switch from Engflow to Buildbuddy

alexeagle:
Our Engflow cluster seems to be down, all RBE builds have been red since last week.

Taking this opportunity to delete Engflow-related TODOs as well - it seems Buildbuddy correctly allows remote actions to use the network for example.


js/private/test/image/structure/BUILD.bazel
```diff
@@ -26,6 +26,8 @@ js_image_layer(
     compression = "none",
     platform = ":amd64_linux",
     root = "/app",
+    # TODO: figure out why the digests_generated.sum is different on buildbuddy executors
```
thesayyn, line 29: this is classic RBE; it has no access to the VCS and therefore has no knowledge of the `modes and bits` for the source files.

thesayyn: and different RBE implementations assign different `permission` bits to the files.

thesayyn (approved):
LGTM

js/private/test/image/structure/BUILD.bazel
alexeagle: okay, there was already a no-remote-exec on the test below, but I think that never would have worked there because it's not the execution of `diff` that fails, it's the action producing the inputs to `diff`...

Issue #1242, opened 2023-08-28
[Bug]: js_run_devserver: if 'tool' consumes stdin stream, js_run_devserver no longer syncs files on ibazel changes


gregjacobs:
### What happened?

Discovered an interesting one. For `js_run_devserver`, if the `tool`/`command` consumes the `stdin` stream, then the tool receives the `"IBAZEL_BUILD_COMPLETED SUCCESS"` messages from ibazel rather than the `js_run_devserver.mjs` script itself.

This causes `js_run_devserver` to no longer sync files to its custom sandbox after ibazel has completed. `webpack-dev-server` seems to be affected btw (including rules_webpack's `webpack_devserver()` rule), when using the Webpack config `watchOptions: { stdin: true }`

I believe the issue is because the child process directly inherits the `stdin` stream. Line 221 of `js_run_devserver.mjs`:

```js
const proc = child_process.spawn(tool, toolArgs, {
    cwd: cwd,
    stdio: 'inherit',  // <----
    env: env,
})
```

### Version

Development (host) and target OS/architectures: MacOS Monterey

Output of `bazel --version`: 6.2.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: `rules_js@1.32.1`

### How to reproduce

Just playing with `e2e/js_run_devserver/src`, and running the `:serve_simple` script.

If I add the following snippet into `simple.js`, then js_run_devserver stops syncing when files are changed:

```js
process.stdin.on('data', chunk => {
    console.log(`Child process's chunk: `, chunk.toString());
});
```

### Any other information?

I think the way to fix this may be to duplicate the `stdin` stream's contents inside `js_run_devserver.mjs`, and pass that to the child process after reading it for `"IBAZEL_BUILD_COMPLETED SUCCESS"` messages.

There's probably some good 3rd party libraries that do this in a one-liner, but just creating a new `stream.Writable` for the child process's `stdin` and writing to it should do the trick.

gregjacobs:
Actually, apologies, the fix is even simpler:

```js
const proc = child_process.spawn(tool, toolArgs, {
    cwd: cwd,
    stdio: ['pipe', 'inherit', 'inherit'],  // <-- changed
    env: env,
})
process.stdin.pipe(proc.stdin)
```

I'll create a PR

gregjacobs:
[...]

PR #1244, opened 2023-08-29
Fix for js_run_devserver to sync file changes if child process consumes stdin

gregjacobs:
Fixes https://github.com/aspect-build/rules_js/issues/1242

### Type of change

The child process of `js_run_devserver` can now consume `stdin` without preventing `js_run_devserver` from syncing upon ibazel's notifications

### Additional info

I'm only alerting the child process of ibazel success after the files have been synced in order to allow the child process to pick up the latest files in the `js_run_devserver` sandbox. This should not be a breaking change because `js_run_devserver` simply doesn't respond to ibazel's success messages currently when the child process adds a 'data' event handler on `process.stdin`.

### Test plan

Have been trying to think of how this can be tested in an automated fashion, but in the meantime it can be tested manually as follows:

1. Add this snippet to `e2e/js_run_devserver/src/simple.js`:
   ```js
   process.stdin.on('data', chunk => {
       console.log(`Child process's chunk: `, chunk.toString());
   });
   ```

2. Start the dev server:
   ```shell
   cd e2e/js_run_devserver/src
   bazel run :serve_simple
   ```

3. Make file content change in `e2e/js_run_devserver/src/index.html`
4. Observe that `js_run_devserver` sync'd the changed index.html file, and afterwards there's the "Child process's chunk" message on stdout

FYI @gregmagolan

### DIFF
--- a/js/private/js_run_devserver.mjs
+++ b/js/private/js_run_devserver.mjs
@@ -199,8 +199,17 @@ async function main(args, sandbox) {
 
         const proc = child_process.spawn(tool, toolArgs, {
             cwd: cwd,
-            stdio: 'inherit',
             env: env,
+
+            // Pipe stdin data to the child process rather than simply letting
+            // the child process inherit the stream and consume the data itself.
+            // If the child process consumes the data itself, then ibazel's
+            // messages like "IBAZEL_BUILD_COMPLETED SUCCESS" won't be seen by
+            // this 'js_run_devserver' process. Furthermore, we want to sync the
+            // files to the js_run_devserver's custom sandbox before alerting
+            // the child process of a successful build to allow it to read the
+            // latest files. This solves: https://github.com/aspect-build/rules_js/issues/1242
+            stdio: ['pipe', 'inherit', 'inherit'],
         })
 
         proc.on('close', (code) => {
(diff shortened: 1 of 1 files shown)


gregmagolan:
Thanks for the fix and details write up here and in the issue @gregjacobs. Makes sense at first glance. I'll try to fully understand it tomorrow with a fresh brain.

gregjacobs:
@gregmagolan Sounds good. It was definitely a weird one, and took me a few days to track down. I basically had this in my webpack config:

```js
watchOptions: {
    stdin: true,
}
```

which I guess causes Webpack to listen on the `stdin` stream, and seems to prevent `js_run_devserver.mjs` from receiving ibazel's "build success" messages instead.

But did notice that if js_run_devserver's child process has any `process.stdin.on('data', ...)` listener, this also prevents js_run_devserver.mjs from receiving ibazel's messages (except on occasion - catches the messages maybe 10% of the time).

Perhaps this is just a Mac thing though? Haven't tried on other platforms.

gregjacobs:
And there's actually a second part to this, which is that if the child process receives the `"IBAZEL_BUILD_COMPLETED SUCCESS\n"` message before js_run_devserver has synced its files, then an ibazel-aware child process may begin re-bundling with stale files

alexeagle:
Thanks @gregjacobs !!

alexeagle (approved):

PR #1245, opened 2023-08-29
fix: make symlink to be relative to path

thesayyn:
Symlinks used to be relative to the root which works for most use cases but fails if there's another symlink to the node_modules tree as symlinks become relative to the tail of the symlink.

### Type of change

### Test plan


alexeagle (approved):

alexeagle:
opt: It would be nice to see a failing test case this fixes

PR #1265, opened 2023-09-13
Update the npm_import documentation to have the correct self-reference.

kylepl:
See the reference just above, it has one less underscore.

---

### Type of change

### DIFF
--- a/docs/npm_import.md
+++ b/docs/npm_import.md
@@ -66,7 +66,7 @@ To consume the downloaded package in rules, it must be "linked" into the link pa
 package's `BUILD.bazel` file:
 
 ```
-load("@npm__at_types_node__15.12.2__links//:defs.bzl", npm_link_types_node = "npm_link_imported_package")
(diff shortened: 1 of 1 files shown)


jbedard:
I believe it's actually the opposite and it should be two. Want to update the one above to two?

kylepl:
Done.

jbedard:
@kylepl looks like that file is generated and you need to change it at the source, see npm/private/npm_import.bzl

kylepl:
Hm, it seems like the source actually has `npm__at_types_node__15` (i.e. double underscores twice). So I'm not sure why this does not match then. If there are steps that generate it, it's likely more involved then I'm willing to patch.

jbedard:
I think I see the bad one in the bzl file, there's also a few that correctly have the double underscore so you might just be looking at the wrong one?

If it's too much work I can try but it would be great if you could give it another shot since you have this going already 👍

PR #1268, opened 2023-09-13
perf: make app and node_modules layer separate actions

### LINKED ISSUE #1259: [Bug]: changes to js_binary recreate both app and node_module in js_image_layer rule
### What happened?

When making code changes to a js_binary target, rebuilding the js_image (with the app and node_modules packaged by js_image_layer) takes a while because both node_modules.tar and app_layer.tar get recreated, even though node_module layer did not change at all.

### Version

Development (host) and target OS/architectures: linux

`WORKSPACE` or `MODULE.bazel` file: rules_js-1.29.2

Language(s) and/or frameworks involved: js

### How to reproduce

_No response_

### Any other information?

_No response_

thesayyn:
Note that with this change, nodejs binaries and fs-patches are put into the node_modules layer which is a less frequently changed layer therefore should improve the dev build round-trip.  At least 2x performance improvement is expected with this change.

Closes #1259

Note: For the next major release we should consider adding a new layer called `binaries` which is almost never rebuilt unless the nodejs version or the rules_js version is changed. 

### Type of change

### Test plan


alexeagle (approved):

js/private/js_image_layer.bzl
```diff
@@ -202,6 +205,37 @@ def _runfiles_dir(root, default_info):
...
+        progress_message = "JsImageLayer %{label}",
+        env = {
+            "BAZEL_BINDIR": ".",
+        },
+    )
+
+    return output
+
+def _should_be_in_node_modules_layer(destination, file):
+    is_node = glob_match("**/*/bin/nodejs/**", destination) and file.owner.workspace_name != ""
```
jbedard, line 234: Avoid globs by checking `file.owner.workspace_name` first?
Avoid globs by short-circuiting?

🤷

jbedard (approved):

Issue #1279, opened 2023-09-20
[Bug]: Bazel managed pnpm can't resolve pnpm-lock.yml file merge conflicts


KatSick:
### What happened?

During merge conflicts, in `package.json` files you may end up with `pnpm-lock.yml` file state where there is merge conflicts in `pnpm-lock.yml`.

`pnpm` already have mechanism to resolve this conflicts https://pnpm.io/git#merge-conflicts, however, it can't be used if you run `pnpm` from the `bazel` (which as I understand is the recommended way of working with pnpm under rules_js) like:

```bash
bazel run -- @pnpm//:pnpm --dir $PWD "$@" install
```
command above will just crash, since Bazel can't read the pnpm lock file in the first place

### Version

bazel version
Bazelisk version: development
Build label: 6.2.1
Build target: bazel-out/darwin_arm64-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Fri Jun 2 17:00:10 2023 (1685725210)
Build timestamp: 1685725210
Build timestamp as int: 1685725210

rules_js: 1.31.0

### How to reproduce

Use pnpm-lock file from https://github.com/pnpm/pnpm/blob/37fffbefa5a9136f2e189c01a5edf3c00ac48018/packages/supi/test/lockfile.ts#L1237

Run `bazel run  -- @pnpm//:pnpm --dir $PWD "$@" install` (see crash)
Run local pnpm `pnpm install` (see successfully resolved conflict)

jbedard:
Fixed in https://github.com/aspect-build/rules_js/pull/1287.

We could potentially *partially* parse the yaml instead of the parser returning nothing which would align with pnpm more, but the crash is fixed.

PR #1287, opened 2023-09-25
fix: do not crash when parsing invalid yaml

jbedard:
See https://github.com/aspect-build/rules_js/issues/1279

Today, rules_js crashes when parsing invalid YAML, such as when there are git conflicts. If a developer uses the rules_js version of pnpm, it will crash when there are merge conflicts (because the repo rule parses the YAML?), unlike when using pnpm outside bazel.

Pnpm has some interesting [merge conflict parsing](https://github.com/pnpm/pnpm/blob/37fffbefa5a9136f2e189c01a5edf3c00ac48018/packages/supi/test/lockfile.ts#L1237). This change just outputs an error and prevents it from crashing.

---

### Type of change

### Test plan

- New test cases added

### DIFF
--- a/npm/private/yaml.bzl
+++ b/npm/private/yaml.bzl
@@ -31,7 +31,7 @@ def parse(yaml):
     """
     yaml = _normalize_yaml(yaml)
 
-    starlark = {"result": None}
+    starlark = {"result": None, "error": None}
     stack = []
     stack.append(_new_CONSUME_SPACE(indent = ""))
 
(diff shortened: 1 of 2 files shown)


npm/private/yaml.bzl
```diff
@@ -55,6 +55,10 @@ def parse(yaml):
             msg = "Unknown state {}".format(state["id"])
             fail(msg)
 
+    if len(starlark["errors"]) > 0:
+        print("YAML parse error(s): ", starlark["errors"])
+        return None
```
kormide, line 60: Do we handle the `None` case where needed in rules_js?

kormide (approved):
👍

npm/private/yaml.bzl
jbedard: We already return `None` in a few cases and the pnpm-lock logic treats it as an empty lockfile. So now an invalid lockfile outputs an error and treats it as empty, instead of crashing.

I'm not sure if that's the best solution but it's better then crashing? 😅

gregmagolan (approved):

PR #1294, opened 2023-09-27
lifecycle_hooks_envs prefixed with npm_config_ should get passed to gyp/lifecycle hook

sfc-gh-ptabor:
TLDR; Without this PR `lifecycle_hooks_envs` cannot be used to configure gyp mirror for download of pre-build artifacts.

     We need to explicitly pass `npm_config_` prefixed env-variables as configuration to the lifecycle hook (or gyp).
     1. rules_js allow to provide per action_type environment using `lifecycle_hooks_envs`.
     2. One of the important use-cases is to able provide mirror where prebuild binaries are stored:
        (see: https://github.com/mapbox/node-pre-gyp#download-binary-files-from-a-mirror
        by {module_name}_binary_host_mirror)
     3. Such flags are taken (by gyp) from environment variables:
        https://github.com/mapbox/node-pre-gyp/blob/a74f5e367c0d71033620aa0112e7baf7f3515b9d/lib/util/versioning.js#L316
     4. Unfortunetely pnpm/lifecycle drops all npm_ prefixed env-variables prior to calling lifecycle hook:
          https://github.com/pnpm/npm-lifecycle/blob/99ac0429025bdf1303879723d3fbd57c585ae8a1/index.js#L351
        and later recreates it based on explicitly given config:
          https://github.com/pnpm/npm-lifecycle/blob/99ac0429025bdf1303879723d3fbd57c585ae8a1/index.js#L408
     5. So we need to perform reversed process: generate rawConfig based on env-variables to preserve them.

### Type of change

- Bug fix / New feature or functionality (change which adds functionality)


sfc-gh-ptabor:
@gregmagolan -> hi. Queuing potentially interesting PR. Thank you.

sfc-gh-ptabor:
@alexeagle: Could you, please, take a look or delegate ?

alexeagle (approved):
Thanks! The comment is helpful

PR #1302, opened 2023-09-29
feat: add package_visibility attribute to npm_translate_lock and npm_import

gregmagolan:
Request from a consulting client.

Since link targets are generated by `npm_translate_lock` and `npm_import` repository rules, customizable visibility for link targets must be an `npm_translate_lock` and `npm_import` attributes. There is no way to set a visibility on an `npm_package` target that could be propagated to the visibility of a link target that uses that `npm_package` target via the virtual store link.

---

### Type of change

### Test plan

- New test cases added


npm/private/test/BUILD.bazel
```diff
@@ -50,30 +53,11 @@ write_source_files(
...
-        ":node_modules/regl",
-        ":node_modules/puppeteer",  # has a bin entry in the transitive closure with two segments: @puppeteer/browsers in https://unpkg.com/@puppeteer/browsers@0.5.0/package.json
-        ":node_modules/segfault-handler",  # segfault-handler has a node-gyp install step
-        ":node_modules/semver-first-satisfied",
-        ":node_modules/syncpack",
-        ":node_modules/typescript",
-        ":node_modules/webpack-bundle-analyzer",
-        # intentionally don't include :node_modules/unused
+        # @kubernetes/client-node: has a "prepare" lifecycle hook that runs tsc
+        # lodash: brought in as a vendored .tgz file: vendored/lodash-4.17.21.tgz
```
jbedard, line 57: Can we no longer directly reference the important ones these comments are for?

npm/private/test/links_defs_checked.bzl
```diff
@@ -0,0 +1,258 @@
...
+    # package directory for use in $(execpath) and $(rootpath)
+    native.filegroup(
+        name = "{}/dir".format(name),
+        srcs = [":{}".format(name)],
+        output_group = _utils.package_directory_output_group,
+        visibility = ["//visibility:public"],
+        tags = ["manual"],
+    )
+
+    return [":{}".format(name)] if True else []
```
jbedard, line 212: Where does the `if True else []` come from?

npm/private/npm_import.bzl
```diff
@@ -229,19 +226,18 @@ def npm_link_imported_package_store(
         name = "{{}}/dir".format(name),
         srcs = [":{{}}".format(name)],
         output_group = _utils.package_directory_output_group,
-        visibility = visibility,
+        visibility = {link_visibility},
         tags = ["manual"],
     )
 
-    return ":{{}}".format(name)
+    return [":{{}}".format(name)] if {public_visibility} else []
```
jbedard, line 233: Oh, this is the spot with the `if True`. I guess if this is a template then it's a bit of a hassle to avoid generating `if True else []`?

npm/private/test/links_defs_checked.bzl
gregmagolan: Looks like you found it below

npm/private/test/links_defs_checked.bzl
```diff
@@ -0,0 +1,258 @@
+"@generated by @aspect_rules_js//npm/private:npm_import.bzl for npm package rollup@2.70.2"
```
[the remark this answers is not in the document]
gregmagolan: Nice catch

PR #1290, opened 2023-09-26
chore(deps): update pnpm versions

vkbansal-rubrik:
---

### Type of change
- Chore

**For changes visible to end-users**

- update pnpm versions

PR #1306, opened 2023-10-04
Fix #1301: Include the entire pnpm package in @pnpm//:pnpm

### LINKED ISSUE #1301: [Bug]: pnpm 8.6.7 doesn't work when invoked through @pnpm//:pnpm js_binary
### What happened?

On rules_js v1.32.5, support for pnpm 8.6.7 was [added](https://github.com/aspect-build/rules_js/pull/1290), but it doesn't work. There appear to be a couple of different failure modes:

```shell
bazel run @pnpm//:pnpm install -- --dir=$PWD
INFO: Build completed successfully, 2 total actions
Scope: all 2 workspace projects
 MODULE_NOT_FOUND  Cannot find module '/Users/mhill/bazel/3c3e59429c86c7585f0ffa57350a75ac/execroot/web/bazel-out/darwin_arm64-fastbuild/bin/external/pnpm/pnpm.sh.runfiles/pnpm/package/dist/worker.js'

Error: Cannot find module '/Users/mhill/bazel/3c3e59429c86c7585f0ffa57350a75ac/execroot/web/bazel-out/darwin_arm64-fastbuild/bin/external/pnpm/pnpm.sh.runfiles/pnpm/package/dist/worker.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)
    at Module._load (node:internal/modules/cjs/loader:922:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at MessagePort.<anonymous> (node:internal/main/worker_thread:164:24)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:741:20)
    at exports.emitMessage (node:internal/per_context/messageport:23:28)
```

```shell
bazel run @pnpm//:pnpm install -- --dir=$PWD
INFO: Build completed successfully, 1 total action
Scope: all 2 workspace projects
Lockfile is up to date, resolution step is skipped
Packages: +3085
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 ERROR  Worker pnpm#7 exited with code 1

pnpm: Worker pnpm#7 exited with code 1
    at Worker.<anonymous> (/Users/mhill/bazel/3c3e59429c86c7585f0ffa57350a75ac/execroot/web/bazel-out/darwin_arm64-fastbuild/bin/external/pnpm/pnpm.sh.runfiles/pnpm/package/dist/pnpm.cjs:81745:27)
    at Worker.emit (node:events:514:28)
    at [kOnExit] (node:internal/worker:289:10)
    at Worker.<computed>.onexit (node:internal/worker:209:20)
Progress: resolved 3085, reused 0, downloaded 0, added 0
```

It looks like at least one of these is caused by the fact that the pnpm js_binary runfiles only includes `package/dist/pnpm.cjs` rather than the whole package:

https://github.com/aspect-build/rules_js/blob/main/npm/private/pnpm_repository.bzl#L40

### Version

Output of `bazel --version`:

6.3.1

http_archive(
        name = "aspect_rules_js",
        sha256 = "8563dd34e75ce989fd2e45ad6dd527c6e6fafacb2dcf6f7dd36ca6e7fb0ca670",
        strip_prefix = "rules_js-1.32.5",
        url = "https://github.com/aspect-build/rules_js/releases/download/v1.32.5/rules_js-v1.32.5.tar.gz",
    )

### How to reproduce

```shell
Delete node_modules
Invoke pnpm through bazel
```

### Any other information?

_No response_

gzm0:
### Type of change

### Test plan

Tested manually:

- In rules_js:

    ```sh
    rules_js/examples/js_binary$ bazel run @pnpm -- install --dir $PWD -D -E react-router --lockfile-only
    ```

- In an external repository affected by #1301 using `local_repository`


gzm0:
@matthewjh seems your analysis was correct.

gzm0:
- Windows failure looks like flaky infra (cannot download Node.js)
- bazel test //... seems to fail. Investigating.

gzm0:
I cannot reproduce the issue in CI locally.

```sh
bazel run //npm/private/test:write_npm_translate_lock
```

Doesn't change anything.

```sh
 bazel test //npm/private/...
```

Passes, :-/

gzm0:
> Do any of the existing example/* or e2e/* tests fail if pnpm is upgraded?

No (otherwise we would have caught this earlier).

> If not maybe you can add your test snippet in a test.sh file there?

It is not entirely clear when `pnpm` actually tries to include the other files. Experimentation suggests "when the workload is large enough" (given the file it is including is called `worker.js` it stands to reason there is some heuristics to parallelize).

I can add a test installing a lot of packages (or increase the number of packages on an existing test), but of course that's at the risk of being slow and flaky.

If you feel we should build up some more understanding of the triggering condition, I'm happy to dig through some pnpm code.

jbedard:
> No (otherwise we would have caught this earlier).

I was assuming it wasn't caught just because the latest pnpm version isn't tested?

If we're unsure when pnpm does this then I don't think it's worth add a test that "might" test it.

Do you know any flags to disable workers? I wonder if we'd want to do that by default when rules_js runs pnpm 🤔

matthewjh:
Yeah, personally I don't think this is worth testing given this pathway is invoked by pnpm only under very specific conditions. Include the whole pnpm package in the js_binary and be done with it, IMO. :) 

@jbedard what would be the argument for disabling workers?

gzm0:
FWIW: The whole worker thing seems to have originated here:

- https://github.com/pnpm/pnpm/pull/6850
- https://github.com/pnpm/pnpm/commit/c8a6ea8ae962ed61a8f19dfbdc6dc4d72eb9a51f (refactor)
- https://github.com/pnpm/pnpm/pull/7022
- https://github.com/pnpm/pnpm/pull/7069

Superficial review of the code suggests that there is no way to turn workers off (the [PNPM_WORKERS env variable](https://github.com/pnpm/pnpm/blob/8605ffa1207867d472f589775717483dde722282/worker/src/index.ts#L27) seems to override worker count, but that doesn't change the inclusion mechanism).

However, this might give us some clues to when workers are required (you'll notice the last paragraph is quite inconsistent with "it sometimes needs workers").

gzm0:
Ugh, sorry about the noise. All these PRs are from pnpm 8.**7**.x. But the upgrade to 8.**6**.x triggered the issue.

gzm0:
Aaaaand more noise: https://github.com/aspect-build/rules_js/pull/1290 upgraded pnpm latest from 8.6.7 to 8.7.6. The version number in #1301 is probably just a typo :shrug:

gzm0:
@jbedard 

> I was assuming it wasn't caught just because the latest pnpm version isn't tested?

You are of course absolutely right. Sorry, I just assumed that:
- A pnpm invocation is tested.
- This is done at the newest pnpm version.

> If we're unsure when pnpm does this then I don't think it's worth add a test that "might" test it.

I think at this point, I understand the conditions. I'll try to add an e2e test (IIUC `example` will not work due to the "old" pnpm version in the root workspace).

> Do you know any flags to disable workers? I wonder if we'd want to do that by default when rules_js runs pnpm 🤔

Looks like no (see my comment above).

gzm0:
I just stumbled upon this failure on `main`:

https://github.com/aspect-build/rules_js/actions/runs/6475850009/job/17583680376

gzm0:
I have added an e2e test and checked manually that it fails, if I revert the fix.

There seems to be still one failure, but it also fails on `main`: https://github.com/aspect-build/rules_js/actions/runs/6475850009/job/17583702216.

In fact, it seems multiple tests fail on main due to what we're fixing with this PR (and they seem to recover in this PR):
- https://github.com/aspect-build/rules_js/actions/runs/6475850009/job/17583680376
- https://github.com/aspect-build/rules_js/actions/runs/6475850009/job/17583697300
- https://github.com/aspect-build/rules_js/actions/runs/6475850009/job/17583697856
- https://github.com/aspect-build/rules_js/actions/runs/6475850009/job/17583703923
- https://github.com/aspect-build/rules_js/actions/runs/6475850009/job/17583711061
- https://github.com/aspect-build/rules_js/actions/runs/6475850009/job/17583711689
- https://github.com/aspect-build/rules_js/actions/runs/6475850009/job/17583716026
- https://github.com/aspect-build/rules_js/actions/runs/6475850009/job/17583725053
- https://github.com/aspect-build/rules_js/actions/runs/6475850009/job/17583725289

jbedard:
Awesome, thanks for the investigation and all the info! 👍 

There are indeed less failures then before your change, I think the one remaining is unrelated to this change.

jbedard (approved):

PR #1311, opened 2023-10-10
chore: bump bazel-lib to 2.0

alexeagle:
Note: users might still bring 1.x version of bazel-lib, and we permit that to the extent that it still works.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -13,7 +13,7 @@ rules_js_dependencies()
 
 load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies", "register_coreutils_toolchains", "register_jq_toolchains")
 
-aspect_bazel_lib_dependencies(override_local_config_platform = True)
+aspect_bazel_lib_dependencies()
 
 register_jq_toolchains()
 
--- a/js/repositories.bzl
+++ b/js/repositories.bzl
@@ -9,14 +9,14 @@ load("//js/private:maybe.bzl", http_archive = "maybe_http_archive")
 def rules_js_dependencies():
     http_archive(
         name = "bazel_skylib",
-        sha256 = "66ffd9315665bfaafc96b52278f57c7e2dd09f5ede279ea6d39b2be471e7e3aa",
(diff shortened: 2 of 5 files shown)


WORKSPACE
```diff
@@ -13,7 +13,7 @@ rules_js_dependencies()
 
 load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies", "register_coreutils_toolchains", "register_jq_toolchains")
 
-aspect_bazel_lib_dependencies(override_local_config_platform = True)
+aspect_bazel_lib_dependencies()
```
kormide, line 16: Do you need to also set up any toolchains?

alexeagle: Maybe? It looks like we'll also have to drop testing with Bazel 5

kormide (approved):

js/repositories.bzl
```diff
@@ -9,14 +9,14 @@ load("//js/private:maybe.bzl", http_archive = "maybe_http_archive")
...
+        sha256 = "b8a1527901774180afc798aeb28c4634bdccf19c4d98e7bdd1ce79d1fe9aaad7",
+        urls = ["https://github.com/bazelbuild/bazel-skylib/releases/download/1.4.1/bazel-skylib-1.4.1.tar.gz"],
     )
 
     http_archive(
         name = "rules_nodejs",
-        sha256 = "8fc8e300cb67b89ceebd5b8ba6896ff273c84f6099fc88d23f24e7102319d8fd",
-        urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.8.4/rules_nodejs-core-5.8.4.tar.gz"],
+        sha256 = "764a3b3757bb8c3c6a02ba3344731a3d71e558220adcb0cf7e43c9bba2c37ba8",
+        urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.8.2/rules_nodejs-core-5.8.2.tar.gz"],
```
Topher-the-Geek, line 19: Was there a reason to downgrade? It looks like v1.33.0 is using rules_nodejs 5.8.4, and then v1.33.1 rolled back to rules_nodejs 5.8.2 (and lost NodJS 18.14+ in the process)

alexeagle: Oops! This definitely looks like a mistake.

PR #1317, opened 2023-10-16
java.io.IOException when vendored tarfile uses path that's not relative to workspace root

jfirebaugh:
This is a bug report in the form of a failing test.

The existing `e2e/vendored_tarfile` tests dependencies of the form `file:<path>.tgz`. The path it uses happens to be relative to the workspace root. Bazel supports such `file:` URLs. Bazel does not support `file:` URLs with paths that are relative to any other location, as may occur when `package.json` is not at the workspace root. Attempting to use such a `package.json` with `npm_translate_lock` produces a java.io.IOException attempting to download the file.

To reproduce, run `bazel clean --expunge && bazel test --repository_cache= //...` in `e2e/vendored_tarfile`. The expunge and `--repository_cache=` ensure that no cached repository or download artifact is used.

Expected result: completes successfully.

Actual result: 

```
ERROR: An error occurred during the fetch of repository 'npm__lodash__4.17.21':
	File "/private/var/tmp/_bazel_john/d601fbcd5200d602e76906b7acaddff2/external/aspect_rules_js/npm/private/npm_import.bzl", line 494, column 38, in _npm_import_rule_impl
		_download_and_extract_archive(rctx)
	File "/private/var/tmp/_bazel_john/d601fbcd5200d602e76906b7acaddff2/external/aspect_rules_js/npm/private/npm_import.bzl", line 452, column 18, in _download_and_extract_archive
		rctx.download(
Error in download: java.io.IOException: Error downloading [file:../a/lodash-4.17.21.tgz] to /private/var/tmp/_bazel_john/d601fbcd5200d602e76906b7acaddff2/external/npm__lodash__4.17.21/package.tgz: ../a/lodash-4.17.21.tgz (No such file or directory)
```

The generated `repositories.bzl` in the `npm` repository contains:

```
        url = "file:../a/lodash-4.17.21.tgz",
```

Instead, this URL should be relativized to the workspace root, i.e.

```
        url = "file:a/lodash-4.17.21.tgz",
```


jbedard:
Can you reproduce it without removing the existing test case? Probably just add a second pnpm-lock in a subdir?

jfirebaugh:
If there are two lockfiles that both reference lodash-4.17.21.tgz, then the bug might not reproduce reliably due to caching effects. Specifically, if the download cache already contains the tarball, then Bazel won't try to download it from the `file:../a/lodash-4.17.21.tgz` URL. That's why my repro instructions include `bazel clean --expunge` and `--repository_cache=`.

I don't necessarily expect this PR to be merged, I just wanted to demonstrate the bug.

jbedard:
This appears to be fixed. Let me know if you ever see similar again though

PR #1338, opened 2023-11-01
fix: propogate npm package store deps from srcs in js_library

gregmagolan:
Users may add npm deps to `srcs` and the generated catch-all `:node_modules` targets has npm packages in `srcs`.

---

### Type of change

### Test plan

- New test cases added

TODO


gregmagolan:
Dup of #1650 which was landed already

PR #1351, opened 2023-11-10
feat: add BAZEL env variable to js_binary

gregmagolan:
---

### Type of change

### Test plan

- New test cases added


alexeagle (commented):
Did you search for a conventional variable set in other languages? What do ppl do there?

gregmagolan:
I haven't seen the `bazel run` case handled in other ecosystems. Cases I've seen, users are keying off of `BAZEL_TEST` for test targets or setting explicit envs such as `BAZEL` on the target to detect if under Bazel.

jbedard:
Bazel itself sets `BAZEL_TEST` though, not rules_js? Should we avoid setting anything without a `JS` pre/suffix? How about `RULES_JS` or `JS_TEST|RUN|BUILD` etc?

gregmagolan:
> Bazel itself sets `BAZEL_TEST` though, not rules_js? Should we avoid setting anything without a `JS` pre/suffix? How about `RULES_JS` or `JS_TEST|RUN|BUILD` etc?

Yes, tho `BAZEL_TEST` is only set when you run `bazel test` so there is a gap for users that run programs with `bazel run` to be able to detect in an easy and ergonomic way they are underly Bazel. An env var such as `JS_TEST` doesn't read well in end-user code since it isn't obvious that you're checking for bazel that way.

There is precedent for setting `BAZEL_*` var in rules_js already in js_run_binary:

```
    fixed_env = {
        "BAZEL_BINDIR": "$(BINDIR)",
        "BAZEL_BUILD_FILE_PATH": "$(BUILD_FILE_PATH)",
        "BAZEL_COMPILATION_MODE": "$(COMPILATION_MODE)",
        "BAZEL_PACKAGE": native.package_name(),
        "BAZEL_TARGET_CPU": "$(TARGET_CPU)",
        "BAZEL_TARGET_NAME": name,
        "BAZEL_TARGET": "$(TARGET)",
        "BAZEL_WORKSPACE": "$(WORKSPACE)",
    }
```

alexeagle:
I'm just nitpicking this because every added variable has the potential to collide with something the user already set, and shorter is more likely.

gregmagolan:
> I'm just nitpicking this because every added variable has the potential to collide with something the user already set, and shorter is more likely.

I do check if it is already set and don't set override it if it is. Perhaps I should create an issue on bazelbuild/bazel so that bazel itself sets `BAZEL` to `1` in all cases including `bazel run`?

alexeagle:
A bit related: https://github.com/bazelbuild/bazel/issues/15470

PR #1354, opened 2023-11-15
fix: add coreutils toolchain to rules that use copy file actions

kormide:
In bazel-lib 2.0 the copy actions now require the coreutils toolchains, so for compatibility with 2.0 we need to declare this.

### Type of change

### Test plan

- Tested locally by swapping the bazel-lib dep with 2.0 and running `test //...`.

### DIFF
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -19,6 +19,7 @@ load("@aspect_bazel_lib//lib:windows_utils.bzl", "create_windows_native_launcher
 load("@aspect_bazel_lib//lib:expand_make_vars.bzl", "expand_locations", "expand_variables")
 load("@aspect_bazel_lib//lib:directory_path.bzl", "DirectoryPathInfo")
 load("@aspect_bazel_lib//lib:utils.bzl", "is_bazel_6_or_greater")
+load("@aspect_bazel_lib//tools:version.bzl", _BAZEL_LIB_VERSION = "VERSION")
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load(":js_helpers.bzl", "LOG_LEVELS", "envs_for_log_level", "gather_runfiles")
 load(":bash.bzl", "BASH_INITIALIZE_RUNFILES")
@@ -604,7 +605,9 @@ js_binary_lib = struct(
         # TODO: on Windows this toolchain is never referenced
         "@bazel_tools//tools/sh:toolchain_type",
         "@rules_nodejs//nodejs:toolchain_type",
-    ],
+    ] + ([
+        "@aspect_bazel_lib//lib:coreutils_toolchain_type",
+    ] if _BAZEL_LIB_VERSION.startswith("2") else []),
 )
 
 js_binary = rule(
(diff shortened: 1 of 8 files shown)


alexeagle (approved):

js/version-visibility.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/tools/BUILD.bazel b/tools/BUILD.bazel
+index 6f86f0a..3058f94 100644
+--- a/tools/BUILD.bazel
++++ b/tools/BUILD.bazel
+@@ -17,7 +17,7 @@ bzl_library(
+ bzl_library(
+     name = "version",
+     srcs = [":version.bzl"],
+-    visibility = ["//lib/private/docs:__pkg__"],
++    visibility = ["//visibility:public"],
```
kormide, line 10: This is already fixed in 2.0 but it's not in the latest 1.x release. We could do another release there rather than include this patch if you prefer.

kormide: So an issue is that `versions.bzl` is in a package that has go deps, which makes this more difficult.

js/private/js_binary.bzl
```diff
@@ -604,7 +605,9 @@ js_binary_lib = struct(
         # TODO: on Windows this toolchain is never referenced
         "@bazel_tools//tools/sh:toolchain_type",
         "@rules_nodejs//nodejs:toolchain_type",
-    ],
+    ] + ([
+        "@aspect_bazel_lib//lib:coreutils_toolchain_type",
+    ] if _BAZEL_LIB_VERSION.startswith("2") else []),
```
alexeagle, line 610: There will be version 3 at some point, probably worth parsing a bit more like digits before first dot

js/private/js_library.bzl
```diff
@@ -25,6 +25,7 @@ js_library(
 
 load(":js_info.bzl", "JsInfo", "js_info")
 load(":js_helpers.bzl", "DOWNSTREAM_LINKED_NPM_DEPS_DOCSTRING", "JS_LIBRARY_DATA_ATTR", "copy_js_file_to_bin_action", "gather_npm_linked_packages", "gather_npm_package_store_deps", "gather_runfiles", "gather_transitive_declarations", "gather_transitive_sources")
+load("@aspect_bazel_lib//tools/version:version.bzl", _BAZEL_LIB_VERSION = "VERSION")
```
alexeagle, line 28: why do you have to make the version-visibility.patch? Can just load from the private location, and deal with stardoc by including some .bzl srcs maybe?

alexeagle (commented):
Is this PR mutually exclusive from https://github.com/aspect-build/bazel-lib/pull/655 ?

kormide:
> Is this PR mutually exclusive from [aspect-build/bazel-lib#655](https://github.com/aspect-build/bazel-lib/pull/655) ?

Yes. It's an attempt to not need that revert, but not successful yet.

js/private/js_library.bzl
kormide: Yeah I think that can work. The problem that I'm dealing with now though is that under bzlmod only, depending on `@aspect_bazel_lib//tools:version.bzl"`, it's expecting bazel-lib to have a dependency on `@io_bazel_rules_go` because it's loaded in `tools/release.bzl`.

kormide: Though oddly all the e2es seem to now be passing.

PR #1355, opened 2023-11-16
feat: import toolchains to use for copy actions

kormide:
Support bazel-lib 1.x and 2.x. Please review the first commit only.

### Type of change

### Test plan

- Second commit (to be removed) tests against bazel-lib 1.x changes.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -224,12 +224,12 @@ jobs:
                 include:
                     - os: windows-latest
                       config: local
-                      bazelversion: 6.2.0
+                      bazelversion: 6.3.2
                       folder: e2e/bzlmod
                       bzlmodEnabled: true
                     - os: macos-latest
                       config: local
-                      bazelversion: 6.2.0
+                      bazelversion: 6.3.2
                       folder: e2e/bzlmod
                       bzlmodEnabled: true
                     - os: ubuntu-latest
--- a/examples/js_library/two/BUILD.bazel
+++ b/examples/js_library/two/BUILD.bazel
@@ -1,5 +1,4 @@
 load(":tsc.bzl", "tsc")
-load("@aspect_bazel_lib//lib:docs.bzl", "stardoc_with_diff_test", "update_docs")
 load("@aspect_rules_js//js:defs.bzl", "js_test")
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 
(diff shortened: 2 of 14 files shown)


.github/workflows/ci.yaml
```diff
@@ -224,12 +224,12 @@ jobs:
                 include:
                     - os: windows-latest
                       config: local
-                      bazelversion: 6.2.0
+                      bazelversion: 6.3.2
```
kormide, line 227: 6.2.0 seems to only be used for windows and osx. Update to 6.3.2, which also allows overriding repos in an e2e that the root overrides (bazel-lib in this case, for testing).

examples/js_library/two/BUILD.bazel
```diff
@@ -32,11 +31,15 @@ bzl_library(
...
+# Remove comments of the released version of bazel-lib succeeds.
+#
+# stardoc_with_diff_test(
+#     name = "tsc-docs",
+#     bzl_library_target = ":tsc",
+# )
+#
+# update_docs(
+#     name = "docs",
+# )
```
kormide, line 45: This can actually go into the second commit which I will remove. The issue only surfaced when using overridden versions which I was doing for testing.

kormide: Nope, this still needs to be commented out.

.github/workflows/ci.yaml
alexeagle: Wow you've learned a lot during this fire 😂

examples/js_library/two/BUILD.bazel
alexeagle: Can you file an issue so we don't forget to fix it? The generated doc will start drifting out of date

examples/macro/BUILD.bazel
```diff
@@ -22,11 +21,15 @@ bzl_library(
...
-stardoc_with_diff_test(
-    name = "mocha-docs",
-    bzl_library_target = ":mocha",
-)
-
-update_docs(
-    name = "docs",
-)
+# TODO(derek): stardoc doesn't work under bzlmod with these
+# targets when working against an overridden version of bazel-lib.
```
alexeagle, line 25: Can we just not run stardoc under bzlmod?

alexeagle (approved):

examples/js_library/two/BUILD.bazel
kormide: Yeah I'll post an issue.

examples/macro/BUILD.bazel
kormide: Only these targets, for some reason. The main docs are fine.

examples/js_library/two/BUILD.bazel
kormide: https://github.com/aspect-build/rules_js/issues/1362

PR #1366, opened 2023-11-19
chore: add a macro for duplicating an e2e for bazel-lib 2.x

kormide:
⚠️  EXPERIMENTAL. I'm not sold on this approach but wanted to get some feedback.

This adds a macro that will take any e2e, swap bazel-lib 1.x for 2.x, and write it to the source tree. The goal here is to make it easy to run the same e2e against bazel-lib 2.x while reducing the test maintenance burden of keeping both version of the test identical.

It uses the buildozer toolchain to modify version in the WORKSPACE and MODULE files of e2es.

I don't think we need to test _every_ e2e with bazel-lib 2.x since our matrix is already large. This just adds a couple.

Related to https://github.com/aspect-build/rules_js/issues/1363.

### Type of change

### Test plan


.bazelignore
```diff
@@ -1,4 +1,38 @@
...
+e2e/rules_foo
+e2e/update_pnpm_lock
+e2e/update_pnpm_lock_with_import
+e2e/vendored_node
+e2e/vendored_tarfile
+e2e/verify_patches
+e2e/webpack_devserver
+e2e/webpack_devserver_esm
+e2e/worker
+e2e/workspace
```
kormide, line 35: We can no longer just ignore `e2e` because the `write_source_files` target which writes the duplicate e2e with bazel-lib 2.x will puke on the analysis phase test where it globs to check for existence.

However I added some additional tests to ensure all e2es are added here.

hidden-files.patch
```diff
@@ -0,0 +1,15 @@
...
+     echo "Copying directory $in to $out in $PWD"
+     rm -Rf "$out"/*
+     mkdir -p "$out"
+-    cp -fRL "$in"/* "$out"
+-    chmod -R ug+w "$out"/*
++    cp -fRL "$in"/. "$out"
++    chmod -R ug+w "$out"/.
+     {executable_dir}
+ fi
+ """.format(
```
kormide, line 15: I found a bug in `write_source_files` while working on this. It doesn't copy over hidden files.

e2e/pnpm_workspace/WORKSPACE
```diff
@@ -1,3 +1,18 @@
...
+    sha256 = "262e3d6693cdc16dd43880785cdae13c64e6a3f63f75b1993c716295093d117f",
+    strip_prefix = "bazel-lib-1.38.1",
+    url = "https://github.com/aspect-build/bazel-lib/releases/download/v1.38.1/bazel-lib-v1.38.1.tar.gz",
+)
+
+load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies", "register_coreutils_toolchains")
+
+aspect_bazel_lib_dependencies()
+
+register_coreutils_toolchains()
```
kormide, line 14: I shouldn't have to add this to the bazel-lib 1.x e2e, but it is needed for the bazel-lib 2.x version of the e2e. So flipping the version isn't always the only thing needed to test both. Perhaps this is a good enough reason not to go with this approach..

kormide: I could add an input to the e2e copying macro that takes buildozer commands to apply to the WORKSPACE and the 2.0 version could be modified that way..

hidden-files.patch
kormide: The patch itself has a bug though where it doesn't preserve executable bits.

kormide:
My overall take is that it's simpler to just make a couple copies of e2es manually and point them to use bazel-lib 2.x This seems like overkill.

.github/workflows/ci.yaml
```diff
@@ -279,7 +283,7 @@ jobs:
               # still bootstraps Aspect CLI from configuration in .bazeliskrc. Aspect CLI will
               # then use .bazelversion to determine which Bazel version to use
               run: |
-                  echo "${{ matrix.bazelversion }}" > .bazelversion
+                  find . -name ".bazelversion" -type f -exec echo "${{ matrix.bazelversion }}" > {}
```
kormide, line 286: The `.bazelversion` symlinks in bazel-lib 2 e2es that are written back to the source tree are not preserved. Find all non-symlink `.bazelversion` files and replace them.

alexeagle:
Yeah this does feel like a bunch of complexity for a case that we don't care that much about, and in 6mo or something we'll expect most users to be on bazel-lib 2.0 simply because there will be some rulesets that take that as a lower-bound (e.g. rules_distroless to get the tar toolchain)

I think we should just mix-and-match for now, most e2e on bazel-lib 2 and a couple on bazel-lib 1. We won't have comprehensive test coverage, so there's some risk, but it will be a smoke test to catch some obvious things. WDYT?

kormide:
> Yeah this does feel like a bunch of complexity for a case that we don't care that much about, and in 6mo or something we'll expect most users to be on bazel-lib 2.0 simply because there will be some rulesets that take that as a lower-bound (e.g. rules_distroless to get the tar toolchain)
> 
> I think we should just mix-and-match for now, most e2e on bazel-lib 2 and a couple on bazel-lib 1. We won't have comprehensive test coverage, so there's some risk, but it will be a smoke test to catch some obvious things. WDYT?

Yeah I agree. I'll just select a few and flip them to use bazel-lib 2.x.

kormide:
Closing in favour of https://github.com/aspect-build/rules_js/pull/1369.

Issue #1373, opened 2023-11-22
[Bug]: jq: error (at packages/api/account/package.json:28): Cannot index array with string "STABLE_SCM_TAG"


supercairos:
### What happened?

While using :
```
stamped_package_json(
    name = "package",
    stamp_var = "STABLE_SCM_TAG",
)
```

Target fails with error: 
`bazel build //packages/api/account:package --stamp`
jq: error (at packages/api/account/package.json:28): Cannot index array with string "STABLE_SCM_TAG"

### Version

Development (host) and target OS/architectures:
macOS (m1) - arm64

Output of `bazel --version`:
bazel 6.4.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
bazel-lib-2.0.1
rules_nodejs-6.0.2
rules_js-1.34.0
rules_ts-2.1.0

Language(s) and/or frameworks involved:
TS

supercairos:
Just saw that in the changelog of bazel-lib 2.0:
```
jq:

We now run jq version 1.7, which includes a Linux arm64 binary.
jq stamp variables are now of the form $stamp[0].MY_VARIABLE rather than $stamp.MY_VARIABLE.
```

PR #1374, opened 2023-11-22
fix: `stamped_package_json` following bazel-lib update to 2.0.x

supercairos:
Fix this issue: #1373


alexeagle (approved):

alexeagle:
Hm, this will be broken if you are still on Bazel-lib 1.x though. We should try to find a solution that can work with both.

supercairos:
How can we check the version of `bazel-lib` ?

We could do this if `bazel-lib.major > 2`

kormide:
> How can we check the version of `bazel-lib` ?
> 
> We could do this if `bazel-lib.major > 2`

I have a fix of this sort going in https://github.com/aspect-build/rules_js/pull/1376. I'll cut a release once it lands.

kormide:
Closing as this is fixed in v1.34.1.

PR #1376, opened 2023-11-22
fix: support stamped_package_json stamping with bazel-lib 2.x

kormide:
Fixes a breakage where stamping for this rule doesn't work under bazel-lib 2.x, which had some breaking changes when stamping using jq.

Closes https://github.com/aspect-build/rules_js/issues/1373

### Type of change

### Test plan


gregmagolan (approved):

gregmagolan:
Should make a note in the RELEASE NOTES when this change is in a release that WORKSPACE users will need to also upgrade to the latest bazel-lib 1.x or 2.x if they are explicitly pulling in a version in their WORKSPACE.

kormide:
Before landing this I want to add an e2e that tests the stamping with bazel-lib 2.x.

gregmagolan (approved):

PR #1377, opened 2023-11-24
Fix broken symlinks during js_run_devserver runs.

duhruh:
This fixes an issue I've been facing for a while now while trying to run the `js_run_devserver` target in any of the examples on a windows machine see https://github.com/bazelbuild/examples/issues/327

The core of the issue is that windows treats symlinks differently than other OS's, for better or worse. The nodejs `fs.symlink` needs to be aware, at least on widows machines, what kind of symlink it's attempting to create. Otherwise at least in my testing it will assume `file` which causes any `node_modules/$PACKAGE` to be symlinked to the run directory as a file, which in turn prevents node from finding anything.

My change here should only affect windows machines and will fallback to the default behavior for all other OS's.

### Type of change

### Test plan

To reproduce 
1. clone https://github.com/bazelbuild/examples 
2.  patch in the follow: we do this because the default `command = "node_modules/.bin/react-scripts",` just wont work on windows machines as that is generated from https://github.com/aspect-build/rules_js/blob/89e9065dc7ad3b456694dc2f3f37cfe4cb0b7778/npm/private/npm_link_package_store.bzl#L69  a different issue than the one i'm attempting to fix here though maybe with the addition of https://github.com/pnpm/pnpm/issues/5131 or a more sophisticated OS aware templating or even an option to the end user to eagerly fetch all bins could fix this.
 ```bazel
# frontend/react/BUILD.bazel
js_run_devserver(
    name = "start",
    args = [
        "node_modules/react-scripts/bin/react-scripts.js", "start",
    ],
    command = "C:\\Program Files\\nodejs\\node.exe", # OR WHERE EVER NODE IS FOR YOU
    chdir = package_name(),
    data = CRA_DEPS,
)
```
3. After patching in the above attempt to run `bazel run react:start` 
4. Here you should receive a failure, scroll up in the terminal and look for a line like
```powershell
Running 'C:\Program Files\nodejs\node.exe node_modules/react-scripts/bin/react-scripts.js start' in C:\Users\SOMEUSER\AppData\Local\Temp\js_run_devserver-SOMEHASH\aspect_rules_js\frontend
```
5. Open that directory and navigate to the `node_modules` you should see something like this
285073565-391e3bb5-435b-4781-8041-e2a7579cec91

Those packages are not valid symlinks, they do point to the correct locations, however from windows perspective and other processes trying to open those, they are invalid.

After reproducing the failure you should be able to patch in this PR with something like

```bazel
# frontend/MODULE.bazel
single_version_override(
    module_name = "aspect_rules_js",
    patch_strip = 1,
    patches = ["//:aspect_rules_js_1.32.2.patch"],
)
```

Then re-run the `bazel run react:start` this should  now succeed and viewing the run directory should now look similar to this:

285073516-808054a4-8a82-4029-ad9e-c23eff9e1cf5


js/private/js_run_devserver.mjs
```diff
@@ -54,6 +54,28 @@ function isNodeModulePath(srcPath) {
     return false;
 }
 
+// Returns the correct symlink type if we are building on windows. This doesn't appear to be an issue
+// on *nix based machines as native fs.symlink without type just works 
+// (see https://nodejs.org/api/fs.html#fssymlinktarget-path-type-callback). However on windows machines
+// we need to be more specific on whether or not we are attempting to symlink a file or directory.
+function getSymlinkArgs(src) {
```
jbedard, line 61: Can we put the fix into the existing `if (isNodeModulePath(src))` by any chance?

jbedard: Or have you seen the issue occur outside the `node_modules`?

duhruh: Well, I ran across this issue in the `node_modules` so that is the immediate thing that needs fixing. However, I don't think it's a good idea to limit the fix to `node_modules` only. Any time we call `fs.symlink` on windows machines we must specify the type otherwise we run the risk of this happening again for some other directory.

jbedard: Can you explain why we need the node_modules specific logic though instead of just `isDirectory`?

js/private/js_run_devserver.mjs
```diff
@@ -54,6 +54,28 @@ function isNodeModulePath(srcPath) {
...
+    const symlinkArgs = [];
+    const isWindows = os.platform() == 'win32';
+    if (!isWindows) {
+        return symlinkArgs;
+    }
+    const isDirectory = p => fs.lstatSync(p).isDirectory();
```
jbedard, line 67: Lets use the `fs.promises` version and `await` similar to other fs calls.

js/private/js_run_devserver.mjs
```diff
@@ -96,7 +118,8 @@ async function syncRecursive(src, dst, writePerm) {
                 // Intentionally synchronous; see comment on mkdirpSync
                 mkdirpSync(path.dirname(dst))
             }
-            await fs.promises.symlink(src, dst)
+            const symlinkArgs = getSymlinkArgs(src);
+            await fs.promises.symlink(src, dst, ...symlinkArgs)
```
jbedard, line 122: Can we change the `getSymlinkArgs` result to be a single value instead of array? What would the default be for non-windows?

jbedard:
This is most likely stale. Please rebase if this is still applicable and we can reopen.

PR #1378, opened 2023-11-25
WIP Use yq to parse yaml file

dzbarsky:
This reduces parsing of the Airtable lockfile from ~10s to <.5s. This clearly needs a bit of work, I wasn't fully sure how to get a toolchain usable inside repo rules or module extensions. The `Label(@@yq)` does work, but requires the userspace repo to declare the yq toolchain, which is undesirable. The homebrew yq was just a quick hack to validate that what we want here is fast.

### Type of change

### Test plan

- New test cases added


gregmagolan:
Thanks for the POC @dzbarsky . Closing this one in favor of https://github.com/aspect-build/rules_js/pull/1458 where I'll take this to completion.

PR #1379, opened 2023-11-28
test: pnpm without npm_translate_lock

jbedard:
Testing the use of the pnpm toolchain without using any pnpm workspaces.

### DIFF
--- a/e2e/pnpm_repo_install/MODULE.bazel
+++ b/e2e/pnpm_repo_install/MODULE.bazel
@@ -22,6 +22,15 @@ npm = use_extension("@aspect_rules_js//npm:extensions.bzl", "npm")
 npm.npm_translate_lock(
     name = "npm",
     pnpm_lock = "//:pnpm-lock.yaml",
+    pnpm_version = "8.6.7",
     verify_node_modules_ignored = "//:.bazelignore",
 )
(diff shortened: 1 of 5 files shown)


e2e/pnpm_repo_install/MODULE.bazel
```diff
@@ -18,10 +19,10 @@ use_repo(node, "nodejs_linux_arm64")
...
-    pnpm_lock = "//:pnpm-lock.yaml",
-    verify_node_modules_ignored = "//:.bazelignore",
+pnpm = use_extension("@aspect_rules_js//npm:extensions.bzl", "pnpm")
+pnpm.pnpm(
+    name = "pnpm_8",
+    pnpm_version = "8.6.7",
+    pnpm_version_integrity = "sha512-vRIWpD/L4phf9Bk2o/O2TDR8fFoJnpYrp2TKqTIZF/qZ2/rgL3qKXzHofHgbXsinwMoSEigz28sqk3pQ+yMEQQ==",
 )
-use_repo(npm, "npm")
+use_repo(pnpm, "pnpm_8")
```
alexeagle, line 28: I think we want to show that you can have your version of pnpm used by npm_translate_lock, so we want both of them in here, and use the same "pnpm" external repository name for both.

PR #1382, opened 2023-11-29
(docs) Added tip regarding unintentional parallelization

hjellek:
After struggling with slow and randomly failing CI builds after migrating our JS applications to Bazel, we finally nailed at least one of the culprits. Tools and plugins trying to run in parallel on top of Bazel parallelizing jobs made our CI runners fail as they ran out of resources (actual communication with the pods inside kubernetes failed due to IO throttling).

Figured it would be worth to give a tip to future devs stumbling onto similar problems.
Improvements and feedback welcome! 

---

### Type of change

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -137,3 +137,12 @@ eslint_bin.eslint_test(
 
 > NB: We plan to add support for the `.npmrc` `public-hoist-pattern` setting to `rules_js` in a future release.
 > For now, you must emulate public-hoist-pattern in `rules_js` using the `public_hoist_packages` attribute shown above.
+
+#### Slower than expected builds (especially in CI)
+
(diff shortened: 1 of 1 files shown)


jbedard:
Want to add a note about things like the rules_jest [run_in_band](https://docs.aspect.build/rulesets/aspect_rules_jest/docs/jest_test/#run_in_band)?

Just something simple such as:

> Some rulesets configure tools to take this into account such as the rules_jest default [run_in_band](https://docs.aspect.build/rulesets/aspect_rules_jest/docs/jest_test/#run_in_band), while other tools (especially those without dedicated rulesets) may need to be configured manually.

The webpack example would then be the next paragraph?

hjellek:
Absolutely, I'll add that.

jbedard (approved):

PR #1392, opened 2023-12-09
chore: run Aspect Workflows setup step on small instances

gregmagolan:
Same as https://github.com/aspect-build/rules_jasmine/pull/102

### DIFF
--- a/.aspect/workflows/config.yaml
+++ b/.aspect/workflows/config.yaml
@@ -1,138 +1,57 @@
 ---
+queue: aspect-small
 workspaces:
     .:
         icon: js
         label: rules_js
+        tasks:
+            format:
+                soft_fail: false
+            buildifier:
+            test:
+                queue: aspect-default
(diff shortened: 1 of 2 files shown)


.aspect/workflows/config.yaml
```diff
@@ -1,138 +1,57 @@
...
             buildifier:
-                without: true
+            test:
+                queue: aspect-default
+    e2e/bzlmod:
+        icon: bazel
     e2e/js_image_docker:
         icon: docker
-        tasks:
-            format:
```
jbedard, line 16: Do these get run at all now for the e2e tests?

gregmagolan: No. I found a better pattern for it that doesn't require the negations

jbedard: Oh I see 👍, I think I misread that

jbedard (approved):

PR #1400, opened 2023-12-12
refactor: remove dependency on internal /pkg npm target

jbedard:
### Type of change

### Test plan


gregmagolan (approved):
🦖

PR #1401, opened 2023-12-12
test: move patch tests to e2e/verify-patches

jbedard:
I think the e2e/verify-patches tests the exact same thing already so I don't think we also need these in the main workspace? They get quite annoying when doing testing in the repo so keeping them in e2e tests is a lot nicer.

### Type of change


gregmagolan (approved):
Nice cleanup

PR #1403, opened 2023-12-13
chore: upgrade to Aspect Workflows 5.9.0-beta.10

gregmagolan:


.aspect/workflows/config.yaml
```diff
@@ -5,53 +5,138 @@ workspaces:
         icon: js
         label: rules_js
         tasks:
-            format:
```
kormide, line 8: Did you intend to remove buildifier and gazelle?

gregmagolan: From this list of `tasks` yes. New bug in rosetta when there are workspaces meant I had to bump these two tasks to the root `tasks` list at the bottom.

.aspect/workflows/config.yaml
```diff
@@ -5,53 +5,137 @@ workspaces:
...
+                  without: true
+            - buildifier:
+                  without: true
     # No test targets. Requires running test.sh.
     # e2e/webpack_devserver:
     # e2e/webpack_devserver_esm:
 tasks:
-    test:
+    - test:
+    # - format:
```
gregmagolan, line 140: bug in beta.10 for format on GCP; will add this task back on the next release

mattem (approved):

PR #1420, opened 2023-12-31
Only emit aspect_rules_js_metadata when it's required

dzbarsky:
### Type of change
**- Performance (a code change that improves performance)**

### Test plan

My understanding is this file is only consumed by actions with a non-empty file, so writing an empty one is unnecessary. This saves 2.5% of files in the npm runfiles tree on typical test actions which translates to a small improvement in test runtime.
As another datapoint, in https://github.com/bazelbuild/bazel/issues/8230#issuecomment-1863733985 I computed that the average NPM package we use has around 45 files. 44/45 = 97.8% :)


jbedard (approved):

gregmagolan (approved):
👍

PR #1422, opened 2024-01-01
WIP: Add an option to bundle NPM packages while installing

DavidZbarsky-at:
I couldn't get ESM bundling to work, only CJS and types. Probably doing something stupid.
Also I would have preferred to bundle with esbuild, but esbuild itself cannot be bundled, it must be marked external. (I even tried with esbuild-wasm). So that would complicate the setup a little bit.


npm/private/npm_import.bzl
```diff
@@ -593,11 +593,18 @@ bin = bin_factory("node_modules")
             rctx_files[build_file].append("""exports_files(["{}"])""".format(_PACKAGE_JSON_BZL_FILENAME))
 
     rules_js_metadata = {}
+
+    custom_postinstalls = rctx.attr.custom_postinstall
+    if rctx.attr.package.startswith("@aws-sdk/client"):
```
DavidZbarsky-at, line 598: this was a hack, really we should check if `custom_postinstalls == ["optimize_package"]` or something

PR #1423, opened 2024-01-02
refactor: sort npm_import repository rule deps

jbedard:
I think I had a diff similar to this when changing repository rules and figured sorting these would prevent code changes from potentially effecting order

### Type of change

### Test plan


gregmagolan (approved):
🌮

PR #1428, opened 2024-01-06
feat: support running js_binary and js_test without runfiles

gregmagolan:
MVP support for js_binary & js_test to run without runfiles (the default mode Bazel runs in on Windows). Resolves long-standing tech debt of not being able to run js_binary and js_test without runfiles. This feature request comes from a discussion at the BazelCon 2023 web birds-of-a-feather session.

This isn't a blanket solution for all use cases when running a js_binary or js_test wihtout runfiles since the underlying tool may need to be runfiles manifest aware to resolve some inputs and/or aware the the PWD that bazel starts the program in is an empty runfiles directory.

---

### Type of change

### Test plan

Windows CI


.bazelrc.common
```diff
@@ -74,3 +60,7 @@ build:release --stamp --workspace_status_command=$(pwd)/workspace_status.sh
 # Config to enable building and testing with bzlmod
 common:bzlmod --enable_bzlmod
 common:bzlmod --@aspect_bazel_lib//lib:flag_bzlmod
+
+common --enable_platform_specific_config
+
+test:windows --test_tag_filters=-no-windows
```
alexeagle, line 66: nit: could we just use `target_compatible_with` to exclude these tests?

alexeagle (approved):
Nice!!

.bazelrc.common
gregmagolan: good call. updated.

js/private/js_binary.sh.tpl
```diff
@@ -232,9 +257,14 @@ if [ "${node:0:1}" = "/" ]; then
         exit 1
     fi
 else
-    export JS_BINARY__NODE_BINARY="$JS_BINARY__RUNFILES/{{node}}"
+    if [ "${JS_BINARY__NO_RUNFILES:-}" ]; then
+        export JS_BINARY__NODE_BINARY
```
jbedard, line 261: What is the reason for exporting before doing the assignment? Not doing the export+assignment on the same line?

gregmagolan: Its a shell check warning to assign `$(...)` to a variable on the same line as exporting it.
```
export FOO=$(bar)
```

jbedard: I see, thanks 👍 

IMO it's odd when we flip back and forth though. Now all within ~10 lines of code we have:
```
export X
X = a
export X=b
```

Can we only export it once at the top (above the `if`) or something like that? Doesn't have to be now, just a thought for next time we change this (or similar)...

PR #1431, opened 2024-01-08
chore: upgrade to Bazel 7.0.0 with Bazel 6 as secondary on CI

gregmagolan:
No longer testing against Bazel 5.

The aspect_bazel_lib bump is required to bring in `is_bazel_7_or_greater()`.

---

### Type of change

### Test plan


alexeagle (commented):
Could you make a README edit somewhere to indicate that we support Bazel 6 and 7, and that it may-or-may-not work with Bazel 5 since we don't test that?

(I think we're missing a README section in every ruleset including the template)

.aspect/workflows/config.yaml
```diff
@@ -18,17 +17,8 @@ workspaces:
...
-        tasks:
-            - format:
-                  without: true
-            - buildifier:
-                  without: true
-            - configure:
-                  without: true
-            - delivery:
-                  without: true
+    # rules_docker not compatible with Bazel 7.
```
alexeagle, line 20: nit: link to issue?

e2e/js_run_devserver/test.sh
```diff
@@ -1,18 +1,26 @@
...
 ./serve_test.sh //src:serve_alt
 ./serve_test.sh //src:serve_simple
 ./serve_test.sh //src:serve_simple_bin
 ./serve_test.sh //src:serve_naughty
 ./serve_test.sh //src:serve_naughty_bin
-./multirun_test.sh //src:serve_multi
+
+bazel_version=$(head -n 1 .bazelversion)
+bazel_major=${bazel_version::1}
+if [[ "$bazel_major" -eq 7 ]]; then
```
alexeagle, line 19: nit: we'll have to update this line again for Bazel 8

MODULE.bazel
```diff
@@ -8,7 +8,7 @@ module(
 
 # Lower-bounds for runtime dependencies.
 # Do not bump these unless rules_js requires a newer version to function.
-bazel_dep(name = "aspect_bazel_lib", version = "1.39.0")
+bazel_dep(name = "aspect_bazel_lib", version = "1.40.0")
```
alexeagle, line 11: thanks for resisting the urge to jump to latest 😉

js/BUILD.bazel
```diff
@@ -59,7 +60,7 @@ bzl_library(
     deps = [
         "@bazel_tools//tools/build_defs/repo:http.bzl",
         "@bazel_tools//tools/build_defs/repo:utils.bzl",
-    ],
+    ] + (["@bazel_tools//tools/build_defs/repo:cache.bzl"] if is_bazel_7_or_greater() else []),
```
alexeagle, line 63: we do have `bazel_features` as a dep here, so if there's something useful there you can use it

alexeagle (approved):
FYI @jbedard this might unblock some of your refactorings?

jbedard:
Do we want to drop bazel 5 on the `next` branch or is this PR officially switching `main` to v2 where bazel5 is dropped?

gregmagolan:
> Do we want to drop bazel 5 on the `next` branch or is this PR officially switching `main` to v2 where bazel5 is dropped?

Technically we're not dropping Bazel 5 support. We're just not testing CI against it. I'll add a section to the README re: CI coverage similar to what [rules_ruby](https://github.com/bazel-contrib/rules_ruby?tab=readme-ov-file#overview) has.

js/BUILD.bazel
gregmagolan: there is from the private API but I'd rather not use the private API. we should expose the bazel version checks to public API in bazel_features repo and then we can use that instead.

MODULE.bazel
gregmagolan: 1.40.0 is the latest 1.x branch anyway 😉

e2e/js_run_devserver/test.sh
gregmagolan: updated to `-ge`

.aspect/workflows/config.yaml
gregmagolan: ✅

PR #1437, opened 2024-01-10
chore: bring in Aspect bazelrc presets

gregmagolan:
Rolling out our presets to all of our open source repositories.

---

### Type of change

### Test plan


.aspect/bazelrc/BUILD.bazel
```diff
@@ -0,0 +1,15 @@
...
+write_aspect_bazelrc_presets(
+    name = "update_aspect_bazelrc_presets",
+    presets = [
+        "ci",
+        "convenience",
+        "correctness",
+        "debug",
+        "javascript",
+        "performance",
+    ],
```
kormide, line 14: Do we want bazel6/7 presets?

gregmagolan: We have them. They are in `.github/workflows` as only CI deals with multiple versions of Bazel. For local development, there is a `.aspect/bazelrc/bazel7.bazelrc` symlink that is load bearing.

kormide (approved):

PR #1440, opened 2024-01-11
chore: disable bzlmod lockfile

gregmagolan:


.bazelrc
```diff
@@ -11,6 +11,9 @@ import %workspace%/.aspect/bazelrc/performance.bazelrc
 # We have some empty globs in rules_js
 common --noincompatible_disallow_empty_glob
 
+# Disable lockfile for now. It is unstable.
```
alexeagle, line 14: Again link to a bug so we know when to remove

alexeagle (approved):

.bazelrc
gregmagolan: There are a number of them right now. Guess I can just list all of the ones outstanding.

PR #1444, opened 2024-01-12
Make sure that auth is respected, even if lockfiles have ports

### LINKED ISSUE #1443: [Bug]: Using a port in a url with a private registry prevents auth from working causing a 403
### What happened?

We have an npmrc file with auth specified in `~.npmrc`,

It looks kind of like
```text
registry=https://private/api/npm/npm
@thing:registry=https://private/api/npm/npm
```

In `yarn.lock` we have URLs that sometimes look like `https://private:443/artifactory/api/npm/npm/...`. Notice the port.

`_get_npm_imports` fails to account for this, which means that auth isn't sent, which means requests 403

### Version

Output of `bazel --version`: 7.0.0

`WORKSPACE` or `MODULE.bazel` file: `rules_js-1.35.0`

Language(s) and/or frameworks involved: JS, `yarn.lock`

### How to reproduce

```shell
Have a yarn.lock which contains ports in urls that have private registries, see that the home npmrc auth is no longer passed through to the download call
```

### Any other information?

PR coming soon

mishazharov:
The `_get_npm_imports` fails to account for the ports in URL, meaning it will not pass through auth if a port is specified. We work around this by checking the URL with and without ports

---

### Type of change

- Closes #1443 

- No breaking changes

### Test plan

  - Create a yarn lock with a private registry (something like `@private`)
  - Add a package with this registry with a port (like 443) in the download URL
  - In the home npmrc, make sure that the registry has auth, and no port in the URL
  - `rules_js` will now fail to download the package

### DIFF
--- a/npm/private/npm_translate_lock_helpers.bzl
+++ b/npm/private/npm_translate_lock_helpers.bzl
@@ -198,6 +198,45 @@ def _get_npm_auth(npmrc, npmrc_path, environ):
     return (registries, auth)
 
 ################################################################################
+def _select_npm_auth(url, npm_auth):
+    registry = url.split("//", 1)[-1]
+
+    # Get rid of the port number
+    registry_no_port = registry
+    if ":" in registry:
+        components = registry.split(":", 1)
+        base = components[0]
+
+        path = components[1].split("/", 1)
+        if len(path) == 2:
+            registry_no_port = base + "/" + path[1]
+        else:
+            registry_no_port = base
+
+    npm_auth_bearer = None
+    npm_auth_basic = None
+    npm_auth_username = None
+    npm_auth_password = None
+    match_len = 0
+    for auth_registry, auth_info in npm_auth.items():
+        if auth_registry == "" and match_len == 0:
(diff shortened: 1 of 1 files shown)


npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -197,6 +197,44 @@ def _get_npm_auth(npmrc, npmrc_path, environ):
...
+
+    # Get rid of the port number
+    registry_no_port = registry
+    if ":" in registry:
+        components = registry.split(":", 1)
+        base = components[0]
+
+        path = components[1].split("/", 1)
+        if len(path) == 2:
+            registry_no_port = base + "/" + path[1]
```
gregmagolan, line 212: any reason for the additional `registry_no_port` here? can just assign back to `registery` instead?

gregmagolan: I see. Both `registery` and `registry_no_port` are needed as they are both checked against on line 228 👍

gregmagolan (approved):
Thanks for the fix

PR #1446, opened 2024-01-17
fix: include npm and yarn lock files in npm_translate_lock(data)

jbedard:
When looking into the bzlmod pnpm-import issue it seems the import sometimes wasn't running at all which seemed to be from the `yarn.lock` not being in the `data`.

WDYT? Worth doing?

### Type of change

### Test plan


alexeagle (approved):
yeah that's a nice ergonomic improvement, it's a pity when we force you to specify the same thing twice.

I notice that we didn't have our own yarn.lock or package-lock.json files declared in data for our e2e so we probably could have reproduced a bug in our repo.

Issue #1445, opened 2024-01-16
[Bug]: update_pnpm_lock fails under Bzlmod, retry fixes it


alexeagle:
### What happened?

Under bzlmod, `update_pnpm_lock` fails on the first attempt.

### Version

HEAD

### How to reproduce

In rules_js,

1. `cd e2e/npm_translate_lock`
2. modify package.json, like `"semver": "7.5.4"`

then `bazel build ...` gives an error:

```shell
~/Projects/rules_js/e2e/npm_translate_lock$ bazel build --enable_bzlmod ...
Starting local Bazel server and connecting to it...
ERROR: Skipping '...': error loading package under directory '': error loading package '': at /shared/cache/bazel/user_base/3b36aba3ddc6cb653f7aaaed8bb07c4d/external/aspect_rules_js~override~npm~npm/defs.bzl:40:6: Unable to find package for @aspect_rules_js~override~npm~npm__semver__registry.npmjs.org_semver_7.5.4__links//:defs.bzl: The repository '@aspect_rules_js~override~npm~npm__semver__registry.npmjs.org_semver_7.5.4__links' could not be resolved: Repository '@aspect_rules_js~override~npm~npm__semver__registry.npmjs.org_semver_7.5.4__links' is not defined.
```

Simply running the same `bazel build` command again fixes it.

alexeagle:
Adding some `print` debugging shows that the module extension isn't executed the way we thought. We call `npm_translate_lock' but it is non-blocking, the `module_ctx.read(attr.pnpm_lock)` later in `extensions.bzl` is already evaluated before we write the changed pnpm-lock.yaml back to the disk. I think this is fixable, but I'm not exactly sure how, and anyway I don't think we have time to prioritize it right now.

jbedard:
Support for this has been removed (bzlmod and workspace) in 2.x: https://github.com/aspect-build/rules_js/commit/ca4a3cb74cf623ef0b5834f2847f093a0bb6f361

nishant-satpathy:
has this bug been fixed? I am using the `2.4.2` version of this rule and still encounter this fail and retry behavior.

jbedard:
See https://github.com/aspect-build/rules_js/issues/1445#issuecomment-2097167353

PR #1447, opened 2024-01-17
fix: exit with message when pnpm-lock.yaml updated in bzlmod

jbedard:
Exit with a message instead of crashing. Maybe we'll fix this in the future, maybe we'll completely remove the `reload_lockfile()` in v2?

---

### Type of change

### Test plan

### DIFF
--- a/npm/private/npm_translate_lock.bzl
+++ b/npm/private/npm_translate_lock.bzl
@@ -76,6 +76,7 @@ _ATTRS = {
     "verify_patches": attr.label(),
     "npm_package_target_name": attr.string(),
     "yarn_lock": attr.label(),
+    "bzlmod": attr.bool(),
 }
 
 npm_translate_lock_lib = struct(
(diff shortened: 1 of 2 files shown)


npm/private/npm_translate_lock.bzl
```diff
@@ -98,8 +99,14 @@ def _npm_translate_lock_impl(rctx):
         if state.action_cache_miss():
             _fail_if_frozen_pnpm_lock(rctx, state)
             if _update_pnpm_lock(rctx, state):
-                # If the pnpm lock file was changed then reload it before translation
-                state.reload_lockfile()
+                if rctx.attr.bzlmod:
+                    msg = """
+
+INFO: {} file updated. Please run your build again.""".format(state.label_store.relative_path("pnpm_lock"))
```
alexeagle, line 105: nit: I'd suggest a TODO here, or maybe include a link to the issue in the message we print, I hope this is a short-term improvement

jbedard:
For reference: https://github.com/aspect-build/rules_js/issues/1445

PR #1450, opened 2024-01-19
chore: disable bzlmod lock file throughout

gregmagolan:
Missed a few spots

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,11 @@ node_modules/
 .pnpm-*
 .DS_Store
 .bazelrc.user
+
+# Disable lockfile for now. It is unstable.
+# https://github.com/bazelbuild/bazel/issues/19026
+# https://github.com/bazelbuild/bazel/issues/19621
+# https://github.com/bazelbuild/bazel/issues/19971
+# https://github.com/bazelbuild/bazel/issues/20272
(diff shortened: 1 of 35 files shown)


.gitignore
```diff
@@ -3,4 +3,12 @@ bazel-*
...
\ No newline at end of file
+.bazelrc.user
+
+# Disable lockfile for now. It is unstable.
+# https://github.com/bazelbuild/bazel/issues/19026
+# https://github.com/bazelbuild/bazel/issues/19621
+# https://github.com/bazelbuild/bazel/issues/19971
+# https://github.com/bazelbuild/bazel/issues/20272
+# https://github.com/bazelbuild/bazel/issues/20369
+MODULE.bazel.lock
```
alexeagle, line 14: nit: this is redundant right? I'd never have this file appear since it's not produced.

I think it's better not to ignore it, because when we do decide to create one, it should get checked in.

alexeagle (approved):

.gitignore
gregmagolan: Yes. Tho its a defence in depth since someone will inevitably add a new e2e test and forget to disable the lock file in that .bazelrc without noticing. OTOH, with this here, when we go to check in the files finally it will be obvious it is .gitignored.

PR #1453, opened 2024-01-22
feat: Support hermetic patching

LaurenceTews:
Allows Bazel native patch to be used with npm_import(). This removes the existing dependency on bash when enabled. The proviso is that patch diff headers must have an unstripped 'package/' prefix when use_native_patch is true.

### Type of change

- use_native_patch flag added to npm_import, allowing for hermetic patching.

### Test plan

Manual testing steps:

* set BAZEL_SH var to a nonexistent path.
* Create valid patch. Patch diff header should contain package prefix, e.g. "diff --git package/lib/font.coffee ..."
* verify that npm_import() with patch applied fails "Error applying patch". System bash cannot be found.
* verify that npm_import() with patch applied and use_native_patch set to 'True' succeeds.


LaurenceTews:
@gregmagolan Just wanted to gauge if there's any interest in this feature? I know Windows is a lesser used platform, but I'm sure there's teams out there (like mine) who would find this useful.

matthewjh:
@gregmagolan  any thoughts on this MR? I tried to run a rules_js project on Windows, and it blew up while resolving NPM packages due to the "patch" dependency.

smocherla-brex:
Ran into this issue as well on our Macs which have an older version of `patch` which is stricter than the one on Linux. Would be nice to use the native patch implementation as opt-in as this PR does, (although for our usecase a note in the docs for `npm_translate_lock` regarding this would also suffice)

PR #1454, opened 2024-01-22
fix: only output warnings on lockfile parsing when regenerating lockfile

jbedard:
Today yaml errors cause failures in the util methods far too soon. The util methods now propagate the errors and are handled more appropriately such as only outputting a warning for lockfile errors if the lockfile is being updated.

---

### Type of change

### Test plan

- New test cases added

### DIFF
--- a/npm/private/yaml.bzl
+++ b/npm/private/yaml.bzl
@@ -27,7 +27,7 @@ def parse(yaml):
         yaml: string, the yaml content to parse
 
     Returns:
-        An equivalent mapping to native starlark types
+        A tuple containing an equivalent mapping to native starlark types and error.
     """
     yaml = _normalize_yaml(yaml)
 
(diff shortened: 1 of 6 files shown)


npm/private/yaml.bzl
```diff
@@ -56,11 +56,9 @@ def parse(yaml):
             fail(msg)
 
         if starlark["error"] != None:
-            # buildifier: disable=print
-            print("YAML parse error: ", starlark["error"])
-            return None
+            return None, starlark["error"]
```
gregmagolan, line 59: nit: update return info in the docstring above

gregmagolan (approved):

npm/private/yaml.bzl
jbedard: You mean prepend the `YAML parse error` string?

jbedard: IMO we should leave that off. When `yaml.parse` returns an error I don't think it's worth prefixing it...? 🤷

gregmagolan: No, I mean in the docstring for this starlark function mention that the return code is now a tuple

gregmagolan: Currently it states,
```
    Returns:
        An equivalent mapping to native starlark types
    """
```
which is now out of date

PR #1458, opened 2024-01-24
feat: use yq to parse yaml file

gregmagolan:
- Use yq to parse yaml file
- wip

---

### Type of change

- Bug fix (change which fixes an issue)
- New feature or functionality (change which adds functionality)
- Style (white-space, formatting, etc...)
- Refactor (a code change that neither fixes a bug or adds a new feature)
- Performance (a code change that improves performance)
- Documentation (updates to documentation or READMEs)
- Chore (any other change that doesn't affect source or test files, such as configuration)

**For changes visible to end-users**

- Breaking change (this change will force users to change their own code or config)
- Relevant documentation has been updated
- Suggested release notes are provided below:

### Test plan

- Covered by existing test cases
- New test cases added
- Manual testing; please provide instructions so we can reproduce:

PR #1459, opened 2024-01-24
chore: switch to a gitpkg.vercel.app that doens't trip https://github.com/pnpm/pnpm/issues/5414

gregmagolan:
Pre-factored out of https://github.com/aspect-build/rules_js/pull/1458 since that change trips on https://github.com/pnpm/pnpm/issues/5414.

---

### Type of change

### Test plan


jbedard:
Should we also add a separate test for that package?

jbedard (approved):

gregmagolan:
> Should we also add a separate test for that package?

All the packages listed in `npm/private/test/package.json` are tested with a `build_test(":node_modules")` target.

PR #1462, opened 2024-01-24
fix: bump minimum aspect_bazel_lib version to 2.2.0 for is_bazel_7_or_greater function

gregmagolan:
The current minimum of 1.40.0 is fine for the 1.x branch but this breaks things for users that have a version >= 2.0.0 and < 2.2.0. This doesn't change the WORKSPACE version pulled in `rules_js_dependencies` which remains as 1.40.0.

---

### Type of change

### Test plan

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -8,7 +8,7 @@ module(
 
 # Lower-bounds for runtime dependencies.
 # Do not bump these unless rules_js requires a newer version to function.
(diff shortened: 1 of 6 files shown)


gregmagolan:
Discussed offline. Better solution is to vendor in the `is_bazel_7_or_greater` function for now until rules_js 2.0 when we can bump the minimum.

PR #1474, opened 2024-02-05
chore: remove Go

alexeagle:
We don't have any Go sources in rules_js, so there's no reason we should install a Go toolchain or worry about upgrading it. 'aspect configure' provides the auto-generation for bzl_library targets so we don't need Gazelle either.


alexeagle:
Note, I confirmed we have

```
tasks:
...
    - configure:
          queue: aspect-small
```
in the workflows config, so we shouldn't lose the CI check that BUILD files are updated.

gregmagolan (changes requested):
Hold on this to tomorrow to discuss

alexeagle:
gazelle is currently load-bearing for testing Aspect Workflows on GCP+Buildkite so closing this.
Can revisit when we stop recommending customers use gazelle.

PR #1481, opened 2024-02-19
feat: add replace_packages to npm_translate_lock

gregmagolan:
Prioritized feature request from an Aspect Workflows customer.

The use case is to be able to replace the npm package sources of an npm package as specified in the pnpm lock file with a custom npm_package target. The motivation for this feature is to allow users to replace non-hermetic lifecycle hooks in npm packages they depend on, such as node-gyp c++ compilation, with hermetic bazel targets. The feature, however, is generic so that it can be used for other use cases where a custom npm_package target to replace the sources specified in the pnpm lock file is useful.

---

### Type of change

### Test plan

- New test cases added


gregmagolan:
Internal slack link for reference: https://aspect-build.slack.com/archives/C05S94VJWNM/p1706724889343889

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -321,6 +321,8 @@ def npm_link_all_packages(name = "node_modules", imported_links = []):
         npm_auth_password = "%s",""" % _import.npm_auth_password) if _import.npm_auth_password else ""
         maybe_dev = ("""
         dev = True,""") if _import.dev else ""
+        maybe_replace_package = """
```
jbedard, line 324: Can we add some `(...)` around this like most of (but not all 😢) the other templates here? There are way too many quotes within quotes here 😅

jbedard (approved):

PR #1486, opened 2024-02-21
fix `testonly` with js_binary so it's not a test

Aghassi:
the current logic is wrong because it tries to test `js_binary` when `testonly` is passed. We actually want the inverse which is we don't want to run js_binary as a test if it is marked as test_only.

Bug discovered by myself and @joeljeske

---

### Type of change

### Test plan

- New test cases added


gregmagolan:
Interesting one. Suppose there is no value of tagging a js_binary as testonly. I'm still confused how you get this to fail since neither `bazel test` or `bazel coverage` should run a js_binary target as test right?

jbedard:
Can you show an example that fails without the change?

joeljeske:
We have a js_binary target that is not a test itself, but a dependency of a test. The testonly here is a guard against accidental usage outside of a test dag.

Presumably you could fix this more properly by having js_test have a _is_test attr that is set to true. This would allow you to ensure all js_tests have a lcov_merger but allow a js_binary to be used with testonly = True.

Aghassi:
@jbedard here is a screenshot from me reproducing the failure with `bazel coverage` using vscode codespaces off of `main` 
image


Here is the code

```
# A simple program that runs the Acorn JS parser to produce an AST
js_binary(
    name = "my_bin",
    # Reference the location where the acorn npm module was linked in the root Bazel package
    data = ["//:node_modules/acorn"],
    entry_point = "require_acorn.js",
    testonly = True
)

js_test(name = "test_testonly_bin", data = ["my_bin"], entry_point = "npm_version_test.js")
```

Aghassi:
I can confirm also that applying this fix with that code fixes the bug

thesayyn:
i believe this is still wrong as _lcov_merger should be present if the coverage is enabled but since we are sharing implementation between js_test and js_binary we need a sentinel attribute for determining where the impl is being invoked so we can still fail if it's js_test.

thesayyn:
Exactly.

js/private/js_binary.bzl
```diff
@@ -641,6 +640,7 @@ the contract between Bazel and a test runner.""",
             default = Label("//js/private/coverage:merger"),
             cfg = "exec",
         ),
+        "_is_test": True
```
Aghassi, line 643:
```suggestion
        "_is_test": attr.bool(
            default = True
        )
```

Aghassi:
@thesayyn Ok I believe I have added the fix as interpreted from @joeljeske's comment. I confirmed with the same test as before that this fix works too

Aghassi:
Looks like this breaks some CI stuff. I'll have to spin up an ubuntu container and debug why `bazel coverage //...` is failing

PR #1490, opened 2024-02-22
chore: update pnpm versions.bzl

jbedard:
### Type of change

### Test plan


alexeagle (approved):
Should add our typical automation for this too

PR #1491, opened 2024-02-23
fix: add bzlmod support for custom pnpm extension in npm_translate_lock

jbedard:
Currently there is no way to override the pnpm version used the `pnpm import` call when using bzlmod, the pnpm extension defined in the rules_js MODULE.bazel is used 100% of the time. The `npm_translate_lock(pnpm_version)` is unused and a noop in bzlmod.

In order to use a pnpm extension defined by the user it must be passed as a label into the extension (the `Label` construction must be in the repository that owns that pnpm extension). That's what is passed as `npm_translate_lock(pnpm_extension)`, or by default it constructs that label *within* rules_js like today to reference the pnpm extension defined in the rules_js MODULE.bazel.

---

### Type of change

### Test plan

- New test cases added


e2e/npm_translate_yarn_lock/MODULE.bazel
```diff
@@ -12,11 +12,19 @@ local_path_override(
     path = "../..",
 )
 
+pnpm = use_extension("@aspect_rules_js//npm:extensions.bzl", "pnpm")
+pnpm.pnpm(
+    name = "pnpm-old-version",
+    pnpm_version = "7.33.3",
```
jbedard, line 18: Note the intentionally old version and the big diff in the `pnpm-lock.yaml` because of the version downgrade.

npm/private/npm_translate_lock.bzl
```diff
@@ -499,6 +501,10 @@ def npm_translate_lock(
 
         pnpm_version: pnpm version to use when generating the @pnpm repository. Set to None to not create this repository.
 
+            Deprecated for bzlmod, use `pnpm_extension` instead.
+
+        pnpm_extension: label of the pnpm extension to use
```
jbedard, line 506: I think I might be getting bzlmod terminology confused here. Let me know if there's a better term or way of phrasing this...

kormide: But if you are passing in a label, why not just make it the full label to a pnpm binary?

jbedard: Is "repo" the correct term and not "extension"?

kormide: Yeah, the extension creates a "repo/workspace". This is a... "label to a repo"?.

kormide: It's essentially the same as how repo rules create repos/workspaces when using WORKSPACE. Now they are just wrapped in module extensions and namespaced.

jbedard: But the full binary vs just the repo(extension?) name is what I'd like to discuss before merging, what the public API looks like...

Currently in this PR it assumes what you pass in is repo name representing the pnpm package, created by the pnpm extension. Then rules_js does `relative(":package/bin/pnpm.cjs")` to point to the specific file within the pnpm package. This means the public API is for the user to specify the pnpm repo name, not the precise .js file being executed. This means the "pnpm repo name" essentially must be a `npm_import` of the `pnpm` package (and I sure hope the pnpm version doesn't change that `bin/pnpm.cjs` path 😓).

kormide: Pnpm could theoretically come from some other repository than the one defined by our bzlmod extension, so I would favour passing in the full label.

npm/private/npm_translate_lock.bzl
```diff
@@ -499,6 +501,10 @@ def npm_translate_lock(
 
         pnpm_version: pnpm version to use when generating the @pnpm repository. Set to None to not create this repository.
 
+            Deprecated for bzlmod, use `pnpm_extension` instead.
```
gregmagolan, line 504: Should the docs be `pnpm_version` for WORKSPACE and `pnpm_extensions` for MODULE.bazel?

We should fail if you set `pnpm_version` in your MODULE.bazel if that is the case

npm/private/npm_translate_lock.bzl
jbedard: If it's the full label then users could pass in whatever they want, which is a good thing, but it also means they need to understand the internal paths of pnpm :/

npm/private/npm_translate_lock.bzl
jbedard: Ideally `pnpm_version` gets dropped from the bzlmod extension. Today it's a noop, so do you think it's ok to actually drop it even though it's sort of a "breaking change"?

jbedard: But the docs for bzlmod + WORKSPACE are both from this same snippet...

npm/private/npm_translate_lock.bzl
kormide: Can you just default the label? Which will always be valid because we always register the pnpm toolchain in the root module?

npm/extensions.bzl
```diff
@@ -215,11 +215,6 @@ def _npm_translate_lock_attrs():
     attrs["name"] = attr.string()
     attrs["lifecycle_hooks_exclude"] = attr.string_list(default = [])
     attrs["lifecycle_hooks_no_sandbox"] = attr.bool(default = True)
-
-    # Note, pnpm_version can't be a tuple here, so we can't accept the integrity hash.
-    # This is okay since you can just call the pnpm module extension below first.
-    # TODO(2.0): drop pnpm_version from this module extension
-    attrs["pnpm_version"] = attr.string(default = LATEST_PNPM_VERSION)
```
jbedard, line 222: @gregmagolan this is technically a breaking change, however previously it was a noop and didn't do anything. So this might break users builds, but exposes the fact that setting this attribute was a pointless noop leading to confusion.

WDYT?

gregmagolan: What we typically do in this case is `fail(...)` with a helpful migration message. For users that have it set and think its working this will be better than a bazel failure with "unrecognized parameter"

npm/private/npm_translate_lock.bzl
gregmagolan: As long as you say which parameter is for which then the docs being common is fine IMO. Long term WORKSPACE will be gone and this will get cleaned up.

npm/private/npm_translate_lock.bzl
jbedard: > I think if pnpm_repo would be clearer. I didn't even know you could have a label to just the repository without referencing a target.. 🤔

Sorry I missed this. It resolves to the default target, then within the rules_js repo it does `.relative` to change to the specific pnpm bin .js file. This is the part that is iffy imo and I want to discuss more.

> Can you just default the label? Which will always be valid because we always register the pnpm toolchain in the root module?

Doesn't that mean it gets defaulted to the label instantiated within the rules_js repo though? So would be scoped to the rules_js repo and not to the users repo. That's what the default value does to reference `@pnpm` within rules_js...

npm/private/npm_translate_lock.bzl
jbedard: I updated this one 👍

npm/extensions.bzl
jbedard: Done, see https://github.com/aspect-build/rules_js/compare/30b5068f59d2895f3c06fc7b94a75efb2fccaf2d..eb9f39f3ce48c2a02b4f8448364734918f7c8938

npm/private/npm_translate_lock_state.bzl
```diff
@@ -155,7 +155,9 @@ def _init_pnpm_labels(label_store, rctx):
     #  Label("@nodejs_host//:bin/node")
     label_store.add("host_node", Label("@{}_{}//:bin/node".format(rctx.attr.node_toolchain_prefix, repo_utils.platform(rctx))))
 
-    label_store.add("pnpm_entry", Label("@pnpm//:package/bin/pnpm.cjs"))
+    pnpm_repo = rctx.attr.use_pnpm if rctx.attr.use_pnpm else Label("@pnpm")
+
+    label_store.add("pnpm_entry", pnpm_repo.relative(":package/bin/pnpm.cjs"))
```
jbedard, line 160: This `":package/bin/pnpm.cjs"` is what I think still needs discussion I think.

Putting the path here means the public API only needs to specify `@pnpm-repo-name`, but then the public API can't change the path at all. I think `@pnpm//:pnpm` (the default label) pointing to the bin would be great, so the default is nice but can still be customized, but currently that default label is [already defined](https://github.com/aspect-build/rules_js/blob/93e53bbb60d966e6b4b9d3b0440739009b62bf5f/npm/private/pnpm_repository.bzl#L40) as something else. I'm trying that out atm, but other ideas/feedback would be great...

jbedard: Since labels can not be evaluated in repo rules I think we either need to either hardcode the `:package/bin/pnpm.cjs"` here when adding it to the label or when the public API is used.

See the diff swapping between my 2 ideas: https://github.com/aspect-build/rules_js/compare/ae7a283ee5285380c3ce0fc749508818ce547ec2..db67dda7077b201ff5323fa040193fac0aa2cfd7

npm/extensions.bzl
```diff
@@ -17,6 +17,10 @@ LATEST_PNPM_VERSION = _LATEST_PNPM_VERSION
 def _extension_impl(module_ctx):
     for mod in module_ctx.modules:
         for attr in mod.tags.npm_translate_lock:
+            # TODO(2.0): remove pnpm_version from bzlmod API
+            if attr.pnpm_version:
+                fail("npm_translate_lock paramater 'pnpm_version' is not supported with bzlmod, use `use_pnpm` instead, recieved value: '{}'")
```
gregmagolan, line 22: nit: i before e except after c

npm/private/npm_translate_lock.bzl
```diff
@@ -499,6 +501,12 @@ def npm_translate_lock(
 
         pnpm_version: pnpm version to use when generating the @pnpm repository. Set to None to not create this repository.
 
+            Use `use_pnpm` for bzlmod.
+
+        use_pnpm: label of the pnpm extension to use
+
+            Use `pnpm_version` for non-bzlmod.
```
gregmagolan, line 508: If this is unset, the version of pnpm comes from the rules_js MODULE.bazel file? Should mention what happens when it is left unset.

gregmagolan: Should mention that the label should be fully qualified as `"@repository_name//:package/bin/pnpm.cjs"` or else take a repository name and add the `//:package/bin/pnpm.cjs` implicitly. Maybe support both so if the label is _not_ fully qualified, for example just `@pnpm`, then `//:package/bin/pnpm.cjs` is added or else it is not.

gregmagolan (approved):
🌮

npm/private/npm_translate_lock.bzl
gregmagolan: Testing for the existence of `//` in the string ought to be enough to assert if you need to append `//:package/bin/pnpm.cjs` or not.

kormide:
Can `use_pnpm` be defaulted in the attrs used by the module extension, where the default is the label pointing to the repo created in rules_js's MODULE.bazel? Otherwise, this LGTM :+1:

npm/private/npm_translate_lock.bzl
jbedard: Since bzlmod extensions are not macros I don't think this is possible. By the time we have access to the `use_pnpm` it has already been evaluated so the default target will be there.

jbedard: If we find a way to do it I think it will be (almost 😅) non-breaking so I think we can keep it as-is for now where the user must specify the full path.

PR #1496, opened 2024-02-27
build: update rules_nodejs to v6.0.5 for tests

jbedard:
To test with the latest rules_nodejs, get the latest updates such as node versions, while keeping the public dependency versions on v5 to avoid breaking changes for now.

---

### Type of change

### Test plan


MODULE.bazel
```diff
@@ -12,7 +12,17 @@ bazel_dep(name = "aspect_bazel_lib", version = "1.40.0")
 bazel_dep(name = "aspect_rules_lint", version = "0.9.1")
 bazel_dep(name = "bazel_features", version = "0.1.0")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
+
+# Keep v5 in to avoid breaking changes.
+# TODO(2.0): change default to v6
```
alexeagle, line 17:
```suggestion
# TODO(2.0): change minimum to v6
```

alexeagle (approved):

PR #1500, opened 2024-02-27
test: add node20 toolchain tests

jbedard:
I don't recall what originally prompted adding node v20 tests, but we probably should have basic node20 tests since it is LTS now.


thesayyn (approved):
LGTM, (i can't tell why we are adding this)

jbedard:
I was looking into the fs patches for newer versions of node (21+) and figured we should at least add the LTS node versions for now

Issue #918, opened 2023-03-02
[Bug]: passing in `//:node_modules` which are empty in a workspace throws an error


Aghassi:
### What happened?

@gregmagolan asked me to make this ticket.

In a project where I'm working on generating/scaffolding new 1st party libraries, I hit this scenario where building a target will fail if you reference node_modules in that target that don't exist (because it has zero deps). Technically this falls under `rules_ts` as a it's a `ts_project` error, but I was asked to post here

```
Error in fail: ts_project '//web/tools/src:src' dependency '//web/tools:node_modules' does does not contain any declarations (.d.ts or other type-check files).
```

Here is the use case:
1. User generates a new internal library with my script
2. They build the library and don't care about node_modules yet (we don't have build file generation) so our macro bulk adds the necessary paths so that we just lay out all node_modules on disk. This is inefficient but it makes onboarding easier for everyone involved.

The current workaround I have is to use `//:node_modules` because the root of the workspace will always resolve dependencies. But without build file generation, it's a massive pain to have folks care about updating what node_modules are needed per target every time they make a change. Easier to bulk add.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: 6.0.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

Language(s) and/or frameworks involved:
Typescript
SWC
rules_js (1.20.0)
rules_ts (1.3.0)

### How to reproduce

```shell
Create a new build file that passes `//path/to:node_modules` in as deps to a `ts_project`. The package.json for this target should have zero deps but exist in a workspace that has node_modules a the root.
```

### Fund our work

- [ ] Sponsor our open source work by donating a [bug bounty](https://opencollective.com/aspect-build/)

gregmagolan:
Seems like it should be fine to depend on an empty `//:node_modules` tree even if there are no deps there. Should just be a bit of tweaking to make this not trip the assertion.

PR #1513, opened 2024-03-04
test: add test of empty node_modules target

jbedard:
Close https://github.com/aspect-build/rules_js/issues/918


Aghassi (approved):
This looks correct given my limited memory of the issue from last year

gregmagolan (approved):
🌮

PR #1521, opened 2024-03-13
perf: cache transitive_closure computations

jbedard:
When manually testing by counting the number of iterations in the "infinite" loop the count is often significantly reduced, although will always depend on the iteration order. For example, `@angular/core` coming before `@angular/material` (which depends on `core`) will result in a cache hit while the opposite order will not due to the the cache persisting being outside the "infinite" loop method.

### Type of change

### Test plan

### DIFF
--- a/npm/private/transitive_closure.bzl
+++ b/npm/private/transitive_closure.bzl
@@ -3,7 +3,7 @@
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load(":utils.bzl", "utils")
 
-def gather_transitive_closure(packages, package, no_optional):
+def gather_transitive_closure(packages, package, no_optional, cache = {}):
     """Walk the dependency tree, collecting the transitive closure of dependencies and their versions.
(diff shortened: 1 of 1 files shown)


npm/private/transitive_closure.bzl
```diff
@@ -45,14 +46,22 @@ def gather_transitive_closure(packages, package, no_optional):
             transitive_closure[name] = transitive_closure.get(name, [])
             if version in transitive_closure[name]:
                 continue
-            transitive_closure[name].insert(0, version)
+            transitive_closure[name].append(version)
```
jbedard, line 49: These get sorted before returning so the insertion-order doesn't matter. I assume appending is more optimal then inserting. Was there any (known) reason for inserting at the start?

gregmagolan: I don't recall

gregmagolan (approved):
🌮

PR #1527, opened 2024-03-18
feat: add npm_package(publishable) to disable npm_publish target generation

SinimaWath:
---

### Type of change
New feature or functionality (change which adds functionality):

Arg `with_npm_publish_target` to `npm_package` to have possibility to prevent generation of `{name}.publish` target.

Added arg `with_npm_publish_target` to `npm_package` to have possibility to prevent generation of `{name}.publish` target.

### Test plan

Manual testing:
- Provide `with_npm_publish_target = False` to `npm_package`
- `bazel query "..." | grep publish`.


SinimaWath:
@gregmagolan Is it ok for you?

jbedard:
@SinimaWath I think we should rename this to `publishable`. WDYT?

I think we should also add a `TODO(2.0): switch to false` and in v2 we can make the breaking change of not creating this target by default. I think most of the time packages are only used internally and this target is not even necessary. Worst case we can drop the `TODO` if we decide we want to keep it as `true`, best case we switch the default in 2.0.

SinimaWath:
@jbedard

Looks good to me, changed that and added TODO

thesayyn:
I proposed this back when we added the `.publish` target, but decided it was unnecessary. I don't think much changed since then. What would be justification for adding this other than making the query result smaller?

SinimaWath:
@thesayyn

Yes, you're right, there are 2 reason for me:
1 - Make query result smaller
2 - Replace `name.publish` with our internal implementation, but keeping DX the same as using `npm_package`, so developers don't need to learn a new target of publishing packages. I think it's small changes, but essential.

gregmagolan (approved):

jbedard (approved):

PR #1529, opened 2024-03-18
feat: js_image_layer support setting owner

thesayyn:
This drops the tests that were added for js_image_layer and adds support for setting `owner` for js_image_layer.

### Type of change

### Test plan

- New test cases added


js/private/image/index.ts
```diff
@@ -354,22 +379,29 @@ export async function build(
...
+    const [
+        entriesPath,
+        outputPath,
+        compression,
+        owner,
+        useLegacySymlinkDetection,
+    ] = process.argv.slice(2)
     const raw_entries = await readFile(entriesPath)
     const entries: Entries = JSON.parse(raw_entries.toString())
+    const [uid, gid] = owner.split(':').map(parseInt)
```
jbedard, line 399: use `Number` instead of `parseInt` so the array index doesn't get passed as the `parseInt(, base)` param.

thesayyn: Done

jbedard (approved):

PR #1533, opened 2024-03-20
fix: don't hardlink the /package source directory in npm_package_store

### LINKED ISSUE #1412: [Bug]: Flaky build failure: npm package directory copy fails with "No such file or directory"
### What happened?

Errors look like

```
ERROR: /__w/monorepo/monorepo/BUILD.bazel:87:22: Copying directory npm__find-up__5.0.0/package failed: Exec failed due to IOException: /github/home/.cache/bazel/_bazel_root/818c72ad5a3a83f42a3d6f48674a13b0/execroot/__main__/external/npm__find-up__5.0.0/package (No such file or directory)
```
https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1702916591850299

```
Copying directory npm_rulesjs__uuid__9.0.0/package failed: Exec failed due to IOException: /root/.cache/bazel/_bazel_root/8937bc5b791d3db8cd10f15350ebd801/execroot/__main__/external/npm_rulesjs__uuid__9.0.0/package (No such file or directory) 
```

They happen on about 2% of builds as observed by one client. They use `--remote_download_minimal`.

### Version

One client observed on rules_js 1.32.2 and bazel 6.4.0

### How to reproduce

```shell
Not sure yet.
```

### Any other information?

_No response_

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/1412.

---

I'm not entirely sure on the reason that this resolves #1412 but I have some loose theories listed below and I've tested it against a consistent and fast repro of the issue in https://github.com/bazelbuild/examples and found that the two consistent fixes are:

1. set `--modify_execution_info=CopyDirectory=+no-remote` on the build
2. don't use hardlink when copying the `/package` source directory in the `copy_directory_bin_action` inside `npm_package_store` (this PR)

Somehow a cache hit for an npm_package_store that has hardlinked files within a tree artifact to the `/package` directory of the external repository leads to the failure:

```
Copying directory npm__some_pkg__1.2.3/package failed: Exec failed due to IOException: /.../npm__some_pkg__1.2.3/package (No such file or directory) 
```

Is bazel is doing something unusual when writing or restoring cached tree artifacts containing hardlinks to a folder in an external repository? Another possibility is some timing issue where the repo rule is being re-run so the `/package` directory indeed does not exist momentarily and somehow that interacts badly with the hardlinks in the tree artifact that is cached.

NB: hardlinks to source file in external repositories mean that Bazel will inadvertently set permissions of those source files in the external repository to read-only after the after the copy action completes, which is long after the repository rule executes. This happens because Bazel will set the files in the output tree to read-only (as it does for all files in the output tree) but, because hardlinks share an underlying file ACL, doing so will also set the files in the external repository to read-only. The changes to the ACL to files contained within a source directory that is an input to a cached action may be related to the flake. NB: the flake in #1412 was reported as being more common in Bazel 7+.

In any case, the hardlinks here were a minor optimization for npm packages with large files so removing them to fix this flake is worth the minor performance penalty on the small subset of npm packages with large files.

### Type of change

### Test plan

The issue does not repro in rules_js. The https://github.com/bazelbuild/examples frontend example was used to confirm this fix in this DNL PR: https://github.com/bazelbuild/examples/pull/429.

### DIFF
--- a/npm/private/npm_package_store.bzl
+++ b/npm/private/npm_package_store.bzl
@@ -150,6 +150,10 @@ If set, takes precendance over the package version in the NpmPackageInfo src.
         - "auto": hardlinks are used for generated files already in the output tree
         - "off": all files are copied
         - "on": hardlinks are used for all files (not recommended)
+
+        NB: Hardlinking source files in external repositories as was done under the hood
+        prior to https://github.com/aspect-build/rules_js/pull/1533 may lead to flaky build
+        failures as reported in https://github.com/aspect-build/rules_js/issues/1412.
         """,
     ),
     "verbose": attr.bool(
@@ -191,13 +195,15 @@ def _npm_package_store_impl(ctx):
             virtual_store_directory = src_directory
         else:
             virtual_store_directory = ctx.actions.declare_directory(virtual_store_directory_path)
-            hardlink = ctx.attr.src[NpmPackageInfo].hardlink if hasattr(ctx.attr.src[NpmPackageInfo], "hardlink") else False
             copy_directory_bin_action(
                 ctx,
                 src = src_directory,
                 dst = virtual_store_directory,
                 copy_directory_bin = ctx.toolchains["@aspect_bazel_lib//lib:copy_directory_toolchain_type"].copy_directory_info.bin,
-                hardlink = "on" if hardlink else ctx.attr.hardlink,
+                # Hardlinking source files in external repositories as was done under the hood
+                # prior to https://github.com/aspect-build/rules_js/pull/1533 may lead to flaky build
+                # failures as reported in https://github.com/aspect-build/rules_js/issues/1412.
+                hardlink = ctx.attr.hardlink,
                 verbose = ctx.attr.verbose,
             )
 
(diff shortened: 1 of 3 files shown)


npm/private/npm_package_store.bzl
```diff
@@ -191,13 +191,12 @@ def _npm_package_store_impl(ctx):
...
         else:
             virtual_store_directory = ctx.actions.declare_directory(virtual_store_directory_path)
-            hardlink = ctx.attr.src[NpmPackageInfo].hardlink if hasattr(ctx.attr.src[NpmPackageInfo], "hardlink") else False
             copy_directory_bin_action(
                 ctx,
                 src = src_directory,
                 dst = virtual_store_directory,
                 copy_directory_bin = ctx.toolchains["@aspect_bazel_lib//lib:copy_directory_toolchain_type"].copy_directory_info.bin,
-                hardlink = "on" if hardlink else ctx.attr.hardlink,
+                hardlink = ctx.attr.hardlink,
```
jbedard, line 199: Is it worth a comment here referencing the bug being fixed and noting that `hardlink` can have potential issues? Or a comment in the docs for `hardlink`?

Partially just so we have something in rules_js referencing the issue, if not a test then a comment in the effected code? 🤷

jbedard (approved):

npm/private/npm_package_store.bzl
gregmagolan: Yes. I'll add some comments here and to the hardlinks docstring.

PR #1534, opened 2024-03-20
chore: add generation file to npm_translate_lock and npm_import

gregmagolan:
---

### Type of change

### Test plan

- New test cases added


jbedard:
Do you think this is worth it? If the issue requiring the cache bust is rare enough maybe users can just do a clean themselves?

PR #1537, opened 2024-03-22
chore: upgrade bazel rulesets

jbedard:
### Type of change

### Test plan


WORKSPACE
```diff
@@ -54,6 +54,10 @@ nodejs_register_toolchains(
     node_version = "20.11.1",
 )
 
+load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
+
+bazel_skylib_workspace()
```
jbedard, line 59: Does this requirement make upgrading `bazel_skylib` a breaking change? 🤔

jbedard: Although this probably isn't required for our use of `bazel_skylib` in our public APIs, only in our tests? So I will say it is not a breaking change despite rules_js tests failing without it...

gregmagolan (approved):

PR #1538, opened 2024-03-22
feat: extract npm package tars into virtual store as action using tar toolchain from bazel-lib

gregmagolan:
Should mostly fix https://github.com/aspect-build/rules_js/issues/1412. I say mostly since this PR only applies for npm package with no lifecycle hooks and no patches. This is the majority of packages. This PR changes these so they will no longer use a CopyDirectory action with a source directory input to copy into the virtual store. Instead they will use a tar toolchain to extract the package `.tgz` directly into the virtual store.

Packages with lifecycle hooks and patches will be handled in follow-up PRs. For lifecycle hooks, the lifecycle hook binary will need to run `tar --extract` itself. This should be fairly straight forward. For packages with patches, we'll likely need a `patch` toolchain in bazel-lib so we have a hermetic `patch` we can call after `tar --extract` in some "extract and patch" action in both `npm_package_store` and the lifecycle hook binary.

---

### Type of change

### Test plan


npm/private/npm_import.bzl
```diff
@@ -501,11 +501,19 @@ def _download_and_extract_archive(rctx):
...
 
 def _npm_import_rule_impl(rctx):
+    has_lifecycle_hooks = not (not rctx.attr.lifecycle_hooks) or not (not rctx.attr.custom_postinstall)
+    has_patches = not (not rctx.attr.patches)
+
+    package_src = _EXTRACT_TO_DIRNAME
     if utils.is_git_repository_url(rctx.attr.url):
         _fetch_git_repository(rctx)
     else:
         _download_and_extract_archive(rctx)
```
gregmagolan, line 511: TODO: if using `package_src = _TARBALL_FILENAME` then we only need to extract the package.json file and not the whole archive

jbedard:
Can tar within `_download_and_extract_archive` also use the toolchain?

gregmagolan:
> Can tar within `_download_and_extract_archive` also use the toolchain?

It could but I'd rather change one thing at a time and do that in a follow-up. `tar` happens to ship on darwin, linux and Windows so its safe to call `tar` in a repo rule. Not causing any problems in the repo rule atm tho switching to use the toolchain would allow us to remote the `_is_gnu_tar` check since we pass a gnu tar specific flag to `tar --extract`.

jbedard:
> It could but I'd rather change one thing at a time and do that in a follow-up.

Yeah 👍 

I just want to make sure it's considered because the `tar` call in `_is_gnu_tar` is a perf issue on cold builds atm.

npm/private/npm_import.bzl
gregmagolan: Trying this now. Ready for review and landing otherwise.

npm/private/npm_import.bzl
```diff
@@ -501,11 +501,19 @@ def _download_and_extract_archive(rctx):
...
 
+        # TODO: support tarball package_src with lifecycle hooks
+        if not has_patches and not has_lifecycle_hooks:
```
jbedard, line 514: What's the issue with patches and hooks atm?

npm/private/npm_package_store.bzl
```diff
@@ -195,17 +195,41 @@ def _npm_package_store_impl(ctx):
...
-                dst = virtual_store_directory,
-                copy_directory_bin = ctx.toolchains["@aspect_bazel_lib//lib:copy_directory_toolchain_type"].copy_directory_info.bin,
-                # Hardlinking source files in external repositories as was done under the hood
-                # prior to https://github.com/aspect-build/rules_js/pull/1533 may lead to flaky build
-                # failures as reported in https://github.com/aspect-build/rules_js/issues/1412.
-                hardlink = ctx.attr.hardlink,
-                verbose = ctx.attr.verbose,
-            )
+            if src_directory.path.endswith(".tgz"):
+                # npm packages are always published with one top-level directory inside the tarball, tho the name is not predictable
```
jbedard, line 199: Is this convention? Are there docs for that somewhere?

gregmagolan: I only know this experimentally. I never found any docs that say why this is.

gregmagolan: For example, https://registry.npmjs.org/@gar/promisify/-/promisify-1.1.3.tgz has a top-level `package` directory while https://registry.yarnpkg.com/@types/js-yaml/-/js-yaml-4.0.5.tgz has a top-level `js-yaml` directory 🤷‍♂️

gregmagolan: Vast majority I've seen have a top-level `package`

npm/private/npm_import.bzl
gregmagolan: Done and working

MODULE.bazel
```diff
@@ -51,6 +52,16 @@ use_repo(bazel_lib_toolchains, "yq_linux_arm64")
 use_repo(bazel_lib_toolchains, "yq_linux_ppc64le")
 use_repo(bazel_lib_toolchains, "yq_linux_s390x")
 use_repo(bazel_lib_toolchains, "yq_windows_amd64")
+use_repo(bazel_lib_toolchains, "bsd_tar_toolchains")
+
+register_toolchains(
```
thesayyn, line 57: do we need this here? toolchains are already registered by bazel-lib

e2e/pnpm_workspace_deps/MODULE.bazel
```diff
@@ -4,7 +4,7 @@ module(
     compatibility_level = 1,
 )
 
-bazel_dep(name = "aspect_bazel_lib", version = "2.4.2")
+bazel_dep(name = "aspect_bazel_lib", version = "2.5.3")
```
gregmagolan, line 7: aspect_bazel_lib v2.4.2 had the bad coreutils ubuntu security url dep in it. Bump to 2.5.3 is unrelated to this change but it is required to green up CI.

npm/private/npm_package_store.bzl
```diff
@@ -195,17 +195,41 @@ def _npm_package_store_impl(ctx):
...
+                args.add("--directory")
+                args.add(virtual_store_directory.path)
+
+                bsdtar = ctx.toolchains["@aspect_bazel_lib//lib:tar_toolchain_type"]
+                ctx.actions.run(
+                    executable = bsdtar.tarinfo.binary,
+                    inputs = depset(direct = [src_directory], transitive = [bsdtar.default.files]),
+                    outputs = [virtual_store_directory],
+                    arguments = [args],
+                    mnemonic = "Tar",
```
thesayyn, line 218:
```suggestion
                    mnemonic = "TarExtract",
```?

File: npm/private/npm_package_store.bzl [hunk 7]
```diff
@@ -195,17 +195,41 @@ def _npm_package_store_impl(ctx):
...
-                src = src_directory,
-                dst = virtual_store_directory,
-                copy_directory_bin = ctx.toolchains["@aspect_bazel_lib//lib:copy_directory_toolchain_type"].copy_directory_info.bin,
-                # Hardlinking source files in external repositories as was done under the hood
-                # prior to https://github.com/aspect-build/rules_js/pull/1533 may lead to flaky build
-                # failures as reported in https://github.com/aspect-build/rules_js/issues/1412.
-                hardlink = ctx.attr.hardlink,
-                verbose = ctx.attr.verbose,
-            )
+            if src_directory.path.endswith(".tgz"):
```
thesayyn, line 198: is it always `.tgz`? can it end with `.tar.gz`?

e2e/gyp_no_install_script/MODULE.bazel
```diff
@@ -1,4 +1,4 @@
-bazel_dep(name = "aspect_bazel_lib", version = "2.4.2")
+bazel_dep(name = "aspect_bazel_lib", version = "2.5.3")
```
thesayyn, line 1: are these upgrades necessary?

thesayyn (approved):

npm/private/npm_package_store.bzl
gregmagolan: We control the name of the archive when we download it so it will always be `.tgz` but its a good call to be more robust... better test would be to assume its an archive to extract if the final segment of the path does not end with `package` which is the name of the directory we extract to in npm_import.

npm/private/npm_package_store.bzl
gregmagolan: I'll make it `NpmPackageExtract` to isolate it to rules_js

npm/private/npm_package_store.bzl
thesayyn: can we just do check if the src_directory (naming?) is a directory by doing `src_directory.is_directory`?

gregmagolan: It doesn't work since `is_directory` only works for output files. Bazel doesn't track if a source file is a directory or a file. Opted to add a is_tarball util function to check against all of the possible common extensions.

MODULE.bazel
gregmagolan: Looks like it is required. Without it I get,
```
ERROR: /Users/greg/aspect/rules/js/rules_js/BUILD.bazel:70:17: While resolving toolchains for target //:.aspect_rules_js/node_modules/@mycorp/pkg-b (3ac6104): No matching toolchains found for types @@aspect_bazel_lib~1.41.0//lib:tar_toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@@aspect_bazel_lib~1.41.0//lib:tar_toolchain_type'
If platforms or toolchains are a new concept for you, we'd encourage reading https://bazel.build/concepts/platforms-intro.
```

gregmagolan: Ohhh. but only on bazel-lib 1.x and not 2.x.. that implies that I'm missing something in bazel-lib 1.x for toolchain registration.

gregmagolan: Fixed with bazel-lib v1.42.0

PR #1541, opened 2024-03-23
[bzlmod] Mark npm extension as reproducible

dzbarsky:
This stops duplicating all the info from the pnpm lockfile into the Bazel one. Tested by upgrading to Bazel 7.1.1 and `bazel test --lockfile_mode=update //...`. Lockfile dropped from 79K lines to 4.4K.

There are still some gains to be had from python and go rules also using deterministic node, but that's outside the scope of this PR.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -10,7 +10,7 @@ module(
 # Do not bump these unless rules_js requires a newer version to function.
 bazel_dep(name = "aspect_bazel_lib", version = "1.40.3")
 bazel_dep(name = "aspect_rules_lint", version = "0.12.0")
-bazel_dep(name = "bazel_features", version = "0.1.0")
+bazel_dep(name = "bazel_features", version = "1.9.0")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 
 # Keep v5 in to avoid breaking changes.
--- a/npm/extensions.bzl
+++ b/npm/extensions.bzl
@@ -2,6 +2,7 @@
 See https://bazel.build/docs/bzlmod#extension-definition
 """
 
+load("@bazel_features//:features.bzl", "bazel_features")
 load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
 load("//npm:repositories.bzl", "npm_import", "pnpm_repository", _LATEST_PNPM_VERSION = "LATEST_PNPM_VERSION")
 load("//npm/private:npm_translate_lock.bzl", "npm_translate_lock", "npm_translate_lock_lib")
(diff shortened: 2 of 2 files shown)


MODULE.bazel
```diff
@@ -57,7 +57,7 @@ use_repo(bazel_lib_toolchains, "yq_windows_amd64")
 bazel_dep(name = "gazelle", version = "0.35.0", dev_dependency = True, repo_name = "bazel_gazelle")
 bazel_dep(name = "buildifier_prebuilt", version = "6.4.0", dev_dependency = True)
 bazel_dep(name = "bazel_skylib_gazelle_plugin", version = "1.5.0", dev_dependency = True)
-bazel_dep(name = "rules_go", version = "0.41.0", dev_dependency = True)
+bazel_dep(name = "rules_go", version = "0.44.0", dev_dependency = True)
```
jbedard, line 60: Are the updates needed?

dzbarsky: They resolve warnings about the version specified being lower than the resolved one, so I think they are a no-op in terms of compat. But I can revert this if you'd prefer

dzbarsky: I guess it might be nice to have the non-bzlmod versions match, although version resolution with WORKSPACE is a "fun time"

jbedard: If it's not needed lets not do it in this PR. I'm trying to do some in https://github.com/aspect-build/rules_js/pull/1537 atm

jbedard:
If we're starting to work on the lockfile maybe we should checkin one that shoes this improvement?

MODULE.bazel
dzbarsky: ok, reverted

dzbarsky:
> If we're starting to work on the lockfile maybe we should checkin one that shoes this improvement?

I switched the bzlmod example to use updated bazel version and check in the lockfile. Is that fine or do you want a second bzlmod example with lockfile?

jbedard:
> I switched the bzlmod example to use updated bazel version and check in the lockfile. Is that fine or do you want a second bzlmod example with lockfile?

I guess it depends how often it's updated and how annoying/noisy that will be. Any ideas?

DavidZbarsky-at:
> > I switched the bzlmod example to use updated bazel version and check in the lockfile. Is that fine or do you want a second bzlmod example with lockfile?
> 
> I guess it depends how often it's updated and how annoying/noisy that will be. Any ideas?

There's nothing really enforcing the lockfile is up to date so it will only get updated when a human runs bazel in that workspace and there are dependency changes. (It would be nice to enforce but there are still os/arch-specific lockfile sections so it was annoying for me to match the lockfile to what CI expected)

jbedard (approved):

jbedard (approved):

PR #1543, opened 2024-03-23
refactor: remove extra npm_translate_lock_repo alias target

jbedard:
I think this alias is a private target and now unnecessary?

---

### Type of change

### Test plan


npm/private/test/fsevents_links_defs_checked.bzl
```diff
@@ -115,13 +115,13 @@ def npm_imported_package_store(name):
     _js_run_binary(
         name = "{}/lc".format(store_target_name),
         srcs = [
-            "@npm//:npm__fsevents__2.3.2_pkg",
```
jbedard, line 118: This is the summary of the change, not generating or using this alias...

gregmagolan:
This was added when bzlmod was first added and it was needed at the time. Maybe that was in Bazel 6 and things changed in Bazel 7.

jbedard:
Note that bazel 6 and 7 are tested on CI atm. We could potentially delay this until v2 if we are worried about bazel 5?

gregmagolan:
Looks like the visibility issues date back to the early days of bzlmod so the alias is indeed no longer needed.

gregmagolan (approved):

PR #1547, opened 2024-03-26
fix: do not generate repositories.bzl for bzlmod

jbedard:
Technically this shouldn't be a breaking change, the `repositories.bzl` should not be used within bzlmod, but maybe we should keep this in the 2.x branch?

### Type of change

### Test plan


npm/private/npm_translate_lock.bzl
```diff
@@ -80,6 +80,7 @@ _ATTRS = {
     "prod": attr.bool(),
     "public_hoist_packages": attr.string_list_dict(),
     "quiet": attr.bool(default = True),
+    # TODO(bzlmod): remove repositories_bzl_filename
```
gregmagolan, line 83: Do you mean remove when we drop support for WORKSPACE?

gregmagolan (approved):
👍

npm/private/npm_translate_lock.bzl
jbedard: Yeah, do we have a keyword I should put here instead of `bzlmod`?

PR #1549, opened 2024-03-27
fix: change order of node_modules and app in `js_image_layer`

SinimaWath:
Usually, application's `node_modules` don't change often, as a application `code`, so I suggest to provide "strong" default: `[node_modules, app]`

### Type of change

Make the `node_modules` layer the first in the `js_image_layer` rule, as the `node_module` don't change often.

### Test plan

### DIFF
--- a/js/private/js_image_layer.bzl
+++ b/js/private/js_image_layer.bzl
@@ -284,7 +284,7 @@ def _js_image_layer_impl(ctx):
     node_modules = _build_layer(ctx, type = "node_modules", entries = node_modules_entries, inputs = node_modules_inputs)
 
     return [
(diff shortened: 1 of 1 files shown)


SinimaWath:
@thesayyn What do you think?

thesayyn:
I'd like to but technically this is a breaking change. I am hesitant.

SinimaWath:
@thesayyn 
So I don't see any problem, let's add that change in 2.x.x release? WDYT?

thesayyn:
I'll add this to 2.x.

gregmagolan (approved):
Merging into 2.x branch. Thanks for the fix.

PR #1554, opened 2024-03-29
perf: don't set JsInfo declarations and transitive_declarations in npm_link_package_store

gregmagolan:
While working on the rules_js 2.0 branch where I thought I needed to make changes to the JsInfo provider to get this perf improvement, I realized that the source of the issue is that `npm_link_package_store` sets `declarations` and `transitive_declarations` which it does not need to set since linked npm packages are already in runfiles for runtime dependencies and ts_project doesn't need them in declarations.

The perf improvement here is that npm_package (when include_runfiles is set to False) will no longer get all npm packages in its manifest. In rules_js 2.0, npm_package include_runfiles can flip to False and this perf improvement will be the default case.

---

### Type of change

### Test plan

### DIFF
--- a/js/private/test/js_library_test.bzl
+++ b/js/private/test/js_library_test.bzl
@@ -39,15 +39,9 @@ def _declarations_test_impl(ctx):
 
     # transitive_declarations should have the source declarations and transitive declarations
     transitive_declarations = target_under_test[JsInfo].transitive_declarations.to_list()
-
-    # the transitive count might be 3 or 4 depending on the type of symlink created.
-    # See utils.make_symlink()
-    asserts.true(env, len(transitive_declarations) == 3 or len(transitive_declarations) == 4)
+    asserts.true(env, len(transitive_declarations) == 2)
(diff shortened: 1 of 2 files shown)


jbedard (approved):

dgp1130:
Just FYI, I think there's a breaking change here which invalidates a `types_only` rule as suggested [here](https://github.com/aspect-build/rules_ts/issues/319). I had a custom implementation of this which is broken when upgrading to `@aspect_rules_js` v1.40.1.

I was able to fix things by depending on the regular NPM package (no limiting to only types) and then retroactively "pruning" the dependency by removing it from the `NpmPackageInfo` provider. It's not exactly elegant, but works well enough for my use case. Manually limiting to `import type` also works, but is harder to guarantee. I added some extra checks to ensure my package doesn't get duplicated on accident, which was the main thing `types_only` was trying to prevent.

https://github.com/dgp1130/rules_prerender/compare/389383417b9e0dbc4c1e1b8c47f2b4ee71648f6e...313314e9d2c0edd2154d73c27f61d61f86d7715c

PR #1564, opened 2024-04-02
fix: add cannonical_id to npm_import

Aghassi:
This prevents npm_import from pulling data with the same integrity.

---

### Type of change

### Test plan

TBD

TL;DR: @joeljeske and I discovered a situation where the SHA mismatched on a url pull through and this needed to be added to fix 

https://docs.bazel.build/versions/3.0.0/repo/http.html#http_archive-canonical_id


jbedard:
This aligns with `http_archive` and `http_file` which set `canonical_id = download_url` by default.

Any thoughts @gregmagolan?

jbedard (approved):

gregmagolan (approved):

PR #1571, opened 2024-04-02
refactor: make unresolved_symlinks_enabled attribute of js_binary mandatory

gregmagolan:


js/private/js_binary.bzl
```diff
@@ -267,8 +267,7 @@ _ATTRS = {
         attribute based on a `config_setting` rule.
         See /js/private/BUILD.bazel in rules_js for an example.
         """,
-        # TODO(2.0): make this mandatory so that downstream binary rules that inherit these attributes are required to set it
```
jbedard, line 270: There are some `hasattr(ctx.attr, "unresolved_symlinks_enabled")` calls that we can maybe delete at the same time?

gregmagolan: They all get deleted further up the stack so it will be easier to unwind the stack by just continuing down it

js/private/js_binary.bzl
```diff
@@ -267,8 +267,7 @@ _ATTRS = {
...
-        mandatory = False,
+        mandatory = True,
     ),
```
jbedard, line 271: Can we provide a default or does that always have to be manually specific with the `select` + `@aspect_rules_js//js:enable_runfiles`?

jbedard (approved):

js/private/js_binary.bzl
gregmagolan: The default is not available in the rules context... but this attribute gets deleted further down the stack anyway.

PR #1579, opened 2024-04-03
refactor: set minimum rules_nodejs version to 6.1.0

gregmagolan:


alexeagle:
Let's just land that first so you don't have to bump again

gregmagolan:
> Let's just land that first so you don't have to bump again

It's just waiting on a stamp

alexeagle:
merged it

jbedard (approved):

PR #1583, opened 2024-04-03
refactor: use Aspect bazel-lib repo_utils.get_env_var to get HOME directory

gregmagolan:


jbedard (approved):

npm/private/utils.bzl
```diff
@@ -404,13 +404,8 @@ if [ ! -f $1 ]; then exit 42; fi
     else:
         fail(INTERNAL_ERROR_MSG)
 
-# TODO(2.0): move this to aspect_bazel_lib
 def _home_directory(rctx):
```
jbedard, line 407: What about moving this full method to bazel-lib instead of hiding the handling of HOME within `get_env_var`?

This PR is good now either way, just a thought and another thing we can do before 2.0...

gregmagolan: I thought about it but decided that `get_env_var(rctx, "HOME")` was already clean enough in bazel-lib instead of a new get_home_directory.

PR #1584, opened 2024-04-03
refactor: use consistent_label_str from Aspect bazel-lib

gregmagolan:


js/private/js_binary.bzl
```diff
@@ -329,18 +329,6 @@ _NODE_OPTION = """JS_BINARY__NODE_OPTIONS+=(\"{value}\")"""
 def _target_tool_path_to_short_path(tool_path):
     return ("../" + tool_path[len("external/"):]) if tool_path.startswith("external/") else tool_path
 
-# Generate a consistent label string between Bazel versions.
-# TODO(2.0): hoist this function to bazel-lib and use from there (as well as the dup in npm/private/utils.bzl)
```
jbedard, line 333: Can we also drop the one in npm/private/utils.bzl that is mentioned here?

gregmagolan: Good catch. I missed a few spots.

gregmagolan: I'll do that after the Bazel 5 removal commit lands

jbedard (approved):

PR #1586, opened 2024-04-03
fix: drop .sh extension from js_binary, merger and js_image_layer launchers

### LINKED ISSUE #845: [Bug]: `js_run_devserver()` leaks `.sh` in file name
### What happened?

The `js_run_devserver()` target (on Linux) creates an executable file with the name `%{name}.sh`, which leaks the fact that this is implemented with Bash under the hood and forces any other scripts which call it via runfiles to add `.sh` to the filenames. This also likely complicates cross-platform support, since the filename is different for both.

```
$ bazel query --output build //common/testing:webdriver_test_server
INFO: Invocation ID: ed977d5a-d7cf-47bb-9839-fd2d99d08555
# /home/doug/Source/rules_prerender/common/testing/BUILD.bazel:268:24
js_run_devserver(
  name = "webdriver_test_server",
  tags = ["ibazel_notify_changes"],
  generator_name = "webdriver_test_server",
  generator_function = "web_resources_devserver",
  generator_location = "common/testing/BUILD.bazel:268:24",
  testonly = True,
  data = ["//common/testing:webdriver_test_resources"],
  entry_point = "@aspect_rules_js//js/private:js_devserver_entrypoint",
  enable_runfiles = select({"@aspect_rules_js//js/private:enable_runfiles": True, "//conditions:default": False}),
  tool = "//common/testing:webdriver_test_server_baked_server",
)

$ bazel build //common/testing:webdriver_test_server
INFO: Invocation ID: 9315b831-1936-4983-9a2e-14a064f1e6fa
INFO: Analyzed target //common/testing:webdriver_test_server (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //common/testing:webdriver_test_server up-to-date:
  dist/bin/common/testing/webdriver_test_server.sh
INFO: Elapsed time: 0.207s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
```

### Version

Host is WSL2 in Windows 10.

```
$ bazel --version
bazel 5.1.1
```

Using `@aspect_rules_js` version `1.16.0` via `WORKSPACE`.

### How to reproduce

```shell
load("@npm_rules_js//:http-server/package_json.bzl", http_server_bin = "bin")

http_server_bin.http_server_binary(
    name = "server",
)

js_run_devserver(
    name = "devserver",
    tool = ":server",
)

# Omitting runfiles setup...
sh_binary(
    name = "executor",
    srcs = ["executor.sh"],
    data = [":devserver"],
)
```

```bash
# Error: `devserver` does not exist in runfiles.
$(rlocation devserver)

# Works.
$(rlocation devserver.sh)
```

### Any other information?

Ideally the binary file should always be generated with the same name as the target. This is more intuitive to use and more accurately abstracts away the underlying implementation so it can be changed without breaking existing API contracts.

### Fund our work

gregmagolan:
Fixes #845

### DIFF
--- a/e2e/worker/BUILD.bazel
+++ b/e2e/worker/BUILD.bazel
@@ -16,7 +16,7 @@ copy_file(
 )
 
 js_binary(
-    name = "worker",
+    name = "worker_binary",
     data = [":copy_worker_js"],
     entry_point = ":copy_dummy_worker",
     visibility = ["//visibility:public"],
--- a/e2e/worker/defs.bzl
+++ b/e2e/worker/defs.bzl
@@ -31,7 +31,7 @@ pi_rule = rule(
         "worker": attr.label(
             executable = True,
             cfg = "exec",
-            default = ":worker",
+            default = ":worker_binary",
         ),
     },
 )
--- a/examples/coverage/BUILD.bazel
+++ b/examples/coverage/BUILD.bazel
@@ -1,7 +1,7 @@
 load("@aspect_rules_js//js:defs.bzl", "js_test")
 
 js_test(
-    name = "coverage",
+    name = "coverage_test",
     data = [
         "coverage.js",
         "//:node_modules/@types/node",
--- a/examples/js_binary/BUILD.bazel
+++ b/examples/js_binary/BUILD.bazel
@@ -471,7 +471,7 @@ write_file(
         { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
         # --- end runfiles.bash initialization v2 ---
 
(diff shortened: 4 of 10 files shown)


alexeagle (approved):
I think we might hit some roadbumps with this, should find out as we roll a release candidate to clients

gregmagolan:
> I think we might hit some roadbumps with this, should find out as we roll a release candidate to clients

Yeah, the `require('foo')` where there is both a `foo.js` and `foo` in the directory prefers `foo` oddly. I doubt that many users will hit that case but you never know.

PR #1588, opened 2024-04-03
refactor: remove unused NpmLinkedPackageInfo provider and corresponding unused npm_linked_packages from JsInfo; rename load bearing npm_linked_package_files to npm_linked_packages

gregmagolan:


js/private/js_helpers.bzl
```diff
@@ -81,44 +81,15 @@ def gather_npm_linked_packages(srcs, deps):
...
         for target in srcs + deps
-        if JsInfo in target and hasattr(target[JsInfo], "transitive_npm_linked_packages")
-    ])
-
-    # transitive_npm_linked_package_files
-    transitive_npm_linked_package_files = depset([], transitive = npm_linked_package_files + [
-        target[JsInfo].transitive_npm_linked_package_files
-        for target in srcs + deps
-        if JsInfo in target and hasattr(target[JsInfo], "transitive_npm_linked_package_files")
+        if JsInfo in target and hasattr(target[JsInfo], "npm_linked_packages")
```
jbedard, line 90: Why is the `hasattr` needed? Shouldn't all `JsInfo` contain `.npm_linked_packages`?

jbedard (approved):

js/private/js_helpers.bzl
gregmagolan: Starlark doesn't enforce that you set all fields when you create a provider so a custom rule in some 3rd party rule set could miss a field.

PR #1589, opened 2024-04-03
refactor: drop support for Bazel 5

gregmagolan:
Dropping Bazel 5 support allows for many simplifications:

- Assume that unresolved symlinks are supported (added in Bazel 6) so remove support for legacy `ctx.actions.declare_file` + `ctx.actions.symlink` in favor of `ctx.actions.declare_symlink`. Rules no longer have to take a `unresolved_symlinks_enabled` or `use_declare_symlink ` parameter.

- Remove legacy node fs patches which were required for how Bazel 5 layed out symlinks in runfiles and the sandbox.

- Remove all uses of consistent_label_str since Bazel 5 was the version with inconsistent labels compared to Bazel 6 & 7

- We can assume labels that start with `@@` are support since that support was added with bzlmod in Bazel 6


npm/private/utils.bzl
```diff
@@ -254,32 +254,18 @@ def _virtual_store_name(name, version):
...
-            # BREAKING CHANGE in Bazel 6 requires you to use declare_directory if your target_file
-            # in ctx.actions.symlink is a directory artifact
-            symlink = ctx.actions.declare_directory(symlink_path)
-        else:
-            symlink = ctx.actions.declare_file(symlink_path)
-        ctx.actions.symlink(
-            output = symlink,
-            target_file = target_file,
-        )
+    if not is_bazel_6_or_greater():
```
jbedard, line 257: Is the `make_symlink` util the best place to put this? 🤷

pnpm-lock.yaml
```diff
@@ -9,7 +9,7 @@ overrides:
   semver-max: file:./npm/private/test/vendored/semver-max
   is-odd: file:./npm/private/test/vendored/is-odd
 
-packageExtensionsChecksum: e978e3e510efb71c3833e1915611f5d9
```
jbedard, line 12: Why did this change? 🤔

.aspect/rules/external_repository_action_cache/npm_translate_lock_LTYxNjU0OTY4OQ==
```diff
@@ -0,0 +1,26 @@
+# @generated
```
jbedard, line 1: Why was this added?

js/private/test/BUILD.bazel
```diff
@@ -35,8 +35,8 @@ write_source_files(
         "shellcheck_launcher.sh": ":shell_launcher_sed",
     },
     target_compatible_with = select({
-        "@aspect_bazel_lib//lib:bzlmod": ["@platforms//:incompatible"],
```
jbedard, line 38: Why switch this one? Should generate+checkin both instead?

.aspect/rules/external_repository_action_cache/npm_translate_lock_LTYxNjU0OTY4OQ==
gregmagolan: The label format that went into the hash generation changed. Fixed it up.

pnpm-lock.yaml
gregmagolan: Good question 🤔

gregmagolan: Unwound it and the delta went away. No idea what happened. 🤷

js/private/test/BUILD.bazel
gregmagolan: Yup. Better to just add both separate targets and test both. ✅

npm/private/npm_translate_lock_state.bzl
```diff
@@ -161,9 +161,11 @@ def _init_pnpm_labels(label_store, rctx):
 def _init_update_labels(priv, rctx, label_store):
     attr = rctx.attr
 
+    pnpm_lock_label = label_store.label("pnpm_lock")
+    pnpm_lock_label_str = "//{}:{}".format(pnpm_lock_label.package, pnpm_lock_label.name)
```
gregmagolan, line 165: We need to maintain the same label format in this commit as `consistent_label_str` generated so that our input string to the hash function below doesn't change else users will get a new action cache lock file on this upgrade which is undesirable

npm/private/utils.bzl
gregmagolan: Not the best place. I'll move it.

gregmagolan: Actually. I'll move this in a follow-up since the best place for the check doesn't exist yet in this commit.

alexeagle (approved):

PR #1590, opened 2024-04-03
chore: write checked bzl files for both WORKSPACE and bzlmod

gregmagolan:
Adds missing snapshot test fixture for bzlmod.

Pre-factored out of https://github.com/aspect-build/rules_js/pull/1589.

---

### Type of change

### Test plan

- New test cases added


alexeagle (commented):
Did you give some thought to whether we could make some assertions about the behavior in addition / instead of all the snapshot files? When they change it's hard for most of us to tell whether that change is desirable/correct

alexeagle (approved):

gregmagolan:
> Did you give some thought to whether we could make some assertions about the behavior in addition / instead of all the snapshot files? When they change it's hard for most of us to tell whether that change is desirable/correct

We have a few "contains" assertions in `e2e/npm_translate_lock/test.sh`. That bash script could probably be replaced with `assert_contains` and `assert_does_not_contain` and we can expand on that pattern with more assertions like that in the future.

We also have some Starlark unit tests on the generated bzl in `npm/private/test/generated_pkg_json_test.bzl`. We can expand on that pattern as well.

cc @jbedard

npm/private/test/BUILD.bazel
```diff
@@ -30,28 +31,46 @@ yaml_tests(name = "test_yaml")
 
 parse_pnpm_lock_tests(name = "test_parse_pnpm_lock")
 
-generated_pkg_json_test(name = "test_generated_pkg_json")
+# generated_pkg_json_test(name = "test_generated_pkg_json")
```
jbedard, line 34: @gregmagolan was this intentional?

gregmagolan: Whoops. That was commented out during testing and should not have landed. I'll update it.

PR #1591, opened 2024-04-03
fix: add package_visibility support to npm_translate_lock and npm_import on bzlmod

gregmagolan:
This was missed when `package_visibility` support was added

---

### Type of change`

### Test plan


alexeagle (approved):

jbedard (commented):
👍

PR #1592, opened 2024-04-03
chore: add a bzlmod variant of the js_binary_sh test for JS_BINARY__WORKSPACE

gregmagolan:
This shows the difference in the env var between WORKSPACE and bzlmod.

---

### Type of change

### Test plan

- New test cases added


jbedard (commented):
👍

PR #1593, opened 2024-04-03
refactor: upgrade to latest rules_nodejs 6.x and move toolchain registration call to new rules_js_configure WORKSPACE function

gregmagolan:
---

### Type of change

### Test plan

- New test cases added


WORKSPACE
```diff
@@ -3,38 +3,23 @@ workspace(
...
-load("//js:repositories.bzl", "rules_js_dependencies")
+load("@aspect_rules_js//js:repositories.bzl", "rules_js_dependencies")
 
 rules_js_dependencies()
 
-load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies", "register_jq_toolchains")
+load("@aspect_rules_js//js:configure.bzl", "rules_js_configure")
 
-aspect_bazel_lib_dependencies()
+rules_js_configure(node_version = "16.14.2")
```
alexeagle, line 16: it bothers me that "configure" is invented here, isn't there an existing convention for that? let's look in some other languages

e2e/js_run_devserver/MODULE.bazel
```diff
@@ -14,6 +14,13 @@ local_path_override(
     path = "../..",
 )
 
+archive_override(
```
alexeagle, line 17: close to a rnjs release so we don't need these?

WORKSPACE
gregmagolan: There is `rules_go_register_toolchains` but that doesn't quite describe what I'm after here since it can be more than just toolchains that are configured.

e2e/js_run_devserver/MODULE.bazel
gregmagolan: yeah... I'm just waiting on a few more PRs to alnd and I'll cut 6.1.0 and update this PR

npm/private/npm_package.bzl
```diff
@@ -506,14 +506,6 @@ def stamped_package_json(name, stamp_var, **kwargs):
 
     For more information on stamping, read https://docs.aspect.build/rules/aspect_bazel_lib/docs/stamping.
 
-    Using this rule requires that you register the jq toolchain in your WORKSPACE:
```
jbedard, line 509: Isn't this still true? However there's now just another way of doing it?

jbedard:
What was the major reason for this? I don't recall if we've discussed it before?

gregmagolan:
> What was the major reason for this? I don't recall if we've discussed it before?

Primarily because we're sneaking in toolchain registrations in WORKSPACE in npm_translate_lock and npm_import which is not a great convention and bad layering. In theory a user could use rules_js without npm_translate_lock and npm_import and they would have to look into the implementation of npm_translate_lock to see what toolchains are needed. Adding an explicity `rules_js_configure` makes rules_js more like other rule sets where toolchains are registered explicitly in WORKSPACE and not implicitly in an unrelated repository rule.

npm/private/npm_package.bzl
gregmagolan: It will get registered implicitly in rules_js_configure now so there is no need to document all of the required toolchains in each rule docstring. jq was a one off here and we don't do this for other toolchains required for other rules.

WORKSPACE
jbedard: What else would we add other then registering toolchains?

npm/repositories.bzl
```diff
@@ -1,16 +1,12 @@
...
 
-load("@bazel_features//:deps.bzl", "bazel_features_deps")
 load("//npm/private:npm_import.bzl", _npm_import = "npm_import")
 load("//npm/private:npm_translate_lock.bzl", _list_patches = "list_patches", _npm_translate_lock = "npm_translate_lock")
 load("//npm/private:pnpm_repository.bzl", _LATEST_PNPM_VERSION = "LATEST_PNPM_VERSION", _pnpm_repository = "pnpm_repository")
 
-npm_import = _npm_import
-
-def npm_translate_lock(**kwargs):
-    bazel_features_deps()
```
jbedard, line 11: Was this never needed? Can it be dropped in 1.x?

WORKSPACE
jbedard: Just to outline and clarify what we have...

Today:
* `rules_js_dev_dependencies` which is internal repository deps, not public API so who cares
* `rules_js_dependencies` which is essentially repository deps
* `nodejs_register_toolchains` from rules_nodejs for node toolchains
* `npm_translate_lock` and `npm_import` have various `register_(copy_directory|copy_to_directory|coreutils|tar|yq)_toolchains` params to register bazel-lib toolchains
* ^ get set to false in many places, such as bzlmod wrappers of those, `pnpm_repository` etc

Future (bzlmod):
* All this goes away. Is anything remaining?

This PR:
* `rules_js_configure` with bool params and a lot of `node*` params to do all this? The `register_*` params go away.

Is that right?

jbedard: I'd like to suggest:
* Add `rules_js_register_toolchains(... all the bool params ...)` and all the `register_*` bools can be dropped
* Keep `nodejs_register_toolchains` as-is and don't put it in this new method, we can always think about it after an initial PR?

npm/repositories.bzl
gregmagolan: The WORKSPACE use of it was dropped on the 2.x branch; still needed on 1.x

WORKSPACE
gregmagolan: Going with `rules_js_register_toolchains`

e2e/js_image_docker/MODULE.bazel
```diff
@@ -0,0 +1,6 @@
+###############################################################################
```
jbedard, line 1: Was this intentional?

gregmagolan: No. Looks like it was auto-generated and snuck in. Will delete.

jbedard (approved):

PR #1594, opened 2024-04-03
perf: reduce size of npm_package_store transitive_files depset

gregmagolan:
I've been waiting to make this optimization for a long time

BREAKING CHANGES:

If using npm packages with genrule, the default output of a target such as `//:node_modules/foo` is _only_ the link directory so if you will also need to add `//:node_modules/foo/dir` which is the virtual store directory to the genrule srcs.

### DIFF
--- a/examples/genrule/BUILD.bazel
+++ b/examples/genrule/BUILD.bazel
@@ -67,7 +67,9 @@ genrule(
     name = "require_acorn",
     srcs = [
         ":require_acorn_js",
+        # reference the location where the "acorn" npm package was linked in our root Bazel package.
         "//:node_modules/acorn",
+        "//:node_modules/acorn/dir",
     ],
     outs = ["out2"],
     cmd = """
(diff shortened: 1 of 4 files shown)


examples/genrule/BUILD.bazel
```diff
@@ -67,7 +67,9 @@ genrule(
     name = "require_acorn",
     srcs = [
         ":require_acorn_js",
+        # reference the location where the "acorn" npm package was linked in our root Bazel package.
         "//:node_modules/acorn",
+        "//:node_modules/acorn/dir",
```
alexeagle, line 72: some users will see this breaking change too? need to document in the commit?

alexeagle (approved):

examples/genrule/BUILD.bazel
gregmagolan: 👍  will mention that.

npm/private/utils.bzl
```diff
@@ -262,9 +261,7 @@ def _make_symlink(ctx, symlink_path, target_file):
         output = symlink,
         target_path = relative_file(target_file.path, symlink.path),
     )
-    files.append(target_file)
```
jbedard, line 265: What allowed this to be dropped?

PR #1601, opened 2024-04-05
chore: bump minimum aspect_bazel_lib version to 2.7.0 for rules_js 2.0

gregmagolan:
All upstream changes needed for rules_js 2.0 are in aspect_bazel_lib 2.7.0.

### DIFF
--- a/e2e/bzlmod/MODULE.bazel
+++ b/e2e/bzlmod/MODULE.bazel
@@ -5,7 +5,7 @@ module(
 )
 
 bazel_dep(name = "aspect_rules_js", version = "0.0.0")
-bazel_dep(name = "aspect_bazel_lib", version = "2.6.1")
+bazel_dep(name = "aspect_bazel_lib", version = "2.7.0")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 
 local_path_override(
--- a/e2e/pnpm_workspace_deps/WORKSPACE
+++ b/e2e/pnpm_workspace_deps/WORKSPACE
@@ -1,20 +1,7 @@
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
-
 local_repository(
     name = "aspect_rules_js",
     path = "../..",
 )
-
-http_archive(
-    name = "aspect_bazel_lib",
-    sha256 = "ac6392cbe5e1cc7701bbd81caf94016bae6f248780e12af4485d4a7127b4cb2b",
-    strip_prefix = "bazel-lib-2.6.1",
-    url = "https://github.com/aspect-build/bazel-lib/releases/download/v2.6.1/bazel-lib-v2.6.1.tar.gz",
-)
-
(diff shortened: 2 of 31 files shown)


e2e/pnpm_workspace_deps/WORKSPACE
```diff
@@ -1,17 +1,8 @@
...
 )
 
-http_archive(
-    name = "aspect_bazel_lib",
-    sha256 = "ac6392cbe5e1cc7701bbd81caf94016bae6f248780e12af4485d4a7127b4cb2b",
-    strip_prefix = "bazel-lib-2.6.1",
-    url = "https://github.com/aspect-build/bazel-lib/releases/download/v2.6.1/bazel-lib-v2.6.1.tar.gz",
-)
-
 load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies")
```
jbedard, line 6: We probably shouldn't be calling this if we aren't loading it anymore? Can we also remove this? Can we also remove it from MODULE if we can from here?

e2e/bzlmod/MODULE.bazel
```diff
@@ -5,7 +5,7 @@ module(
 )
 
 bazel_dep(name = "aspect_rules_js", version = "0.0.0")
-bazel_dep(name = "aspect_bazel_lib", version = "2.6.1")
+bazel_dep(name = "aspect_bazel_lib", version = "2.7.0")
```
jbedard, line 8: Shouldn't bzlmod take the latest version in the deep tree and therefore none of the e2e changes are necessary?

e2e/pnpm_workspace_deps/WORKSPACE
gregmagolan: Good catch

jbedard:
What are the core changes this brings in that 2.0 requires? Maybe add that to the commit message?

e2e/pnpm_workspace_deps/WORKSPACE
gregmagolan: Some of the MODULE.bazel instances will still be needed if there are load statements to aspect_bazel_lib in that e2e test.. I'll reduce down to the minimal set.

gregmagolan:
> What are the core changes this brings in that 2.0 requires? Maybe add that to the commit message?

I'll mention it in the release notes for rules_js 2.0 since I'll be compiling all the BREAKING CHANGES there

e2e/bzlmod/MODULE.bazel
jbedard: That doesn't mean the MODULE for that test needs the version instrumented though, does it?

gregmagolan: It has to specify the dependency as a direct dependency same as you would have to in package.json for a direct dep if you're not hoisting. Do you mean that it can specify the dep but with version 0.0.0?

PR #1603, opened 2024-04-05
feat: support rules_nodejs improved toolchain attributes in rules_nodejs 6.1.0 with backward compat for older versions

gregmagolan:
rules_nodejs 6.1.0 toolchain exports canonical and more ergonomic File objects instead of paths. This change supports those new attributes while keeping backward compat for older versions of rules_nodejs. In the future, after some sufficient grace period, support for the older less ergonomic attributes can be dropped.


alexeagle (approved):

gregmagolan:
Replaced by https://github.com/aspect-build/rules_js/pull/1603

PR #1604, opened 2024-04-06
feat: support rules_nodejs improved toolchain attributes in rules_nodejs 6.1.0 with backward compat for older versions

gregmagolan:
rules_nodejs 6.1.0 toolchain exports canonical and more ergonomic File objects instead of paths. This change supports those new attributes while keeping backward compat for older versions of rules_nodejs. In the future, after some sufficient grace period, support for the older less ergonomic attributes can be dropped.

---

### Type of change

### Test plan

### DIFF
--- a/js/private/coverage/merger.bzl
+++ b/js/private/coverage/merger.bzl
@@ -16,17 +16,23 @@ _ATTRS = {
 # Do the opposite of _to_manifest_path in
 # https://github.com/bazelbuild/rules_nodejs/blob/8b5d27400db51e7027fe95ae413eeabea4856f8e/nodejs/toolchain.bzl#L50
 # to get back to the short_path.
-# TODO(2.0): fix toolchain so we don't have to do this
-def _target_tool_path_to_short_path(tool_path):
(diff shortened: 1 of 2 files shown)


alexeagle (approved):

js/private/coverage/merger.bzl
```diff
@@ -16,17 +16,23 @@ _ATTRS = {
 # Do the opposite of _to_manifest_path in
 # https://github.com/bazelbuild/rules_nodejs/blob/8b5d27400db51e7027fe95ae413eeabea4856f8e/nodejs/toolchain.bzl#L50
 # to get back to the short_path.
-# TODO(2.0): fix toolchain so we don't have to do this
-def _target_tool_path_to_short_path(tool_path):
+# TODO(3.0): remove this after a grace period for the DEPRECATED toolchain attributes
```
jbedard, line 19: Should this be 2.0?

gregmagolan: Nope. I'm punting removing support for rules_nodejs < 6.1.0 to a future 3.0 release

PR #1610, opened 2024-04-06
refactor: remove js_image_layer bazel5 legacy symlink detection

jbedard:
### Type of change

### Test plan


thesayyn (commented):
you need to update rule impl as well

jbedard:
> you need to update rule impl as well

Good point, thanks! (and updated...)

gregmagolan (approved):
🌮

PR #1615, opened 2024-04-07
refactor: rename js_filegroup to js_info_files

gregmagolan:
This was badly named originally and its not clear what it does from its naming. `js_info_files` is better since it describes the tie to `JsInfo` and what it produces and the naming won't be associated with the built-in `filegroup`.

This rules is public API but it is not generally used explicitly. It primary use is under the hood of `js_run_binary`.


js/private/js_info_files.bzl
```diff
@@ -17,9 +17,9 @@ def _js_filegroup_impl(ctx):
         include_npm_linked_packages = ctx.attr.include_npm_linked_packages,
     ))
 
-js_filegroup = rule(
+js_info_files = rule(
```
jbedard, line 20: Normally "JsInfo" is one word, like "filegroup" and many other things. But our rules already start with `js_*` so 🤷

jbedard (approved):

js/private/js_info_files.bzl
gregmagolan: I think the JsInfo camel case translates over to js_info fine. There is already a `js_info` helper function so this will be consistent with that.

PR #1618, opened 2024-04-07
refactor: rename npm_link_package_deps to npm_deps and npm_package_store_deps to npm_package_store_infos

gregmagolan:
Improved naming for these so they are self-documenting.

Backward-compat handled in js_binary for the deprecated `include_npm_linked_packages` attribute.


js/defs.bzl
```diff
@@ -26,15 +26,29 @@ load(
     _js_image_layer = "js_image_layer",
 )
 
+# buildifier: disable=function-docstring
 def js_binary(**kwargs):
+    include_npm_deps = kwargs.pop("include_npm_deps", False)
+
+    # For backward compat
+    # TODO(3.0): remove backward compat handling
+    include_npm_linked_packages = kwargs.pop("include_npm_linked_packages", None)
```
jbedard, line 35: Why not drop it in v2? Maybe add this warning in v1? I don't think we we necessarily *need* a full semver warning people?

js/private/js_info.bzl
```diff
@@ -4,8 +4,8 @@ JsInfo = provider(
     doc = "Encapsulates information provided by rules in rules_js and derivative rule sets",
     fields = {
         "declarations": "A depset of declaration files produced by the target",
-        "npm_linked_packages": "A depset of files in npm linked package dependencies of the target and the target's transitive deps",
-        "npm_package_store_deps": "A depset of NpmPackageStoreInfo providers from non-dev npm dependencies of the target and the target's transitive dependencies to use as direct dependencies when linking downstream npm_package targets with npm_link_package",
+        "npm_deps": "A depset of files in npm dependencies of the target and the target's transitive deps",
```
jbedard, line 7: Why add the term `deps` here while removing it elsewhere? None of the other fields here use the term after changing `npm_package_store_*` from "deps" to "infos"...

gregmagolan:
Splitting these two renames into two PRs starting with: https://github.com/aspect-build/rules_js/pull/1620

js/defs.bzl
gregmagolan: Adding the warning in V1 means that if someone updates from an older version of v1 to v2 then they will get a fail instead of a WARNING. Keeping backward compat with a warning is generally good practice where possible. Was not possible for all of the breaking changes in V2 but for this one we can give users a nice hint of what field to upgrade to instead of getting the Bazel failure of "invalid attribute".

js/private/js_info.bzl
gregmagolan: It felt out of place on `npm_package_store_deps`. More reasoning on that here: https://github.com/aspect-build/rules_js/pull/1620

`npm_linked_packages` doesn't seem like the right name either since the depset includes both linked package and transitive deps in the packages store.

Might be able to find something better than `npm_deps`. I'll open up another PR and we can bikeshed there.

PR #1619, opened 2024-04-07
refactor: rename declarations to types in JsInfo & throughout rules_js

gregmagolan:
`declarations` are a subset of types. For example, we include all `.json` files in "declarations" currently since typescript can resolve types from JSON files and may need package.json "types" field to resolve typings.

https://github.com/aspect-build/rules_js/blob/fb366b1f47eacea59f283ba4ebd1fc1bed33721a/js/private/js_library.bzl#L143


alexeagle (approved):

alexeagle:
SGTM, "declarations" was indeed a subset of typings.

PR #1630, opened 2024-04-08
perf: construct single list of linked packages

jbedard:
### Type of change

### Test plan


gregmagolan (approved):
😎

PR #1638, opened 2024-04-10
chore: bring in llvm toolchain so we can update //js/private/test/image snapshots on darwin

gregmagolan:


e2e/npm_link_package-esm/MODULE.bazel
```diff
@@ -0,0 +1,6 @@
+###############################################################################
```
jbedard, line 1: Delete...

gregmagolan: ✅

gregmagolan:
Re-based and targeted `main` for this one. I'll consolidate the duplicate http_archives before landing.

e2e/npm_link_package-esm/MODULE.bazel
jbedard: It's still there...

jbedard (approved):

e2e/npm_link_package-esm/MODULE.bazel
gregmagolan: 🤦

gregmagolan: huh... git commit says they are removed
```$ gd HEAD~ --name-only
.bazelrc
WORKSPACE
WORKSPACE.bzlmod
js/dev_repositories.bzl
```
I think the GH UI is being buggy here

gregmagolan: There we go. Pushed a few more times and it updated correctly.

jbedard (approved):

PR #1642, opened 2024-04-11
fix: set default internal npm_translate_lock state

jbedard:
When no `pnpm_lock` attribute is set the `patched_dependencies` never gets initialized. This fix just sets a default initial value for `patched_dependencies`, essentially the exact same as `importers` which has a default and is only overridden if a `pnpm_lock` is set.

I'm also dropping the `lockfile_is_loaded = "importers" in priv` since `importers` is set 100% of the time. The alternative (for `importers` and `patched_dependencies`) is to set the initial value to `None` and change how this works a bit. WDYT?

The only way I can reproduce this is having an e2e test with no `pnpm-lock.yaml`, then running the test updates the lockfile in the root of the repo. I'm not sure if that's it's own bug?

### Type of change

### Test plan

- Verified manually

### DIFF
--- a/npm/private/npm_translate_lock_state.bzl
+++ b/npm/private/npm_translate_lock_state.bzl
@@ -549,23 +549,16 @@ def _load_root_package_json(priv, rctx, label_store):
         priv["root_package_json"] = json.decode(rctx.read(root_package_json_path))
         return priv["root_package_json"]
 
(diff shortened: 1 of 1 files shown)


gregmagolan (approved):

gregmagolan:
lgtm

Issue #1092, opened 2023-05-31
[Bug]: "Cannot find module" error for dependency that is a peerDep of a first-party package


etlovett:
### What happened?

We have two `rules_js`-based libraries; let's call them `leaf` and `consumer`. `consumer` depends on `leaf`, and each have a set of third-party dependencies. Applications might choose to depend on either or both. Our expectation is that we should be able to do the following:
```
// leaf/package.json
{
  "peerDependencies": {
    "third-party-dep-1": "whatever version",
    "third-party-dep-2": "whatever version",
  },
  "devDependencies": {
    "third-party-dep-1": "whatever version",
    "third-party-dep-2": "whatever version",
  }
}

// consumer/package.json
{
  "peerDependencies": {
    "@first-party/leaf": "workspace:*",
    "third-party-dep-1": "whatever version",
    "third-party-dep-2": "whatever version",
  },
  "devDependencies": {
    "@first-party/leaf": "workspace:*",
    "third-party-dep-1": "whatever version",
    "third-party-dep-2": "whatever version",
  }
}

// some-app/package.json
{
  "dependencies": {
    "@first-party/consumer": "workspace:*",
    "@first-party/leaf": "workspace:*", // optional
    "third-party-dep-1": "whatever version",
    "third-party-dep-2": "whatever version",
  }
}
```

I’m trying to set up just the two libraries right now, using essentially the `package.json` files above. However, when I do all the `pnpm install`s and then run the Jest tests in `consumer` (`bazel test //path/to/consumer/src:test`), I get the following two errors in the output:
```
Cannot find module 'third-party-dep-1' from '../../../../node_modules/.aspect_rules_js/@first-party+leaf@0.0.0/node_modules/@first-party/leaf/src/some-file.js'
Cannot find module 'third-party-dep-2' from '../../../../../../../../../../../../../../../../../../execroot/rh/bazel-out/darwin_arm64-fastbuild/bin/bazel/js/jest/config/src/jest-env.mjs'
```

Changing all of `consumer`’s peerDeps/devDeps to (regular) deps doesn’t fix anything. Changing all of `leaf`’s peerDeps to (regular) deps does fix it, since then it can find its own dependencies, but this doesn’t seem to us like the right philosophical approach for a monorepo.

It also seems odd that utils has itself in its own node_modules, and we’re not sure why that would be.

Are the first-party package's peerDeps not getting properly passed along the Bazel graph?

### Version

[...]

PR #1646, opened 2024-04-13
feat: add support for linking js_library as 1p npm deps

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/1092.

This PR adds the ability to link `js_library` targets at 1p deps in addition to `npm_package` targets. Using `js_library` to link results in a symlink into the `js_library` package directory in the output tree instead of a `copy_to_directory` of the `npm_package` TreeArtifact. This more closely matches how pnpm handles 1p linked deps and it fixes the long standing bug with 1p peer deps reported in #1092.

The new test case `//examples/npm_deps:test9` illustrates how the `//examples/npm_package/packages/pkg_d:pkg_d` `js_library` that is linked `examples/npm_deps/package.json` as a `"@mycorp/pkg-d": "workspace:*"` 1p npm package in the pnpm lock file.

```
bazel-bin/examples/npm_deps/node_modules/@mycorp/pkg-d -> ../../../../node_modules/.aspect_rules_js/@mycorp+pkg-d@0.0.0/node_modules/@mycorp/pkg-d

bazel-bin/node_modules/.aspect_rules_js/@mycorp+pkg-d@0.0.0/node_modules/@mycorp/pkg-d -> ../../../../../examples/npm_package/packages/pkg_d
```

This is one extra level of indirection vs. what pnpm does but the result is the same in that the npm package 1p symlink `examples/npm_deps/node_modules/@mycorp/pkg-d` can be resolved back to the output-tree package of the `js_library` target `examples/npm_package/packages/pkg_d`. In pnpm, the symlinks goes directly there and not through the package store and, of course, it points into the source tree since there is no concept of an bazel output tree outside of bazel. Under rules_js the extra level of indirection fits nicer into the current support for `npm_package` and so makes for simpler code to support both  modes.

---

### Type of change

### Test plan

- New test cases added


npm/private/npm_package_store.bzl
```diff
@@ -303,10 +342,14 @@ deps of npm_package_store must be in the same package.""" % (ctx.label.package,
 
     files_depset = depset(files)
 
+    for transitive_package_store_infos_depset in transitive_package_store_infos_depsets:
+        for npm_package_store_info in transitive_package_store_infos_depset.to_list():
```
jbedard, line 346: I thought the `to_list` could be avoided when we're just putting `npm_package_store_infos` into `depset(transitive)`?

gregmagolan: This is an unusual case as its a depset of NpmPackageStoreInfo providers and we need to get at the underlying list so we can get at the underlying providers to pull `.transitive_files` out of them.

On the bright side, this code path is _only_ followed for js_library links which should be little to none in most cases.

The code that populates it,

```
transitive_package_store_infos_depsets.append(ctx.attr.src[JsInfo].npm_package_store_infos)
```

is on in the `elif ctx.attr.src and JsInfo in ctx.attr.src:` path

js/private/js_library.bzl
```diff
@@ -196,7 +196,7 @@ def _js_library_impl(ctx):
     )
 
     npm_package_store_infos = gather_npm_package_store_infos(
-        targets = ctx.attr.data + ctx.attr.deps,
+        targets = ctx.attr.srcs + ctx.attr.data + ctx.attr.deps,
```
gregmagolan, line 199: `srcs` was notably missing here. this is a related bug fix

jbedard: I was wondering about that. Was this required for this PR? Should this go into 1.x?

jbedard:
> bazel-bin/node_modules/.aspect_rules_js/@mycorp+pkg-d@0.0.0/node_modules/@mycorp/pkg-d -> ../../../../../examples/npm_package/packages/pkg_d

Doesn't that expose the entire `packages/pkg_d` directory? What if we have 2 `js_library` in there that both end up in the sandbox... won't they be able to access eachother, the package can access files outside the package etc?

js/private/js_library.bzl
```diff
@@ -214,12 +214,14 @@ def _js_library_impl(ctx):
 
     return [
         js_info(
+            target = ctx.label,
+            bin_dir = ctx.bin_dir,
```
jbedard, line 218: Can't `ctx.bin_dir` be used without having to copy it here? The only place I see this used it does:
```
ctx.attr.src[JsInfo].bin_dir.path
```
Would `ctx.bin_dir` not be correct there?

js/private/js_library.bzl
gregmagolan: It could but it's a rare corner case that no one has hit yet so may not be worth the effort

js/private/js_library.bzl
gregmagolan: It's not clear if it would be the same bin_dir in the rule context with transitions. To be safe the bin dir of the target that created the provider should always be correct

js/private/js_library.bzl
gregmagolan: split it out to https://github.com/aspect-build/rules_js/pull/1650

js/private/js_library.bzl
gregmagolan: In theory, the two bin_dirs _should_ always be the same since if the npm_package_store is transitioned so should all of js_library targets that it depends on. Lets go with `ctx.bin_dir` for now for simplicity and if it ends up being an issue then it will be bug fix in the future.

jbedard (approved):

gregmagolan:
> > bazel-bin/node_modules/.aspect_rules_js/@mycorp+pkg-d@0.0.0/node_modules/@mycorp/pkg-d -> ../../../../../examples/npm_package/packages/pkg_d
> 
> 
> 
> Doesn't that expose the entire `packages/pkg_d` directory? What if we have 2 `js_library` in there that both end up in the sandbox... won't they be able to access eachother, the package can access files outside the package etc?

Yes. That is the intention and it more closely follows how pnpm links 1p packages. Controlling what is in the sandbox in that directory is a user space concern.

PR #1659, opened 2024-04-19
refactor: remove references to pnpm-lock specifiers field

jbedard:
This seems unused?

### DIFF
--- a/npm/private/utils.bzl
+++ b/npm/private/utils.bzl
@@ -110,16 +110,13 @@ def _convert_v6_importers(importers):
     # Convert pnpm lockfile v6 importers to a rules_js compatible format.
     result = {}
     for import_path, importer in importers.items():
-        result[import_path] = {
-            "specifiers": {},
-            "dependencies": {},
-            "optionalDependencies": {},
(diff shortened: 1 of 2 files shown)


npm/private/utils.bzl
```diff
@@ -110,16 +110,13 @@ def _convert_v6_importers(importers):
     # Convert pnpm lockfile v6 importers to a rules_js compatible format.
     result = {}
     for import_path, importer in importers.items():
-        result[import_path] = {
-            "specifiers": {},
-            "dependencies": {},
```
jbedard, line 115: Removing the empty defaults here might be a bit more then the PR description says, but it doesn't seem to effect anything...

gregmagolan: Yeah. That's fine. The use site has defaults:
```
        prod_deps = {} if dev else lock_importer.get("dependencies", {})
        dev_deps = {} if prod else lock_importer.get("devDependencies", {})
        opt_deps = {} if no_optional else lock_importer.get("optionalDependencies", {})
```

PR #1661, opened 2024-04-19
refactor: remove unreachable code

jbedard:
This seems impossible...


gregmagolan (approved):
Yup. Don't know why that condition is there 🤷‍♂️

PR #1662, opened 2024-04-19
test: update e2e/pnpm_lockfiles setup to update existing lockfiles

jbedard:
Removing `sharp` which doesn't seem to add any test cases not covered by others.

Adding `rollup` which contains an `optionalDependencies` to demonstrate the `optionalDependencies` key within the `packages`.

### DIFF
--- a/e2e/pnpm_lockfiles/lockfile-test.bzl
+++ b/e2e/pnpm_lockfiles/lockfile-test.bzl
@@ -12,11 +12,33 @@ def lockfile_test(name = "lockfile", node_modules = "node_modules"):
     build_test(
         name = name,
         targets = [
+            # The full node_modules target
             ":%s" % node_modules,
+
+            # Direct deps
             ":%s/@aspect-test/a" % node_modules,
             ":%s/@aspect-test/b" % node_modules,
             ":%s/@aspect-test/c" % node_modules,
-            ":%s/sharp" % node_modules,
+            ":%s/rollup" % node_modules,
             ":%s/uvu" % node_modules,
+
+            # Targets within the virtual store...
+            # Direct dep targets
+            ":.aspect_rules_js/node_modules/@aspect-test+a@5.0.2",
+            ":.aspect_rules_js/node_modules/@aspect-test+a@5.0.2/dir",
+            ":.aspect_rules_js/node_modules/@aspect-test+a@5.0.2/pkg",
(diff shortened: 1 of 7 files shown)


npm/private/utils.bzl
```diff
@@ -132,11 +132,6 @@ def _convert_v6_packages(packages):
             dependencies[dep_name] = _convert_pnpm_v6_package_name(dep_version)
         package_info["dependencies"] = dependencies
 
-        # optionalDependencies
```
jbedard, line 135: FWICT `optionalDependencies` are only in the `importers`, not here in the `packages`

gregmagolan (approved):

npm/private/utils.bzl
gregmagolan: Hmm. I'm not sure about that since importers only deals with direct dependencies and not optional dependencies of transitives. Lets hold on this one for now.

gregmagolan (changes requested):
...

npm/private/utils.bzl
jbedard: This is the type of edge case I want in that pnpm_lockfiles e2e test...

jbedard: I've updated this PR to demonstrate this content in the lockfile, although doesn't really test it.

e2e/pnpm_lockfiles/lockfile-test.bzl
```diff
@@ -12,11 +12,33 @@ def lockfile_test(name = "lockfile", node_modules = "node_modules"):
...
+            ":.aspect_rules_js/node_modules/@aspect-test+a@5.0.2/pkg",
+            ":.aspect_rules_js/node_modules/@aspect-test+a@5.0.2/ref",
+
+            # Direct deps with lifecycles
+            ":.aspect_rules_js/node_modules/@aspect-test+c@2.0.2/lc",
+            ":.aspect_rules_js/node_modules/@aspect-test+c@2.0.2/pkg_lc",
+
+            # TODO: differs across lockfile versions
+            # Direct deps from custom registry
+            # ":.aspect_rules_js/node_modules/@types+node@registry.npmjs.org+@types+node@16.18.11",
```
jbedard, line 38: These ones are interesting where we generate different targets because the lockfiles are formatted different. Do we care?

gregmagolan (approved):

PR #1663, opened 2024-04-19
refactor: don't gather files from NpmPackageStoreInfo providers in gather_files_from_js_info

gregmagolan:
This was a bad pattern to pull from NpmPackageStoreInfo in the helper. Instead npm_package_store should just put its transitive files closure in DefaultInfo runfiles and also provide a JsInfo for when it is used directly in a js_run_binary or other js rules.

### DIFF
--- a/npm/private/npm_package_store.bzl
+++ b/npm/private/npm_package_store.bzl
@@ -4,7 +4,7 @@ load("@aspect_bazel_lib//lib:copy_directory.bzl", "copy_directory_bin_action")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 
 # buildifier: disable=bzl-visibility
-load("//js/private:js_info.bzl", "JsInfo")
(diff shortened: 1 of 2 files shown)


npm/private/npm_package_store.bzl
```diff
@@ -364,9 +362,16 @@ deps of npm_package_store must be in the same package.""" % (ctx.label.package,
             for npm_package_store_info in npm_package_store_infos
         ])
 
+    files_depset = depset(files)
+
     providers = [
+        js_info(
```
jbedard, line 368: Thank you for simplifying the diff so it is 4 lines added instead of 50 changed 😅 👍

jbedard:
So we added the `transitive_files_depset` in 2 spots (`JsInfo` and `DefaultInfo(runfiles)`). Do we have tests verifying those are both required?

jbedard (approved):

PR #1673, opened 2024-04-21
chore: cleanup WORKSPACE and MODULE.bazel files

gregmagolan:
Consistent formatting, structure & set `dev_dependency = True` everywhere it applies.


e2e/bzlmod/MODULE.bazel
```diff
@@ -1,18 +1,12 @@
...
-    name = "e2e_bzlmod",
-    version = "0.0.0",
-    compatibility_level = 1,
-)
-
-bazel_dep(name = "aspect_rules_js", version = "0.0.0")
-bazel_dep(name = "aspect_bazel_lib", version = "2.7.1")
-bazel_dep(name = "bazel_skylib", version = "1.5.0")
-
+bazel_dep(name = "aspect_rules_js", version = "0.0.0", dev_dependency = True)
```
jbedard, line 1: This is because `e2e/bzlmod` doesn't publish rules so this is a dev dependency? Which I guess is true for 99% of projects so e2e tests should do the same?

e2e/js_image_docker/WORKSPACE
```diff
@@ -28,11 +28,10 @@ load("@npm//:repositories.bzl", "npm_repositories")
 
 npm_repositories()
 
-# TODO: remove when rules_?  upgrades to modern version supporting
 http_archive(
     name = "bazel_gazelle",
```
jbedard, line 32: Why is this needed at all? I guess you don't have to answer or solve that in this PR though 😅

e2e/npm_translate_lock/MODULE.bazel
```diff
@@ -1,20 +1,18 @@
...
     module_name = "aspect_rules_js",
     path = "../..",
 )
 
-npm = use_extension("@aspect_rules_js//npm:extensions.bzl", "npm")
+bazel_dep(name = "aspect_bazel_lib", version = "2.7.1", dev_dependency = True)
+bazel_dep(name = "bazel_skylib", version = "1.5.0", dev_dependency = True)
+bazel_dep(name = "platforms", version = "0.0.9", dev_dependency = True)
+
+npm = use_extension(
```
jbedard, line 11: Why were some of these added?

e2e/js_image_docker/WORKSPACE
gregmagolan: Looked up git blame on that comment which confused me here. It came in on a renovate PR. Context here: https://github.com/aspect-build/rules_js/pull/1002/files#r1189105423.

gregmagolan: Added a detailed comment instead of the TODO. If it is a transitive of rules_docker that is the issue then its never going to be resolved since rules_docker is now archived.

e2e/npm_translate_lock/MODULE.bazel
gregmagolan: These are moved from above, re-ordered, and `dev_dependency = True` is added to them

e2e/bzlmod/MODULE.bazel
gregmagolan: Right. The pattern for a terminal module is that everything is a dev dep because the module is never published and used by other modules. In practice, clients and customers don't mark all their deps as `dev_dependency = True` in the MODULE.bazel files but technically they are. In client & customer repos there is no ambiguity because everything is a dev dep. In our rule sets I think its helpful and clearer to be explicit and it makes it easier to see at glance (and in a search of "bazel_dep") which deps are load bearing to users and which are internal only.

e2e/npm_translate_lock/MODULE.bazel
jbedard: Oh I see, adding `dev_dependency` made it 3 lines so I didn't see the single line removed while quickly scanning it 😅

jbedard (approved):

PR #1677, opened 2024-04-23
chore: update workflows config

gregmagolan:
Some churn in the config for workspaces after a task inheritance change was reverted


jbedard:
That's a lot of duplication :/

jbedard (approved):

PR #1679, opened 2024-04-23
feat: support pnpm lock v9

jbedard:
With the lockfile `requiresBuild` flag removed (https://github.com/pnpm/pnpm/issues/7707) lifecycle hooks must now be manually specified using [pnpm.onlyBuiltDependencies](https://pnpm.io/package_json#pnpmonlybuiltdependencies).

---

### Type of change

### Test plan

- New test cases added

### DIFF
--- a/e2e/pnpm_lockfiles/base/package.json
+++ b/e2e/pnpm_lockfiles/base/package.json
@@ -20,6 +20,9 @@
     "pnpm": {
         "patchedDependencies": {
             "meaning-of-life@1.0.0": "patches/meaning-of-life@1.0.0-pnpm.patch"
-        }
+        },
+        "onlyBuildDependencies": [
+            "@aspect-test/c"
+        ]
     }
 }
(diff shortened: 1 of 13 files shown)


e2e/pnpm_lockfiles/base/package.json
```diff
@@ -20,6 +20,9 @@
     "pnpm": {
         "patchedDependencies": {
             "meaning-of-life@1.0.0": "patches/meaning-of-life@1.0.0-pnpm.patch"
-        }
+        },
+        "onlyBuildDependencies": [
+            "@aspect-test/c"
```
jbedard, line 25: This is required for v9, and we have a test asserting the lifecycle targets exist.

We currently have no tests asserting that other packages *do not* have lifecycle targets though, that would be more then the simple `build_test`...

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -278,7 +278,6 @@ def _get_npm_imports(importers, packages, patched_dependencies, only_built_depen
         optional_deps = package_info.get("optional_dependencies")
         dev = package_info.get("dev")
         optional = package_info.get("optional")
-        pnpm_patched = package_info.get("patched")
```
jbedard, line 281: This property came directly from the lockfile and it no longer exists in v9. However the `patched_dependencies` dict has already been computed before this method is called so we can just check if the package exists there. Assuming `patched_dependencies` was constructed correctly, but we probably make that assumption already? 😅

gregmagolan (approved):
🚀

PR #1683, opened 2024-04-24
refactor: default npm_package(publishable) to false

jbedard:
The majority of `npm_package`s are not published and this target goes unused.

---

- Suggested release notes appear below:

`npm_package` `.publish` targets are no longer generated by default. Set `npm_package(publishable = True)` to enable.

### Test plan


gregmagolan (approved):
🌮

PR #1684, opened 2024-04-25
refactor: change default linked target name to 'pkg'

gregmagolan:
`pkg` is a better default for this compared to `{dirname}`, especially when running gazelle since it has trouble switching rules types if the default target of package is already a different rule and then that package is turned into a linkable package


jbedard:
I think a test somewhere needs to be updated to use `{dirname}` so it's still tested?

npm/extensions.bzl
```diff
@@ -221,7 +221,7 @@ def _npm_translate_lock_attrs():
     attrs["run_lifecycle_hooks"] = attr.bool(default = True)
 
     # Args defaulted differently by the macro
-    attrs["npm_package_target_name"] = attr.string(default = "{dirname}")
```
jbedard, line 224: Lets just make sure the `{dirname}` var is used in a test somewhere still.

jbedard (approved):

npm/extensions.bzl
gregmagolan: Already used in `rules_js/e2e/npm_link_package`:

```
npm_translate_lock(
    name = "npm",
    lifecycle_hooks_envs = {
        "*": [
            "aspect_test_id=testc",
            "BUILD_VERSION=1",
        ],
        # Without APPDATA environment variable, sharp install will fail because of the binary can not
        # be copied to ~/.npm folder which is readonly for bazel.
        "sharp": ["APPDATA=."],
    },
    lifecycle_hooks_execution_requirements = {
        "sharp": ["requires-network"],
    },
    npm_package_target_name = "{dirname}_pkg",
)
```

PR #1690, opened 2024-05-02
refactor: update js_binary executable path to {name}_/{name}

gregmagolan:
Follow-up to #1586.

The trouble with the change in #1586 with the name of the executable in the output tree the same as the target name is that a target name such as `test` could be in a BUILD file where a subdirectory named test exists which can lead to build failures such as:

```
ERROR: One of the output paths 'bazel-out/k8-fastbuild/bin/foo/test' (belonging to //foo:test) and 'bazel-out/k8-fastbuild/bin/foo/test/active_file_users_test.ts' (belonging to //foo:foo) is a prefix of the other. These actions cannot be simultaneously present; please rename one of the output files or build just one of them
```

This change resolves that issue without bringing back the `.sh` suffix that #1586 was addressing.

### DIFF
--- a/e2e/worker/BUILD.bazel
+++ b/e2e/worker/BUILD.bazel
@@ -16,7 +16,7 @@ copy_file(
 )
 
 js_binary(
-    name = "worker_binary",
+    name = "worker",
     data = [":copy_worker_js"],
     entry_point = ":copy_dummy_worker",
     visibility = ["//visibility:public"],
--- a/e2e/worker/defs.bzl
+++ b/e2e/worker/defs.bzl
@@ -31,7 +31,7 @@ pi_rule = rule(
         "worker": attr.label(
             executable = True,
             cfg = "exec",
-            default = ":worker_binary",
+            default = ":worker",
         ),
     },
 )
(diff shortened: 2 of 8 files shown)


examples/js_binary/BUILD.bazel
```diff
@@ -457,21 +458,26 @@ js_binary(
...
-        # --- end runfiles.bash initialization v2 ---
-
-        $(rlocation aspect_rules_js/examples/js_binary/bin10-js_binary) "$@"
+    data = [":bin10-js_binary"],
+    is_executable = True,
+    substitutions = {
+        "{rlocationpath}": "$(rlocationpath :bin10-js_binary)",
+    },
+    template = ["""\
+# --- begin runfiles.bash initialization v2 ---
```
alexeagle, line 470: nit: we have the bash runfiles lib in a starlark constant exposed by bazel-lib, you could probably load that instead of copy-paste

jbedard:
Was this convention copied from somewhere?

gregmagolan:
> Was this convention copied from somewhere?

Copied from rules to

examples/js_binary/BUILD.bazel
gregmagolan: good idea

jbedard (approved):

PR #1693, opened 2024-05-03
fix: do not use npm_translate_lock patch_args without patches

### LINKED ISSUE #1427: [Bug]: Error from npm_translate_lock when using pnpm.patchedDependencies
### What happened?

If you have a [`pnpm.patchedDependencies`](https://pnpm.io/package_json#pnpmpatcheddependencies) entry in `package.json`, then `npm_translate_lock` will produce an error like:

```
Error in fail: ERROR: patch_args for package @jest/environment@29.7.0 contains a strip prefix that is incompatible with a patch applied via `pnpm.patchedDependencies`.

`pnpm.patchedDependencies` requires a strip prefix of `-p1`. All applied patches must use the same strip prefix.
```

`npm_translate_lock`'s support for `pnpm.patchedDependencies` should extend to supplying a default `patch_args` value of `["-p1"]` for packages that appear in `pnpm.patchedDependencies`, since that's the format pnpm uses.

### Version

`WORKSPACE` or `MODULE.bazel` file: rules_js-1.33.1

Language(s) and/or frameworks involved: Javascript

### How to reproduce

_No response_

### Any other information?

_No response_

jbedard:
Fix #1427
An alternative to #837 from @kormide 

This makes pure `pnpm.patchedDependencies` nicer like #837, normally having the same effect, I think?

I think the main difference is that #837 removes `-p0` if there are no `npm_translate_lock(patches)` but still applies the rest of `npm_translate_lock(patch_args)`. While this PR does not apply `npm_translate_lock(patch_args)` when there are no `npm_translate_lock(patches)`. 

Maybe this PR is actually more confusing? 🤦 
Maybe we should just drop support for mixing these 2 types of patches in 2.x? See #1694

---

### Test plan

### DIFF
--- a/e2e/npm_translate_lock_subdir_patch/MODULE.bazel
+++ b/e2e/npm_translate_lock_subdir_patch/MODULE.bazel
@@ -12,9 +12,6 @@ npm.npm_translate_lock(
         "//subdir:package.json",
         "//subdir:patches/debug@4.3.4.patch",
     ],
-    patch_args = {
-        "*": ["-p1"],
-    },
     pnpm_lock = "//subdir:pnpm-lock.yaml",
     verify_node_modules_ignored = "//:.bazelignore",
 )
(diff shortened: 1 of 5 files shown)


jbedard:
I think we should just do https://github.com/aspect-build/rules_js/pull/1694 instead of this or #837

WDYT?

jbedard:
Done with https://github.com/aspect-build/rules_js/pull/1694

PR #1694, opened 2024-05-03
fix: do not support mixing pnpm.patchedDependencies and npm_translate_lock(patches)

### LINKED ISSUE #1427: [Bug]: Error from npm_translate_lock when using pnpm.patchedDependencies

jbedard:
Fix #1427

An alternative to #837 or #1693. This removes the option of mixing `pnpm.patchedDependencies` + `npm_translate_lock(patches, patch_args)` on a single package and instead providing simple defaults to patch args when using `pnpm.patchedDependencies`.

---

Patches from both `pnpm.patchedDependencies` and `npm_translate_lock(patches)` can no longer be applied to the same npm package.

### Test plan

### DIFF
--- a/npm/private/npm_translate_lock_helpers.bzl
+++ b/npm/private/npm_translate_lock_helpers.bzl
@@ -326,12 +326,30 @@ def _get_npm_imports(importers, packages, patched_dependencies, root_package, rc
             unfriendly_name = None
 
         # gather patches & patch args
-        patches, patches_keys = _gather_values_from_matching_names(True, attr.patches, name, friendly_name, unfriendly_name)
+        patches = []
+        patch_args = []
+
+        translate_patches, patches_keys = _gather_values_from_matching_names(True, attr.patches, name, friendly_name, unfriendly_name)
+
+        if len(translate_patches) > 0 and pnpm_patched:
+            msg = """\
+ERROR: can not apply both `pnpm.patchedDependencies` and `npm_translate_lock(patches)` to the same package {pkg}.
+""".format(
+                pkg = name,
+            )
+            fail(msg)
 
         # Apply patch from `pnpm.patchedDependencies` first
         if pnpm_patched:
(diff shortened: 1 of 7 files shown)


thesayyn (approved):
SGTM, reasonable

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -326,12 +326,28 @@ def _get_npm_imports(importers, packages, patched_dependencies, root_package, rc
...
+                pkg = name,
+            )
+            fail(msg)
 
         # Apply patch from `pnpm.patchedDependencies` first
         if pnpm_patched:
             patch_path = "//%s:%s" % (attr.pnpm_lock.package, patched_dependencies.get(friendly_name).get("path"))
-            patches.insert(0, patch_path)
+            patches.append(patch_path)
+            patch_args = ["-p1"]
```
gregmagolan, line 346: nit: add comment that we hard-code to `-p1` since pnpm patches always require that

gregmagolan (approved):
:taco:

PR #1695, opened 2024-05-03
fixup! refactor: set minimum aspect_bazel_lib to 2.7.1 (#1580)

jbedard:
When you upgraded to 2.7.1 `aspect_bazel_lib` was removed as a direct dep of these so we shouldn't be invoking `aspect_bazel_lib_dependencies` anymore.

This is a `!fixup` on that commit, assuming we'll rebase/squash and `push -f` at some point still?

### DIFF
--- a/e2e/pnpm_workspace_deps/WORKSPACE
+++ b/e2e/pnpm_workspace_deps/WORKSPACE
@@ -3,10 +3,6 @@ local_repository(
     path = "../..",
 )
 
-load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies")
-
-aspect_bazel_lib_dependencies()
-
 load("@aspect_rules_js//js:repositories.bzl", "rules_js_dependencies")
 
 rules_js_dependencies()
--- a/e2e/pnpm_workspace/WORKSPACE
+++ b/e2e/pnpm_workspace/WORKSPACE
@@ -3,10 +3,6 @@ local_repository(
     path = "../..",
 )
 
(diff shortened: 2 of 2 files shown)


e2e/pnpm_workspace_deps/WORKSPACE
```diff
@@ -3,10 +3,6 @@ local_repository(
     path = "../..",
 )
 
-load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies")
-
-aspect_bazel_lib_dependencies()
```
gregmagolan, line 8: You missed a spot. I counted three when I greened up my other PR: https://github.com/aspect-build/rules_js/pull/1690/commits/28fdfa00b2df78bfcdd6a7f10df845efc030eed4

gregmagolan (approved):

gregmagolan:
Yea, we can squash it with the commit on the 2.x stack that makes most sense. Maybe the one that introduces rules_js_toolchains?

e2e/pnpm_workspace_deps/WORKSPACE
jbedard: This PR is green though. I don't think that 3rd one is required?

jbedard:
This commit is marked as a `fixup!` on the commit that made it red so I think that's the one that it should be squashed into.

e2e/pnpm_workspace_deps/WORKSPACE
gregmagolan: Not required to green it up but it is a superfluous call to `aspect_bazel_lib_dependencies` in the 3rd case

PR #1696, opened 2024-05-03
refactor: remove support for pnpm v6

jbedard:
This lockfile version is for pnpm v6 which last had a patch released over 2y ago. This version is missing some features such as listing patches in the lockfile which is most likely buggy today and not worth maintaining.

---

Remove support for pnpm v6.

### Test plan

- New test cases added


e2e/js_image_oci/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "image",
     "dependencies": {
-        "chalk": "^5.2.0"
+        "chalk": "^4.0.0"
```
jbedard, line 4: 5.2 actually causes an error

I think renovate probably updated this but because there was no pnpm-workspace.yaml it updated the root pnpm-lock.yaml and had no effect on this e2e test. I added the pnpm-workspace in this PR

gregmagolan: Yeah, chalk 5 is esm only and that breaks things if I recall

npm/private/test/utils_tests.bzl
```diff
@@ -61,12 +61,19 @@ def test_package_store_name(ctx):
...
     env = unittest.begin(ctx)
-    utils.assert_lockfile_version(5.3)
-    utils.assert_lockfile_version(5.4)
+
+    # Unsupported versions + msgs
+    msg = utils.assert_lockfile_version(5.3, testonly = True)
+    asserts.equals(env, "npm_translate_lock requires lock_version at least 5.4, but found 5.3. Please upgrade to pnpm v6 or greater.", msg)
     msg = utils.assert_lockfile_version(1.2, testonly = True)
-    asserts.equals(env, "npm_translate_lock requires lock_version at least 5.3, but found 1.2. Please upgrade to pnpm v6 or greater.", msg)
+    asserts.equals(env, "npm_translate_lock requires lock_version at least 5.4, but found 1.2. Please upgrade to pnpm v6 or greater.", msg)
```
gregmagolan, line 69: "please upgrade to pnpm v7 or greater"

jbedard: Good catch 👍

alexeagle (commented):
Can you add a README change that indicates minimum and maximum versions of pnpm we support?

jbedard:
I've added a sentence about pnpm versions to docs/README.md. WDYT?

e2e/js_image_docker/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "image",
     "dependencies": {
-        "chalk": "^5.0.0"
+        "chalk": "^4.0.0"
```
gregmagolan, line 4: This can land on main? I'll put up a PR there

e2e/js_image_oci/pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
```
gregmagolan, line 1: Should go on main as well

e2e/js_image_docker/package.json
jbedard: Yes it can, and should

gregmagolan (approved):
🌮

PR #1697, opened 2024-05-03
refactor: simplify collection of lifecycle hooks

jbedard:
---

### Test plan

- New test cases added


npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -444,11 +449,10 @@ ERROR: patch_args for package {package} contains a strip prefix that is incompat
             patch_args = patch_args,
             patches = patches,
             root_package = root_package,
-            run_lifecycle_hooks = run_lifecycle_hooks,
```
jbedard, line 447: This is the summary of the change. The struct passed around no longer has this flag and instead the `lifecycle_hooks` is `False` which should have already been the case anyway.

The rule attribute `run_lifecycle_hooks` is different, that is what you use to opt-out of lifecycle hooks completely (which results in `lifecycle_hooks` being `False` due to a `lifecycle_hooks if run_lifecycle_hooks else False` statement that you can find in this PR...).

PR #1700, opened 2024-05-03
refactor: always read pnpm.patchedDependencies from lockfile

jbedard:
Why load them different when running pnpm update? That makes it inconsistent, ~~especially with lockfile v5.3 where the lockfile vs package.json will give different results~~ (edit: but we no longer support 5.3).

Unlike other `pnpm.*` properties from the package.json this one does end up in the lockfile with additional metadata, so we will always read it from the lockfile and not package.json.

---

### Test plan

### DIFF
--- a/npm/private/npm_translate_lock_state.bzl
+++ b/npm/private/npm_translate_lock_state.bzl
@@ -39,14 +39,6 @@ WARNING: `update_pnpm_lock` attribute in `npm_translate_lock(name = "{rctx_name}
 
     _init_root_package(priv, rctx, label_store)
 
-    root_package_json_declared = label_store.has("package_json_root")
-    if root_package_json_declared and _should_update_pnpm_lock(priv):
-        # If we need to read the patches list from the package.json since
-        # update_pnpm_lock is enabled and the user has declared the root package.json
-        # as an input file then we can read it right away.
-        _init_patched_dependencies_labels(priv, rctx, label_store)
-        priv["patched_dependencies_labels_initialized"] = True
-
     if _should_update_pnpm_lock(priv) or not rctx.attr.pnpm_lock:
         # labels only needed when updating or bootstrapping the pnpm lock file
         _init_pnpm_labels(label_store, rctx)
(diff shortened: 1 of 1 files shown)


npm/private/npm_translate_lock_state.bzl
```diff
@@ -193,15 +180,10 @@ def _init_patches_labels(priv, rctx, label_store):
 
 ################################################################################
 def _init_patched_dependencies_labels(priv, rctx, label_store):
-    if rctx.attr.update_pnpm_lock:
-        # Read patches from package.json `pnpm.patchedDependencies`
```
gregmagolan, line 197: There is another instance of this divergence where `_load_root_package_json` is called in `_copy_unspecified_input_files`

gregmagolan: If we remove it from there then we can drop the entire `_load_root_package_json` block

gregmagolan:
> Why load them different when running pnpm update? That makes it inconsistent, especially with lockfile v5.3 where the lockfile vs package.json will give different results.
> Or should we do the opposite? Note that some package.json pnpm.* properties are never in the lockfile, so maybe all those pnpm.* properties should be read from the package.json... 🤔

I think this divergence was needed back before we failed the build when the lock file is updated. Now that we do that we should be able to _just_ load the list of patches from the lock file

npm/private/npm_translate_lock_state.bzl
jbedard: I'm wondering if we should actually do the opposite and always load `pnpm.*` from the package.json instead of lockfile.

Some things like the `pnpm.onlyBuiltDependencies` seem to not end up in the lockfile at all. Others like this exact one (`pnpm.patchedDependencies`) are in lockfile 5.4+ but not 5.3. But unless one of these `pnpm.*` options effects package versions there really is no reason for it to be required in a lockfile...

jbedard:
I think we have to do the opposite to support other `pnpm.*` properties.

npm/private/npm_translate_lock_state.bzl
jbedard: I'm going to leave `_load_root_package_json` as-is since we will need it for other `pnpm.*` properties in a followup.

npm/private/npm_translate_lock_state.bzl
```diff
@@ -62,12 +54,6 @@ WARNING: `update_pnpm_lock` attribute in `npm_translate_lock(name = "{rctx_name}
     pnpm_lock_exists = is_windows or utils.exists(rctx, label_store.path("pnpm_lock"))
     if pnpm_lock_exists:
         _load_lockfile(priv, rctx, label_store)
-
-    if not priv["patched_dependencies_labels_initialized"]:
-        # If we didn't initialize the patched dependency labels above then try again now. We'll either
-        # read the list of patches from the lock file now that it has been read or, if update_pnpm_lock
-        # is enabled, we can derive the existance of a root package.json by looking for a `.` importer
-        # in the `importers` list.
         _init_patched_dependencies_labels(priv, rctx, label_store)
```
jbedard, line 57: Note this is in the `if pnpm_lock_exists` and it was not before. I think that is correct though...

gregmagolan (approved):
:taco:

PR #1701, opened 2024-05-06
refactor: remove support for reloading lockfile on update

jbedard:
Reloading isn't supported with bzlmod so we should not support it with non-bzlmod just to make things consistent.

Close https://github.com/aspect-build/rules_js/issues/1445

---

When `npm_translate_lock(update_pnpm_lock)` updates the lockfile it will exit with a non-0 exit code and must be rerun with the updated lockfile.

### Test plan

### DIFF
--- a/e2e/update_pnpm_lock/test.sh
+++ b/e2e/update_pnpm_lock/test.sh
@@ -53,22 +53,30 @@ ASPECT_RULES_JS_DISABLE_UPDATE_PNPM_LOCK=
 # Have to make another change to package.json to invalidate the repository rule
 _sedi 's#"@types/node": "16"#"@types/node": "14"#' package.json
 
-# Trigger the update of the pnpm lockfile
-if ! bazel run "$BZLMOD_FLAG" @npm//:sync; then
-    echo "ERROR: expected 'bazel run $BZLMOD_FLAG @npm//:sync' to pass"
+# Trigger the update of the pnpm lockfile which should fail
+if bazel run "$BZLMOD_FLAG" @npm//:sync; then
+    echo "ERROR: expected 'update_pnpm_lock' to exit with non-zero exit code on update"
     exit 1
 fi
 
+# The lockfile should be updated
 diff="$(git diff pnpm-lock.yaml)"
 if [ -z "$diff" ]; then
     echo "ERROR: expected 'git diff pnpm-lock.yaml' to not be empty"
     exit 1
 fi
 
+# The action cache file should be updated
 action_cache_file=".aspect/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU="
 diff="$(git diff "$action_cache_file")"
 if [ -z "$diff" ]; then
     echo "ERROR: expected 'git diff $action_cache_file' to not be empty"
+    exit 1
+fi
+
+# The lockfile has been updated and sync should now succeed
+if ! bazel run "$BZLMOD_FLAG" @npm//:sync; then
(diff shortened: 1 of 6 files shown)


e2e/update_pnpm_lock/test.sh
```diff
@@ -54,8 +54,8 @@ ASPECT_RULES_JS_DISABLE_UPDATE_PNPM_LOCK=
 _sedi 's#"@types/node": "16"#"@types/node": "14"#' package.json
 
 # Trigger the update of the pnpm lockfile
-if ! bazel run "$BZLMOD_FLAG" @npm//:sync; then
-    echo "ERROR: expected 'bazel run $BZLMOD_FLAG @npm//:sync' to pass"
+if bazel test "$BZLMOD_FLAG" //...; then
+    echo "ERROR: expected 'update_pnpm_lock' to exit with non-0 exit code"
```
gregmagolan, line 58: nit: "non-zero" or "non 0"

e2e/update_pnpm_lock/test.sh
```diff
@@ -54,8 +54,8 @@ ASPECT_RULES_JS_DISABLE_UPDATE_PNPM_LOCK=
 _sedi 's#"@types/node": "16"#"@types/node": "14"#' package.json
 
 # Trigger the update of the pnpm lockfile
-if ! bazel run "$BZLMOD_FLAG" @npm//:sync; then
```
gregmagolan, line 57: Does `bazel run @npm//:sync` not work under bzlmod??

jbedard: It should work fine. I think I could revert this actually, I must have been doing something weird to go down the path of changing this (e2e tests that modify source are odd...) 🤔

e2e/update_pnpm_lock/test.sh
```diff
@@ -54,8 +54,8 @@ ASPECT_RULES_JS_DISABLE_UPDATE_PNPM_LOCK=
...
-    echo "ERROR: expected 'bazel run $BZLMOD_FLAG @npm//:sync' to pass"
+if bazel run "$BZLMOD_FLAG" @npm//:sync; then
```
gregmagolan, line 57: Maybe run it again to test that `bazel run` exits 0 on the 2nd run?

e2e/update_pnpm_lock_with_import/test.sh
```diff
@@ -47,8 +47,8 @@ fi
 ASPECT_RULES_JS_FROZEN_PNPM_LOCK=
 
 print_step "It should update the lockfile after a running the invalide target with ASPECT_RULES_JS_FROZEN_PNPM_LOCK unset"
-if ! bazel run "$BZLMOD_FLAG" @npm//:sync; then
-    echo "ERROR: expected 'bazel run $BZLMOD_FLAG @npm//:sync' to pass"
+if bazel run "$BZLMOD_FLAG" @npm//:sync; then
```
gregmagolan, line 50: Same here. Run it twice and assert the exit code is 0 on the 2nd attempt.

e2e/update_pnpm_lock/test.sh
jbedard: It runs `bazel test //...` below and asserts 0 which is similar...? The mix of `bazel test` and `bazel run` might be the reason I previously had switched this to `bazel test`...

gregmagolan: Similar but I think its still worth asserting that `bazel run @npm//:sync` exits 0 on the 2nd go

gregmagolan: Defence in-depth

PR #1707, opened 2024-05-07
refactor: use BASH_RLOCATION_FUNCTION from @aspect_bazel_lib in examples/js_binary/BUILD.bazel

gregmagolan:
---

### Test plan

- New test cases added


jbedard (approved):
Can this go into 1.x?

gregmagolan:
> Can this go into 1.x?

Was going to say no because this depends on expand_template out of a newer bazel-lib... but because this is under //examples and we depend on the newer bazel-lib as a dev dep then I think yes. I'll point it there instead.

alexeagle (approved):

Issue #1563, opened 2024-04-01
[Bug]: bzlmod extension performance


jbedard:
### What happened?

The rules_js bzlmod npm extensions are slower then the repository rules today due to parsing and iterating over the lockfile+npmrc file content multiple times.

### How to reproduce

Run any build using the bzlmod `npm.npm_translate_lock` API and a large lockfile.

### Known issues/reasons:
* parsing the pnpm lock twice (https://github.com/aspect-build/rules_js/pull/2480)
* ✅ generating `npm_import` targets twice (https://github.com/aspect-build/rules_js/pull/1547)

PR #1709, opened 2024-05-07
refactor: extract repositories.bzl generation

jbedard:
In 2.x this is not generated at all with bzlmod, this makes the `if bzlmod:` cleaner in 2.x.

This may be required for #1563 as well as pnpm v9.

---

### Test plan


jbedard:
FYI @gregmagolan, after this merges the following 2.x rebase have conflicts and I should do it.

thesayyn (approved):

PR #1710, opened 2024-05-07
chore: add 1p linked deps to e2e/js_image_oci

gregmagolan:
Adds test coverage for indirect 1p dep links (via npm_package).

Pre-factor for adding support in js_image_layer for direct 1p dep links (js_library links).


jbedard:
@thesayyn do you have suggestions for how to add some assertions to this?

jbedard (approved):

gregmagolan:
> @thesayyn do you have suggestions for how to add some assertions to this?

I added a basic one for the npm_package 1p dep that it's peer and direct deps can be required and are the expected versions. Good enough for now. More per layer assertions could be added in a follow-up.

Issue #1641, opened 2024-04-11
[Bug]: js_image_layer puts both first-party and third-party code into same layer


matthewjh:
### What happened?

Hi. js_image_layer has a separation of "app" and "node_modules" layers. The problem is that when 1st-party libraries are used as linked node_modules via the npm_link rules or pnpm workspaces - probably very common for rules_js users - these 1st party libraries are included amongst all the 3rd party node_modules in the node_modules layer, meaning you have 1 layer containing most of the code (100-200mb) and that whole layer changing whenever one of the 1st party libraries changes even slightly.

See https://github.com/aspect-build/rules_js/blob/main/js/private/js_image_layer.bzl#L241

I ended up patching this to check for our "scope" prefix, so our first-party packages would go into the app layer instead of node_modules. However, under this issue it probably makes sense to revisit this bifurcation. Given that rules_js can distinguish between packages that originated in the workspace and externally (e.g. from npm), it would be perhaps nice to see a separate layer for everything "external" and "internal", which appears to better capture the intent of the "app" and "node_modules" layers to separate "infrequently changing" from "frequently changing".

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:
7.1.1
Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

Language(s) and/or frameworks involved:

    http_archive(
        name = "aspect_rules_js",
        sha256 = "d7500d59712accca9622bce723b8ea596b92fb9b4ff4a4dbd77ed353b9d29f34",
        strip_prefix = "rules_js-1.40.1",
        url = "https://github.com/aspect-build/rules_js/releases/download/v1.40.1/rules_js-v1.40.1.tar.gz",
        patches = [
            "//tools/bazel-patches:aspect_rules_js-image-layer.patch",
        ],
        patch_args = ["-p1"],
    )

gregmagolan:
Tricky one. From the rule's perspective when it looks in the JsInfo provider all node_modules is just a depset of Files under that path. Being able to discern between the two would likely require some provider changes... tho as it happens we're currently working on the 2.x branch where we are free to make breaking provider changes.

One way to go would be to make the provider changes so that js_image_layer can automatically separate out 1p and 3p npm deps. The other possibility is to add the functionality to js_image_layer to create multiple node_modules layers based on a regex or glob of the package name. You could argue the 2nd way is more generic and would allow users to also split their 3p deps into multiple layers if needed.

[...]

PR #1712, opened 2024-05-08
feat: split js_image_layer into fine grained layers

gregmagolan:
Resolves https://github.com/aspect-build/rules_js/issues/1641.

This change splits the output of js_image_layer into 5 layers:

1. `node` layer contains Node.js and the node fs patches
2. `package_store_3p` layers containers all 3p deps in the `.aspect_rules_js` package store
3. `package_store_1p` layers containers all 1p deps in the `.aspect_rules_js` package store
4. `node_modules` layer contains all `node_modules/foo` symlinks into the packages tore
5. `app` layer contains everything that doesn't fall into any of the above layers


thesayyn:
I think we should put node and node_modules into same layer as those are less likely to change.

thesayyn:
hmm how about; `node_modules_links  + package_store_3p + node ` they all go into the same layer, as they won't change frequently?

gregmagolan:
> hmm how about; `node_modules_links + package_store_3p + node ` they all go into the same layer, as they won't change frequently?

They could; node_modules_links include 1p links but won't change as much as the actual contents of the 1p links. Is there a downside to more fine grained layers?

thesayyn:
Other than network round trip not really, more the better, but i am afraid that we are splitting it too much that it will be confusing.

gregmagolan:
> Other than network round trip not really, more the better, but i am afraid that we are splitting it too much that it will be confusing.

Its transparent for users of rules_oci tho right since `tars = [":layers",],` in `oci_image`. Just the rules_docker example has them listed explicitly and we don't care about rules_docker ease of use since that rule set is abandonware now.

thesayyn:
Makes sense, i guess this is fine, fine grained layering will give users more power.

matthewjh:
Completely agree with fine-grained layering along these lines - it should save a lot of bandwidth and time for orgs that have 100+ services.

Thanks @gregmagolan!

js/private/js_image_layer.bzl
```diff
@@ -209,6 +279,9 @@ def _runfile_path(ctx, file, runfiles_dir):
     return paths.join(runfiles_dir, to_rlocation_path(ctx, file))
 
 def _build_layer(ctx, type, all_entries_json, entries, inputs):
+    if not entries:
+        return None
```
gregmagolan, line 283: short-circuit and don't create the layer if it has no entries

thesayyn:
I think this PR has a bad rebase, @gregmagolan could you fix that before i can take a look?

gregmagolan:
> I think this PR has a bad rebase, @gregmagolan could you fix that before i can take a look?

Ahh yes, the base branch was rebased which messed up the commits. All clean again now.

thesayyn (approved):

PR #1714, opened 2024-05-10
feat: add generate_empty_layers attribute to js_image_layer

gregmagolan:
Useful for when using js_image_layer with rules_docker in consistent manner for all images regardless of which layers are empty:

```
RULES_JS_LAYERS = [
    "node",
    "package_store_3p",
    "package_store_1p",
    "node_modules",
    "app",
]

def js_container_layers(name, binary, root, tags = []):
    js_image_layer(
        name = "{}/layers".format(name),
        binary = binary,
        generate_empty_layers = True,
        root = root,
        visibility = ["//visibility:__pkg__"],
    )

    [
        [
            native.filegroup(
                name = "{}/layers/{}.tar".format(name, layer),
                srcs = ["{}/layers".format(name)],
                output_group = layer,
            ),
            container_layer(
                name = "{}/{}_layer".format(name, layer),
                tars = [":{}/layers/{}.tar".format(name, layer)],
                tags = tags + ["manual"],
            ),
        ]
        for layer in RULES_JS_LAYERS
    ]
```

with `generate_empty_layers = False` the above example may fail with a message such as:

```
(18:01:23) ERROR: /home/runner/work/rules_js/rules_js/e2e/js_image_docker/BUILD.bazel:78:16: in tars attribute of container_layer_ rule //:package_store_1p_layer: '//:package_store_1p_tar' does not produce any container_layer_ tars files (expected .tar, .tar.xz, .tar.gz or .tgz). Since this rule was created by the macro 'container_layer', the error might have been caused by the macro implementation
```

if there are empty layers

### DIFF
--- a/js/private/js_image_layer.bzl
+++ b/js/private/js_image_layer.bzl
@@ -279,7 +279,7 @@ def _runfile_path(ctx, file, runfiles_dir):
     return paths.join(runfiles_dir, to_rlocation_path(ctx, file))
 
 def _build_layer(ctx, type, all_entries_json, entries, inputs):
-    if not entries:
+    if not entries and not ctx.attr.generate_empty_layers:
         return None
 
     entries_json = ctx.actions.declare_file("{}_{}_entries.json".format(ctx.label.name, type))
@@ -489,6 +489,12 @@ js_image_layer_lib = struct(
         ),
         "platform": attr.label(
             doc = "Platform to transition.",
+        ),
+        "generate_empty_layers": attr.bool(
+            doc = """Generate layers even if they are empty.
+
+Helpful when using js_image_layer with rules_docker.
+See https://github.com/aspect-build/rules_js/pull/1714 for more info""",
         ),
     },
 )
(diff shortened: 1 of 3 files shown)


js/private/js_image_layer.bzl
```diff
@@ -490,6 +490,9 @@ js_image_layer_lib = struct(
         "platform": attr.label(
             doc = "Platform to transition.",
         ),
+        "generate_empty_layers": attr.bool(
+            doc = "Generate layers even if they are empty.",
```
alexeagle, line 494: should it say "only intended for rules-docker" and/or "deprecated, to be removed in 3.0" or sth?

alexeagle (approved):

thesayyn (approved):

js/private/js_image_layer.bzl
gregmagolan: its a simple enough option that it can stick around;; but I will mention what it is intended for

PR #1715, opened 2024-05-10
feat: add backward compat handling for rules_js 1.x js_info parameters declarations, transitive_declarations, npm_package_store_deps and transitive_npm_linked_package_files

gregmagolan:
Makes it easier for custom downstream rule sets to support both rules_js 1.x and rules_js 2.x

### DIFF
--- a/js/private/js_info.bzl
+++ b/js/private/js_info.bzl
@@ -16,11 +16,12 @@ JsInfo = provider(
 def js_info(
         target,
         sources = depset(),
-        types = depset(),
+        types = None,
         transitive_sources = depset(),
-        transitive_types = depset(),
-        npm_sources = depset(),
-        npm_package_store_infos = depset()):
+        transitive_types = None,
+        npm_sources = None,
+        npm_package_store_infos = None,
+        **kwargs):
     """Construct a JsInfo.
 
     Args:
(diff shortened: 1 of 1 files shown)


js/private/js_info.bzl
```diff
@@ -31,10 +32,33 @@ def js_info(
...
     """
+
+    # Handle backward compat for rules_js 1.x js_info factory parameters:
+    # - declarations
+    # - transitive_declarations
+    # - npm_package_store_deps
+    # - transitive_npm_linked_package_files
+    declarations = kwargs.pop("declarations", None)
+    if declarations != None:
+        types = declarations
```
jbedard, line 48: What if `types` was also specified?

js/private/js_info.bzl
```diff
@@ -31,10 +32,33 @@ def js_info(
...
+    npm_package_store_deps = kwargs.pop("npm_package_store_deps", None)
+    if npm_package_store_deps != None:
+        npm_package_store_infos = npm_package_store_deps
+    transitive_npm_linked_package_files = kwargs.pop("transitive_npm_linked_package_files", None)
+    if transitive_npm_linked_package_files != None:
+        npm_sources = transitive_npm_linked_package_files
+    if len(kwargs):
+        msg = "Invalid js_info parameter '{}'".format(kwargs.keys()[0])
+        fail(msg)
+
```
jbedard, line 61: Should we record which legacy names we `pop`ed and output a warning here?

gregmagolan: Updated. Warning per attribute will work.

js/private/js_info.bzl
gregmagolan: Add fail for the mixed case

jbedard (approved):

PR #1721, opened 2024-05-14
refactor: reuse npm_translate_lock_state parse logic in bzlmod

jbedard:
This way common logic such as parsing `pnpm.*` from the package.json can live in the `npm_translate_lock_state.bzl` util.

---

### Test plan

### DIFF
--- a/npm/extensions.bzl
+++ b/npm/extensions.bzl
@@ -3,12 +3,12 @@ See https://bazel.build/docs/bzlmod#extension-definition
 """
 
 load("@bazel_features//:features.bzl", "bazel_features")
-load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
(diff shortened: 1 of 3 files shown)


npm/extensions.bzl
```diff
@@ -148,7 +126,7 @@ WARNING: Cannot determine home directory in order to load home `.npmrc` file in
     imports = npm_translate_lock_helpers.get_npm_imports(
         importers = importers,
         packages = packages,
-        patched_dependencies = lock_patched_dependencies,
+        patched_dependencies = state.patched_dependencies(),
```
jbedard, line 129: The goal in 2.x is adding:
```
only_built_dependencies = state.only_built_dependencies()`
```

This PR is to allow the parsing of the root package.json to remain only in the `state` util and not bleed into this bzlmod specific file.

gregmagolan (approved):
🧙

Issue #1671, opened 2024-04-20
2.0 RELEASE


gregmagolan:
rules_js 2.x branch: https://github.com/aspect-build/rules_js/tree/2.x

# FEATURES

- perf: reduce size of `npm_package_store` transitive_files depset (#1594)
- feat: add support for linking js_library as 1p npm deps (#1646)
- feat: split js_image_layer into fine grained layers (https://github.com/aspect-build/rules_js/pull/1712)
- feat: support pnpm v9 (https://github.com/aspect-build/rules_js/pull/1679)
- feat: support pnpm.onlyBuiltDependencies (https://github.com/aspect-build/rules_js/pull/1723)

# RCs ARE OUT

These are ready to try,

https://github.com/aspect-build/rules_js/releases/tag/v2.0.0-rc2

https://github.com/aspect-build/rules_cypress/releases/tag/v0.6.0-rc0
https://github.com/aspect-build/rules_esbuild/releases/tag/v0.21.0-rc1
https://github.com/aspect-build/rules_jasmine/releases/tag/v2.0.0-rc0
https://github.com/aspect-build/rules_jest/releases/tag/v0.22.0-rc0
https://github.com/aspect-build/rules_rollup/releases/tag/v2.0.0-rc0
https://github.com/aspect-build/rules_swc/releases/tag/v2.0.0-rc0
https://github.com/aspect-build/rules_ts/releases/tag/v3.0.0-rc0
https://github.com/aspect-build/rules_webpack/releases/tag/v0.16.0-rc0

# BLOCKING TODOS (before 2.0 final)

- [x] https://github.com/aspect-build/rules_js/issues/1445 / https://github.com/aspect-build/rules_js/pull/1701
- [x] https://github.com/aspect-build/rules_js/issues/1641
- [x] https://github.com/aspect-build/rules_js/pull/1694
- [x] https://github.com/aspect-build/rules_js/pull/1690
- [x] https://github.com/aspect-build/rules_js/pull/1701
- [x] https://github.com/aspect-build/rules_js/issues/1652
- [x] https://github.com/aspect-build/rules_js/issues/1689
- [x] https://github.com/aspect-build/rules_js/pull/1731
- [x] https://github.com/aspect-build/rules_js/pull/1734
- [x] https://github.com/aspect-build/rules_js/pull/1730
- [x] https://github.com/aspect-build/rules_js/issues/1753.
- [x] https://github.com/aspect-build/rules_js/issues/1530
[...]

PR #1723, opened 2024-05-14
feat: support pnpm.onlyBuiltDependencies

jbedard:
---

The [pnpm.onlybuiltdependencies](https://pnpm.io/package_json#pnpmonlybuiltdependencies) is now respected for lifecycle hooks. If specified only the packages listed in `pnpm.onlybuiltdependencies` will run lifecycle hooks.

### Test plan

- New test cases added


docs/pnpm.md
```diff
@@ -258,7 +260,15 @@ npm packages have "lifecycle scripts" such as `postinstall` which are documented
 
 We refer to these as "lifecycle hooks".
 
-> You can disable this feature completely by setting all packages to have no hooks, using
+Determining which packages have lifecycle hooks is determined by, in priority order:
+
+1. The workspace `package.json` [`pnpm.onlyBuiltDependencies` attribute](https://pnpm.io/package_json#pnpmonlybuiltdependencies).
+2. The pnpm-lock.yaml `requiresBuild` attribute. Note this is only available in pnpm _before_ v9, see [pnpm #7707](https://github.com/pnpm/pnpm/issues/7707) for reasons why this attribute was removed.
```
gregmagolan, line 266: Reading the code it looks like if `pnpm.onlyBuiltDependencies` is set then `requiresBuild` is ignored:

```
run_lifecycle_hooks = name in only_built_dependencies if only_built_dependencies != None else requires_build
```

That sounds right since if the user wants to use `requilresBuild` from pre-pnpm v9 then they don't set any `onlyBuiltDependencies`. I think this could be made clearer here in the docs.

What happens in the corner case of the user setting `onlyBuiltDependencies` to an empty list `[]` in the package.json?

gregmagolan: Also, if the user sets `onlyBuiltDependencies` in pnpm v9, does the `requiresBuild` in the lock file reflect the user's choices or is that unchanged?

jbedard: > Reading the code it looks like if pnpm.onlyBuiltDependencies is set then requiresBuild is ignored:

Correct. Using `onlyBuiltDependencies` sets an explicit list of packages that have lifecycle hooks, falling back to the `requiresBuild` flag if the `onlyBuiltDependencies` is not set.

Where `npm_translate_lock(lifecycle_hooks)` is a *filter* of those lifecycle hooks.

> What happens in the corner case of the user setting onlyBuiltDependencies to an empty list [] in the package.json?

Leaving `onlyBuiltDependencies` unset implies using the "default" behaviour, which for pnpm v9+ means no lifecycle hooks at all 😢

gregmagolan: Is there a difference for pnpm v8 if you omit `onlyBuiltDependencies` entirely and if you set it to an empty list `[]`. From reading the code I think in the former you'd get a `None` and in the latter you'd get an empty `[]` list out of the `package.json` read which would be different behaviors?

jbedard: In the `e2e/pnpm_lockfiles` if I add `onlyBuiltDependencies: []` all 3 lockfile versions seem to do the same thing: all lifecycle hook targets are gone.

jbedard: I'm thinking once we had v9 we. can edit that test to specify `onlyBuiltDependencies: ["@aspect-test/c"]`. Then it might be nice to assert that things like `.aspect_rules_js/node_modules/fsevents@2.3.3/lc` do *not* have lifecycle hooks, but asserting targets do not exist is a bit more of a hassle 🤷

gregmagolan: Cool. That answer my question 👍

docs/pnpm.md
```diff
@@ -258,8 +260,19 @@ npm packages have "lifecycle scripts" such as `postinstall` which are documented
...
+
+If the `pnpm.onlyBuiltDependencies` attribute is set, even when empty, it is considered to be the full list of
+all packages containing lifecycle hooks. This is recommended going forward for pnpm v9+ compatibility.
+
+When a package has lifecycle hooks the `lifecycle_*` attributes are applied to filter which hooks are run and how they are run.
+
+For example, you can restrict lifecycle hooks across all packages to only run `postinstall`:
+
+> `lifecycle_hooks = { "*": ["postinstall"] }` in `npm_translate_lock`.
```
gregmagolan, line 275: Hmm. Should we error out if you try to specify a lifecycle_hooks entry for a package that is not in an explicity `onlyBuiltDependencies` or does not have a `requiresBuilds` (pre-v9)?

For example, you set

```
lifecycle_hooks = { "foobar": ["postinstall"] }
```

but `foobar` doesn't have lifecycle hooks generated?

gregmagolan: Talked about it offline

gregmagolan (approved):
🌮

docs/pnpm.md
jbedard: Discussed and will be a followup, has been added to the 2.0 release issue: https://github.com/aspect-build/rules_js/issues/1671

PR #1726, opened 2024-05-14
chore: bump to stardoc 0.6.2

gregmagolan:


WORKSPACE
```diff
@@ -202,6 +202,28 @@ npm_import(
...
 )
 
+############################################
+# Stardoc
+load("@io_bazel_stardoc//:setup.bzl", "stardoc_repositories")
+
+stardoc_repositories()
+
+load("@rules_jvm_external//:repositories.bzl", "rules_jvm_external_deps")
+
```
jbedard, line 212: Why are all the jvm things required now?

jbedard (approved):

WORKSPACE
gregmagolan: Stardoc 0.6.2 has really bad transitive deps when using WORKSPACE 🤷

PR #1733, opened 2024-05-15
refactor: align pnpm lock v6+ with v5 target names

jbedard:
Lockfile v5 had `1.2.3_hash` where v6+ has `1.2.3_(patch_hash=1.2.3)`. In 2.x we can add the extra test case that is broken in lockfile v5.3 (which we dropped in 2.x).

---

### Test plan

### DIFF
--- a/npm/private/test/snapshots/wksp/repositories.bzl
+++ b/npm/private/test/snapshots/wksp/repositories.bzl
@@ -14516,21 +14516,21 @@ def npm_repositories():
     )
 
     npm_import(
-        name = "npm__meaning-of-life__1.0.0__-1287509853",
+        name = "npm__meaning-of-life__1.0.0__o3deharooos255qt5xdujc3cuq",
         root_package = "",
         link_workspace = "",
         link_packages = {
             "examples/npm_deps": ["meaning-of-life"],
         },
         package = "meaning-of-life",
-        version = "1.0.0_-1287509853",
+        version = "1.0.0_o3deharooos255qt5xdujc3cuq",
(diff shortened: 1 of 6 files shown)


npm/private/test/snapshots/wksp/repositories.bzl
```diff
@@ -14516,21 +14516,21 @@ def npm_repositories():
...
-        version = "1.0.0_-1287509853",
+        version = "1.0.0_o3deharooos255qt5xdujc3cuq",
         url = "https://registry.npmjs.org/meaning-of-life/-/meaning-of-life-1.0.0.tgz",
         package_visibility = ["//visibility:public"],
         dev = True,
         generate_bzl_library_targets = True,
         integrity = "sha512-fVA4xSydqtK9owabGcYw1r4EKEsMOVVeYQLeCXPu77Z+8Y2j2B2I16UqZlKIOHnYkJ4RSvpJ00ywy9IWjmuxYw==",
         transitive_closure = {
-            "meaning-of-life": ["1.0.0_-1287509853"],
+            "meaning-of-life": ["1.0.0_o3deharooos255qt5xdujc3cuq"],
```
gregmagolan, line 14533: that's an unexpected looking diff given the change `version = version.replace("(patch_hash=", "(")`. How did that happen?

gregmagolan (approved):

npm/private/test/snapshots/wksp/repositories.bzl
jbedard: When a version string is >32 chars we use the `-hash` (which looks ugly after an '_' 😅). After dropping the `patch_hash=` it must no longer be >32...

PR #1734, opened 2024-05-15
fix: fail lockfile translation for pnpm pnpm v9+ when no onlyBuildDependencies specified

jbedard:
To ensure users always set `onlyBuildDependencies` for pnpm 9+ where we can not automatically detect lifecycle events from only the lockfile.

---

### Test plan

### DIFF
--- a/npm/private/npm_translate_lock_helpers.bzl
+++ b/npm/private/npm_translate_lock_helpers.bzl
@@ -565,6 +565,31 @@ def _normalize_bazelignore(lines):
     return result
 
 ################################################################################
+def _verify_lifecycle_hooks_specified(_, state):
+    # lockfiles <9.0 specify the `requiresBuild` flag in the lockfile.
+    #
+    # lockfiles >=9.0 no longer specify if packages have lifecycle hooks,
+    # and declaration of hooks must be done manually in the `pnpm.onlyBuiltDependencies`.
+    if state.lockfile_version() < 9.0:
+        return
+
+    if state.only_built_dependencies() == None:
+        fail("""\
+ERROR: pnpm.onlyBuiltDependencies required in pnpm workspace root package.json when using pnpm v9 or later
+
+As of pnpm v9, the lockfile no longer specifies if packages have lifecycle hooks.
+
+Packages that rules_js should generate lifecycle hook actions for must now be declared in
+pnpm.onlyBuiltDependencies in the pnpm workspace root package.json. See
+https://pnpm.io/package_json#pnpmonlybuiltdependencies for more information.
+
+Prior to pnpm v9, rules_js keyed off of the requiresBuild attribute in the pnpm lock
+file to determine if a lifecycle hook action should be generated for an npm package.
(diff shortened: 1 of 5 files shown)


npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -564,6 +564,27 @@ def _normalize_bazelignore(lines):
...
+    # lockfiles >=9.0 no longer specify if packages have lifecycle hooks,
+    # and declaration of hooks must be done manually in the `pnpm.onlyBuiltDependencies`.
+    if state.lockfile_version() < 9.0:
+        return
+
+    if state.only_built_dependencies() == None:
+        fail("""\
+ERROR: in verify_lifecycle_hooks_specified:
+
+As of pnpm v9 the lockfile no longer specifies if packages have lifecycle hooks.
```
gregmagolan, line 580: Lets add a little some more "why" and a link to that pnpm issue re: requiresBuild to this error message. The more verbose the better here as this is part of the upgrade documentation.

```
As of pnpm v9, the lockfile no longer specifies if packages have lifecycle hooks.

Packages that rules_js should generate lifecycle hook actions for must now be declared in pnpm.onlyBuiltDependencies in the pnpm workspace root package.json. See https://pnpm.io/package_json#pnpmonlybuiltdependencies for more information.

Prior to pnpm v9, rules_js keyed off of the requiresBuild attribute in the pnpm lock file to determine if a lifecycle hook action should be generated for an npm package. See <link> for the reasons that pnpm removed the requiredBuild attribute from the lockfile in v9.
```

gregmagolan: When we add the "strict" check we can mention that in this error message as well.

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -564,6 +564,27 @@ def _normalize_bazelignore(lines):
...
+    # lockfiles <9.0 specify the `requiresBuild` flag in the lockfile.
+    #
+    # lockfiles >=9.0 no longer specify if packages have lifecycle hooks,
+    # and declaration of hooks must be done manually in the `pnpm.onlyBuiltDependencies`.
+    if state.lockfile_version() < 9.0:
+        return
+
+    if state.only_built_dependencies() == None:
+        fail("""\
+ERROR: in verify_lifecycle_hooks_specified:
```
gregmagolan, line 578: "verify_lifecycle_hooks_specified" won't be very helpful for users. Lets just summarize:

```
ERROR: pnpm.onlyBuiltDependencies required in pnpm workspace root package.json when using pnpm v9 or later
```

gregmagolan (approved):
🌮

npm/private/npm_translate_lock_helpers.bzl
jbedard: For some reason all the messages in this file start with `ERROR: method_name` 🤷

gregmagolan: 🤷

PR #1735, opened 2024-05-15
refactor: normalize pnpm lockfile data in parser

jbedard:
This way the data is standardized to a format for rules_js at parse time instead of elsewhere throughout the code.

Today the lockfile parsing normalizes to a version similar to pnpm lock v5, then `_gather_package_info` normalizes that a bit to a more rules_js specific format. The idea of this PR is to directly convert to that rules_js format as part of parsing, instead of `pnpm vX => pnpm v5 => rules_js` (where X = 5,6,9) we will have `pnpm vX => rules_js` and ideally have less code...

---

### Test plan

### DIFF
--- a/npm/private/test/parse_pnpm_lock_tests.bzl
+++ b/npm/private/test/parse_pnpm_lock_tests.bzl
@@ -15,6 +15,38 @@ def _parse_empty_lock_test_impl(ctx):
 
     return unittest.end(env)
 
+expected_importers = {
+    ".": {
+        "dependencies": {
+            "@aspect-test/a": "5.0.0",
+        },
+        "dev_dependencies": {},
+        "optional_dependencies": {},
+    },
+}
+expected_packages = {
+    "@aspect-test/a@5.0.0": {
+        "id": None,
+        "name": "@aspect-test/a",
+        "dependencies": {
+            "@aspect-test/b": "5.0.0",
+            "@aspect-test/c": "1.0.0",
+            "@aspect-test/d": "2.0.0_at_aspect-test_c_1.0.0",
+        },
+        "optional_dependencies": {},
+        "peer_dependencies": {},
+        "dev": False,
+        "has_bin": True,
+        "optional": False,
+        "requires_build": False,
+        "version": "5.0.0",
+        "friendly_version": "5.0.0",
+        "resolution": {
+            "integrity": "sha512-t/lwpVXG/jmxTotGEsmjwuihC2Lvz/Iqt63o78SI3O5XallxtFp5j2WM2M6HwkFiii9I42KdlAF8B3plZMz0Fw==",
+        },
+    },
+}
+
 def _parse_lockfile_v5_test_impl(ctx):
     env = unittest.begin(ctx)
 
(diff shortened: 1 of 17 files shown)


npm/private/test/parse_pnpm_lock_tests.bzl
```diff
@@ -44,30 +76,15 @@ def _parse_lockfile_v5_test_impl(ctx):
 }
 """)
 
+    # v5 formatted peer deps different in many ways including how peer packages were escaped
+    v5_expected_packages = dict(expected_packages)
```
jbedard, line 80: This highlights the differences between the "norm" and v5. I think these v5 vs v6 diffs is a bug and v5 is the one which should change, but that might not be worth our time at this point.

Note the v9 version of this are correct exceptions, data no longer in the lockfile (`requiresBuild` being the major one).

npm/private/utils.bzl
```diff
@@ -151,36 +303,70 @@ def _convert_v6_packages(packages):
...
+            requires_build = package_snapshot.get("requiresBuild", False),
+            resolution = package_snapshot.get("resolution"),
+        )
 
         if package_key in result:
             msg = "ERROR: duplicate package: {}\n\t{}\n\t{}".format(package_key, result[package_key], package_info)
-            fail(msg)
+
+            # buildifier: disable=print
+            print(msg)
```
gregmagolan, line 352: no longer a failure?

jbedard: I'll have to check again, but I think when upgrading the rules_js pnpm-lock it actually hit this case, there's a bug (which this PR neither introduces or fixes) where an oddly configured package has two entries with the same key.

jbedard: I'll have to check again, it's possible I switched it to `print` while debugging 🤔

jbedard: If you do a build on 2.x right now you'll see:
```
DEBUG: /Users/jason/dev/rules_js/npm/private/transitive_closure.bzl:164:18: ERROR: duplicate package info: lodash/4.17.21
	{"name": "lodash", "id": None, "version": "4.17.21", "friendly_version": "4.17.21", "resolution": {"integrity": "sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg=="}, "dependencies": {}, "optional_dependencies": {}, "dev": True, "optional": False, "has_bin": False, "requires_build": False}
	{"name": "lodash", "id": None, "version": "4.17.21", "friendly_version": "4.17.21", "resolution": {"integrity": "sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg==", "tarball": "file:npm/private/test/vendored/lodash-4.17.21.tgz"}, "dependencies": {}, "optional_dependencies": {}, "dev": True, "optional": False, "has_bin": False, "requires_build": False}
```

This PR now normalizes all the data in the parser, so that duplicate key is now caught here so needs to just be a warning like it was previously in `transitive_closure.bzl`. Basically the bug got pushed into the parser and I don't want this refactor to turn the warning into a failure. For lockfile v9 I hope to keep it as an error, it's just the v6 that needs to be non-breaking.

gregmagolan (approved):
🌮

PR #1744, opened 2024-05-17
refactor: detect and log duplicate package keys

jbedard:
Some of these were useful when debugging pnpm 9, some name changes just for consistency, one WARNING for a case that I think is a bug today.

---

### Test plan

### DIFF
--- a/npm/private/npm_translate_lock_helpers.bzl
+++ b/npm/private/npm_translate_lock_helpers.bzl
@@ -269,8 +269,8 @@ def _get_npm_imports(importers, packages, patched_dependencies, only_built_depen
         importer_links[import_path] = links
 
     patches_used = []
-    result = []
-    for package, package_info in packages.items():
+    result = {}
+    for package_key, package_info in packages.items():
         name = package_info.get("name")
         version = package_info.get("version")
         friendly_version = package_info.get("friendly_version")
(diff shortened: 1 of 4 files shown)


npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -467,7 +467,15 @@ ERROR: can not apply both `pnpm.patchedDependencies` and `npm_translate_lock(pat
             package_info = package_info,
             dev = dev,
             replace_package = replace_package,
-        ))
+        )
+
+        if repo_name in result:
```
jbedard, line 472: This is an error/bug that exists with the lockfile on main right now.

gregmagolan (approved):
🦖

npm/private/transitive_closure.bzl
```diff
@@ -155,12 +155,19 @@ def translate_to_transitive_closure(lock_importers, lock_packages, prod = False,
     package_version_map = {}
 
     for package_path, package_snapshot in lock_packages.items():
-        package, package_info = _gather_package_info(package_path, package_snapshot)
-        packages[package] = package_info
+        package_key, package_info = _gather_package_info(package_path, package_snapshot)
+
+        if package_key in packages:
```
jbedard, line 160: This is an error/bug that exists with the lockfile on main right now.

The key `_gather_package_info` produces is not unique for at least 1 case where there are 2 versions of a dependency with the exact same name+version but a different URL.

(Also note I renamed `package` to `package_key` since it is supposed to be a unique key...)

PR #1745, opened 2024-05-17
test: add snapshots of lockfile repositories across pnpm versions

jbedard:
### Test plan

- New test cases added


e2e/pnpm_lockfiles/base/package.json
```diff
@@ -3,7 +3,6 @@
     "private": true,
     "dependencies": {
         "@aspect-test/a": "^5.0.2",
-        "@aspect-test/a2": "npm:@aspect-test/a",
```
jbedard, line 6: This actually fails with pnpm9 on a afresh install. Where it was working when run with an existing lockfile!? Sounds like a pnpm9 bug...

gregmagolan: Fails as in `pnpm install` exits non-zero?

gregmagolan (approved):

e2e/pnpm_lockfiles/base/package.json
jbedard: Yeah

Issue #1753, opened 2024-05-22
[Bug]: pnpm v9 lockfile issues


jbedard:
Not all pnpm features are supported with lockfiles generated with pnpm9

Issues:
* `npm:` deps (examples: [1749](https://github.com/aspect-build/rules_js/issues/1749#issuecomment-2126342101)) (https://github.com/aspect-build/rules_js/pull/1754, https://github.com/aspect-build/rules_js/pull/1767)
* `http:` deps (and related `https`, `http+git` etc) (https://github.com/aspect-build/rules_js/pull/1765)
* custom named packages depending on `https`, `http+git` etc (https://github.com/aspect-build/rules_js/pull/1765, https://github.com/aspect-build/rules_js/pull/1770)
* `npm:` aliases (https://github.com/aspect-build/rules_js/pull/1770)

jbedard:
I think these are all fixed, I'll create new issues for anything else.

PR #1754, opened 2024-05-23
fix(pnpm): url and git package specifiers

jbedard:
Ref #1753

---

### Test plan

- New test cases added


e2e/pnpm_lockfiles/base/package.json
```diff
@@ -3,6 +3,9 @@
     "private": true,
     "dependencies": {
         "@aspect-test/a": "^5.0.2",
+        "debug": "ngokevin/debug#9742c5f383a6f8046241920156236ade8ec30d53",
+        "hello": "https://gitpkg.vercel.app/EqualMa/gitpkg-hello/packages/hello",
+        "jsonify": "https://github.com/aspect-build/test-packages/releases/download/0.0.0/@foo-jsonify-0.0.0.tgz",
```
gregmagolan, line 8: Are there more from `npm/private/test/package.json` that should be pulled into `e2e/pnpm_lockfiles`?

jbedard: Yeah there probably are. I need to get this PR actually working before I add much more though.

The `debug` and `hello` ones are unique cases (odd git syntax, https syntax). The `jsonify` one is also plain http but has a more interesting URL which requires escaping.

jbedard:
What was originally intended to be a prefactor for this has turned into the fix: https://github.com/aspect-build/rules_js/pull/1765

PR #1757, opened 2024-05-26
refactor: simplify handling of #423, package aliases

jbedard:
---

### Test plan

### DIFF
--- a/npm/private/utils.bzl
+++ b/npm/private/utils.bzl
@@ -734,7 +734,6 @@ utils = struct(
     parse_pnpm_lock_json = _parse_pnpm_lock_json,
     friendly_name = _friendly_name,
     package_store_name = _package_store_name,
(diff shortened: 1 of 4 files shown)


npm/private/utils.bzl
```diff
@@ -755,4 +752,9 @@ utils = struct(
     exists = _exists,
     replace_npmrc_token_envvar = _replace_npmrc_token_envvar,
     is_tarball_extension = _is_tarball_extension,
+
+    # Only for testing
+    npm_registry_url = _npm_registry_url,
```
gregmagolan, line 757: export via another struct named `test_utils` so that it is enforced by the load statement?

gregmagolan (approved):

npm/private/utils.bzl
jbedard: I was pondering that. Means the tests will import two things (normal + test-only) but makes this clearer 👍

PR #1758, opened 2024-05-27
refactor: move npm_registry_url to utils

jbedard:
I'm not sure why I missed this, but it looks like this path is not tested on CI until merging to main? See https://github.com/aspect-build/rules_js/actions/runs/9248267733/job/25438315100

### Test plan


e2e/pnpm_lockfiles/cases/tarball-no-url-v54.yaml
```diff
@@ -0,0 +1,19 @@
...
+        dependencies:
+            '@aspect-build/a': 1.0.0
+
+packages:
+    # This was found in https://github.com/aspect-build/rules_js/blob/795138561e878c09f0b62a25572358d64960d732/e2e/npm_translate_lock_auth/pnpm-lock.yaml
+    /@aspect-build/a/1.0.0:
+        resolution:
+            {
+                integrity: sha512-MYeL/yqAPYJXOnnSEOAdtQSu/8tiifFtKN6Jg/rgpKRqxKL8NVsQXrX9H2dlJ4mS23pu7VS0+i9mZNiRoCUYwg==,
+                tarball: '@aspect-test/a/-/a-1.0.0.tgz',
```
jbedard, line 16: This is the edge case that was not tested until merging to `main` where `e2e/npm_translate_lock_auth` is actually run. The case where the `tarball` is a path relative to the default registry instead of a full URL.

Note this file does not have the normal `pnpm-lock.yaml` name, it was autoformatted, I added comments, renovate won't run on it etc. This is all intentional since it is a very specfiic edge case. We can debate where that test case should live but this `e2e/pnpm_lockfiles` seems the best atm 🤷

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -426,15 +426,16 @@ ERROR: can not apply both `pnpm.patchedDependencies` and `npm_translate_lock(pat
                 # in pnpm-lock.yaml which we must replace with the desired registry in the `registry` field:
                 #   tarball: https://registry.npmjs.org/@types/cacheable-request/-/cacheable-request-6.0.2.tgz
                 #   registry: https://registry.yarnpkg.com/
-                if registry and tarball.startswith(utils.default_registry()):
-                    url = registry + tarball[len(utils.default_registry()):]
+                if registry and tarball.startswith(default_registry):
```
jbedard, line 429: This should have no change since `utils.default_registry()` is passed into this method already, but we should use the param instead of invoking that util again.

e2e/pnpm_lockfiles/cases/tarball-no-url-v54.yaml
gregmagolan: In that e2e test the tarball is `download/@aspect-build/a/1.0.0/95cfc2dab39ada95aec1736778d545403edfbf9e`

```
  /@aspect-build/a/1.0.0:
    resolution: {integrity: sha512-TQ8SpBPjKkHmLJoxVTU+TRU+9utF9hABiFuql5zJWPAtyUGAc3KM5+VsgpNHytqyXe5AyOUV2Ff+ScsQLNEQ7Q==, tarball: download/@aspect-build/a/1.0.0/95cfc2dab39ada95aec1736778d545403edfbf9e}
    hasBin: true
    dev: false
```

jbedard: That's because the npmrc in that e2e test changes URL to one which requires auth. I changed it here so it works with the default `registry.npmjs.org` while keeping the odd path-without-domain `tarball` field.

WDYT?

gregmagolan: sgtm

gregmagolan (approved):

PR #1759, opened 2024-05-28
refactor: do not include registry url in package keys

jbedard:
Pre-v9 pnpm lock files are a bit of a mess, as I keep learning.

It seems like when a non-default registry is used on a package the `package.*` key in pnpm-lock.yaml then gets *the default registry* as a prefix. Then the `resolution` gets `{registry: https://registry.yarnpkg.com/, tarball: https://registry.npmjs.org}` (that's right, the `tarball` points to the default registry, despite our npmrc configuring the registry to the yarnpkg).  See https://github.com/aspect-build/rules_js/blob/v2.0.0-rc1/npm/private/npm_translate_lock_helpers.bzl#L424-L429

This PR drops that default registry from the `package.*` keys, which means all the generated bzl files, all the arrays of dep versions etc etc.

See the diffs on all the snapshot tests.

---

### Test plan


gregmagolan (approved):
🧙

PR #1767, opened 2024-05-29
fix(pnpm): support npm: alias dependencies with pnpm lockfile v5

jbedard:
It's possible this bug has existed with lockfile v5 for quite some time, although I'm not certain.

--

### Test plan

- New test cases added


gregmagolan (approved):
🌮

PR #1770, opened 2024-05-31
fix(pnpm): support npm: aliases to packages in alternate registries

jbedard:
A few cases not covered, at least one which was introduced when I dropped the `registry.npmjs.com/prefix` from the labels:
* aliases to a package with that `registry.npmjs.com/` prefix
* aliases to packages not directly listed in the package.json

Other fixes:
* v90 versions that include `@` such as URLs containing random `@`s in them

---

### Test plan

- New test cases added


npm/private/utils.bzl
```diff
@@ -66,26 +66,10 @@ def _parse_pnpm_package_key(pnpm_name, pnpm_version):
...
 
-    if not pnpm_version.startswith("/"):
-        if not pnpm_name:
-            fail("parse_pnpm_package_key: pnpm_name is empty for non-versioned package %s" % pnpm_version)
+    if pnpm_version.startswith("npm:"):
+        return pnpm_version[4:].rsplit("@", 1)
 
-        return pnpm_name, pnpm_version
-
-    # Parse a package key such as:
```
jbedard, line 75: These are all handled and normalized at parse-time, most from before this PR and now the final `npm:` case.

npm/private/utils.bzl
```diff
@@ -434,29 +491,27 @@ def _convert_v9_packages(packages, snapshots):
...
         package_info = _new_package_info(
-            id = package_data.get("id", None),
+            id = package_data.get("id", None),  # TODO: does v9 have "id"?
             name = name,
             version = version,
             friendly_version = friendly_version,
             dependencies = _convert_pnpm_v9_package_dependency_map(package_snapshot.get("dependencies", {})),
             optional_dependencies = _convert_pnpm_v9_package_dependency_map(package_snapshot.get("optionalDependencies", {})),
-            peer_dependencies = _convert_pnpm_v9_package_dependency_map(package_data.get("peerDependencies", {})),
+            peer_dependencies = package_data.get("peerDependencies", {}),
```
jbedard, line 514: Peer data is from the static `packages` data, not the `snapshots` where all the peer data has been mixed in so the callback isn't needed.

npm/private/utils.bzl
```diff
@@ -379,16 +405,47 @@ def _convert_v6_packages(packages):
...
+            version = "npm:{}@{}".format(alias, version)
+        else:
+            # Transition version[(patch)(peer)(data)] to a rules_js version format
+            version = _convert_pnpm_v6_v9_version_peer_dep(version)
 
+        result[name] = version
     return result
 
-# v9 importers are the same as v6 importers
-_convert_v9_importers = _convert_v6_importers
```
jbedard, line 391: v9 has less edge cases so I copy+pasted and then deleted some code basically, mixing the two was getting too confusing

gregmagolan (approved):

PR #1772, opened 2024-05-31
fix(pnpm): npm aliases to url specifiers 

jbedard:
---

### Test plan

- New test cases added

### DIFF
--- a/npm/private/utils.bzl
+++ b/npm/private/utils.bzl
@@ -194,7 +194,7 @@ def _convert_v5_v6_file_package(package_path, package_snapshot):
     else:
         friendly_version = package_snapshot["version"] if "version" in package_snapshot else version
 
-    return _to_package_key(name, version), name, version, friendly_version
+    return name, version, friendly_version
 
 def _convert_v5_packages(packages):
     result = {}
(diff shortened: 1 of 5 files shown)


npm/private/utils.bzl
```diff
@@ -194,7 +194,7 @@ def _convert_v5_v6_file_package(package_path, package_snapshot):
     else:
         friendly_version = package_snapshot["version"] if "version" in package_snapshot else version
 
-    return _to_package_key(name, version), name, version, friendly_version
```
jbedard, line 197: This evolved a bit and no longer needs special cases in this `file:` method.

npm/private/utils.bzl
```diff
@@ -501,9 +501,6 @@ def _convert_v9_packages(packages, snapshots):
         # package_data can have the resolved "version" for things like https:// deps
         friendly_version = package_data["version"] if "version" in package_data else static_key[version_index + 1:]
 
-        # Convert the package_key to a rules_js compatible format
-        package_key = _to_package_key(name, version)
```
jbedard, line 505: pnpm9 aliases such as `my-name: http://with/other/name` result in a package key such as `real-name@http://with/other/name`.

In this case `_to_package_key` checking if the first character is a digit and prepending the `my-name` was the bug.  With pnpm9 the `package_key` is already correct here and we don't need anything (where v5-6 still have this).

gregmagolan (approved):

PR #1775, opened 2024-06-03
fix(pnpm): pnpm9 transitive npm: dependenies

jbedard:
This is for transitive `npm:` dependencies, previously we only handled direct `npm:` deps with v9.

---

### Test plan

- New test cases added


gregmagolan (approved):
🫠

PR #1776, opened 2024-06-04
fix(pnpm): pnpm9 npm: dependencies with peers

jbedard:
Transitive `npm:` deps that contain peers were not being escaped.

---

### Test plan

- New test cases added

### DIFF
--- a/e2e/pnpm_lockfiles/cases/BUILD.bazel
+++ b/e2e/pnpm_lockfiles/cases/BUILD.bazel
@@ -1,4 +1,6 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
+load("@docusaurus-direct-peer-v6//:defs.bzl", docusaurus_direct_with_peers_v6_link_all = "npm_link_all_packages")
+load("@docusaurus-direct-peer-v9//:defs.bzl", docusaurus_direct_with_peers_v9_link_all = "npm_link_all_packages")
 load("@isaacs-cliui-v90//:defs.bzl", isaacs_cliui_v90_link_all = "npm_link_all_packages")
 load("@override-with-alias-url-v9//:defs.bzl", override_with_alias_link_all = "npm_link_all_packages")
 load("@tarball-no-url-v54//:defs.bzl", tarball_no_url_link_all = "npm_link_all_packages")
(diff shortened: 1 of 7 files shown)


e2e/pnpm_lockfiles/cases/BUILD.bazel
```diff
@@ -34,3 +36,19 @@ build_test(
...
+
+docusaurus_direct_with_peers_v6_link_all(name = "docusaurus_direct_with_peers_v6-modules")
+
+build_test(
+    name = "docusaurus_direct_with_peers",
+    targets = [
+        ":docusaurus_direct_with_peers_v9-modules",
+        ":docusaurus_direct_with_peers_v9-modules/@docusaurus/module-type-aliases",
+
+        # TODO: causes 'conflicting actions' error
```
jbedard, line 50: @gregmagolan this seems like a different issue with the virtual store. Shouldn't 2 `npm_link_all_packages()` with different names be independent of each other? But they seem to both be writing to the parent `.aspect_rules_js`...

gregmagolan: Under the hood both of `docusaurus_direct_with_peers_v9_link_all` and `docusaurus_direct_with_peers_v6_link_all` are going to create targets that output to `node_modules/.aspect_rules_js` when they are called in the root of the pnpm workspace. The `name = "docusaurus_direct_with_peers_v9-modules"` and `name = "docusaurus_direct_with_peers_v6-modules"` only control the label names and not the output paths.

jbedard: fwiw this is the error:
```
PrimaryOutput: File:[[<execution_root>]bazel-out/darwin_arm64-fastbuild/bin]cases/node_modules/.aspect_rules_js/inline-style-parser@0.1.1/node_modules/inline-style-parser
Owner information: ConfiguredTargetKey{label=//cases:.aspect_rules_js/docusaurus_direct_with_peers_v5-modules/inline-style-parser@0.1.1/pkg, config=BuildConfigurationKey[5ad37220f5139d1297b85ba7bb63120be5a04113f839838c453bdb47ae9f3712]}, ConfiguredTargetKey{label=//cases:.aspect_rules_js/docusaurus_direct_with_peers_v9-modules/inline-style-parser@0.1.1/pkg, config=BuildConfigurationKey[5ad37220f5139d1297b85ba7bb63120be5a04113f839838c453bdb47ae9f3712]}
MandatoryInputs: Attempted action contains artifacts not in previous action (first 5): 
	package.tgz
Previous action contains artifacts not in attempted action (first 5): 
	package.tgz
Outputs: are equal
```

jbedard: I guess these tests just need to be in separate directories though 🤷

gregmagolan: Yeah, they need to go in separate directories since both lock files have the same package from the looks of it

gregmagolan (approved):

e2e/pnpm_lockfiles/cases/BUILD.bazel
jbedard: It's the exact same packages, just different lockfile versions since this `npm:` + peer stuff changed quite a bit between versions.

gregmagolan: From Bazel's perspective it is two different targets wanting to the create the same output. If all the inputs were the same then Bazel would merge them in the graph but something about the inputs is mismatched and I doubt its worth the time to track down since this setup will only show up for test cases

PR #1777, opened 2024-06-04
refactor(pnpm): move pnpm parsing logic to pnpm.bzl

jbedard:
Is this worth loosing the `git blame` history on all this code? The `utils.bzl` is getting getting pretty big and I keep thinking this stuff should live on its own...

Some of the moved code was reordered to group things by pnpm lockfile version a bit better, I think I added `_v5` to one method name to highlight that it's only for v5 lockfiles, otherwise the moved code is unchanged.

---

### Test plan


npm/private/utils.bzl
```diff
@@ -670,7 +104,8 @@ def _npm_registry_download_url(package, version, registries, default_registry):
         registry.removesuffix("/"),
         package,
         package_name_no_scope,
-        _strip_v5_peer_dep_or_patched_version(version),
+        # Strip the rules_js peer/patch metadata off the version. See pnpm.bzl
+        version[:version.find("_")] if version.find("_") != -1 else version,
```
jbedard, line 108: This is the one actual change in this file, inlining that `_strip_v5_peer_dep_or_patched_version` method to here. It's inlined so no functional change, but this method is called with data in the rules_js format and not in the pnpm lockfile format so I think this is better anyway 🤷

PR #1781, opened 2024-06-08
refactor: add option for optimized runfiles collection

gregmagolan:
---

### Test plan

- New test cases added


js/private/js_helpers.bzl
```diff
@@ -114,6 +114,73 @@ this option is not needed.
...
+    transitive_files_depsets = [sources]
+
+    files_runfiles = []
+    for d in data_files:
+        if copy_data_files_to_bin and d.is_source and d not in no_copy_to_bin:
+            files_runfiles.append(copy_js_file_to_bin_action(ctx, d))
+        else:
+            files_runfiles.append(d)
+
+    if len(files_runfiles) > 0:
```
jbedard, line 172: See https://github.com/aspect-build/rules_js/pull/1782 and drop it here as well if it lands

js/private/js_helpers.bzl
```diff
@@ -114,6 +114,73 @@ this option is not needed.
...
+        A [runfiles](https://bazel.build/rules/lib/runfiles) object created with [ctx.runfiles](https://bazel.build/rules/lib/ctx#runfiles).
+    """
+
+    # Includes sources
+    if type(sources) == "list":
+        sources = depset(sources)
+    transitive_files_depsets = [sources]
+
+    files_runfiles = []
+    for d in data_files:
```
jbedard, line 166: See https://github.com/aspect-build/rules_js/pull/1783 and apply it here as well if it lands

js/private/js_binary.bzl
```diff
@@ -259,6 +259,16 @@ _ATTRS = {
         """,
         default = True,
     ),
+    "optimized_runfiles_collection": attr.bool(
```
jbedard, line 262: How about "include legacy runfiles" and `False` by default? Something to make it sound old/bad/deprecated...

gregmagolan: Perhaps... tho "legacy" doesn't quite describe it in my mind. It is an overbroad, defence-in-depth runfiles collection mode that some users may need to use if they are using deps from other rule sets or custom rules that don't define their runfiles correctly.

jbedard: I guess I mean to to highlight the "don't define their runfiles correctly" part of it...

gregmagolan:
Superseded by https://github.com/aspect-build/rules_js/pull/2615

PR #1785, opened 2024-06-08
perf: avoid js_info param validation on default values

jbedard:
If we're using default params then there's no reason to do `type(p)` to validate them.

---

### Test plan


gregmagolan (approved):
🌮

PR #1786, opened 2024-06-08
perf: do not check JsInfo hasattr for always-set attributes

jbedard:
If we state that all these attributes are "required" on `JsInfo` and we enforce that in the `js_info` factory then we can avoid checking if the attributes are present everywhere we access them.

See alternative: https://github.com/aspect-build/rules_js/pull/1791

---

### Test plan


gregmagolan:
This is conflicting now

gregmagolan (approved):
Looks good to me. Lets assume JsInfos are well formed in rules_js 2

PR #1787, opened 2024-06-08
perf: change js_helpers gather_* utils to accept list, not depset

jbedard:
In almost all use cases these are lists, or for `gather_runfiles` always depsets. rules_ts seems to follow this pattern already

---

### Test plan

### DIFF
--- a/js/private/js_helpers.bzl
+++ b/js/private/js_helpers.bzl
@@ -8,39 +8,35 @@ def gather_transitive_sources(sources, targets):
     """Gathers transitive sources from a list of direct sources and targets
 
     Args:
-        sources: list or depset of direct sources which should be included in `transitive_sources`
+        sources: list of direct sources which should be included in `transitive_sources`
         targets: list of targets to gather `transitive_sources` from `JsInfo`
 
     Returns:
         A depset of transitive sources
     """
-    if type(sources) == "list":
-        sources = depset(sources)
     transitive = [
         target[JsInfo].transitive_sources
         for target in targets
         if JsInfo in target
     ]
-    return depset(transitive = [sources] + transitive)
+    return depset(sources, transitive = transitive)
 
 def gather_transitive_types(types, targets):
     """Gathers transitive types from a list of direct types and targets
 
     Args:
-        types: list or depset of direct sources which should be included in `transitive_types`
+        types: list of direct sources which should be included in `transitive_types`
         targets: list of targets to gather `transitive_types` from `JsInfo`
 
     Returns:
         A depset of transitive sources
     """
-    if type(types) == "list":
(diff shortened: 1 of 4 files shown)


js/private/js_helpers.bzl
```diff
@@ -177,8 +173,6 @@ def gather_runfiles(
     """
 
     # Includes sources
-    if type(sources) == "list":
```
DavidZbarsky-at, line 180: recommend to make it optional so caller isn't forced to construct an empty depset and can pass None instead

jbedard: I guess I would prefer that as well, just means changing the method signature 🤷

js/private/js_helpers.bzl
```diff
@@ -138,7 +134,7 @@ def gather_runfiles(
     Args:
         ctx: the rule context
 
-        sources: list or depset of files which should be included in runfiles
+        sources: depset which should be included in runfiles
```
jbedard, line 137: Note that this one switched to `depset` instead of `list` like the others. It seems like all use cases of `gather_runfiles` in rules_js+ts are already passing a `depset` (except the one in rules_js that was just passing an empty list...).

js/private/js_helpers.bzl
```diff
@@ -8,39 +8,35 @@ def gather_transitive_sources(sources, targets):
     """Gathers transitive sources from a list of direct sources and targets
 
     Args:
-        sources: list or depset of direct sources which should be included in `transitive_sources`
+        sources: list of direct sources which should be included in `transitive_sources`
```
gregmagolan, line 11: should be "depset of direct sources"

js/private/js_library.bzl
```diff
@@ -207,20 +207,23 @@ def _js_library_impl(ctx):
     sources = depset(transitive = [sources, additional_sources])
     types = depset(transitive = [types, additional_sources, additional_types])
 
-    transitive_sources = gather_transitive_sources(
-        sources = sources,
-        targets = ctx.attr.srcs + ctx.attr.types + ctx.attr.deps,
-    )
-
-    transitive_types = gather_transitive_types(
-        types = types,
```
jbedard, line 216: I think this (and the `gather_transitive_sources` above) are a unique case where the first param is a depset instead of list of files.

Is this because `js_library` is unique where it accepts `JsInfo`-producing targets as `srcs` + `types`?

This is the reason removing the use of these helpers in this unique spot is required for this PR...

DavidZbarsky-at:
This would be great! It saves around 1.5s of analysis time for us (close to 10%)

gregmagolan (approved):
🦖

PR #1791, opened 2024-06-08
perf: persist None instead of empty depsets in JsInfo

jbedard:
This way when we collect transitive depsets of these it will be a significantly smaller tree in some cases. Smaller, but only because empty `depset`s are not there. Do empty depsets make any difference?

Alternative to https://github.com/aspect-build/rules_js/pull/1786

---

### Test plan


jbedard:
Closed in favor of https://github.com/aspect-build/rules_js/pull/1786

Issue #1637, opened 2024-04-10
[Bug]: NpmPackageExtract extraction of malformed packages causes errors in RBE/Remote Cache


qtica:
### What happened?

When depending on an npm package that has an archive with directories that do not contain the searchable permission (e.g. 
 `drw-r--r--`), the bazel cache fails to write the action output to the cache getting a permission denied on the files in the extracted non-searchable directory. If you are using remote execution, you may have the build exit due to the executors failing with the same permission denied. This started with the latest feature if extracting tars vis bsdtar. The prior method of  extracting packages did not seem to have any issue.

I also think it would be nice if rules_js could gracefully handle this scenario considering npm and pnpm appear to handle extracting the package okay. However, at the end of the day, I don't really know how frequent this is, nor if it's really worth the effort of maintaining on your side. I mostly just wanted to report it for visibility.

Related issue on package where I noticed this: https://github.com/pngjs/pngjs/issues/381

I've also reported this to the remote executions service I am using, as it crashes the remote executor.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: `7.1.1`

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: `1.4.0`

Language(s) and/or frameworks involved:
* npm package [pngjs](https://github.com/pngjs/pngjs)

### How to reproduce

A full reproducible repo can be found here: https://github.com/qtica/bazel-rules-js-cache-error-with-bad-package?tab=readme-ov-file

In short, depend on `pngjs`, have RBE / remote cache (--disk_cache works too) setup and build your target. The remote cache will provide a warning with permission denied. The remote execution service I use results in a crashed executor with the error permission denied resulting in a failed build.

The full details are in the README.md of the reproducible repo.

### Any other information?

My current workaround is to vendor the package and provide overrides in the `package.json`.

```json5
{
  // <redacted>
  "pnpm": {
    "overrides": {
      "pngjs": "file:./third_party/npm/pngjs-5.0.0"
    }
  }
}
```

gregmagolan:
For future reference, one our of consulting clients hit something similar. Here is the internal Slack thread link for that: https://aspect-build.slack.com/archives/C04PGV0U79P/p1715100472949349?thread_ts=1715093034.381059&cid=C04PGV0U79P

gregmagolan:
[...]

PR #1794, opened 2024-06-08
fix: handle malformed npm packages gracefully in extract action

gregmagolan:
Fixes https://github.com/aspect-build/rules_js/issues/1637.

Inspired by https://github.com/frc971/971-Robot-Code/blob/master/third_party/rules_js/0001-Fix-package-permissions.patch. Thanks @AustinSchuhBRT.

Similar to what is already done in `npm_import.bzl`: https://github.com/aspect-build/rules_js/blob/f52cc4c4e9fd3350661ab18e3b995ccce3b90894/npm/private/npm_import.bzl#L489

`pngjs@5.0.0`, added to `npm/private/test/package.json`, is a package known to have a malformed tarball where the directories don't have execute/list permissions.


.aspect/rules/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU=
```diff
@@ -2,26 +2,26 @@
...
 examples/js_binary/package.json=-41174383
+examples/linked_empty_node_modules/package.json=-1039372825
 examples/macro/package.json=857146175
 examples/npm_deps/package.json=-1377141392
 examples/npm_package/libs/lib_a/package.json=-1377103079
 examples/npm_package/packages/pkg_a/package.json=1006424040
 examples/npm_package/packages/pkg_b/package.json=1041247977
 examples/webpack_cli/package.json=1911342006
 js/private/coverage/bundle/package.json=-1543718929
+js/private/image/package.json=-1260474848
```
jbedard, line 18: Why did the order of these change? 🤔

npm/private/npm_package_store.bzl
```diff
@@ -205,25 +205,20 @@ def _npm_package_store_impl(ctx):
...
+                # `chmod -R a+X` to fix up these packages (https://stackoverflow.com/a/14634721).
+                # See https://github.com/aspect-build/rules_js/issues/1637 for more info.
                 bsdtar = ctx.toolchains["@aspect_bazel_lib//lib:tar_toolchain_type"]
-                ctx.actions.run(
-                    executable = bsdtar.tarinfo.binary,
+                ctx.actions.run_shell(
+                    tools = [bsdtar.tarinfo.binary],
                     inputs = depset(direct = [src], transitive = [bsdtar.default.files]),
                     outputs = [package_store_directory],
-                    arguments = [args],
```
jbedard, line 226: Why can't we still use `arguments`?

npm/private/test/snapshots/bzlmod/npm_defs.bzl
```diff
@@ -745,238 +745,239 @@ load("@@_main~npm~npm__pify__4.0.1__links//:defs.bzl", store_741 = "npm_imported
 load("@@_main~npm~npm__pkg-dir__4.2.0__links//:defs.bzl", store_742 = "npm_imported_package_store")
 load("@@_main~npm~npm__pkg-types__1.0.2__links//:defs.bzl", store_743 = "npm_imported_package_store")
 load("@@_main~npm~npm__plotly.js__2.25.2__links//:defs.bzl", link_744 = "npm_link_imported_package_store", store_744 = "npm_imported_package_store")
-load("@@_main~npm~npm__point-in-polygon__1.1.0__links//:defs.bzl", store_745 = "npm_imported_package_store")
```
jbedard, line 748: I've been very tempted to try to reduce how much these change just from adding a single package :/

.aspect/rules/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU=
gregmagolan: Not sure

npm/private/npm_package_store.bzl
gregmagolan: There is an && in the command which makes args confusing since there are two commands here

jbedard: Is there any risk in manually concatenating args? Is it worth fixing a rare edge case for that?

Would it be better patching the bad package instead? etc...

jbedard: I'd also be curious if this can be fixed with a tar command instead?

gregmagolan: No way that I know of. We ran into the same issue in npm_import.bzl after extracting packages in the repo rule.

gregmagolan: I don't think putting args in to the `command` is an issue. Actually, reading the docstring on arguments for run_shell, you're still required to set `$1` `$2` in the command to reference them either way: https://bazel.build/rules/lib/builtins/actions#run_shell

gregmagolan: Yea. That works nicely.

npm/private/npm_package_store.bzl
```diff
@@ -205,25 +205,23 @@ def _npm_package_store_impl(ctx):
...
+                    command = "$1 --extract --no-same-owner --no-same-permissions --strip-components 1 --file $2 --directory $3 && chmod -R a+X $3",
+                    arguments = [
+                        bsdtar.tarinfo.binary.path,
```
jbedard, line 221: I think we can drop the `.path` and bazel will expand it when the action actually runs? See https://bazel.build/rules/performance#ctx-actions-args

npm/private/npm_package_store.bzl
jbedard: Yes that's much better imo.

`ctx.action.args()` should be used whenever possible though. In this case when `command` is just a string (no concatenation, no depset `.to_list` etc) it might be fine. See https://bazel.build/rules/performance#ctx-actions-args

jbedard:
I'm still curious how frequent this is and if it's better to just patch the bad package instead? I don't think we should be writing workarounds for other peoples bugs within rules_js unless it is very common (like circular deps :/).

gregmagolan:
> I'm still curious how frequent this is and if it's better to just patch the bad package instead? I don't think we should be writing workarounds for other peoples bugs within rules_js unless it is very common (like circular deps :/).

I know of two packages so rare but impossible to know how many unless you check all npm packages on the npm registry. The failure mode is so bad that user's would have no idea that the problem is a missing `x` bit on directories within the npm package. IMO the defence-in-depth here with an inexpensive `chmod` is the right solution. Circular deps as well doesn't feel like a bug in npm packages but rather a property of the registry which allows circular deps. Unless Bazel is in the loop, circular deps don't matter to the package managers.

npm/private/npm_package_store.bzl
gregmagolan: Will try it out

jbedard:
Maybe circular deps is a bad example then, I just mean something very command that we can't ignore. Where this seems so rare I wonder if it's better to just patch or open PRs for the bad packages 🤷 

However with the use of `$1 ... $2` and `args` I'm not so worried anymore 👍

jbedard (approved):

gregmagolan:
Even if you fixed the package at head you won't be able to fix bad tarballs in the registry for versions already published 🤷‍♂️

.aspect/rules/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU=
jbedard: I'm going to try this, once your pr mergess: https://github.com/aspect-build/rules_js/pull/1796

jbedard:
```
Error in add: Cannot add directories to Args#add since they may expand to multiple values. Either use Args#add_all (if you want expansion) or args.add(directory.path) (if you do not).
```

I guess we can't do that. Up to you if you want to try doing `.path` only for directories vs everything 🤷

gregmagolan:
> I guess we can't do that. Up to you if you want to try doing `.path` only for directories vs everything 🤷

Lame. It should just expand to the path of the directory. I'll just special case that one.

.aspect/rules/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU=
gregmagolan: It looks like it is sorted now

PR #1796, opened 2024-06-09
refactor: sort lock translation action cache content

jbedard:
`dict.items` has no guaranteed order, even if it did I assume the items are added in a non-guaranteed order.

This sorts the elements to ensure the order listed in the action cache is consistent which will minimize diffs when updated.

---

### Test plan


.aspect/rules/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU=
```diff
@@ -2,26 +2,26 @@
 # Input hashes for repository rule npm_translate_lock(name = "npm", pnpm_lock = "@@//:pnpm-lock.yaml").
 # This file should be checked into version control along with the pnpm-lock.yaml file.
 .npmrc=-2065072158
-pnpm-lock.yaml=-1309835144
```
jbedard, line 5: FWIW I kind of faked these. Ran the updates, reverted everything other then the sorting and `@generated` header being added to some. Things reverted:
* updates to the lockfile + lockfile hash
* changes to the `@@//:pnpm-lock.yaml` label since the tests were run with bzlmod

gregmagolan (approved):

PR #1797, opened 2024-06-10
DRAFT: Optimize npm_package_store_impl

DavidZbarsky-at:
Before:
image

After:
image

The way the symlink prefix is constructed is a little hacky, open to better suggestions

### Test plan

### DIFF
--- a/npm/private/npm_package_store.bzl
+++ b/npm/private/npm_package_store.bzl
@@ -151,6 +151,33 @@ If set, takes precendance over the package version in the NpmPackageInfo src.
     ),
 }
 
+def _fast_symlink(ctx, symlink_path, target_path, prefix):
+    symlink = ctx.actions.declare_symlink(symlink_path)
+    symlink_path = symlink.path
+
+    total = len(symlink_path)
+    for i in range(total):
+        if symlink_path[i] != target_path[i]:
+            for j in range(i, 0, -1):
+                if symlink_path[j] == "/":
+                    break
+            break
+    new_target_path = prefix + target_path[j:]
+
+    ctx.actions.symlink(
+        output = symlink,
+        target_path = new_target_path,
+    )
+    return symlink
+
+def _fast_join(path, *others):
+    result = path
+
+    for p in others:
+        result += "/" + p
+
+    return result
+
 def _npm_package_store_impl(ctx):
     if ctx.attr.src:
         if NpmPackageInfo in ctx.attr.src:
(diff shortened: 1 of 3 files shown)


npm/private/utils.bzl
```diff
@@ -72,13 +71,12 @@ def _package_store_name(pnpm_name, pnpm_version):
...
-    if not bazel_lib_utils.is_bazel_6_or_greater():
-        # ctx.actions.declare_symlink was added in Bazel 6
-        fail("A minimum version of Bazel 6 required to use rules_js")
     symlink = ctx.actions.declare_symlink(symlink_path)
+    new_target_path = relative_file(target_path, symlink.path)
+
     ctx.actions.symlink(
         output = symlink,
-        target_path = relative_file(target_path, symlink.path),
+        target_path = new_target_path,
```
jbedard, line 79: There's no actual change here?

npm/private/npm_package_store.bzl
```diff
@@ -151,6 +151,33 @@ If set, takes precendance over the package version in the NpmPackageInfo src.
...
+        output = symlink,
+        target_path = new_target_path,
+    )
+    return symlink
+
+def _fast_join(path, *others):
+    result = path
+
+    for p in others:
+        result += "/" + p
```
jbedard, line 177: I think this would be slower then `path.join`? String concat is generally not recommended because it has to allocate a new string every time where `.format()` or `path.join` etc will do it optimally...

npm/private/npm_translate_lock.bzl
```diff
@@ -505,6 +506,10 @@ def npm_translate_lock(
         **kwargs: Internal use only
     """
 
+    if not bazel_lib_utils.is_bazel_6_or_greater():
```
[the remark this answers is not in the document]
dzbarsky: sure, https://github.com/aspect-build/rules_js/pull/1800

npm/private/npm_package_store.bzl
dzbarsky: We were previously calling `paths.join`; this code is a stripped-down [version of it](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/paths.bzl#L70-L98). The key bit is to remove the call to `_is_absolute` since that is pretty slow.

I'm not sure what `path.join` is, but I can give `str.format` a shot, that's a good idea

DavidZbarsky-at: I pulled this out into https://github.com/aspect-build/rules_js/pull/1801

npm/private/utils.bzl
jbedard: Can we revert this and rebase the PR to minimize the diff here?

DavidZbarsky-at: yes, I'll do that once we decide if we're landing the paths -> str.format PR

npm/private/npm_translate_lock.bzl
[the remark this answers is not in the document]
jbedard: Rebase needed?

DavidZbarsky-at:
I think this is too hacky

PR #1798, opened 2024-06-10
perf: share single empty depset when using js_info factory

jbedard:
Avoid constructing empty depsets in this common factory method case.

---

### Test plan

### DIFF
--- a/js/private/js_info.bzl
+++ b/js/private/js_info.bzl
@@ -13,6 +13,8 @@ JsInfo = provider(
     },
 )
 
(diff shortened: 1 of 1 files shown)


DavidZbarsky-at:
Here's the PR I remember, this may not matter anymore, will check https://github.com/bazelbuild/bazel/issues/19380

jbedard:
That seems to indicate that this PR is not required and the `depset()` factory method already does it...

jbedard:
This is done within bazel as of 6.4, not worth thinking of in rulesets.

PR #1805, opened 2024-06-13
perf: do not double process lifecycle_hooks* attributes in bzlmod

jbedard:
These are processed differently in bzlmod in extensions.bzl where the bzlmod invocation of `npm_import` is done.

See what is no longer unnecessarily duplicated in bzlmod: https://github.com/aspect-build/rules_js/blob/b6e94d3d1b09c34cbdb70e1515f1a147b3b0fa9a/npm/private/npm_translate_lock.bzl#L512-L562

---

### Test plan


gregmagolan (approved):
🦖

PR #1807, opened 2024-06-14
Optimize _is_gnu_tar check

DavidZbarsky-at:
Before:
image

After:
image

I have another version of this PR switching to the hermetic bsdtar toolchain (basically resolving it in the repo rule/module extension and passing it down in the same way), but this seems safer for now.

I also threw in some minor optimizations along the way while I was in the code.

---

### Test plan


npm/private/tar.bzl
```diff
@@ -0,0 +1,22 @@
+load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
+
+def check_is_gnu_tar(rctx):
+    """# TODO: use a hermetic tar from aspect_bazel_lib and we can drop the "if _is_gnu_tar" branch
```
jbedard, line 4: Move this comment out of the doc string?

npm/private/npm_import.bzl
```diff
@@ -450,17 +437,19 @@ def _download_and_extract_archive(rctx, package_json_only):
...
         msg = "Failed to create package directory. '{}' exited with {}: \nSTDOUT:\n{}\nSTDERR:\n{}".format(" ".join(mkdir_args), result.return_code, result.stdout, result.stderr)
         fail(msg)
 
     # npm packages are always published with one top-level directory inside the tarball, tho the name is not predictable
     # so we use tar here which takes a --strip-components N argument instead of rctx.download_and_extract
-    tar_args = ["tar", "-xf", _TARBALL_FILENAME, "--strip-components", str(1), "-C", _EXTRACT_TO_DIRNAME, "--no-same-owner", "--no-same-permissions"]
+    tar_args = ["tar", "-xf", _TARBALL_FILENAME, "--strip-components", "1", "-C", _EXTRACT_TO_DIRNAME, "--no-same-owner", "--no-same-permissions"]
 
-    if _is_gnu_tar(rctx):
+    if rctx.attr.is_gnu_tar == "True" or (rctx.attr.is_gnu_tar == "" and check_is_gnu_tar(rctx)):
```
jbedard, line 452: Maybe assign this to a var? 🤷

DavidZbarsky-at: Is this what you meant?
```
    is_gnu_tar = rctx.attr.is_gnu_tar
    if is_gnu_tar == "True" or (is_gnu_tar == "" and check_is_gnu_tar(rctx)):
```

The if condition needs to stay like this to short-circuit the evaluation

jbedard: Shouldn't it also be short-circuited when assigned to a var?

DavidZbarsky-at: I split out a var to make it a bit shorter. Splitting another one will just make it longer again I think? Check the current code

jbedard: I guess I'm just being picky/opinionated, but imo it just reads nicer when "compute `is_gnu_tar`" and `if is_gnu_tar` are two separate statements 🤷 

```
is_gnu_tar = ...
if is_gnu_tar:
  ...
```

DavidZbarsky-at: ok sure, i don't feel that strongly

jbedard (approved):

npm/extensions.bzl
```diff
@@ -133,6 +134,8 @@ WARNING: Cannot determine home directory in order to load home `.npmrc` file in
         npm_auth = npm_auth,
     )
 
+    is_gnu_tar = str(check_is_gnu_tar(module_ctx))
```
jbedard, line 137: Can we do the `str()` in the `npm_import` macro instead of here? So the macro is `True|False|None` which converts it to the underling `"True"|"False"|""`...?

DavidZbarsky-at: that might be slightly less efficient, but sure

jbedard: Just because it does the conversion essentially at repo-rule runtime (N times) instead of repo-rule generation time (~1 time)?

DavidZbarsky-at: yep. probably doesn't matter

jbedard: It's just so much prettier 😅

PR #1808, opened 2024-06-15
WIP Avoid double-parsing lockfile in bzlmod

DavidZbarsky-at:
I think this takes a slightly different approach from https://github.com/jbedard/rules_js/commit/2ba13116bd33ab0ea598376003fbbb9ced2caf36, so we can pass back the state and reuse it. I'm stuck on the following error, I can't figure out how this repo is created:
`ERROR: module extension "npm" from "//npm:extensions.bzl" does not generate repository "npm", yet it is imported as "npm"`

Also, this is skipping some of the coercion logic in `npm_translate_lock` macro because I was a bit lazy; I think factoring that into helpers should work or maybe duplicating it isn't the end of the world.

Does this seem sane?


jbedard:
Closed in favour of https://github.com/aspect-build/rules_js/pull/2480

PR #1811, opened 2024-06-20
refactor: change npm_import(is_gnu_tar) to system_tar

jbedard:
### Test plan


alexeagle (commented):
nit: I think `check_*` implies no return value and throws if violated. To make the type more intuitive I think it ought to be `is_system_tar` (or `uses_system_tar`)

jbedard:
Good point, another thing we should have done originally.

I've changed it to `is_system_tar`. I'm not sure if that's the greatest name but at least it implies it returns something related to "system tar".

npm/private/tar.bzl
```diff
@@ -3,7 +3,7 @@
 load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
 
 # TODO: use a hermetic tar from aspect_bazel_lib and remove this.
-def check_is_gnu_tar(rctx):
+def is_system_tar(rctx):
```
gregmagolan, line 6: `detect_system_tar` a better name for this since it returns the system tar type and not a boolean

jbedard: Yes that's better, thank you 👍

gregmagolan (approved):
🦖

PR #1818, opened 2024-06-26
fix: use .cjs extension for fs patches

### LINKED ISSUE #1756: [Bug]: error from aspect_rules_js/js/private/node-patches/register.js with type=module in package.json
### What happened?

When using a package.json file within packages with `"type":"module"`, a js_binary will error on startup:

```js
[Error [ERR_REQUIRE_ESM]: require() of ES Module /execroot/web/bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/tools/tw-k8s-deployment-pipeline/bin-packaged__js_binary.sh.runfiles/aspect_rules_js/js/private/node-patches/register.js not supported.
register.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename register.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /execroot/web/bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/tools/tw-k8s-deployment-pipeline/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

Fix is probably to use the .cjs extension for register.js

### Version

Output of `bazel --version`:

### How to reproduce

_No response_

### Any other information?

_No response_

jbedard:
Fix #1756

---

### Test plan


thesayyn (approved):
LGTM,

e2e/npm_link_package-esm/MODULE.bazel
```diff
@@ -0,0 +1,6 @@
+###############################################################################
```
thesayyn, line 1: Do we need this file?

jbedard: Opps, I guess that was me running tests with bzlmod and this e2e test isn't enabled with bzlmod? 🤷 

I'll drop the file before merging, can add it another day in its own PR if needed.

PR #1824, opened 2024-06-27
fix(pnpm): npm: aliases to packages with peer deps

### LINKED ISSUE #1814: [Bug]: 2.x regression with aliased package names
### What happened?

with a `package.json` containing something like this:

```
"artillery-engine-des": "npm:@des-e2e/artillery-engine-des@12.1.0",
```

We get an error from `ts_project` which looks like the linker created a file where it expected a folder?

```
ERROR: /services/docengine-e2e-service/BUILD.bazel:127:11: 
in deps attribute of ts_project rule //services/docengine-e2e-service:tsc: 
source file '//services/docengine-e2e-service:node_modules/artillery-engine-des' is misplaced here (expected no files). 
Since this rule was created by the macro 'ts_project', the error might have been caused by the macro implementation 
```

### Version

rules_js 2.0 (rc5)

### How to reproduce

_No response_

### Any other information?

_No response_

jbedard:
Fix #1814

---

### Test plan

- New test cases added


thesayyn (approved):
This looks okay to me. Added test cases makes sense.

PR #1825, opened 2024-06-27
perf: prioritize package version type checks

jbedard:
The rare edge case should be checked last.

---

### Test plan


jbedard:
... or not

PR #1830, opened 2024-06-28
chore: fix rules_lint in non-bzlmod

jbedard:
### Test plan


jbedard:
Reverting instead for now https://github.com/aspect-build/rules_js/pull/1831

PR #1836, opened 2024-07-02
fix: transitive js_library npm package deps

jbedard:
### Test plan

- New test cases added


e2e/pnpm_workspace_rerooted/lib/d/index.js
```diff
@@ -1,9 +1,13 @@
...
-const a2 = require('alias-2/package.json')
 module.exports = {
     id: () =>
         `${packageJson.name}@${
             packageJson.version ? packageJson.version : '0.0.0'
         }`,
-    test: () => [a2.name].join('\n'),
+    test: () =>
+        [
+            require('@aspect-test/a').version,
```
jbedard, line 9: Is this actually valid? It is depending on a package in `//:node_modules/@aspect-test/a`, not declared in the local `package.json`...

gregmagolan (approved):

e2e/pnpm_workspace_rerooted/lib/d/index.js
jbedard: Just removing this for now since I think it's "invalid but works"

PR #1838, opened 2024-07-03
feat(js_image_layer): Add 'pure' option suitable for distroless image (without bash)

SinimaWath:
In our company, we use distroless images (with installed NodeJS), so basic `js_image_layer` isn't suitable for our needs, as it creates a Bazel-like structure with nodejs, nodejs patches, and runfiles.

We have added the ability to generate layers similar to a native/pure NodeJS application, including only:
- Sources
- Node_modules

I believe this is a common scenario where it is necessary to generate a layer without Bazel-specific sources/wrappers. 

The `pure` option does the following:
- Removes NodeJS
- Removes NodeJS patches
- Removes js_binary wrappers
- Preserves the structure without the runfiles tree

---

### Test plan

- New test cases added

### DIFF
--- a/docs/js_image_layer.md
+++ b/docs/js_image_layer.md
@@ -19,7 +19,7 @@ js_image_layer(
 ## js_image_layer
 
 <pre>
-js_image_layer(<a href="#js_image_layer-name">name</a>, <a href="#js_image_layer-binary">binary</a>, <a href="#js_image_layer-compression">compression</a>, <a href="#js_image_layer-generate_empty_layers">generate_empty_layers</a>, <a href="#js_image_layer-owner">owner</a>, <a href="#js_image_layer-platform">platform</a>, <a href="#js_image_layer-root">root</a>)
+js_image_layer(<a href="#js_image_layer-name">name</a>, <a href="#js_image_layer-binary">binary</a>, <a href="#js_image_layer-compression">compression</a>, <a href="#js_image_layer-generate_empty_layers">generate_empty_layers</a>, <a href="#js_image_layer-owner">owner</a>, <a href="#js_image_layer-platform">platform</a>, <a href="#js_image_layer-pure">pure</a>, <a href="#js_image_layer-root">root</a>)
 </pre>
 
 Create container image layers from js_binary targets.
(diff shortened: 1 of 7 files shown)


SinimaWath:
@alexeagle @jbedard Could you do review pls?

jbedard:
If you don't want any of the node specific content would using pure rules_oci not be an option?

SinimaWath:
> If you don't want any of the node specific content would using pure rules_oci not be an option?

The problem is that I still want to have tree `node_modules` structure, with symlinks. And there is no tools, which can keep:
- Symlinks in tar as in js_image_layer.
- Right structure as in rules_js/pnpm.

thesayyn:
My take on this change is that "pure" is highly subjective and there's no "one-size-fits-all" for it.  Plus this change breaks many of the js_binary features that depends on these removed layers. 

js_image_layer today already splits files into several layers that you can pick from to get the outcome you want without changing js_image_layer drastically. 

Eg: create a tar archive that only uses `app_layer.tar` `node_modules_layer.tar` generated by js_image_layer and put them into the location that you want by rewriting some paths with bsdtar toolchain from bazel-lib.

SinimaWath:
> Eg: create a tar archive that only uses `app_layer.tar` `node_modules_layer.tar` generated by js_image_layer and put them into the location that you want by rewriting some paths with bsdtar toolchain from bazel-lib.

I got the idea. The main drawback which I don't want to face is performance:
1. js_image_layers - generating `app.tar` and `node_modules.tar`
2. Unziping tar which was generated
3. Rewriting(and copying ) paths and files
4. Creating new tars

I'm sure, that on large projects with 100 - 500 MB of node_modules it will take 2-3 times longer to create a image

@thesayyn 

That's why I wanted to have an in-place, "pure" image creation without any overhead from zipping/unzipping.

thesayyn:
for `node_modules.tar` you don't need to rewrite anything, you can simply create a symlink from the location of node_modules directory created by js_image_layer to the destination directory. 

As for `app.tar`, you can use  `tar` with the combination of `mtree_spec` and `mtree_mutate` to create an mtree(not a tar file therefore really fast) to your liking.

thesayyn:
Going to close this as it goes against the design principles of js_image_layer, See above for my reasoning.

Issue #1843, opened 2024-07-11
[Bug]: version in pnpm9 lockfile is incorrectly parsed


rnons:
### What happened?

Error in npm_translate_lock

```
        File "***/external/aspect_rules_js/npm/private/pnpm.bzl", line 396, column 28, in _convert_pnpm_v9_importer_dependency_map
                alias, version = version.rsplit("@", 1)
Error: too few values to unpack (got 1, want 2)
```

The lines in pnpm-lock.yaml that triggered this error:
```
      '@sentio/truffle-debugger':
        specifier: npm:@sentio/truffle-debugger@^0.0.56
        version: 0.0.56(bufferutil@4.0.8)(utf-8-validate@5.0.10)
```
I tried to print out specifier and version
```
specifier: npm:@sentio/truffle-debugger@^0.0.56 
version: 0.0.56_1408705619
```

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: bazel 7.2.1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: rules_js-2.0.0-rc6

Language(s) and/or frameworks involved:

### How to reproduce

```shell
Sorry, it's a private repo
```

jbedard:
Looks like an issue with `"pkg": "npm:pkg@1.2.3"`, where an `npm:` alias just references the package of the same name, then the resolved `version` is formatted like a non-`npm:` dep.

See https://github.com/aspect-build/rules_js/pull/1846

PR #1846, opened 2024-07-11
fix(pnpm): npm: aliases to self

jbedard:
When a specifier uses `npm:` to reference itself the resolved version is just the raw version, where normally `npm:` specifiers result in `actual-pkg@version`.

Fix https://github.com/aspect-build/rules_js/issues/1843

---

### Test plan

- New test cases added

### DIFF
--- a/e2e/pnpm_lockfiles/base/package.json
+++ b/e2e/pnpm_lockfiles/base/package.json
@@ -13,7 +13,7 @@
         "tslib": "^2.6.3",
         "typescript": "^5.4.5",
         "meaning-of-life": "1.0.0",
-        "is-odd": "3.0.1",
+        "is-odd": "npm:is-odd@3.0.1",
         "is-odd-alt-version": "npm:is-odd@^2.0.0",
         "uvu": "0.5.6",
         "@scoped/a": "workspace:*",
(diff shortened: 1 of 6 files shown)


e2e/pnpm_lockfiles/base/package.json
```diff
@@ -3,19 +3,27 @@
     "private": true,
     "dependencies": {
         "@aspect-test/a": "^5.0.2",
+        "// @aspect-test/a2": "npm: reference to another dependency, not specifying the version number",
```
gregmagolan, line 6: you can have comment deps?

jbedard: Ahhh, is it only outside `dependencies`? I think you're right 😢

gregmagolan (approved):
🌮

e2e/pnpm_lockfiles/base/package.json
jbedard: I was hoping to comment **somewhere** to say why we have `"is-odd": "npm:is-odd"`, and explaining the other test cases is also valuable. Mainly to ensure no one accidentally removes the test cases in the future...

PR #1850, opened 2024-07-18
fix: do not put sources+types into NpmPackageStoreInfo.files

### LINKED ISSUE #1847: [Bug]: Missing test for js_library "sandwich" no-type-check
### What happened?

verify we don't eager-type-check a test/devserver that deps through to a 1p package

### Version

Output of `bazel --version`:

### How to reproduce

_No response_

### Any other information?

slack context: https://aspect-build.slack.com/archives/C02KYA0UZ4K/p1720813137299539

jbedard:
The `NpmPackageStoreInfo` `files` and `transitive_files` now only contain the files/symlinks required to put the package into the store. All sources + types are only in the `JsInfo` which was already passed around throughout these targets but now must be read in addition to the `NpmPackageStoreInfo`.

Close #1847

---

### Test plan

- New test cases added, see rules_ts: https://github.com/aspect-build/rules_ts/pull/651

### DIFF
--- a/npm/private/npm_link_package_store.bzl
+++ b/npm/private/npm_link_package_store.bzl
@@ -22,7 +22,7 @@ https://github.com/npm/rfcs/blob/main/accepted/0042-isolated-mode.md.
 _ATTRS = {
     "src": attr.label(
         doc = """The npm_package_store target to link as a direct dependency.""",
-        providers = [NpmPackageStoreInfo],
+        providers = [NpmPackageStoreInfo, JsInfo],
         mandatory = True,
     ),
     "package": attr.string(
@@ -61,6 +61,7 @@ exec node "$basedir/{bin_path}" "$@"
 
 def _npm_link_package_store_impl(ctx):
     store_info = ctx.attr.src[NpmPackageStoreInfo]
+    store_js_info = ctx.attr.src[JsInfo]
 
     package_store_directory = store_info.package_store_directory
     if not package_store_directory:
@@ -91,11 +92,19 @@ def _npm_link_package_store_impl(ctx):
         )
         files.append(bin_file)
 
-    files_depset = depset(files, transitive = [store_info.files])
-
-    transitive_files_depset = depset(files, transitive = [store_info.transitive_files])
+    files_depset = depset(files, transitive = [
+        store_info.files,
+        store_js_info.npm_sources,
+        store_js_info.sources,
+    ])
+    transitive_files_depset = depset(files, transitive = [
+        store_info.transitive_files,
+        store_js_info.npm_sources,
+        store_js_info.transitive_sources,
+    ])
 
     providers = [
+        # TODO: delete? all rules_js+tests pass without...
         DefaultInfo(
             # Only provide direct files in DefaultInfo files
             files = files_depset,
@@ -105,6 +114,11 @@ def _npm_link_package_store_impl(ctx):
         ),
         js_info(
             target = ctx.label,
+            # TODO: if no other JsInfo properties are set, where do they come from?
+
+            # Add the additional files required for linking
+            # TODO: why does this need store_js_info.sources for rules_ts tests!?
+            # ... shouldn't sources=store_js_info.sources be used instead?
             npm_sources = transitive_files_depset,
             # only propagate non-dev npm dependencies to use as direct dependencies when linking downstream npm_package targets with npm_link_package
             npm_package_store_infos = depset([store_info]) if not store_info.dev else depset(),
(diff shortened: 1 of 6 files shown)


npm/private/npm_package_store.bzl
```diff
@@ -360,30 +375,30 @@ deps of npm_package_store must be in the same package.""" % (ctx.label.package,
...
-
-    files_depset = depset(files)
+    npm_sources = depset(files, transitive = [jsinfo.npm_sources for jsinfo in js_infos])
 
     providers = [
         js_info(
             target = ctx.label,
-            npm_sources = transitive_files_depset,
-        ),
-        DefaultInfo(
```
jbedard, line 375: Why was `DefaultInfo` here? Or maybe the question is why are tests still passing without it?

npm/private/npm_link_package_store.bzl
```diff
@@ -105,6 +111,11 @@ def _npm_link_package_store_impl(ctx):
         ),
         js_info(
             target = ctx.label,
+            # TODO: if no other JsInfo properties are set, where do they come from?
```
jbedard, line 114: All questions I came across while debugging. Can just drop the comments if we want to merge and answer questions/cleanup later.

jbedard: Note that the types from `ctx.attr.src[JsInfo]` are no longer passed through *at all* in this rule.

Is there a missing test for this? Or why would types not have to go through here?

npm/private/npm_link_package_store.bzl
```diff
@@ -91,9 +92,16 @@ def _npm_link_package_store_impl(ctx):
         )
         files.append(bin_file)
 
-    files_depset = depset(files, transitive = [store_info.files])
-
-    transitive_files_depset = depset(files, transitive = [store_info.transitive_files])
+    files_depset = depset(files, transitive = [
```
jbedard, line 95: This should be the exact same as previously but without `store_js_info.types`.

npm/private/npm_package_store.bzl
gregmagolan: I don't think it is load bearing in any test cases and likely not in any use cases in the wild so we can remote it... probably just added so that if you `bazel build` the target you get some default output and if you depend on it directly in some binary rule that doesn't understand `js_info` then things still work. Safe to remove IMO.

npm/private/npm_link_package_store.bzl
gregmagolan: Hmm. Guessing we're missing coverage here for a ts_project downstream of a linked JsInfo that requires the types to typecheck.

The changes in npm_package_store.bzl look correct. I think the trick is to add the missing test coverage for types required && populate a full js_info here.

npm/private/npm_link_package_store.bzl
```diff
@@ -91,11 +92,19 @@ def _npm_link_package_store_impl(ctx):
...
+        store_js_info.sources,
+    ])
+    transitive_files_depset = depset(files, transitive = [
+        store_info.transitive_files,
+        store_js_info.npm_sources,
+        store_js_info.transitive_sources,
+    ])
 
     providers = [
+        # TODO: delete? all rules_js+tests pass without...
```
jbedard, line 107: Should we also delete this `DefaultInfo`? Then we can drop `files_depset` 🤷

npm/private/npm_link_package_store.bzl
```diff
@@ -91,11 +92,19 @@ def _npm_link_package_store_impl(ctx):
...
-    files_depset = depset(files, transitive = [store_info.files])
-
-    transitive_files_depset = depset(files, transitive = [store_info.transitive_files])
+    files_depset = depset(files, transitive = [
+        store_info.files,
+        store_js_info.npm_sources,
+        store_js_info.sources,
+    ])
+    transitive_files_depset = depset(files, transitive = [
+        store_info.transitive_files,
```
jbedard, line 101: If I delete this line only `//examples/js_binary:test4-b` fails 🤷

npm/private/npm_link_package_store.bzl
jbedard: It data in:
```
        store_info.files,
        store_js_info.npm_sources,
        store_js_info.sources,
```
Only contain references to the store symlinks such as:
```
<generated file node_modules/.aspect_rules_js/gl-text@1.3.1/node_modules/font-measure>
```

So the package contents (sources, types, random files etc) are no where in here. Those get linked/copied elsewhere?

gregmagolan (approved):

PR #1855, opened 2024-07-20
fix(npm): invalid export of package/resolved.json

jbedard:
The generated `@npm//v61/BUILD.bazel` had:
```
export_files(["v61/is-odd/resolved.json"])
```

Which should not have the `v61/` because this is the `v61/BUILD`. The majority of use cases the lockfile is in the root and we don't see this problem.

---

### Test plan

- New test cases added

### DIFF
--- a/e2e/pnpm_lockfiles/lockfile-test.bzl
+++ b/e2e/pnpm_lockfiles/lockfile-test.bzl
@@ -19,6 +19,9 @@ BZL_FILES = {
     # global
     "defs.bzl": "@REPO_NAME//:defs.bzl",
 
+    # resolved.json reference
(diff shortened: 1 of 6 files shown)


e2e/pnpm_lockfiles/lockfile-test.bzl
```diff
@@ -19,6 +19,9 @@ BZL_FILES = {
     # global
     "defs.bzl": "@REPO_NAME//:defs.bzl",
 
+    # resolved.json reference
+    "is-odd_resolved.json": "@REPO_NAME//VERSION:is-odd/resolved.json",
```
jbedard, line 23: This just asserts `resolved.json` is generated in the expected spot, previously it was generated in `:VERSION/is-odd/resolved.json`

JesseTatasciore (approved):

PR #1857, opened 2024-07-20
fix(npm): expost package_json.bzl and resolved.json for each package alias

jbedard:
Different aliases of a package, potentially being different versions, should each have their own `@npm//:pkg-alias/resolved.json` and `@npm//:pkg-alias/package_json.bzl`.

Previously each alias of a package would write to the same files of the real package name causing a last-write-win and aliases not having these two files. 

---

### Test plan

- New test cases added


e2e/pnpm_lockfiles/MODULE.bazel
```diff
@@ -79,7 +81,7 @@ npm = use_extension(
         # Deep with many peers
         "lock-%s__at_rollup_plugin-typescript__8.2.1__%s" % (
             version,
-            "ommloj5qql5ba6x5wuiluawhoi" if version == "v54" else "1662662592",
+            "ommloj5qql5ba6x5wuiluawhoi" if version == "v54" else "1662662592" if version == "v90" else "1813138439",
```
jbedard, line 84: pnpm9 did not change the `@rollup/plugin-typescript` peer `rollup` version, pnpm <9 did resolve the peer differently 🤷

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -71,7 +71,7 @@ _FP_DIRECT_TARGET_TMPL = \
 
 _BZL_LIBRARY_TMPL = \
```
jbedard, line 72: I'd recommend [ignoring whitespace](https://github.com/aspect-build/rules_js/pull/1857/files?w=1) in the diff for this, it's mainly just putting things into a "for each alias" loop (which will normally have 1 alias and make no difference).

gregmagolan (approved):
🌮

PR #1858, opened 2024-07-20
refactor(npm): assert generated files are never overwritten

jbedard:
This assertion would prevent things like https://github.com/aspect-build/rules_js/pull/1857

---

### Test plan


jbedard:
Doesn't cover most possible cases and too much extra data being persisted.

PR #1859, opened 2024-07-20
perf(npm): reduce array concatenation

jbedard:
### Test plan

### DIFF
--- a/npm/private/test/snapshots/bzlmod/npm_defs.bzl
+++ b/npm/private/test/snapshots/bzlmod/npm_defs.bzl
@@ -1009,7 +1009,9 @@ def npm_link_all_packages(name = "node_modules", imported_links = []):
         new_link_targets, new_scope_targets = link_fn(name)
         link_targets.extend(new_link_targets)
         for _scope, _targets in new_scope_targets.items():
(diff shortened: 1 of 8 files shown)


npm/private/test/snapshots/bzlmod/npm_defs.bzl
```diff
@@ -2121,7 +2163,6 @@ def npm_link_all_packages(name = "node_modules", imported_links = []):
             link_targets.append("//{}:{}/hello".format(bazel_package, name))
             link_503(name = "{}/handlebars-helpers/helper-date".format(name))
             link_targets.append("//{}:{}/handlebars-helpers/helper-date".format(bazel_package, name))
-            scope_targets["handlebars-helpers"] = scope_targets["handlebars-helpers"] + [link_targets[-1]] if "handlebars-helpers" in scope_targets else [link_targets[-1]]
```
jbedard, line 2124: Note an actual change here. This package was declared in package.json as `"handlebars-helpers/helper-date": "file:...` or something like that... so it's not actually a scoped package but is an alias with a `/` in it.

I think the change here is correct to not treat it as a scoped package?

thesayyn (approved):
Great change!

PR #1871, opened 2024-07-26
fix(npm): npm_link_package() in subdirectory to js_library() package

jbedard:
This aligns the `e2e/npm_link_package[-esm]` tests, adds a use case within those to test linking `js_library` packages.

A 3rd e2e test is also added which is the exact same but with the lockfile in a `root/` subdir which is where the bug occurs.

---

### Test plan

- New test cases added


e2e/npm_link_package-rerooted/root/BUILD.bazel
```diff
@@ -0,0 +1,54 @@
...
+)
+
+aspect_test_a_bin.bin_a_test(
+    name = "aspect_test_a_bin_test",
+)
+
+output_files(
+    name = "e2e/lib-package_store_entry",
+    paths = [
+        "root/node_modules/.aspect_rules_js/@e2e+lib@0.0.0/node_modules/@e2e/lib",
```
jbedard, line 35: This is the path that had the extra `root/`.

npm/private/npm_package_store.bzl
```diff
@@ -292,10 +292,12 @@ deps of npm_package_store must be in the same package.""" % (ctx.label.package,
...
 
         # Symlink to the directory of the target that created this JsInfo
-        if ctx.label.workspace_name:
+        if ctx.label.workspace_name and ctx.label.package:
             symlink_path = "external/{}/{}/{}".format(ctx.label.workspace_name, ctx.label.package, package_store_directory_path)
+        elif ctx.label.workspace_name:
+            symlink_path = "external/{}/{}".format(ctx.label.workspace_name, package_store_directory_path)
         else:
-            symlink_path = "{}/{}".format(ctx.label.package or ".", package_store_directory_path)
+            symlink_path = package_store_directory_path
```
jbedard, line 300: This is the fix. This symlink is already relative to `ctx.label.package` and does not need the prefix when it is a relative prefix instead of absolute into `external`, if I'm understanding that correctly...? 🤔

gregmagolan: Sounds right

gregmagolan (approved):

PR #1883, opened 2024-08-09
perf: avoid large array concatenation

jbedard:
Avoid doing the same `+` operations multiple times and use `+=` instead of `.extends` to avoid the starlark method invocation.

---

### Test plan


DavidZbarsky-at:
I couldn't measure a difference on our repo, it's too noisy. I'm sure this doesn't hurt though

jbedard:
I think I'll just stick with the diff that we are more certain about for now. We can investigate the method vs operator another time.

gregmagolan (approved):

PR #1885, opened 2024-08-12
fix(npm): use labels for npm_translate_lock(replace_packages)

jbedard:
This way real labels are passed through to the underlying rule which should handle bzlmod repository mappings.

---

### Test plan


jbedard:
This attempt to avoid breaking changes will not work. The wksp macro can change without breaking changes by making the `npm_translate_lock` macro switch to `label:pkgs` but that is not possible with bzlmod.

See https://github.com/aspect-build/rules_js/pull/1886 for alternate bzlmod API.

PR #1886, opened 2024-08-12
fix(npm): use labels for npm_translate_lock(replace_packages)

jbedard:
Alternative to #1885 for bzlmod.

---

### Test plan

- New test cases added


npm/extensions.bzl
```diff
@@ -26,13 +26,22 @@ def _npm_extension_impl(module_ctx):
         fail("A minimum version of Bazel 6 required to use rules_js")
 
     for mod in module_ctx.modules:
+        replace_packages = {}
```
jbedard, line 29: Should this be computed outside the `module_ctx.modules` loop?

MODULE.bazel
```diff
@@ -206,13 +206,24 @@ npm.npm_translate_lock(
...
-        "chalk@5.0.1": "@chalk_501//:pkg",
-    },
     update_pnpm_lock = True,
     verify_node_modules_ignored = "//:.bazelignore",
     verify_patches = "//examples/npm_deps/patches:patches",
 )
+
+http_archive = use_repo_rule("@bazel_tools//:http.bzl", "http_archive")
+
+http_archive(
```
alexeagle, line 216: nit: should be dev only so rules_js users don't have chalk 5.0.1

npm/extensions.bzl
alexeagle: probably, since module extensions have a global view over the build, in *theory* you could have a translate lock from module A and a replacement coming from module B.

alexeagle (approved):

jbedard:
Replaced with https://github.com/aspect-build/rules_js/pull/2289

PR #1895, opened 2024-08-17
test: add @bazel/runfiles test

jbedard:
A copy of the old rules_nodejs runfiles tests that are broken in that repo with additional cross-module bzlmod tests.

---

### Test plan

- New test cases added


gregmagolan (changes requested):
Need to add the new e2e/runfiles to the list of e2e test to run on CI. Two different places for GHA and for Workflows.

gregmagolan (approved):

PR #1901, opened 2024-08-20
fix: traverse extracted package using the find utility

plobsing:
Recursive `chmod` would work fine, but we want to skip the top-level directory. Using a shell glob to enumerate the first-level children, as was done with https://github.com/aspect-build/rules_js/pull/1894, runs into difficulties when importing large packages: we can run afoul of `ARG_MAX`. This is observed, for instance, when extracting `@material-ui/icons`:

```
Extracting npm package @material-ui/icons@4.11.3_1407018657 failed: (Exit 126): bash failed: error executing NpmPackageExtract command (from target //web/web-ui:.aspect_rules_js/node_modules/@material-ui+icons@4.11.3_1407018657/pkg) /bin/bash -c '$1 --extract --no-same-owner --no-same-permissions --strip-components 1 --file $2 --directory $3 && chmod -R a+X $3/*' '' external/aspect_bazel_lib~~toolchains~bsd_tar_darwin_arm64/tar ... (remaining 2 arguments skipped)

: /bin/chmod: Argument list too long
```

The number of characters in the enumeration of all top-level entries in the package's archive is 492 KiB:

```
$ tar tf aspect_rules_js\~\~npm\~webui_npm__at_material-ui_icons__4.11.3_1407018657/package.tgz | wc -c
  504139
```

This is a sizeable fraction of the `ARG_MAX` (1024 KiB on MacOS):

```
$ sysctl kern.argmax
kern.argmax: 1048576
```

While the relative filenames are still under the limit, we go over once we prefix each filename with `package_store_directory`.

---

### Test plan

  - Confirmed the failure that motivated https://github.com/aspect-build/rules_js/pull/1894 is still fixed
    in a Buildkite-based CI that was observed to be affected by the issue when running an unpatched version
    of `aspect_rules_js` v2.0.0.

### DIFF
--- a/npm/private/npm_package_store.bzl
+++ b/npm/private/npm_package_store.bzl
@@ -222,7 +222,7 @@ def _npm_package_store_impl(ctx):
                 # tho the name is not predictable we can use the --strip-components 1 argument with
                 # tar to strip one directory level. Some packages have directory permissions missing
                 # executable which make the directories not listable (pngjs@5.0.0 for example). Run
-                # `chmod -R a+X` to fix up these packages (https://stackoverflow.com/a/14634721).
+                # `chmod a+X` to fix up these packages (https://stackoverflow.com/a/14634721).
                 # See https://github.com/aspect-build/rules_js/issues/1637 for more info.
                 bsdtar = ctx.toolchains["@aspect_bazel_lib//lib:tar_toolchain_type"]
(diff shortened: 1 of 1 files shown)


plobsing:
Withdrawing proposal. `find` is not universally available. https://github.com/aspect-build/rules_js/pull/1904 is a better proposal.

PR #1902, opened 2024-08-21
refactor(npm): improve pnpm.onlyBuiltDependencies error message

jbedard:
### Test plan

### DIFF
--- a/npm/private/npm_translate_lock_helpers.bzl
+++ b/npm/private/npm_translate_lock_helpers.bzl
@@ -586,6 +586,9 @@ def _verify_lifecycle_hooks_specified(_, state):
         fail("""\
 ERROR: pnpm.onlyBuiltDependencies required in pnpm workspace root package.json when using pnpm v9 or later
 
+The root package.json must be alongside the pnpm-lock.yaml file and contain a pnpm.onlyBuiltDependencies and
+will be automatically detected even if not explicitly added to data attribute.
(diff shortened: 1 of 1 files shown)


npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -586,6 +586,8 @@ def _verify_lifecycle_hooks_specified(_, state):
         fail("""\
 ERROR: pnpm.onlyBuiltDependencies required in pnpm workspace root package.json when using pnpm v9 or later
 
+The root package.json must be alongside the pnpm-lock.yaml file and contain a pnpm.onlyBuiltDependencies.
```
gregmagolan, line 589: Also mention that the rule will automatically read the root `package.json` file next to the `pnpm-lock.yaml` file without it having to be explicitly set as an input, tho you can add it to `data` to be explicit.

gregmagolan: Oh sweet. https://pnpm.io/package_json#pnpmonlybuiltdependencies is already linked. So the only thing I'd add then is that the rule discovered the `package.json` next to the `pnpm-lock.yaml` file automatically even if not added to `data` explicitly. This is what the user tripped up on when trying to add it as an input which is more atypical of bazel rules.

gregmagolan (approved):

PR #1904, opened 2024-08-21
fix: remove problematic chmod extract command in npm_package_store in favor of using a custom postinstall

gregmagolan:
Addresses https://github.com/aspect-build/rules_js/issues/1637 (the original issue that brought in `chmod -R a+X *` as the fix). See Bazel Slack thread https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1723821643734399 for more context.

- Breaking change (forces users to change their own code or config): for some users yes but unavoidable as this is fixing a blocking bug

Backing out of the blanket chmod after npm package tarball extract since it can lead to exceeding MAX_ARGS on the command with the glob expansion in npm packages with a very large number of files.

Instead, users should opt-in packages such as `pngjs@5.0.0` that have malformed tarball content file permissions and require fixing with `chmod` with `custom_postinstalls`. For example,

```
npm_translate_lock(
    name = "npm",
    custom_postinstalls = {
        "pngjs": "chmod -R a+X *",
    },
    ...
```

### Test plan

CI

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -122,6 +122,7 @@ npm.npm_translate_lock(
     custom_postinstalls = {
         "@aspect-test/c": "echo moo > cow.txt",
         "@aspect-test/c@2.0.2": "echo mooo >> cow.txt",
+        "pngjs": "chmod -R a+X *",  # fixes malformed tarball content file permissions in this package after extraction (see https://github.com/aspect-build/rules_js/issues/1637 for more details)
     },
     data = [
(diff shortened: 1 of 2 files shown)


jbedard (approved):
Add an example to some docs somewhere?

gregmagolan:
> Add an example to some docs somewhere?

I thought about adding something to the FAQ but I think the issue with the error message itself is more discoverable anyway. I commented on the bottom of the issue. https://github.com/aspect-build/rules_js/issues/1637

PR #1908, opened 2024-08-26
chore: use aspect workflows RBE

kormide:


.aspect/workflows/config.yaml
```diff
@@ -5,11 +5,20 @@ workspaces:
         label: rules_js
         tasks:
             - test:
+                  bazel:
+                        flags:
+                              - --config=aspect_rbe
+            - configure:
+                  bazel:
+                        flags: [] # TODO: Aspect CLI does not support --config for configure cmd
```
gregmagolan, line 13: no longer need to strip the flags since you're now applying specifically to test tasks?

gregmagolan (approved):

PR #1909, opened 2024-08-27
chore: now testing on Aspect Workflows RBE

gregmagolan:


kormide:
Can you also remove the existing `common:rbe` configs for buildbuddy and rename `aspect_rbe` to just `rbe`?

jbedard (approved):

Issue #1915, opened 2024-09-11
[Bug]: npm package patches are stricter then pnpm


jbedard:
### What happened?

Patch files could be out of date. With pnpm it seems like minor errors when applying patches are ignored while the bazel `ctx.patch` or the bash `patch` seem to be stricter and fail instead of partially applying patches.

### Version

rules_js: 2
bazel-lib: 2.7

### How to reproduce

https://github.com/samsternatretool/aspect-patch-repro

PR #1916, opened 2024-09-11
docs(pnpm): add note about pnpm vs bazel patching

jbedard:
Close https://github.com/aspect-build/rules_js/issues/1915

### Test plan

### DIFF
--- a/docs/pnpm.md
+++ b/docs/pnpm.md
@@ -194,7 +194,9 @@ git config --global merge.ours.driver true
 
 ### Patching via pnpm.patchedDependencies
 
-Patches included in [pnpm.patchedDependencies](https://pnpm.io/next/package_json#pnpmpatcheddependencies) are automatically applied. These patches must be included in the `data` attribute of `npm_translate_lock`, for example:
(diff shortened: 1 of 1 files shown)


gregmagolan (approved):

docs/pnpm.md
```diff
@@ -216,6 +218,11 @@ npm_translate_lock(
 )
 ```
 
+Patching applied by rules_js may slightly deviate from pnpm patching behavior due to the use
+of the `patch` command instead of the internal patching mechanism used by pnpm.
```
gregmagolan, line 222: Can you provide more details about which `patch` command is used? I can't recall if we're relying on a non-hermetic local patch tool on the machine here or a hermetic patch tool that comes with bazel

jbedard: It actually delegates to a bazel-lib util to do it so maybe I can just link to docs there.

gregmagolan: The implementation of that util has a if / else that ends up using different tools if I recall correctly. rules_js may be always picking one of those code paths because of the constraints of extracting to a directory

Issue #1918, opened 2024-09-15
[Bug]: No longer able to reference js_library's after https://github.com/aspect-build/rules_js/pull/1850


jrandolf:
### What happened?

Before https://github.com/aspect-build/rules_js/pull/1850, `js_library` was linked automatically and could be referenced as `:node_modules/package_name`. This no longer works.

### Version

Development (host) and target OS/architectures: N/A

Output of `bazel --version`: N/A

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: N/A

Language(s) and/or frameworks involved: JS/TS

jrandolf:
Resolved.

jrandolf:
After some debugging, it looks like `package.json`s of linked packages are missing at build times if specified in the `srcs` attribute of `js_library`.

jbedard:
Have you found a way to workaround this? Using `js_library` seems to work fine everywhere I've used it so you must have a more specific issue happening here?

jrandolf:
The workaround is the attached PR, but as you've written in the PR, it can probably be written better.

The issue arises when the javascript files in a `js_library` require the `package.json` to be interpreted. To be precise, if you don't use `.mjs` or `.cjs`, Node determines the JS file as cjs unless either (1) the es cli flag is enabled or (2) the package.json says the package is of type "module". In the latter case, `js_library` does carry the information if you add it as a `srcs` dependency, but when you try to use a dependency targeting (e.g. `:node_modules:library-name`) instead of direct targets (e.g. `//path/to:library-name`, the `package.json` gets lost for some reason.

jbedard:
I've see that used in several repositories and it seems to work fine though so there must be something unique you're doing. You'll have to reproduce it in a test to show what's going on here.

PR #1919, opened 2024-09-16
fix: include package.json in npm_sources

jrandolf:
Fixed: https://github.com/aspect-build/rules_js/issues/1918


jbedard:
Can you write a test demonstrating the issue and confirming the change fixes it?

I think we'll want to find a solution not involving `.to_list()` but lets start with a test...

jrandolf:
@jbedard Sorry, it's a little time-consuming to write a test for this one; I don't have capacity at the moment.

More specifically, the issue occurs when you try to use a linked `js_library` as a `ts_library` dependency. Because `package.json` is excluded (due to https://github.com/aspect-build/rules_js/pull/1850), the `ts_library` will not resolve imports properly because there is no package.json to look for.

npm/private/npm_package_store.bzl
```diff
@@ -402,6 +402,12 @@ deps of npm_package_store must be in the same package.""" % (ctx.label.package,
         for npm_package_store_info in npm_package_store_infos:
             transitive_files_depsets.append(npm_package_store_info.files)
 
+    for jsinfo in js_infos:
+        for source in jsinfo.sources.to_list():
```
jbedard, line 406: We can't use `to_list()` like this and will need a better solution if a fix like this is required

PR #1924, opened 2024-09-26
Match v6 lockfile behaviour in v9 around injected deps

Silic0nS0ldier:
This PR makes a few tweaks to how package references are acquired and processed such that handling of injected dependencies (unsupported, but non-fatal with v6 lockfile) works as expected with the new v9 lockfile.

Additionally;

- Adds `.git` to `.bazelignore` (editors see `.git` when inspecting test logs otherwise)
- Fixed typo `file` -> `find` in `npm/private/npm_translate_lock_generate.bzl`
- Documented params for `_new_package_info` in `npm/private/pnpm.bzl`
- Documented and tested `_convert_pnpm_v6_v9_version_peer_dep` in `npm/private/pnpm.bzl`
- `.strip` over `.(r|l)strip` in `npm/private/pnpm.bzl`
- Added tests for v6 and v9 lockfiles covering current behaviour around injected deps (unsupported feature that is non-fatal when used in v6 and now v9)
- Remove inlined JSON in favour of Starlark + `json.encode` in `npm/private/test/parse_pnpm_lock_tests.bzl`
- Added test suite target for `//npm/private/test:test_parse_pnpm_lock`

---

### Changes are visible to end-users: yes(-ish)

- Breaking change (forces users to change their own code or config):no

### Test plan

- New test cases added

### DIFF
--- a/npm/private/pnpm.bzl
+++ b/npm/private/pnpm.bzl
@@ -28,11 +28,33 @@ def _new_import_info(dependencies, dev_dependencies, optional_dependencies):
         "optional_dependencies": optional_dependencies,
     }
 
-# Metadata about a package.
-#
-# Metadata may come from different locations depending on the lockfile, this struct should
-# have data normalized across lockfiles.
 def _new_package_info(id, name, dependencies, optional_dependencies, dev, has_bin, optional, requires_build, version, friendly_version, resolution):
+    """
+    Metadata about a package.
+
+    Metadata may come from different locations depending on the lockfile, this struct should
+    have data normalized across lockfiles.
+
+    Args:
+        id: The package id, if present.
+            TODO Remove. Used for to resolve path of local packages, however `resolution` is a better source of truth.
+        name: The package name.
+        dependencies: A map of package dependencies.
+        optional_dependencies: A map of optional package dependencies.
+        dev: True if the package is a dev dependency, None otherwise.
+        has_bin: True if the package has a bin field.
+        optional: True if the package is an optional dependency.
+            Determines if package should be omitted `no_optional = True` specified.
+        requires_build: True if the package requires a build.
+            NOTE: With pnpm v9, this cannot be known ahead of time.
+        version: The resolved package version.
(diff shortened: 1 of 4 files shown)


npm/private/pnpm.bzl
```diff
@@ -485,8 +508,20 @@ def _convert_v9_packages(packages, snapshots):
         # package_data can have the resolved "version" for things like https:// deps
         friendly_version = package_data["version"] if "version" in package_data else static_key[version_index + 1:]
 
+        package_id = package_snapshot.get("id", None)
```
jbedard, line 511: Do we have any tests that show the side-effects of this change? Can we see that in a diff?

npm/private/pnpm.bzl
```diff
@@ -499,6 +534,11 @@ def _convert_v9_packages(packages, snapshots):
             resolution = package_data.get("resolution"),
         )
 
+        # Match v6 lockfile behaviour with local dependency keys
+        if "file:" in package_key:
```
jbedard, line 538: This as well. I'd like to see this change along with a test demonstrating the side-effects...

Silic0nS0ldier:
Quick update as I've not been able to return to this as soon as I would have liked, and likely won't be able to until the internal project this change was made for is complete.

As for the questions, I don't believe any of the existing tests demonstrate the issue. New tests cases in this PR do (partially, they don't show the final label but they do show the malformed data that leads generation of the bad label), however since this PR also aligns v9 with v6 that isn't obvious.

Splitting the test changes into a separate PR and merging that first (with the erroneous outputs) would it a lot more obvious exactly what this PR changes. If this is acceptable, I'll be happy to implement the split once I'm back on this.

jbedard:
Splitting the PR into multiple would be great and make this a lot easier to merge one thing at a time.

The minor cleanup/refactoring/comments can probably be done in a few isolated PRs?
I think some of the tests can be added in their own PRs?

Then the actual fixes + tests (which failed before the fix) in their own PRs.

Also checkout https://github.com/aspect-build/rules_js/pull/1989, does that overlap with your `file:` changes in here at all?

PR #1926, opened 2024-09-30
docs: Clarify that $npm_package.publish requires publishable = True

fweikert:
This fact is already mentioned further down in the docs, but this commit makes it more apparent when only scanning the summary.

The default value for `publishable` was changed in https://github.com/aspect-build/rules_js/pull/1683.


alexeagle (approved):
thank you!

PR #1927, opened 2024-10-01
Add support for `+` in canonical repo names

fmeum:
Also add a missing Bzlmod dependency that otherwise fails the example with `--noenable_workspace`.

---

### Test plan

```
bazel build //examples/... --incompatible_use_plus_in_repo_names
```


fmeum:
@alexeagle Merging and releasing this should get you Bazel 8 compatibility.

alexeagle (approved):
thank you!

alexeagle:
will be released in v2.0.2

Issue #1907, opened 2024-08-23
[FR]: support `latest` keyword when loading `pnpm` in MODULE.bazel


macourteau:
### What is the current behavior?

Currently, when loading `pnpm` in `MODULE.bazel` like [this](https://github.com/aspect-build/rules_js/blob/v2.0.1/MODULE.bazel#L27-L33), we can either not specify the version, in which case it will [use the default version](https://github.com/aspect-build/rules_js/blob/v2.0.1/npm/extensions.bzl#L294), or specify a version.

### Describe the feature

`extensions.bzl` already [defines `LATEST_PNPM_VERSION`](https://github.com/aspect-build/rules_js/blob/v2.0.1/npm/extensions.bzl#L19), and it doesn't seem like there's a way to load constants from that file in the context of a `MODULE.bazel` file.

Could we consider adding a way to use the latest version of pnpm when invoking `pnpm.pnpm` instead of needing to have a specific version, perhaps with a `latest` keyword or something like that?

PR #1932, opened 2024-10-04
test: unittest pnpm extension version resolution

gzm0:
This is in preparation for #1907: As dependency resolution gets more complex, we should test it.

Note: No updates to to bzl_library targets, because it seems `extensions.bzl` isn't in a bzl_library.

---

### Test plan

- Covered by existing test cases (e2e extension execution)
- New test cases added

### DIFF
--- a/npm/private/pnpm_extension.bzl
+++ b/npm/private/pnpm_extension.bzl
@@ -1,0 +1,59 @@
+"""pnpm extension logic (the extension itself is in npm/extensions.bzl)."""
+
+DEFAULT_PNPM_REPO_NAME = "pnpm"
+
+# copied from https://github.com/bazelbuild/bazel-skylib/blob/b459822483e05da514b539578f81eeb8a705d600/lib/versions.bzl#L60
+# to avoid taking a dependency on skylib here
+def _parse_version(version):
+    return tuple([int(n) for n in version.split(".")])
+
+def resolve_pnpm_repositories(modules):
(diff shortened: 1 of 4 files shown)


npm/private/pnpm_extension.bzl
```diff
@@ -0,0 +1,59 @@
...
+        # Note, the 'sorted(list)' function in starlark doesn't allow us to provide a custom comparator
+        if len(versions) > 1:
+            selected = versions[0]
+            selected_tuple = _parse_version(selected)
+            for idx in range(1, len(versions)):
+                if _parse_version(versions[idx]) > selected_tuple:
+                    selected = versions[idx]
+                    selected_tuple = _parse_version(selected)
+
+            result.notes.append("NOTE: repo '{}' has multiple versions {}; selected {}".format(name, versions, selected))
```
jbedard, line 53: Why collect "notes" instead of just printing here?

jbedard (approved):

npm/private/pnpm_extension.bzl
gzm0: So we can test it easily (unless there is a mechanism in unittest to do that that I don't know about).

jbedard:
@gzm0 I think this was all correctly merged in the other PR?

gzm0:
Yep, looks like it (seems it got squashed, but the changes are there).

PR #1933, opened 2024-10-04
feat(pnpm): allow `latest` as version to get the latest version

### LINKED ISSUE #1907: [FR]: support `latest` keyword when loading `pnpm` in MODULE.bazel

gzm0:
Fixes #1907.

---

### Test plan

- New test cases added

### DIFF
--- a/npm/extensions.bzl
+++ b/npm/extensions.bzl
@@ -11,14 +11,13 @@ load("//npm/private:npm_translate_lock.bzl", "npm_translate_lock_lib", "npm_tran
 load("//npm/private:npm_translate_lock_helpers.bzl", npm_translate_lock_helpers = "helpers")
 load("//npm/private:npm_translate_lock_macro_helpers.bzl", macro_helpers = "helpers")
 load("//npm/private:npm_translate_lock_state.bzl", "npm_translate_lock_state")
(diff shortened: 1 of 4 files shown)


gregmagolan (approved):
🌮

gregmagolan:
Looks good to me. Thanks @gzm0 !

@jbedard want to give it pass?

jbedard:
My only real question was in the prefactor: https://github.com/aspect-build/rules_js/pull/1932#discussion_r1788014000

But otherwise both PR LGTM 👍

jbedard (approved):

PR #1974, opened 2024-10-13
npm package untar to not use remote-cache

fa93hws:
It's pretty meaningless to use remote-cache for npm untar action. All npm packages are small.
- If there's a few files in the tarball, untar would be instant anyway
- If there's a lot of files in the barball, untar would be instant, but it takes long time to upload to remote cache

image

### DIFF
--- a/npm/private/npm_package_store.bzl
+++ b/npm/private/npm_package_store.bzl
@@ -238,6 +238,9 @@ def _npm_package_store_impl(ctx):
                         "--directory",
                         package_store_directory.path,
                     ],
+                    execution_requirements = {
+                        "no-remote-cache": "1",
+                    }
                     mnemonic = "NpmPackageExtract",
                     progress_message = "Extracting npm package {}@{}".format(package, version),
                 )


npm/private/npm_package_store.bzl
```diff
@@ -238,6 +238,9 @@ def _npm_package_store_impl(ctx):
                         "--directory",
                         package_store_directory.path,
                     ],
+                    execution_requirements = {
+                        "no-remote-cache": "1",
+                    }
```
fa93hws, line 243:
```suggestion
                    },
```

npm/private/npm_package_store.bzl
```diff
@@ -238,6 +238,13 @@ def _npm_package_store_impl(ctx):
                         "--directory",
                         package_store_directory.path,
                     ],
+                    # no need to cache tarball extraction for npm package remotely. Untar is faster than uploading/downloading
+                    # cache in most of the cases, and untar creates mutliple outputs and sometimes the number of the output can be massive
+                    # meaning it can takes much more time to upload the cache compared to the actual untar.
+                    # see https://github.com/aspect-build/rules_js/pull/1974
+                    execution_requirements = {
```
dzbarsky, line 245: I'd believe this is a speedup for local builds, but I think it's a pessimization for remote builds. Same exact reasoning as the execution requirements on bazel-lib `CopyFile` applies here. So let's make this configurable if we do land it

matthewjh: I remembered the original CopyFile/Directory execution requirements didn't play well with BWOB and was worried about the same thing here.

alexeagle: Right, the problem is:
- if inputs are local, and subsequent actions that depend on the outputs will run local, you want this action to run local and skip writing the result to remote cache
- if inputs are remote, or subsequent actions are remote, you want this action to run remote and store result in the remote cache

So there isn't a reasonable default we can use for everyone. Perhaps a better answer is to document that you can use `--modify_execution_info= NpmPackageExtract=+no-remote` if you find your builds are better tuned with that setting

alexeagle: another possibility is to add that flag to a preset, like https://github.com/bazel-contrib/bazel-lib/blob/main/.aspect/bazelrc/javascript.bazelrc - this way more devs will discover it

PR #1975, opened 2024-10-13
docs: rules_js 2.x migration guide moved to docsite

gregmagolan:


docs/migrate_2.md
alexeagle, line 1: please search for incoming links, or even better leave behind a MOVED redirect

gregmagolan: No incoming links I can find. This is just a recently added single page .md file.

jennymagolan (approved):

PR #1979, opened 2024-10-16
refactor(npm): add validation of parsed pnpm-lock data

jbedard:
This has been useful debugging and I think should always be done. For now it outputs errors to stdout, maybe we can change that to `fail` if/when we are more confident.

---

### Test plan


dzbarsky (approved):
seems fine but can you verify this doesn't show up in profiles? (We should probably make some "big lockfile" benchmarks in this repo now that we're losing access to my normal testing setup)

Issue #1045, opened 2023-05-04
[Bug]: Make variable substitution corrupts regex (or escaping rules undocumented!)


ahmedneilhussain:
### What happened?

Porting some code from `build_bazel_rules_nodejs` to use `rules_js`, I replaced a `genrule` with `js_run_binary`. Command line flags containing regexes that worked fine with `genrule` get corrupted when passed to the `args` attribute of `js_run_binary`.

These used to work fine using `genrule`:

        "--ignore-custom-fragments='/{{[{]?(.*?)[}]?}}/'",
        "--ignore-custom-fragments='/{%(.*?)%}/'",

In `js_run_binary` they get corrupted, and in fact as it stands cause the shell script runner to blow up before even invoking the binary, I think because the input gets truncated after the first `%` of the second CLI arg and you get a shell complaint about the unmatched quote.

I presume makefile variable expansion is causing the problem but I would have expected the substitution to behave the same as native genrule?

### Version

Development (host) and target OS/architectures: MacOS Ventura 13.3.1a/x86

Output of `bazel --version`:
bazel 7.0.0-pre.20230316.2

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
https://github.com/aspect-build/rules_js/archive/002397780d51735f1a98b362fca7018df31e85ac.tar.gz
(needed to use a commit because a fix for newer syntax of `pnpm-lock.yaml` files isn't/wasn't yet in a released build).

Language(s) and/or frameworks involved:

### How to reproduce

```shell
Build a simple node binary that echoes its arguments and pass those in - should be malformed.
```

PR #1987, opened 2024-10-17
fix: upgrade bazel-lib to include run_binary param parsing fix

jbedard:
This is to include https://github.com/bazel-contrib/bazel-lib/commit/73d021fb368c67be40ae204a62cfa789e14515a7 which fixed a bug in `run_binary` and therefore `js_run_binary`.

Currently the rules_js WORKSPACE deps have this fix but bzlmod does not. This change upgrades the minimum bazel-lib version to include this fix.

The updated assertions align with running node outside bazel on the cli:
```
> echo 'console.log(`${process.argv[2]} + ${process.argv[3]}`)' | node - --x='3' '4'

--x=3 + 4
```

The stdout has no quotes despite the arguments being quoted. The tests updated in this change show rules_js now does the same.

---

### Test plan

### DIFF
--- a/.aspect/bazelrc/convenience.bazelrc
+++ b/.aspect/bazelrc/convenience.bazelrc
@@ -3,7 +3,7 @@
 build --keep_going
 
 # Output test errors to stderr so users don't have to `cat` or open test failure log files when test
-# fail. This makes the log noiser in exchange for reducing the time-to-feedback on test failures for
+# fail. This makes the log noisier in exchange for reducing the time-to-feedback on test failures for
 # users.
 # Docs: https://bazel.build/docs/user-manual#test-output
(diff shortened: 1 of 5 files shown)


jbedard:
Here's a CI run at HEAD with the current failure: https://github.com/aspect-build/rules_js/actions/runs/11374410329/job/31643056254#step:7:1543

jbedard:
@ahmedneilhussain you logged the original issue (https://github.com/aspect-build/rules_js/issues/1045) the tests changed in this PR were testing. Do you know if you're already using bazel-lib >= 2.8.1 and can confirm your issue is still fixed?

dzbarsky (approved):

PR #1988, opened 2024-10-17
chore: enable bazel6 ci tests

jbedard:
This seems to have been an accident and disabled all bazel6 testing by default.

---

### Test plan


dzbarsky (approved):
oh yikes, classic yaml

PR #1999, opened 2024-10-22
fix(js_run_devserver): ensure files are synced before waiting for next build event

jbedard:
I think this removes the possibility of processing multiple build events in parallel.

---

### Test plan

### DIFF
--- a/js/private/js_run_devserver.mjs
+++ b/js/private/js_run_devserver.mjs
@@ -447,38 +447,44 @@ async function main(args, sandbox) {
             process.exit(code)
         })
 
+        // Process stdin data in order using a promise chain.
         let syncing = Promise.resolve()
         process.stdin.on('data', async (chunk) => {
+            return (syncing = syncing.then(() => processChunk(chunk)))
+        })
+
+        async function processChunk(chunk) {
             try {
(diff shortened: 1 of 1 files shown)


js/private/js_run_devserver.mjs
```diff
@@ -447,38 +447,44 @@ async function main(args, sandbox) {
...
+                    // sync to be complete before writing to stdin of the child
+                    // process
+                    await Promise.all([
                         // Remove files that were previously synced but are no longer in the updated list of files to sync
-                        deleteFiles(
-                            config.data_files,
-                            updatedDataFiles,
-                            sandbox
-                        )
+                        deleteFiles(oldFiles, updatedDataFiles, sandbox),
```
jbedard, line 476: Note that this returns a promise, which this PR has not changed, but previously we were never doing `await` anywhere. So the `deleteFiles` and `syncFiles` were running async and concurrently (and they still are), but nothing was doing `await` to wait for them.

This means the `await syncing` below was essentially not waiting for anything. Then this code could start to be executed multiple times in parallel, defeating hte purpose of `syncing`...

gregmagolan (approved):

PR #2001, opened 2024-10-22
perf(js_run_devserver): run fs delete operations in parallel

jbedard:
See individual commits. Primarily changing async fs operations to be run in parallel instead of sequentially.

---

### Test plan

### DIFF
--- a/js/private/js_run_devserver.mjs
+++ b/js/private/js_run_devserver.mjs
@@ -302,7 +302,7 @@ async function deleteFiles(previousFiles, updatedFiles, sandbox) {
 
 // Sync list of files to the sandbox
 async function syncFiles(files, sandbox, writePerm) {
-    console.error(`+ Syncing ${files.length} files && folders...`)
+    console.error(`+ Syncing ${files.length} files & folders...`)
     const startTime = perf_hooks.performance.now()
 
     const [nodeModulesFiles, otherFiles] = partitionArray(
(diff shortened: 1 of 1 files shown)


js/private/js_run_devserver.mjs
```diff
@@ -358,15 +359,18 @@ async function syncFiles(files, sandbox, writePerm) {
...
-    ).reduce((s, t) => s + t, 0)
+    syncPromises = syncPromises.concat(
+        otherNodeModulesFiles.map((file) => {
+            const src = path.join(RUNFILES_ROOT, file)
+            const dst = path.join(sandbox, file)
+            return syncRecursive(src, dst, sandbox, writePerm)
+        })
+    )
+
+    const totalSynced = (await Promise.all(syncPromises)).reduce(
```
gregmagolan, line 370: Doing all of these in parallel could cause problems since Node.js doesn't like symlinks being created to files that don't exist. I think we need to `Promise.all` three times for the three tronches.
1)
```
    // Sync non-node_modules files first since syncing 1p js_library linked node_modules symlinks
    // requires the files they point to be in place.
```
2)
```
    // Sync first-party package store files before other node_modules files since correctly syncing
    // direct 1p node_modules symlinks depends on checking if the package store synced files exist.
```
3)
```
// Finally sync all remaining node_modules files
```

jbedard: I've dropped this change for now.

gregmagolan (approved):

js/private/js_run_devserver.mjs
gregmagolan: 👍

PR #2004, opened 2024-10-29
Expanded test coverage for pnpm lockfile processing

Silic0nS0ldier:
Split from #1924

This PR primarily serves to add test cases to cover the malformed package label generation that #1924 aims to fix, such that the changes in that other PR are more obvious.

Additionally;

- Adds `.git` to `.bazelignore` (editors see `.git` when inspecting test logs otherwise)
- Fixed typo `file` -> `find` in `npm/private/npm_translate_lock_generate.bzl`
- Documented params for `_new_package_info` in `npm/private/pnpm.bzl`
- Documented and tested `_convert_pnpm_v6_v9_version_peer_dep` in `npm/private/pnpm.bzl`
- Added tests for v6 and v9 lockfiles covering current behaviour around injected deps (unsupported feature that is non-fatal when used in v6 and now v9)
- Remove inlined JSON in favour of Starlark + `json.encode` in `npm/private/test/parse_pnpm_lock_tests.bzl`
- Added test suite target for `//npm/private/test:test_parse_pnpm_lock`

---

### Changes are visible to end-users: yes(-ish)

### Test plan

NA, PR is adding new tests with little/no behaviour changes.

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -22,3 +22,4 @@ node_modules/
 npm/private/test/node_modules/
 npm/private/test/npm_package/node_modules/
 npm/private/test/npm_package_publish/node_modules
+.git
--- a/npm/private/test/parse_pnpm_lock_tests.bzl
+++ b/npm/private/test/parse_pnpm_lock_tests.bzl
@@ -46,34 +46,92 @@ expected_packages = {
     },
 }
 
+expected_imports_injected = {
+    ".": {
+        "dependencies": {},
+        "dev_dependencies": {},
+        "optional_dependencies": {},
+    },
+    "packages/a": {
+        "dependencies": {
+            "b": "file:packages/b_typescript_5.6.2",
+        },
+        "dev_dependencies": {},
+        "optional_dependencies": {},
+    },
+    "packages/b": {
+        "dependencies": {
+            "typescript": "5.6.2",
+        },
+        "dev_dependencies": {},
+        "optional_dependencies": {},
+    },
+}
+expected_packages_injected = {
+    "file:packages/b_typescript_5.6.2": {
+        "id": "file:packages/b",
+        "name": "b",
+        "dependencies": {
+            "typescript": "5.6.2",
+        },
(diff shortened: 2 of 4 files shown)


.bazelignore
```diff
@@ -23,3 +23,4 @@ node_modules/
 npm/private/test/node_modules/
 npm/private/test/npm_package/node_modules/
 npm/private/test/npm_package_publish/node_modules
+.git
```
jbedard, line 26: I don't think I've ever seen this in any bazel repo, do you have any more info on this?

npm/private/test/parse_pnpm_lock_tests.bzl
```diff
@@ -46,34 +46,92 @@ expected_packages = {
...
-        "@aspect-test/b": "5.0.0",
-        "@aspect-test/c": "1.0.0",
-        "@aspect-test/d": "2.0.0_@aspect-test+c@1.0.0"
-      },
-      "dev": false
-    }
-  }
-}
-""")
+    parsed_json = pnpm.parse_pnpm_lock_json(json.encode({
```
jbedard, line 110: Why switch to JSON? When I'm writing these tests I always generate a pnpm-lock and then paste it into here, converting to json sounds like extra work when writing tests?

Silic0nS0ldier: Happy to switch this back.

I've forgotten the exact reason I did this, my guess is the syntax highlighting just made it easier to scan when comparing inputs to outputs. Could also just be habit, in a few internal Rust projects we use a `json!` macro so the data is expressed in the "host" language, meaning it'll be processed by existing tooling (e.g. formatters).

I _do_ remember being annoyed that `repr` has no way of returning indented output.

jbedard: I'd prefer just switching it back. If you really want to debate it I'd rather debate it in a different PR so we can merge the rest of your changes sooner rather then later.

PR #2008, opened 2024-11-02
fix(npm): add missing patch copy action key

jbedard:
I have no idea if this will have any side-effects or fixes, but my ide highlighted that the args were wrong 😨 

---

### Test plan


dzbarsky (approved):
Lol

gregmagolan (approved):

Issue #1993, opened 2024-10-18
rules_js v3 breaking changes


jbedard:
* remove pnpm lockfile version 5 (maybe 6, 6.1)
* remove bazel v6 support

PR #2014, opened 2024-11-07
docs: document exclude_types_from_npm_packages as deprecated

jbedard:
With `js_library` always being recommended I think `npm_package` should be used almost exclusively for creating packages ready to be published and not for use as a runtime/type-checking dependency within the workspace.

The only remaining use case I can see for using `npm_package` instead of `js_library` would be to properly test a package in the same state it would be published. In that case this flag may still be useful, but is it worth maintaining for that use case?

---

Deprecate `--@aspect_rules_js//npm:exclude_types_from_npm_packages`.

### Test plan

### DIFF
--- a/docs/npm_package.md
+++ b/docs/npm_package.md
@@ -64,8 +64,6 @@ using the `include_sources`, `include_transitive_sources`, `include_types`, `inc
 
 The two `include*_types` options may cause type-check actions to run, which slows down your
 development round-trip.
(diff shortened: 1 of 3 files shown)


alexeagle (approved):

alexeagle:
Need a TODO(3.0) so we remember to do the removal

jbedard:
I've added the TODO where the flag is defined, it's also listed in https://github.com/aspect-build/rules_js/issues/1993

PR #2020, opened 2024-11-15
fix: properly report errors when lock file cannot be parsed

devversion:
The function returns only 4 arguments upon errors, but 5 are expected and returned when everything succeeded.

### DIFF
--- a/npm/private/pnpm.bzl
+++ b/npm/private/pnpm.bzl
@@ -531,12 +531,12 @@ def _parse_lockfile(parsed, err):
         A tuple of (importers dict, packages dict, patched_dependencies dict, error string)
     """
     if err != None or parsed == None or parsed == {}:
-        return {}, {}, {}, err
+        return {}, {}, {}, -1, err if err != None else "could not parse Yaml config"
 
     if not types.is_dict(parsed):
-        return {}, {}, {}, "lockfile should be a starlark dict"
+        return {}, {}, {}, -1, "lockfile should be a starlark dict"
     if not parsed.get("lockfileVersion", False):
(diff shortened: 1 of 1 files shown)


npm/private/pnpm.bzl
```diff
@@ -531,12 +531,12 @@ def _parse_lockfile(parsed, err):
         A tuple of (importers dict, packages dict, patched_dependencies dict, error string)
     """
     if err != None or parsed == None or parsed == {}:
-        return {}, {}, {}, err
+        return {}, {}, {}, -1, err if err != None else "could not parse Yaml config"
```
alexeagle, line 534: I'm confused, you now return a 5-tuple instead of a 4-tuple, doesn't that require changes at the callsites?

devversion: The call sites already expect 5-tuple 😄 — they never see the invalid 4-tuple though because of no errors most of the time

alexeagle: I see, do you mind correcting the `Returns` doc above in that case?

jbedard: This change looks correct, where the `-1` is the `lockfileVersion`.

However I don't think we should have the default `"could not parse Yaml config"` since I think an empty lockfile is valid? The `-1` version number is probably still ok in that case.

Can we add a test for that case that failed previously?

jbedard:
@devversion have a chance to fix the build errors? Potentially add a test if you have time...

jbedard:
This was done at some point or another, probably while refactoring or adding pnpm 10

PR #2037, opened 2024-12-13
refactor: reduce runtime string format calls, use local target paths

jbedard:
Was there any reason these were not relative? In all cases we are already checking if we are in the `native.package_name()` so I think this can all be relative.

---

### Test plan


dzbarsky (approved):
Nice!

PR #2044, opened 2024-12-18
fix: override the locale when running bsdtar actions

### LINKED ISSUE #2039: [Bug]: package tar extraction fails with unicode characters in paths
### What happened?

Packages may contain files or directories containing interesting characters, for example [the snowman](https://www.unicodepedia.com/unicode/miscellaneous-symbols/2603/snowman/) in [@fastify/send tests](https://unpkg.com/browse/@fastify/send@3.3.0/test/fixtures/snow%20%E2%98%83/)

Failure when extracting the package:
```
rules_js/BUILD.bazel:14:22: Extracting npm package @fastify/send@3.3.0 failed: (Exit 1): tar failed: error executing NpmPackageExtract command (from target //:.aspect_rules_js/node_modules/@fastify+send@3.3.0/pkg) external/aspect_bazel_lib~~toolchains~bsd_tar_darwin_arm64/tar --extract --no-same-owner --no-same-permissions --strip-components 1 --file external/_main~npm~npm__at_fastify_send__3.3.0/package.tgz ... (remaining 2 arguments skipped)

tar: Pathname can't be converted from UTF-8 to current locale.
tar: Error exit delayed from previous errors.
```

### Version

bazel 7.4
rules_js 2.1.2

### How to reproduce

```shell
https://github.com/aspect-build/rules_js/commit/580086b75c825dbc9cbcefe9ebd2f6f531c064ae
```

### Any other information?

_No response_

alexeagle:
Fixes #2039 
Workaround https://github.com/bazel-contrib/bazel-lib/issues/1018

---

### Test plan

Figma had a repro and confirmed this one-line change fixes it for them.


npm/private/npm_package_store.bzl
```diff
@@ -238,6 +238,7 @@ def _npm_package_store_impl(ctx):
                         "--directory",
                         package_store_directory.path,
                     ],
+                    env = {"LC_ALL": "en_US.UTF-8"},
```
jbedard, line 241: Does every platform in the world have `en_US`? 😅

jbedard (approved):

npm/private/npm_package_store.bzl
alexeagle: I think this can be anything ending in UTF-8 actually. I've seen `C.UTF-8` too. Lemme research what's the most correct value.

jbedard: `rules_js.UTF-8`? 😄

alexeagle: The libarchive tests definitely handle cases where `en_US.UTF-8` isn't on the machine. I'm not a big fan of trying to invent something here - and I think it's likely we're just fixing one specific case of a general problem. Users may be better off fixing the locale of their Bazel environment since this isn't the only spot that can go wrong.

jbedard: As long as you looked into it 👍

alexeagle: https://github.com/bazelbuild/rules_java/blob/767e4410850453a10ccf89aa1cededf9de05c72e/toolchains/utf8_environment.bzl has a well-reasoned answer

npm/private/npm_package_store.bzl
```diff
@@ -240,6 +242,16 @@ def _npm_package_store_impl(ctx):
                     ],
                     mnemonic = "NpmPackageExtract",
                     progress_message = "Extracting npm package {}@{}".format(package, version),
+                    # Workaround https://github.com/bazelbuild/bazel-central-registry/issues/2256
+                    # Always override the locale to give better hermeticity.
+                    # See https://github.com/bazelbuild/rules_java/blob/767e4410850453a10ccf89aa1cededf9de05c72e/toolchains/utf8_environment.bzl
+                    env = {
+                        "LC_CTYPE":
```
alexeagle, line 249: Note @jbedard I repeated the testing to verify that this environment variable also controls the bsdtar behavior. Based on some reading, `LC_ALL` is overkill, this is the surgical value to choose, and also what rules_java is using, see link above.

npm/private/npm_import.bzl
```diff
@@ -452,7 +452,7 @@ def _download_and_extract_archive(rctx, package_json_only):
 
     # npm packages are always published with one top-level directory inside the tarball, tho the name is not predictable
     # so we use tar here which takes a --strip-components N argument instead of rctx.download_and_extract
-    tar_args = ["tar", "-xf", _TARBALL_FILENAME, "--strip-components", "1", "-C", _EXTRACT_TO_DIRNAME, "--no-same-owner", "--no-same-permissions"]
+    tar_args = ["tar", "--extract", "--file", _TARBALL_FILENAME, "--strip-components", "1", "--directory", _EXTRACT_TO_DIRNAME, "--no-same-owner", "--no-same-permissions"]
```
jbedard, line 455: Is this change necessary? This might be causing the windows e2e/bzlmod failure:
```
tar.exe: Error opening archive: Can't initialize filter; unable to run program "gzip -d"
```

alexeagle: It's not necessary - I just wanted to use the long form as a matter of style. Will put that in a separate change.

jbedard: I agree with that, but it broke windows...

alexeagle: reverted. CI is still red, but so is `main`. Want me to force-merge?

jbedard: Just confirm it's only the auth issues first. With this tar args change windows was failing with a tar error

PR #2046, opened 2024-12-18
fix: cleanup js_run_devserver directory on SIGINT

### LINKED ISSUE #2033: [Bug]: Sandbox folders created by js_run_devserver are not always deleted
### What happened?

We noticed that sandbox folders created by `js_run_devserver` are not deleted and they can take quite a lot of space. If the process is interrupted by the user, then the cleanup of those folders will not performed. https://github.com/aspect-build/rules_js/blob/cd48714c7fe445a6bce6e2bcc1db36c4eefcffdf/js/private/js_run_devserver.mjs#L531-L538

### Version

Output of `bazel --version`: `7.4.1`

`WORKSPACE` or `MODULE.bazel` file: `aspect_rules_js version 2.0.2`

### How to reproduce

```shell
In my case, I was running a `webpack_devserver` target and killed the process with ctrl+c.
```

### Any other information?

_No response_

jbedard:
Fix #2033

---

Cleanup the `js_run_devserver` temp directory on SIGINT in addition to standard process exit.

### Test plan

- Manual testing: run the devserver in `e2e/js_run_devserver` and `cmd-c` to trigger a SIGINT

### DIFF
--- a/js/private/js_run_devserver.mjs
+++ b/js/private/js_run_devserver.mjs
@@ -518,6 +518,10 @@ async function main(args, sandbox) {
         // short-circuit for unit tests
         return
     let sandbox
+
+    // Callback to cleanup the sandbox if it exists, once and only once.
+    onProcessEnd(() => sandbox && removeSandbox(sandbox) && (sandbox = null))
+
     try {
         sandbox = path.join(
             await fs.promises.mkdtemp(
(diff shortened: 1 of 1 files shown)


js/private/js_run_devserver.mjs
```diff
@@ -518,6 +518,10 @@ async function main(args, sandbox) {
         // short-circuit for unit tests
         return
     let sandbox
+
+    // Callback to cleanup the sandbox if it exists, once and only once.
+    onProcessEnd(() => sandbox && removeSandbox(sandbox) && (sandbox = null))
```
alexeagle, line 523: what's the final clause `sandbox = null` doing here? Just freeing the pointer out of fear?

alexeagle (approved):

js/private/js_run_devserver.mjs
jbedard: To avoid the `fs.rmDir` multiple times. Clear the var on success, only remove if it is set.

js/private/js_run_devserver.mjs
```diff
@@ -531,15 +535,30 @@ async function main(args, sandbox) {
...
-            )
-        }
     }
 })()
+
+function removeSandbox(sandbox) {
+    try {
+        if (sandbox) {
+            // Must be synchronous when invoked from process exit handler
+            fs.rmSync(sandbox, { force: true, recursive: true })
```
jbedard, line 545: Note the `force` was also added here.

PR #2047, opened 2024-12-19
fix: honor shell env on tar extract

### LINKED ISSUE #2039: [Bug]: package tar extraction fails with unicode characters in paths

alexeagle:
Allows users to override the locale, needed for tarballs with filenames not supported by the system locale

Fixes #2039 

---

### Test plan

Similar to #2044

### DIFF
--- a/npm/private/npm_package_store.bzl
+++ b/npm/private/npm_package_store.bzl
@@ -240,6 +240,10 @@ def _npm_package_store_impl(ctx):
                     ],
                     mnemonic = "NpmPackageExtract",
                     progress_message = "Extracting npm package {}@{}".format(package, version),
+                    # Allow users to set --action_env=LC_ALL=C.UTF-8 for example,
+                    # on systems where default locale is wrong.
+                    # See https://github.com/aspect-build/rules_js/issues/2039
+                    use_default_shell_env = True,
                 )
             else:
                 copy_directory_bin_action(


jbedard:
I'm thinking this is better then #2044 because it aligns with a lot of other tools that respect `--action_env`?

alexeagle:
Based on reasoning here https://github.com/bazel-contrib/bazel-lib/pull/1019#issuecomment-2554794112

I think https://github.com/aspect-build/rules_js/pull/2044 is the better approach

PR #2048, opened 2024-12-19
Add patch_tool argument to npm_translate_lock

MichaelMitchell-at:
We encountered a bug with MacOS's implementation with `patch`, likely due to it being based on an ancient version of `patch`.
```sh
$ patch --version
patch 2.0-12u11-Apple
```

One downstream effect of `patch` being broken locally is that it leads to remote cache misses since our remote cache warming machines are using a non-broken implementation of `patch`.

This PR adds support for specifying a custom patch tool so that we can use a non-broken implementation.

### Test plan

Manual testing; please provide instructions so we can reproduce:

- Add the following to MODULE.bazel:
```python
http_file(
    name = "patch",
    executable = True,
    integrity = "sha256-0RgbIckQNYDwdh4kyrCgStgN3Dqj8XnnvIwOMJRRoks=",
    url = "https://cosmo.zip/pub/cosmos/bin/patch",
)
```
- Pass the following argument to `npm.npm_translate_lock`
```
    patch_tool = "@patch//file:downloaded",
```


npm/private/npm_import.bzl
```diff
@@ -1095,6 +1102,8 @@ def npm_import(
 
         package_visibility: Visibility of generated node_module link targets.
 
+        patch_tool: The patch tool.
```
alexeagle, line 1105: mention "will use `patch` from the PATH if not specified?"

npm/private/npm_import.bzl
```diff
@@ -503,7 +503,12 @@ def _npm_import_rule_impl(rctx):
 
     # apply patches to the extracted package before reading the package.json incase
     # the patch targets the package.json itself
-    patch(rctx, patch_args = rctx.attr.patch_args, patch_directory = _EXTRACT_TO_DIRNAME)
+    patch(
+        rctx,
+        patch_tool = rctx.path(rctx.attr.patch_tool) if rctx.attr.patch_tool else "patch",
```
alexeagle, line 508: we should consider ALWAYS using a hermetic patch implementation. I recall we looked into the one Bazel itself uses, but it wasn't exposed for repository rules to use? Maybe @gregmagolan remembers. Can be in a later PR though.

alexeagle (approved):
thanks!

alexeagle:
Hm, I don't understand why this isn't triggering our GitHub Actions CI. Is it rebased on a really old base commit?

MichaelMitchell-at:
> Hm, I don't understand why this isn't triggering our GitHub Actions CI. Is it rebased on a really old base commit?

Oh right. I had this on top of v2.1.0 which is what we're still currently using. Let me rebase.

PR #2050, opened 2024-12-20
test: add repositories.bzl snapshot tests for each lockfile version

jbedard:
This is valuable when looking at the dev vs prod dependencies, finding differences between versions and unexpected diffs when the lockfile changes.

---

### Test plan

- New test cases added


jbedard:
Adding tests in https://github.com/aspect-build/rules_js/pull/2057 for now

Issue #1879, opened 2024-08-07
[FR]: `npm_link_all_packages` should provide shorthands for "all non-devDependencies" and "all devDependencies"


jfirebaugh:
### What is the current behavior?

`npm_link_all_packages` provides you with a `:node_modules` target that represents all dependencies in package.json.

### Describe the feature

There should also be some sort of automatically generated shorthand target for ":node_modules excluding dev deps" and "only dev deps". It's common to not want to micro-manage package dependencies in both package.json and BUILD.bazel, but still have targets to which dev deps are not accessible (most non-test targets), to avoid accidental inclusion of dev libraries in production code.

See https://bazelbuild.github.io/rules_rust/crate_universe.html#all_crate_deps for one potential API approach.

sin-ack:
This would be awesome in combination with `js_image_layer`. My use-case is a TypeScript project running on Node which needs to be packaged into an image. Unfortunately, `ts_project` will forward _all_ of its dependencies in `JsInfo` including development dependencies and will cause huge amounts of bloat in the final image (because it includes stuff like the entire TypeScript compiler). It would be possible to strip the Node.js dependencies from the output JavaScript files with `js_info_files`, but then there's no way for me to add only the production dependencies (if I try to split `npm_translate_lock` into two with `prod = True` and `dev = True` it will still fail with output collisions because some packages are shared between both).

jfirebaugh:
I just noticed that [the documentation for `ts_project(deps)`](https://github.com/aspect-build/rules_ts/blob/main/docs/rules.md#ts_project-deps) implies that rules_ts wanted to make this feature unnecessary, at least for the case of doing `ts_project(..., deps = [":node_modules"])`, which is all I really want it for.

> NB: Linked npm package targets that are "dev" dependencies do not forward their underlying `npm_package_store` target(s) through `npm_package_store_deps` and will therefore not be propagated to the direct dependencies of downstream linked `npm_package` targets.

However, this is followed by:

> npm packages that come in from `npm_translate_lock` are considered "dev" dependencies if they are have `dev: true` set in the pnpm lock file. This should be all packages that are only listed as "devDependencies" in all package.json files within the pnpm workspace.

[...]

PR #2051, opened 2024-12-20
feat: add `prod` and `dev` attributes to `npm_link_targets`

jfirebaugh:
Fixes https://github.com/aspect-build/rules_js/issues/1879

---

- Searched for relevant documentation and updated as needed: TODO

### Test plan

- (TODO) New test cases added


npm/private/npm_translate_lock_generate.bzl
```diff
@@ -219,7 +222,7 @@ def npm_link_targets(name = "node_modules", package = None):
     npm_link_all_packages_bzl = [
         """\
 # buildifier: disable=function-docstring
-def npm_link_all_packages(name = "node_modules", imported_links = []):
+def npm_link_all_packages(name = "node_modules", imported_links = [], prod = False, dev = False):
```
jbedard, line 225: Any other ideas for this API?

It might be nice if we didn't have this scenario where 1/4 of the states is basically invalid and has an odd non-intuitive default. Maybe just throw when `false/false` and then the other 3 states are valid and understandable?

`dev = True|False|None` or something like that is one alternative, although the word "dev" in the API without "prod" in the API is confusing imo. Or `type = "dev"|"prod"|None` but I'd want a term better then "type".

npm/extensions.bzl
```diff
@@ -134,6 +134,9 @@ WARNING: Cannot determine home directory in order to load home `.npmrc` file in
     system_tar = detect_system_tar(module_ctx)
 
     for i in imports:
+        link_packages = {}
+        for link_package, link_names in i.link_packages.items():
+            link_packages[link_package] = [link_name["pkg"] for link_name in link_names]
         npm_import(
```
jbedard, line 140: (this is probably beyond this scope of this PR, but I'd like to start the discussion here...)

I think the `npm_import(dev)` flag is flawed and should actually be removed completely. The dev vs prod logic needs to be 100% at package link time based on the package.json dev vs non-dev dependencies, not the lockfile `dev` flag which is currently misused in rules_js (and non-existent in pnpm9).

Does that sound right to you?

npm/private/test/snapshots/bzlmod/npm_defs.bzl
```diff
@@ -2048,371 +2048,369 @@ def npm_link_all_packages(name = "node_modules", imported_links = []):
     if link:
         if bazel_package == "js/private/worker/src":
             link_2(name = "{}/abortcontroller-polyfill".format(name))
-            link_targets.append("//{}:{}/abortcontroller-polyfill".format(bazel_package, name))
+            link_targets.append("//{}:{}/abortcontroller-polyfill".format(bazel_package, name)) if (not prod and not dev) or (prod and not True) or (dev and True) else None
```
jbedard, line 2051: This generated `if ...` is ugly and I'm pondering how else we can do it. I think we can generate something simpler.

Can't we just compute a few vars at the top of the method such as `include_dev` and `include_prod`? Then this will be `if include_dev else None` or `if include_prod else None` or blank (or `if True else None` instead of blank).

jbedard:
IIUC this solves the issue where rules_js currently treats the lockfile `dev` flag incorrectly, but does not add the API suggested in #1879?

npm/private/npm_translate_lock_generate.bzl
jbedard: Actually is this API simply a copy of `npm_translate_lock`? In which case we should do exactly as you have and my suggestion can be considered in rules_js v3 or something like that...

jfirebaugh: Yeah, it's just following the `prod`/`dev` API from `npm_translate_lock`. I agree that an API that excludes invalid states by design would be better for v3.

npm/extensions.bzl
jfirebaugh: I agree that the behavior of lockfile `dev` flag makes it unsuitable for really anything that rules_js wants to do with regard to dev dependencies.

npm/private/test/snapshots/bzlmod/npm_defs.bzl
jfirebaugh: Yeah, something like that should be possible I think.

npm/private/npm_translate_lock_generate.bzl
jbedard: This current API does not allow a single BUILD to have a mix of these though, correct? If you wanted to include dev deps for some targets (such as tests) but not for others (such as libraries).

Would something such as `:node_modules[__{dev,prod}]` as the API also solve this?

jfirebaugh: It does allow a mix. You could write `npm_link_targets()` and `npm_link_targets(prod = True)` and npm_link_targets(dev = True)` in the same BUILD if you wanted.

jbedard: I was thinking you can't have multiple `npm_link_all_packages()` targets though...?

jbedard: Do we need the `npm_link_all_packages` API changes in this PR at all?

npm/private/transitive_closure.bzl
```diff
@@ -115,6 +115,7 @@ def translate_to_transitive_closure(importers, packages, prod = False, dev = Fal
             # deps this importer should pass on if it is linked as a first-party package; this does
             # not include devDependencies
             "deps": deps,
+            "dev_deps": dev_deps,
             # all deps of this importer to link in the node_modules folder of that Bazel package and
             # make available to all build targets; this includes devDependencies
             "all_deps": all_deps,
```
jbedard, line 121: Is this still used?

jbedard:
@jfirebaugh have you had a chance to make any updates here? Are you using the patch in your repo?

WDYT of `npm_link_all_packages(name = "node_modules")` simply producing additional targets instead of accepting the flags?
* `:node_modules__dev`
* `:node_modules__prod`

Then `npm_link_targets(dev, prod)` like you already have here.

npm/private/test/snapshots/bzlmod/npm_defs.bzl
jfirebaugh: After taking another look, I don't think it is possible. The contents of this fragment depend on whether the package in question is a production or dev dependency for the `bazel_package` in question. I.e., it's not a simple `include_dev` / `include_prod` boolean for each package, but rather a matrix of `bazel_package` x `link_alias`, which we're encoding as a nested conditional.

There might be some other way to structure this, but it would require a more invasive change.

npm/private/transitive_closure.bzl
jfirebaugh: No, will remove.

npm/private/npm_translate_lock_generate.bzl
jfirebaugh: I can't remember why I included the `npm_link_all_packages` changes originally; I guess just for parallelism with `npm_link_targets`. We're only using `npm_link_targets(prod = True)`.

jfirebaugh:
Still using this in our repo. Just updated the patch for 2.3.8 and removed the change to `npm_link_all_packages`'s API.

jfirebaugh:
Hmm, removing that breaks something for us. Will have to investigate later.

jbedard:
@jfirebaugh this has been picked up in https://github.com/aspect-build/rules_js/pull/2298, would you be able to test that out to see if it works for you?

jbedard:
https://github.com/aspect-build/rules_js/pull/2298 has merged, let us know if that is missing anything

PR #2057, opened 2025-01-05
test: add repositories.bzl snapshot tests for each lockfile version

jbedard:
Adding the test snapshot files to ensure we see unexpected changes when changing related code.

---

### Test plan

- New test cases added


alexeagle:
I wish we didn't have so many of these - they are big and brittle, and it's hard to tell whether the changes to these "golden" files are desirable or not. Would it be just as good to test just one of them?

jbedard:
I explicitly want to see the diff between the pnpm versions and make sure changes to them align. These ones (e2e/pnpm_lockfiles) are easy to update with 1 sh script, it's all just for me to look at the diffs for each version.

dzbarsky (approved):
while the huge amount of goop is annoying, given how pnpm behavior differs between the versions it seems reasonable to have the various goldens. Perhaps the right path forward is to start dropping support for the older versions, v 5.4 was released 5 years ago ;)

PR #2065, opened 2025-01-09
fix(npm): correct use of pnpm lockfile dev flag

jbedard:
Fix the usage of the lockfile `dev` flag, rename vars where possible to better indicate the meaning.

---

### Test plan

- Manual testing; please provide instructions so we can reproduce: https://github.com/jbedard/rules_js/commit/cf3f12a7194e62b8194afa596fc0213b23df7bd8 and toggling `npm_translate_lock` `dev` and `prod` flags.


dzbarsky (approved):
nit: Maybe also consider "dev_only" or "only_dev", its a bit closer to `testonly`?

PR #2072, opened 2025-01-15
docs: add troubleshooting note and example of using node_options for nicer stack traces

jbedard:
An example and note in the troubleshooting for how to make nicer stack traces.

---

Add an example and troubleshooting notes for nicer stack traces when running node, such as `js_test` or `js_binary` targets.

### Test plan

- New test cases added


alexeagle (approved):
Thanks @jfirebaugh - okay with you for us to use your work? Want attribution?

jfirebaugh:
Go for it. No attribution necessary.

PR #2080, opened 2025-01-30
refactor: do not add js_library(types) to JsInfo.sources

jbedard:
No "types" should ever make it into runfiles or `JsInfo.[transitive_]sources` as it might trigger type-producing actions such as tsc.

I'm uncertain if packages in `js_library(types)` should be passed forward to `JsInfo.{npm_sources,npm_package_store_infos}` though.

---

Do not propagate `js_library(types)` or `JsInfo.[transitive_]types` to `JsInfo.sources`.

### Test plan

### DIFF
--- a/js/private/js_library.bzl
+++ b/js/private/js_library.bzl
@@ -198,15 +198,14 @@ def _js_library_impl(ctx):
         files = ctx.files.srcs,
     )
 
-    additional_sources, additional_types = _gather_sources_and_types(
+    additional_types_sources, additional_types = _gather_sources_and_types(
         ctx = ctx,
         targets = ctx.attr.types,
         files = ctx.files.types,
     )
 
     # Direct sources and types
-    sources = depset(transitive = [sources, additional_sources])
-    types = depset(transitive = [types, additional_sources, additional_types])
+    types = depset(transitive = [types, additional_types_sources, additional_types])
 
     # Transitive sources and types
     transitive_sources = [sources]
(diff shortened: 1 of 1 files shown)


js/private/js_library.bzl
```diff
@@ -216,18 +215,24 @@ def _js_library_impl(ctx):
...
 
-    # Collect transitive sources, types and npm providers from srcs+types+deps
-    for target in srcs_types_deps:
+    # Collect transitive sources, types and npm providers from js_library(srcs+deps)
+    for target in srcs_deps:
         if JsInfo in target:
             jsinfo = target[JsInfo]
             transitive_sources.append(jsinfo.transitive_sources)
             transitive_types.append(jsinfo.transitive_types)
             npm_sources.append(jsinfo.npm_sources)
```
gregmagolan, line 227: Targets in `types` may have npm deps they depend on so I think it is correct to propagate the `npm_package_store` infos as well.

Technically, this PR is a breaking change as we're changing the contract for the `types` attribute:

```
"""Same as `srcs` except all files are also provided as "types" available to downstream rules for type checking.
For example, a js_library with only `.js` files that are intended to be imported as `.js` files by downstream type checking
rules such as `ts_project` would list those files in `types`:
js_library(
    name = "js_lib",
    types = ["index.js"],
)
"""
```

With this change it is no longer "Same as `srcs`" but instead deps added to `types` only propogate their types and not their sources. That does seem like a better pattern for that attribute... essentially if you wanted something that wouldn't be normally be a type to be included in types then you would put it in `types`. If you also wanted it in sources then you would put it in both `types` and `sources`. With this change you can pass a dep with both sources and types and intentionally strip out the sources.

However, the question is how breaking is this for users? My guess is that very few ppl are using the `types` attribute and this could be marked as a "fix" under the reasoning that the past behaviour was not correct.

gregmagolan: Semver police might overlook this one :)

jbedard:
Since this is a rarely used API I think I'll drop this for now...

PR #2082, opened 2025-02-05
Use Label.repo_name instead of Label.workspace_name

EdSchouten:
According to the documentation, Label.workspace_name is deprecated:

https://bazel.build/rules/lib/builtins/Label.html#workspace_name

Label.repo_name behaves identically and is available in Bazel 6 and later.

### Test plan


jbedard (approved):

jbedard:
Thank you! 👍

PR #2087, opened 2025-02-06
chore: remove unused rules_go dev dep

jbedard:
I don't know why this was here? The e2e rules_docker test specifies it's there to fix a rules_docker bug so I left it there.

---

### Test plan


alexeagle (approved):
🧹

PR #2093, opened 2025-02-07
[bugfix]: fix exclude_package_contents with multiple patterns

Mivr:
[feat]: exclude_package_contents handle * as a matching pattern
[bugfix]: fix exclude_package_contents with multiple patterns
[docs] for exclude_package_contents

Attribute exclude_package_contents now handles multiple exclude patterns per package and it allows for the use of * as a pattern for matching all packages, matches will be NOT accumulative between * and all other package name patterns.

### Test plan
N/A


docs/troubleshooting.md
```diff
@@ -165,3 +165,23 @@ If you are experiencing slower than expected builds, you can try disabling or re
 #### Jest
 
 See [rules_jest](https://github.com/aspect-build/rules_jest) specific [troubleshooting](https://docs.aspect.build/rulesets/aspect_rules_jest/docs/troubleshooting#performance).
+
+### Too much is downloaded from a package
```
jbedard, line 169: => "Unnecessary npm package content"?

There might also be issues with the number of files, not only download size. More files means more bazel actions to copy them etc.

docs/troubleshooting.md
```diff
@@ -165,3 +165,23 @@ If you are experiencing slower than expected builds, you can try disabling or re
...
+
+In some cases NPM dependencies bring too much not needed files in them, for example test code and images for documentation and so on.
```
jbedard, line 171: > Npm packages sometimes include unnecessary files such as tests, test data etc. Large files or a large number of files can effect performance and are sometimes worth explicitly excluding content.
> For example:

docs/troubleshooting.md
```diff
@@ -165,3 +165,23 @@ If you are experiencing slower than expected builds, you can try disabling or re
...
+
+In these cases you can add such packages and the respective files/folders you want to exclude to your npm_translate_lock rule in the exclude_package_contents attribute like so:
+
+
+```starlark
+npm.npm_translate_lock(
+    ...
+    exclude_package_contents = {
+        "chalk": ["**/README*"],
```
jbedard, line 180: Lets use an example such as `**/tests/*` which I think is the most common and realistic example. I don't think there's any real reason to exclude README or LICENSE files...

npm/private/exclude_package_contents_default.bzl
```diff
@@ -0,0 +1,6 @@
+"A default files exclude list for common packages."
+
+exclude_package_contents_default = {
+    #    TODO: add more packages here with their not needed contents defined
+    "is-odd": ["**/README*", "**/LICENSE*"],
```
jbedard, line 5: I don't think we should be excluding licenses, that probably breaks some rules.

For now maybe just leave this blank?

docs/troubleshooting.md
jbedard: It seems like the `resolve` package is popular and has a `test` directory in it, that might be a better example?

docs/troubleshooting.md
Mivr: Fixed

docs/troubleshooting.md
Mivr: Fixed

docs/troubleshooting.md
Mivr: Applied

npm/private/exclude_package_contents_default.bzl
Mivr: Removed

docs/troubleshooting.md
```diff
@@ -165,3 +165,23 @@ If you are experiencing slower than expected builds, you can try disabling or re
 #### Jest
 
 See [rules_jest](https://github.com/aspect-build/rules_jest) specific [troubleshooting](https://docs.aspect.build/rulesets/aspect_rules_jest/docs/troubleshooting#performance).
+
+### Unnecessary npm package content
```
jbedard, line 169: Lets move it above `Jest` though, so it's a subcategory under `## Performance`

docs/troubleshooting.md
```diff
@@ -165,3 +165,23 @@ If you are experiencing slower than expected builds, you can try disabling or re
...
+
+Npm packages sometimes include unnecessary files such as tests, test data etc. Large files or a large number of files can effect performance and are sometimes worth explicitly excluding content.
+
+In these cases you can add such packages and the respective files/folders you want to exclude to your npm_translate_lock rule in the exclude_package_contents attribute like so:
```
jbedard, line 173: backticks around "exclude_package_contents"?

docs/troubleshooting.md
```diff
@@ -165,3 +165,23 @@ If you are experiencing slower than expected builds, you can try disabling or re
...
+```starlark
+npm.npm_translate_lock(
+    ...
+    exclude_package_contents = {
+        "resolve": ["**/test/*"],
+    },
+)
+```
+
+This example will remove the README file, regardless of its extension and its position inside any version of the library chalk.
```
jbedard, line 185: "This example will remove the README file" => ...

docs/troubleshooting.md
Mivr: Fixed

docs/troubleshooting.md
Mivr: Fixed

docs/troubleshooting.md
Mivr: I left only: This example will remove the README file
Is this what you meant?

npm/private/npm_package_store.bzl
```diff
@@ -231,7 +233,13 @@ def _npm_package_store_impl(ctx):
                 # executable which make the directories not listable (pngjs@5.0.0 for example).
                 bsdtar = ctx.toolchains["@aspect_bazel_lib//lib:tar_toolchain_type"]
                 is_macos = ctx.target_platform_has_constraint(ctx.attr._macos_constraint[platform_common.ConstraintValueInfo])
-                tar_exclude_package_contents = (["--exclude"] + ctx.attr.exclude_package_contents) if ctx.attr.exclude_package_contents else []
+
+                tar_exclude_package_contents = []
```
jbedard, line 237: Can we add a test that previously failed without this fix?

docs/troubleshooting.md
Mivr: Fixed

npm/private/npm_import.bzl
```diff
@@ -819,15 +829,15 @@ _ATTRS_LINKS = dicts.add(_COMMON_ATTRS, {
     "transitive_closure": attr.string_list_dict(),
     "package_visibility": attr.string_list(),
     "replace_package": attr.string(),
-    "exclude_package_contents": attr.string_list(),
+    "exclude_package_contents": attr.string_list(default = attributes_dummy_list),
```
jbedard, line 832: Can't this be the `exclude_package_contents_default`?

Mivr: Yeah, fixed, thanks

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -47,6 +47,35 @@ Check the public_hoist_packages attribute for duplicates.
...
+                    for item in v:
+                        result[item] = []
+                elif type(v) == "string":
+                    result[v] = []
+                else:
+                    fail("expected value to be list or string")
+            elif type(v) == "list":
+                result = v
+            elif type(v) == "string":
+                result = [v]
```
jbedard, line 69: Won't assigning an array here break the `.keys()` in the `return`?

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -361,7 +390,7 @@ ERROR: can not apply both `pnpm.patchedDependencies` and `npm_translate_lock(pat
         patches = [("@" if patch.startswith("//") else "") + patch for patch in patches]
 
         # gather exclude patterns
-        exclude_package_contents, _ = _gather_values_from_matching_names(True, attr.exclude_package_contents, name, friendly_name, unfriendly_name)
```
jbedard, line 364: Is it possible and maybe less code to just use `_gather_values_from_matching_names` and then follow it with:
```
if exclude_package_contents = None and "*" in attr.exclude_package_contents:
   exclude_package_contents = attr.exclude_package_contents["*"]
```
?

npm/private/npm_translate_lock_helpers.bzl
Mivr: Fixed

npm/private/npm_translate_lock_helpers.bzl
Mivr: It is shorter, but I think it will change the behavior in this case:
{
   "*":[<common patterns>],
   "packageA": [<specific patterns>],
}

The current solution will have common and specific patterns for packageA, while the if statement will give only the specific. And I think the user would expect an additive behavior here.

WDYT?

jbedard: I don't think we want additive because a user needs to be able to override the default.

The default might include something like `*/test/*` but maybe a package has a testing API in that directory etc. That does mean they'd need to clone the values into every package but that's better then being unable to override.

Mivr: Ok, for the default now it will be overriden and not merged with any values from the user.

However for the case of multiple matches should we pick one or merge them all?

An example:
{
"*":["README"],
"react": ["test-one"],
"react@18": ["test-two"],
}

In this case should we exclude both the react and react@18 matches and the * matches or something else?

e2e/npm_translate_lock_exclude_package_contents/MODULE.bazel
```diff
@@ -17,7 +17,11 @@ npm.npm_translate_lock(
     name = "npm",
     data = ["//:package.json"],
     exclude_package_contents = {
-        "is-odd@3.0.1": ["**/README*"],
+        "is-odd@3.0.1": [
+            "**/README*",
+            "",
```
jbedard, line 22: Why a blank string?

Mivr: We had a failure when passing a blank string, thus some code was added to filter it out, and this line is testing that the filtering still works

Mivr:
Broken down to several branches and PRs, first PR: https://github.com/aspect-build/rules_js/pull/2131

PR #2099, opened 2025-02-11
devserver: use readline instead of reading from raw stdin

dmiller-figma:
Without this change, it is theoretically possible for the `IBAZEL_*` messages to be combined in to one chunk, or have one message be split across other chunks. This would cause the `.includes` checks to either fail to see a message, or only see one message when in reality both were present.

### Test plan

This is somewhat difficult to test as it depends on timing. It does seem to reliably happen when the devserver is first starting up and receives a bunch of messages quickly as it hasn't processed any of them yet.

### DIFF
--- a/js/private/js_run_devserver.mjs
+++ b/js/private/js_run_devserver.mjs
@@ -4,6 +4,7 @@ import * as fs from 'node:fs'
 import * as os from 'node:os'
 import * as child_process from 'node:child_process'
 import * as crypto from 'node:crypto'
+import * as readline from 'node:readline'
 
 // Globals
 const RUNFILES_ROOT = path.join(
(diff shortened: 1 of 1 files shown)


thesayyn (approved):
LGTM, i'll defer to @jbedard for final look

js/private/js_run_devserver.mjs
```diff
@@ -452,10 +453,11 @@ async function main(args, sandbox) {
         })
 
         // Process stdin data in order using a promise chain.
-        let syncing = Promise.resolve()
-        process.stdin.on('data', async (chunk) => {
```
jbedard, line 456: I think both the `async` and `return` were pointless? 🤔

jbedard (approved):

jbedard:
Also LGTM 👍

alexeagle:
@dmiller-figma do you mind clicking through the CLA?

dmiller-figma:
Hey @alexeagle, waiting on legal approval from my end. Will ping them again

PR #2101, opened 2025-02-11
feat: support pnpm v10

### LINKED ISSUE #2100: [FR]: pnpm 10

Pnpm 9.15 is the latest version of PNPM available.

### Describe the feature

Can we please add pnpm versions until 10.3.0: https://www.npmjs.com/package/pnpm 
to https://github.com/aspect-build/rules_js/blob/main/npm/private/versions.bzl

jbedard:
Close #2100

---

Add tests + official support for pnpm v10+. Note there have been reports of bugs with pnpm-lock.yaml files generated with pnpm v10.0 where the [lockfiles can not be parsed by jq](https://github.com/aspect-build/rules_js/issues/2100#issuecomment-2668453889). The known scenarios of this occurring seem to no longer exist in pnpm v10.1+.

### Test plan

- New test cases added


e2e/pnpm_lockfiles/lockfile-test.bzl
```diff
@@ -198,8 +201,8 @@ def lockfile_test(npm_link_all_packages, name = None):
             ":.aspect_rules_js/node_modules/lodash@4.17.21/dir",
 
             # Patched dependencies
-            ":.aspect_rules_js/node_modules/meaning-of-life@1.0.0_o3deharooos255qt5xdujc3cuq",
-            "@%s__meaning-of-life__1.0.0_o3deharooos255qt5xdujc3cuq//:pkg" % lock_repo,
```
jbedard, line 202: pnpm v10 changed hashes from 26 characters to 64 characters, this causes it to go above the [arbitrary 32 limit we have in rules_js](https://github.com/aspect-build/rules_js/blob/f81d70e46eb83d8c8937cbc2ae20151d0317c065/npm/private/pnpm.bzl#L273-L277) so this becomes a hash (of a hash in this case).

e2e/pnpm_lockfiles/projects/peer-types/package.json
```diff
@@ -2,11 +2,18 @@
     "name": "test-peer-types",
     "private": true,
     "description": "A package with peer dependencies including: workspace, link, also-in-dev, odd version specifiers",
-    "peerDependencies": {
+    "devDependencies": {
         "@scoped/a": "workspace:*",
         "@scoped/b": "link:../b",
```
jbedard, line 7: All of these odd version specified (except `workspace:*`) are no longer valid `peerDependencies` in v10 so I'm changing the test

e2e/pnpm_lockfiles/setup.sh
```diff
@@ -4,13 +4,17 @@
...
-pushd base && npx pnpm@8.6.0 install --lockfile-only && mv pnpm-lock.yaml ../v61/ && popd
+pushd base && npx -y pnpm@8.5.1 install --lockfile-only && mv pnpm-lock.yaml ../v60/ && popd
+pushd base && npx -y pnpm@8.6.0 install --lockfile-only && mv pnpm-lock.yaml ../v61/ && popd
 
 # pnpm v9.0.0 bumped the lockfile version to 9.0
-pushd base && npx pnpm@^9.0 install --lockfile-only && mv pnpm-lock.yaml ../v90 && popd
+pushd base && npx -y pnpm@~9 install --lockfile-only && mv pnpm-lock.yaml ../v90 && popd
+
+# pnpm v10
+pushd base && npx -y pnpm@~10 install --lockfile-only && mv pnpm-lock.yaml ../v10 && popd
```
jbedard: Note that ~10.0 has bugs with URLs being fully resolved within `{name}@{version}` values in the yaml, so this is `~10` instead of `~10.0` like I originally had it. I've made the same change to the v9 line above (which did not cause any changes to the v9 lockfile, but I'd like to change it so we are aware of future changes).

e2e/pnpm_lockfiles/MODULE.bazel
```diff
@@ -10,6 +10,7 @@ PNPM_LOCK_VERSIONS = [
     "v60",
     "v61",
     "v90",
+    "v10",
```
alexeagle: naming convention left something to be desired. Seems this should be v100 so you have a way to express v10.2

e2e/pnpm_lockfiles/lockfile-test.bzl
alexeagle: why not just increase that arbitrary limit to 64? It feels wrong to me for us to throw away bits of entropy here, since it raises the question "can we have collisions"

alexeagle (approved):

e2e/pnpm_lockfiles/lockfile-test.bzl
jbedard: Technically that would be a "breaking change" if we simply change 32 to 64 because it will change repository names of packages. That is the type of thing I want to explore more while looking into rules_js v3 and I have similar changes in https://github.com/aspect-build/rules_js/pull/2177 atm

e2e/pnpm_lockfiles/MODULE.bazel
jbedard: I'll change that to v100 👍

PR #2102, opened 2025-02-12
chore: mirror external pnpm v10 releases

jbedard:
There may still be lockfile issues (see #2101) but those may be edge cases and we can add v10 to `versions.bzl` since pnpm v8 is still the default.

---

Add pnpm v10 

### Test plan


alexeagle (approved):

alexeagle:
Isn't there a spot where the automation should be updated to mirror these?

jbedard:
I forget why I did this manually 🤷

PR #2106, opened 2025-02-17
fix(coverage): ensure js_library#srcs are instrumented for coverage

alexeagle:
---

### Test plan

See comment in the source file where I added repro

### DIFF
--- a/js/private/js_library.bzl
+++ b/js/private/js_library.bzl
@@ -256,7 +256,7 @@ def _js_library_impl(ctx):
     return [
         coverage_common.instrumented_files_info(
             ctx,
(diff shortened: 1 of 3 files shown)


js/private/js_library.bzl
```diff
@@ -254,6 +254,16 @@ def _js_library_impl(ctx):
     )
 
     return [
+        coverage_common.instrumented_files_info(
+            ctx,
+            source_attributes = ["srcs"],
```
thesayyn, line 259: We need to add deps to dependency attributes

thesayyn (approved):
One minor change, otherwise lgtm

PR #2115, opened 2025-03-03
devserver: skip unlink when possible

dmiller-figma:
Without this change we incur the cost of an extra unnecessary syscall on every file we need to sync. In my testing this didn't make a perceivable difference with small sets of changed files, however it could make a difference on large ones.

This change works just fine on Linux. However, on macOS you get a permission error if you try to copy over an existing file without `grant_sandbox_write_permissions = True`. So on macOS we only enable this optimization if `grant_sandbox_write_permissions == True`, otherwise we `unlink` the file like we did before.

---

### Test plan
Existing tests

If it is possible to write a test that only runs on macOS, we could test that code path. However, I'm not sure how feasible that is.

### DIFF
--- a/js/private/js_run_devserver.mjs
+++ b/js/private/js_run_devserver.mjs
@@ -219,9 +219,14 @@ async function syncRecursive(src, dst, sandbox, writePerm) {
                     )} (${friendlyFileSize(lstat.size)})`
                 )
             }
-            if (exists) {
+            // unlink file if it exists, unless `writePerm` is true or we are not on Linux
+            // on macOS, and possibly other operating systems, you get errors if you copy
+            // over a file without the necessary permissions
+            let unlinkIsNecessary = writePerm || process.platform !== "linux";
+            if (exists && unlinkIsNecessary) {
                 await fs.promises.unlink(dst)
-            } else {
+            }
+            if (!exists) {
                 // Intentionally synchronous; see comment on mkdirpSync
                 mkdirpSync(path.dirname(dst))
             }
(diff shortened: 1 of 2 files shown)


js/private/js_run_devserver.mjs
```diff
@@ -219,9 +219,14 @@ async function syncRecursive(src, dst, sandbox, writePerm) {
                     )} (${friendlyFileSize(lstat.size)})`
                 )
             }
-            if (exists) {
+            // unlink file if it exists, unless `writePerm` is true or we are not on Linux
+            // on macOS, and possibly other operating systems, you get errors if you copy
+            // over a file without the necessary permissions
+            let unlinkIsNecessary = writePerm || process.platform !== "linux";
```
jbedard, line 225: `const`? 🤷

js/private/js_run_devserver.mjs
```diff
@@ -214,14 +214,11 @@ async function syncRecursive(src, dst, sandbox, writePerm) {
...
                 console.error(
-                    `Syncing file ${src.slice(
-                        RUNFILES_ROOT.length + 1
-                    )} (${friendlyFileSize(lstat.size)})`
+                    `Syncing file ${src} (${friendlyFileSize(lstat.size)}) to ${dst}`
                 )
+                console.error(`Dst exists: ${exists}`)
             }
-            if (exists) {
-                await fs.promises.unlink(dst)
```
jbedard: Don't we need to keep this when `!grant_sandbox_write_permissions`?

PR #2116, opened 2025-03-04
chore: remove non-bzlmod unit testing

jbedard:
### Test plan


WORKSPACE
```diff
@@ -1,260 +1,3 @@
 workspace(name = "aspect_rules_js")
```
alexeagle, line 1:
```suggestion
# This file is not used by Bazel 8 and should be removed.
#
# However its presence works around issues with some tools:
# https://github.com/bazel-contrib/bazel-gazelle/issues/2012
# https://github.com/bazelbuild/bazel-watcher/issues/646
```

WORKSPACE
```diff
@@ -1,260 +1,3 @@
...
-    urls = {
-        "darwin-aarch64": ["https://github.com/aspect-forks/llvm-project/releases/download/aspect-release-14.0.0/clang+llvm-14.0.0-arm64-apple-darwin.tar.xz"],
-        "linux-x86_64": ["https://github.com/aspect-forks/llvm-project/releases/download/aspect-release-14.0.0/clang+llvm-14.0.0-x86_64-linux-gnu.tar.xz"],
-    },
-)
-
-load("@llvm_toolchain//:toolchains.bzl", "llvm_register_toolchains")
-
-llvm_register_toolchains()
+fail("Local WORKSPACE development unsupported")
```
alexeagle, line 3: interesting, I haven't seen a `fail` in a WORKSPACE file before.

What does "local" mean here?
I think it's enough to just upgrade to Bazel 8 and not expect anyone will pass `--enable_workspace`

alexeagle (approved):

WORKSPACE
jbedard: Just means you'll see that `fail` message instead of random bazel errors saying things our undefined.

I guess I meant "running bazel in this directory". You can still use rules_js from a different workspace, it's just development within the `rules_js` repository that won't work...

PR #2127, opened 2025-03-12
Add new idea bazel plugin folder to gitignore

Mivr:
### Test plan

Just a developer improvement if the rules_js dev is using the new idea Bazel plugin: https://plugins.jetbrains.com/plugin/22977-bazel-eap-

This plugin creates this folder for keeping the project view. I think it should be ignored as it is specific to the used IDE of the rules_js dev.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ node_modules/
 # [MODULE.bazel.lock file "reads through" already-locked package manager](https://github.com/bazelbuild/bazel/issues/20272)
 # [moduleFileHash in MODULE.bazel.lock causes frequent Git merge conflicts](https://github.com/bazelbuild/bazel/issues/20369)
 MODULE.bazel.lock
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -16,3 +16,5 @@ node_modules/
 # [MODULE.bazel.lock file "reads through" already-locked package manager](https://github.com/bazelbuild/bazel/issues/20272)
 # [moduleFileHash in MODULE.bazel.lock causes frequent Git merge conflicts](https://github.com/bazelbuild/bazel/issues/20369)
 MODULE.bazel.lock
+
+.bazelbsp
```
jbedard, line 20: Can we move this up alongside `.idea` and the other ide stuff?

Mivr: Moved

jbedard (approved):

PR #2130, opened 2025-03-12
feat: implement layer_groups

thesayyn:
Implements the `layer_groups` feature for js_image_layer.  This is a major overhaul for js_image_layer as well. We no longer have a custom tar builder program, instead we use bsdtar toolchain to turn `mtree` generated by js_image_layer_splitter into an archive.

Add `js_image_layer(layer_groups)` to allow adding and customizing js image layers.

### Test plan

- New test cases added


js/private/js_image_layer.mjs.tpl
```diff
@@ -0,0 +1,353 @@
+import { readdir, readFile, readlink, writeFile } from 'node:fs/promises'
```
jbedard, line 1: This being a `.tpl` is really annoying, and breaks the diff (or maybe there is no usable diff, I can't tell atm).

Which part is a template? Can we do some other type of magic to keep the `.mjs` extension despite using it as a template?

js/private/js_image_layer.bzl
```diff
@@ -500,18 +531,45 @@ js_image_layer_lib = struct(
...
+JavaScript identifier (alphanumeric characters),  and <value> MAY be either an empty string (signifying a universal match)
+or a valid regular expression.
+ 
+For better performance, it is recommended to split the large parts of a js_binary to have a separate layer.
+
+The matching order for layer groups is as follows:
 
-Helpful when using js_image_layer with rules_docker.
-See https://github.com/aspect-build/rules_js/pull/1714 for more info""",
+1. `layer_groups` are checked in order first
```
jbedard, line 557: These newlines aren't showing up in the .md? It looks ugly when the .md is rendered and this is all one line...

js/private/js_image_layer.bzl
```diff
@@ -500,18 +531,45 @@ js_image_layer_lib = struct(
...
+2. If no match is found for `layer_groups`, the `default layer groups` are checked.
+3. Any remaining files are placed into the app layer.
+
+The default layer groups are as follows and always created.
+
+```
+{
+    "node": ".*\\/js\\/private\\/node-patches.*|.*\\/bin\\/nodejs\\/.*",
```
jbedard, line 565: This `*`s here make things bold in the rendered .md. Do `*`s need to be escaped even when they are in the triple-backtick?

js/private/js_image_layer.bzl
```diff
@@ -500,18 +531,45 @@ js_image_layer_lib = struct(
...
+            default = False,
+        ),
+        "layer_groups": attr.string_dict(
+            doc = """Layer groups to create.
+These are utilized to categorize files into distinct layers, determined by their respective paths. 
+The expected format for each entry is "<key>": "<value>", where <key> MUST be a valid Bazel and 
+JavaScript identifier (alphanumeric characters),  and <value> MAY be either an empty string (signifying a universal match)
+or a valid regular expression.
+ 
+For better performance, it is recommended to split the large parts of a js_binary to have a separate layer.
```
jbedard, line 553: Backticks on `js_binary`?

jbedard, line 553: What does "large parts of a js_binary" mean? Does that mean the `js_binary(data)` such as transitive js+npm paths?

js/private/js_image_layer.mjs.tpl
jbedard: The formatting is also odd, with tabs/spaces mixed all over the place etc

js/private/test/image/BUILD.bazel
```diff
@@ -69,3 +71,21 @@ assert_js_image_layer_listings(
     name = "custom_owner_test",
     js_image_layer = ":custom_owner",
 )
+
```
jbedard, line 74: Can we see an example of excluding things? Similar to my use case wanting to exclude `@types`...

js/private/js_image_layer.bzl
thesayyn: eh startdoc is stupid.

js/private/test/image/custom_owner_test_package_store_1p.bzlmod.listing
```diff
@@ -1,13 +1,13 @@
...
-drwxr-xr-x  0 100    0           0 Jan  1  1970 app/js/private/test/image/bin.runfiles
-drwxr-xr-x  0 100    0           0 Jan  1  1970 app/js/private/test/image/bin.runfiles/_main
-drwxr-xr-x  0 100    0           0 Jan  1  1970 app/js/private/test/image/bin.runfiles/_main/node_modules
-drwxr-xr-x  0 100    0           0 Jan  1  1970 app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js
-drwxr-xr-x  0 100    0           0 Jan  1  1970 app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/@mycorp+pkg-d@0.0.0
-drwxr-xr-x  0 100    0           0 Jan  1  1970 app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/@mycorp+pkg-d@0.0.0/node_modules
-drwxr-xr-x  0 100    0           0 Jan  1  1970 app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/@mycorp+pkg-d@0.0.0/node_modules/@mycorp
-lrwxrwxr-x  0 100    0           0 Jan  1  1970 app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/@mycorp+pkg-d@0.0.0/node_modules/@mycorp/pkg-d -> ../../../../../examples/npm_package/packages/pkg_d
+drwxr-xr-x  0 100    0           0 Jan  1  1970 ./app/
```
jbedard, line 1: Is the `./` + `/` change intentional? Why?

js/private/js_image_layer.bzl
thesayyn: Anything, its up to the user to determine if something "large" and should go into its own layer.

js/private/test/image/BUILD.bazel
thesayyn: I'll do that as a follow-up.

js/private/test/image/custom_owner_test_package_store_1p.bzlmod.listing
thesayyn: this something bsdtar does by default. i guess to maintain compatibility, i should revert it back to original.  `./app` and `app` mean the same thing for overlayfs

js/private/js_image_layer.bzl
thesayyn: stardoc...

js/private/test/image/custom_owner_test_package_store_1p.bzlmod.listing
thesayyn: on second thought, i think i will leave this as is, unless somebody complains.

js/private/js_image_layer.bzl
```diff
@@ -13,9 +13,17 @@ js_image_layer(
 ```
 """
 
-load("@aspect_bazel_lib//lib:paths.bzl", "to_rlocation_path")
+load("@aspect_bazel_lib//lib:tar.bzl", "tar_lib")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 
+_DEFAULT_LAYER_GROUPS = {
+    "node": ".*\\/js\\/private\\/node-patches.*|.*\\/bin\\/nodejs\\/.*",
```
jbedard, line 20: I think you can drop all the `.*`.
Can you add a `/` at the end of the `node-patches` directory?

js/private/js_image_layer.mjs
```diff
@@ -1,9877 +1,339 @@
...
-    /**
-     * The target of this event.
-     * @type {EventTarget}
-     */
-    get currentTarget() {
-        return pd(this).currentTarget
-    },
+// Resolve symlinks while staying inside the sandbox.
+async function resolveSymlink(p) {
+    let prevHop = path.resolve(p)
```
jbedard, line 56: I think the `path.resolve` could be dropped?

js/private/js_image_layer.mjs
```diff
@@ -1,9877 +1,339 @@
...
-            const isFunc = typeof descriptor.value === "function";
-            Object.defineProperty(
-                CustomEvent.prototype,
-                key,
-                isFunc
-                    ? defineCallDescriptor(key)
-                    : defineRedirectDescriptor(key)
-            );
+function add_parents(
+	mtree,
```
jbedard, line 131: This file needs formatting run on it

js/private/js_image_layer.mjs
```diff
@@ -1,9877 +1,339 @@
...
 
-    CustomEventTarget.prototype = Object.create(EventTarget.prototype, {
-        constructor: {
-            value: CustomEventTarget,
-            configurable: true,
-            writable: true,
-        },
-    });
+async function split() {	
+    const UID = "{{UID}}"
```
jbedard, line 195: I think these are unused because `{{UID}}` is used everywhere, so it's always inlined...

js/private/js_image_layer.mjs
```diff
@@ -1,9877 +1,339 @@
...
-	];
-	const classRegExp = /^([A-Z][a-z0-9]*)+$/;
-	const nodeInternalPrefix = '__node_internal_';
-	const codes = {};
-	function assert(value, message) {
-	  if (!value) {
-	    throw new codes.ERR_INTERNAL_ASSERTION(message)
-	  }
-	}
+    await Promise.all([
```
jbedard, line 332: If we're doing `await` on the `WRITE_STATEMENTS` shouldn't we also do `await` on the `{}unusedinputs.close()` calls?

Otherwise we may as well just remove this `Promise.all` since we're not waiting for "all" things anyway.

thesayyn: we can't since we need stat calls to finish before we can sort the array. so this is working as intended.

js/private/js_image_layer.mjs
thesayyn: Its already formatted. I don't see precommit yelling me so i'll just leave it like that.

jbedard: It looks horrible 😲

js/private/js_image_layer.mjs
jbedard: Should we also `await` all the `*unusedinputs.close()` or do you want to just assume they are written before node exits? I think we can assume it but I thought it was clearer when we close them, see [CLOSE_STATEMENTS](https://github.com/jbedard/rules_js/blob/daf19490b832c87486dde357be3c6d5cc9a78f7d/js/private/js_image_layer.bzl#L356C9-L356C25) in my branch.

docs/js_image_layer.md
```diff
@@ -268,10 +269,12 @@ container_image(
...
...les into distinct layers, determined by their respective paths. The expected format for each entry is "<key>": "<value>", where <key> MUST be a valid Bazel and JavaScript identifier (alphanumeric characters),  and <value> MAY be either an empty string (signifying a universal match) or a valid regular expression.<br><br>For better performance, it is recommended to split the large parts of a `js_binary` to have a separate layer.<br><br>The matching order for layer groups is as follows:<br><br>1. `layer_groups` are checked in order first 2. If no match is found for `layer_groups`, the `default layer groups` are checked. 3. Any remaining files are placed into the app layer.<br><br>The default layer groups are as follows and always created.<br><br><pre><code>{&#10;    "node": ".*\/js\/private\/node-patches.*\|.*\/bin\/nodejs\/.*",&#10;    "package_store_1p": "\.aspect_rules_js\/.*@0\.0\.0\/node_modules",&#10;    "package_store_3p": "\.aspect_rules_js\/.*\/node_modules",&#10;    "node_modules": "\/node_modules\/",&#10;    "app": "", # empty means just match anything.&#10;}</code></pre>   | <a href="https://bazel.build/rules/lib/dict">Dictionary: String -> String</a> | optional |  `{}`  |
```
alexeagle, line 274: nit: the attribute docstring doesn't render well in the table, and even if it did, it's not a great place to give longer explanation. In these cases I've been moving this long-form doc up to the top, then just reference it in the attribute

js/private/js_image_layer.mjs
thesayyn: they are closed automatically.

js/private/js_image_layer.mjs
thesayyn: i don't remember why it was put there in the first place, so i will just leave it as is.

js/private/js_image_layer.bzl
```diff
@@ -254,210 +262,295 @@ container_image(
...
+            map_each = json.encode,
+        )
+    entries.add("}")
+
+    entries_json = ctx.actions.declare_file("{}_entries.json".format(ctx.label.name))
+    ctx.actions.write(entries_json, content = entries)
+
+    # Ordering of these matter.
+    layer_groups = dict()
+    for key in ctx.attr.layer_groups:
```
jbedard, line 502: @alexeagle do you have any opinions on this API? I haven't used `js_image_layer` enough to have a good opinion here. I find it a bit odd that custom layers go before the node layer, but I can't think of any better API to overcome that.

jbedard (approved):

js/private/js_image_layer.bzl
```diff
@@ -254,210 +262,297 @@ container_image(
...
+
+    layer_groups_gen = _run_splitter(ctx, runfiles_dir, runfiles_plus_files, entries_json, layer_groups)
+
+    tarinfo = ctx.toolchains[tar_lib.toolchain_type].tarinfo
+
+    outputs = []
+    output_groups = dict()
+    compress = "" if ctx.attr.compression == "none" else ctx.attr.compression
+    for typ, mtree, unused_inputs in layer_groups_gen:
+        ext = tar_lib.common.compression_to_extension[compress] if compress else ""
```
fa93hws, line 524: Why the `ext` is empty string when compress is `None` instead of `.tar`?

PR #2133, opened 2025-03-13
[feat] handle star pattern in exclude package contents

Mivr:

feat: exclude package contents now accepts also star pattern for selecting all npm packages in the package-lock.json

### Test plan

Tested with manually adding * as a pattern to the e2e test for exclude patterns and checking the result.

TODO: Discuss if we want to merge the * pattern excludes with the specific modules excludes and make a test for the decided approach.

### DIFF
--- a/npm/private/npm_translate_lock_helpers.bzl
+++ b/npm/private/npm_translate_lock_helpers.bzl
@@ -73,9 +73,9 @@ def _gather_unique_values_from_matching_names(additive, keyed_lists, *names):
 
     # in case the key has not been met even once, we return None, instead of empty list as empty list is a valid value
     if not keys:
-        return (None, keys)
+        return None
 
-    return (result.keys(), keys)
+    return result.keys()
 
 ################################################################################
 def _gather_values_from_matching_names(additive, keyed_lists, *names):
@@ -391,7 +391,7 @@ ERROR: can not apply both `pnpm.patchedDependencies` and `npm_translate_lock(pat
         patches = [("@" if patch.startswith("//") else "") + patch for patch in patches]
 
         # gather exclude patterns
-        exclude_package_contents, _ = _gather_unique_values_from_matching_names(True, attr.exclude_package_contents, name, friendly_name, unfriendly_name)
+        exclude_package_contents = _gather_unique_values_from_matching_names(True, attr.exclude_package_contents, name, friendly_name, unfriendly_name)
 
         # gather replace packages
         replace_packages, _ = _gather_values_from_matching_names(True, attr.replace_packages, name, friendly_name, unfriendly_name)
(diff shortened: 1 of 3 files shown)


npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -47,6 +47,36 @@ Check the public_hoist_packages attribute for duplicates.
...
+            elif type(v) == "string":
+                result[v] = []
+            else:
+                fail("expected value to be list or string")
+
+    # in case the key has not been met even once, we return None, instead of empty list as empty list is a valid value
+    if not keys:
+        return (None, keys)
+
+    return (result.keys(), keys)
```
jbedard, line 78: Why return the `keys` if we never use it? I assume it was used in the other `_gather_*` methods...?

Mivr: Yup, copy-paste error from other methods

jbedard:
Lets just add a test, maybe after https://github.com/aspect-build/rules_js/pull/2134

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -361,7 +391,7 @@ ERROR: can not apply both `pnpm.patchedDependencies` and `npm_translate_lock(pat
         patches = [("@" if patch.startswith("//") else "") + patch for patch in patches]
 
         # gather exclude patterns
-        exclude_package_contents, _ = _gather_values_from_matching_names(True, attr.exclude_package_contents, name, friendly_name, unfriendly_name)
+        exclude_package_contents = _gather_unique_values_from_matching_names(True, attr.exclude_package_contents, name, friendly_name, unfriendly_name)
```
jbedard, line 394: Can we drop the first param since we only use this method once? Maybe give the method a more specific name for now so we can simplify it such as `_gather_package_content_excludes(attr.exclude_package_contents, name, friendly_name, unfriendly_name)`?

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -47,6 +47,36 @@ Check the public_hoist_packages attribute for duplicates.
...
+            v = keyed_lists[name] if name in keyed_lists else keyed_lists["*"]
+            if additive:
+                if type(v) == "list":
+                    for item in v:
+                        result[item] = []
+                elif type(v) == "string":
+                    result[v] = []
+                else:
+                    fail("expected value to be list or string")
+            elif type(v) == "list":
```
jbedard, line 66: Lets make sure we have tests for each of this `type()`s...

npm/private/npm_translate_lock_helpers.bzl
Mivr: Fixed

npm/private/npm_translate_lock_helpers.bzl
Mivr: Added tests

jbedard (approved):

PR #2134, opened 2025-03-13
[chore] remove not needed empty lists from bazel files generation

Mivr:
Removed some not needed empty lists in the generation and prepared the code for the default list, that will come in the next PR so it can be discussed separately.

---

### Test plan


npm/private/npm_import.bzl
```diff
@@ -1180,6 +1182,7 @@ def npm_import(
             ```
             exclude_package_contents = ["**/tests/**"]
             ```
+            The default value is a list of common excludes. You can use [] to override the default.
```
jbedard, line 1185: This should be added in the PR that actually adds the default list later

Mivr: Removed

npm/private/exclude_package_contents_default.bzl
```diff
@@ -0,0 +1,5 @@
+"A default files exclude list for common packages."
+
+exclude_package_contents_default = [
```
jbedard, line 3: is it easy to leave this out for now? This PR was to remove the default `exclude_package_contents` from the generated code, can the default stuff all be in 1 PR later?

npm/private/test/snapshots/bzlmod/fsevents_links_defs.bzl
```diff
@@ -36,7 +36,6 @@ def npm_imported_package_store(name):
         name = "{}/ref".format(store_target_name),
         package = "fsevents",
         version = "2.3.2",
-        exclude_package_contents = [],
```
jbedard, line 39: Can we add a test where there is some generated code that does not get deleted with this PR?

npm/private/exclude_package_contents_default.bzl
Mivr: Good point moving it up the PRs

npm/private/test/snapshots/bzlmod/fsevents_links_defs.bzl
Mivr: I can modify a test, or add a snapshot to the e2e test we have, I will do the later and we can discuss if thats ok

jbedard (approved):

Issue #2110, opened 2025-02-20
[FR]: provide default exclude_package_contents similar to yarn autoclean


jbedard:
The [exclude_package_contents](https://github.com/aspect-build/rules_js/commit/ef8e187423bcb520641aa6d0183fafd3b342c607) feature should have reasonable defaults similar to yarn autoclean, see https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/commands/autoclean.js#L16

`exclude_package_contents` could have a default `["*"]` entry similar to the yarn defaults that users can then override with another `"*"` entry or with package specific excludes.

PR #2136, opened 2025-03-19
feat: add opt-in default npm package excludes from yarn

Mivr:
Closes https://github.com/aspect-build/rules_js/issues/2110

Added option to use default package content exclusions similar to `yarn autoclean`.

With bzlmod extensions:
```
npm.npm_package_content_excludes(
    package = "*",
    use_defaults = True,
)

# Override for `my-pkg@1.2.3`
npm.npm_package_content_excludes(
    package = "my-pkg@1.2.3",
    excludes = ["README.md", "*.spec.js"]
)
```

For legacy WORKSPACE:
```
npm_translate_lock(
    package_content_excludes = {
        "*": True,
        "my-pkg@1.2.3": ["README.md", "*.spec.js"]
    )
)
```

### Test plan


npm/private/exclude_package_contents_default.bzl
```diff
@@ -0,0 +1,49 @@
+"A default files exclude list for common packages."
```
jbedard, line 1: Lets document where this list came from, probably just a link to the yarn docs and/or code?

dzbarsky (commented):
So I love this idea (and in fact I implemented the same thing last time I worked on a big JS monorepo), but I'm a bit worried about the perf of all these globs. Alternately, we could drop these tiles as part of the extraction step. @jbedard WDYT?

jbedard:
These are globs evaluated by `tar` which I would think is the best place to do it. What do you mean "we could drop these tiles as part of the extraction step"?

dzbarsky:
> These are globs evaluated by `tar` which I would think is the best place to do it. What do you mean "we could drop these tiles as part of the extraction step"?

Oh right we invoke tar directly. In my fork we had a script action around the tar so I was thinking to put the expansion there and hide it from analysis.

Disregard me this approach is probably fine

jbedard:
@Mivr can you confirm if these are actually enabled by *default* in yarn? Or is it only when you explicitly call `yarn autoclean`? That makes me wonder if this should be opt-in.

If it is opt-in, maybe should also make this feature opt-in and then we can merge this now (without waiting for rules_js v3). Can we do something like replacing `True` with the default list? So users can do `exclude_package_contents = {"*": True}` then it replaces `True` with this default list...

Mivr:
So in yarn it has to be enabled. It is disabled by default.

I added a magic string for now, not sure it will work with a Boolean and the string does allow for more flexibility like extending the defaults further for some packages.

It's interesting to discuss which option will be more user friendly.

I tried also with a constant but Module.Bazel does not work with a constant as it does not allow for load statements.

It can also be a rule, which by syntax for the user is a close to the magic string.

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -57,15 +58,38 @@ def _gather_package_content_excludes(keyed_lists, *names):
...
             else:
                 fail("expected value to be list or string")
 
     # in case the key has not been met even once, we return None, instead of empty list as empty list is a valid value
     if not found and "*" in keyed_lists:
-        excludes = keyed_lists["*"] if type(keyed_lists["*"]) == "list" else [keyed_lists["*"]]
+        v = keyed_lists["*"]
+        if type(v) == "list":
+            for e in v:
+                if e == "yarn-autoclean":
```
alexeagle: instead of a magic string, how about `True` means you want the yarn-autoclean default list

Mivr:
The closest I could get this is:
In Module bazel the users will need to migrate to a string that contains a complex object:
'{ "*": True, "some-package": "pattern" }'
Instead of the current syntax with a dictionary.

For WORKSPACE it will just work both ways.

I am open to other approaches for this, for now this was was least intrusive way I could find for the users.

e2e/npm_translate_lock_exclude_package_contents/MODULE.bazel
```diff
@@ -16,15 +16,20 @@ npm = use_extension(
...
+    packages = ["*"],
+    use_defaults = True,
+)
+npm.exclude_package_contents(
+    packages = ["is-odd@3.0.1"],
+    patterns = [
+        "**/README*",
+        "**/LICENSE*",
+    ],
+    use_defaults = True,
```
dzbarsky, line 33: I would think use_defaults propagates from the * above, is that surprising?

Mivr: It does propagate, the test is misleading here. It does test that we will not crash if we have defaults and patterns, but it misses the point on being meaningful.

e2e/npm_translate_lock/WORKSPACE
```diff
@@ -17,7 +17,7 @@ npm_translate_lock(
     name = "npm",
     data = ["//:package.json"],
     exclude_package_contents = {
-        "is-odd@3.0.1": ["**/README*"],
+        "is-odd@3.0.1": "**/README*",  # Single pattern as string
```
jbedard: Do we need to add "single pattern" support? I'd prefer keeping it as just `True|False|[]string`...

e2e/npm_translate_lock_exclude_package_contents/MODULE.bazel
```diff
@@ -16,15 +16,20 @@ npm = use_extension(
...
-            "",
-            "**/LICENSE*",
-        ],
-    },
     npmrc = "//:.npmrc",
     pnpm_lock = "//:pnpm-lock.yaml",
     verify_node_modules_ignored = "//:.bazelignore",
 )
+npm.exclude_package_contents(
+    packages = ["*"],
```
jbedard: Why an array instead of `package`? I guess an array simplifies many scenarios like a single array of packages to opt-out of the defaults?

e2e/npm_translate_lock_exclude_package_contents/WORKSPACE
```diff
@@ -16,9 +16,10 @@ load("@aspect_rules_js//npm:repositories.bzl", "npm_translate_lock")
 npm_translate_lock(
     name = "npm",
     exclude_package_contents = {
+        "*": [True],  # Use default exclusions for all packages
```
jbedard: IMO this should be `True|False|[]string` and in workspace we don't support extending the defaults? Or what would be less code? For workspace I want the smallest+simplest code, for bzlmod I want the best API for long term

e2e/npm_translate_lock_exclude_package_contents/MODULE.bazel
jbedard: @dzbarsky any opinions here?

npm/extensions.bzl
```diff
@@ -237,11 +264,28 @@ def _npm_import_attrs():
...
+            default = False,
+        ),
+    }
+
 npm = module_extension(
     implementation = _npm_extension_impl,
     tag_classes = {
         "npm_translate_lock": tag_class(attrs = _npm_translate_lock_attrs()),
         "npm_import": tag_class(attrs = _npm_import_attrs()),
+        "exclude_package_contents": tag_class(attrs = _exclude_package_contents_attrs()),
```
jbedard: I would vote to put the `npm_` prefix just to be consistent. We can debate dropping them all in v3 maybe?

e2e/npm_translate_lock_exclude_package_contents/MODULE.bazel
dzbarsky: I think gazelle uses a separate default_override tag and then module_override for individual ones. Not sure if it's better or worse, just pointing out precedent :)

dzbarsky: Oh that's not quite what you were asking, I think usually rulesets will do single one and user can do a list comprehension if desired. (Again just mentioning common practice for pip/rust/go, not necessarily opposed to this)

jbedard: Ok, so switch this to `package = "*"` and if a user wants to have multiple they can do `[... for p in [...]]` 👍 

> I think gazelle uses a separate default_override tag and then module_override for individual ones. Not sure if it's better or worse, just pointing out precedent :)

That is an interesting point... so instead of 2 attributes to 1 tag we can have 2 tags? That means you can no longer do "defaults + my-excludes" but that reduces confusion...

e2e/npm_translate_lock/WORKSPACE
Mivr: Addressed

e2e/npm_translate_lock_exclude_package_contents/MODULE.bazel
Mivr: Removed, now its package

e2e/npm_translate_lock_exclude_package_contents/WORKSPACE
Mivr: Fixed

npm/extensions.bzl
Mivr: Added

npm/extensions.bzl
```diff
@@ -237,11 +264,28 @@ def _npm_import_attrs():
...
+        "package": attr.string(
+            doc = "Package name to apply exclusions to. Supports wildcards like '*' for all packages.",
+            mandatory = True,
+        ),
+        "patterns": attr.string_list(
+            doc = "List of glob patterns to exclude from the specified package.",
+            default = [],
+        ),
+        "use_defaults": attr.bool(
+            doc = "Whether to use default exclusion patterns for the specified package.",
```
jbedard: Lets mention that the list is from yarn?

npm/extensions.bzl
```diff
@@ -44,7 +48,27 @@ def _npm_extension_impl(module_ctx):
...
+def _build_exclude_package_contents_config(module_ctx):
+    """Build exclude_package_contents configuration from tags across all modules."""
+    exclusions = {}
+
+    for mod in module_ctx.modules:
+        for exclude_tag in mod.tags.npm_exclude_package_contents:
+            # Process the package in the tag
+            package = exclude_tag.package
+            if package not in exclusions:
+                exclusions[package] = []
```
jbedard: Lets overwrite 100% of the time. Maybe add a `print()` warning if we are overwriting and a value is getting lost though?

jbedard: Actually lets throw an error instead of `print()`, and if bugs are logged we can decide on overwrite vs merge later...

npm/private/npm_translate_lock.bzl
```diff
@@ -284,18 +310,27 @@ def npm_translate_lock(
...
-            to a list of patterns to exclude from the package's generated node_modules link targets. Multiple matches are additive.
+        exclude_package_contents: Configuration for excluding package contents (WORKSPACE only).
+
+            For MODULE.bazel, use the `exclude_package_contents` tag class instead.
+
+            The configuration is a dictionary that maps package names (or package names with their version, e.g., "my-package" or "my-package@v1.2.3") to exclusion rules.
+
+            Values can be:
+            - `True`: Use default exclusions
+            - String: Single exclusion pattern
```
jbedard: We removed support for this?

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -48,27 +48,39 @@ Check the public_hoist_packages attribute for duplicates.
                 fail(msg)
 
 ################################################################################
-def _gather_package_content_excludes(keyed_lists, *names):
+def _gather_package_content_excludes(config, *names):
+    # Handle empty or default configuration
+    if not config:
+        return None
+
+    if type(config) != "dict":
```
jbedard: Delete? Or is this check needed?

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -383,7 +395,15 @@ ERROR: can not apply both `pnpm.patchedDependencies` and `npm_translate_lock(pat
         patches = [("@" if patch.startswith("//") else "") + patch for patch in patches]
 
         # gather exclude patterns
-        exclude_package_contents = _gather_package_content_excludes(attr.exclude_package_contents, name, friendly_name, unfriendly_name)
+        if exclude_package_contents_config != None:
+            # MODULE.bazel mode: use only tag-based configuration
+            exclude_package_contents = _gather_package_content_excludes(exclude_package_contents_config, name, friendly_name, unfriendly_name)
+        elif attr.bzlmod:
+            # MODULE.bazel mode with no tags: no exclusions
```
jbedard:
```
if attr.bzlmod:
           exclude_package_contents = _gather_package_content_excludes(exclude_package_contents_config, name, friendly_name, unfriendly_name)
else:
            exclude_package_contents = _gather_package_content_excludes(attr.exclude_package_contents, name, friendly_name, unfriendly_name)
```
?

npm/extensions.bzl
```diff
@@ -44,7 +48,29 @@ def _npm_extension_impl(module_ctx):
...
-def _npm_translate_lock_bzlmod(attr):
+def _build_exclude_package_contents_config(module_ctx):
+    """Build exclude_package_contents configuration from tags across all modules."""
+    exclusions = {}
+
+    for mod in module_ctx.modules:
+        for exclude_tag in mod.tags.npm_exclude_package_contents:
+            # Process the package in the tag
+            package = exclude_tag.package
+            if package not in exclusions:
```
jbedard:
```suggestion
            if package in exclusions:
                fail("Duplicate exclude_package_contents tag for package: {}".format(package))
                
            exclusions = []
```

PR #2137, opened 2025-03-19
fix: `js_run_binary` doesn't accept `select()`

Aghassi:
I have a situation internally where I need to pass a different environment variable based on CI or remote development. Right now, `select` is not a `dict` so this fails. This is not an issue with skylibs `run_binary` or `bazel-lib`'s `run_binary`. By flipping this to a `|` merger, the issue goes away.

---

### Test plan

- New test cases added


jbedard:
@Aghassi this is great!

Can we get a 1 line test added somewhere though? Just a `select()` statement added to 1 BUILD file with a comment explaining what it's testing would be more then enough 👍

jbedard (approved):

jbedard (approved):

PR #2142, opened 2025-03-21
fix: use exec nodejs for splitter

thesayyn:
Fixes an issue where target node is used for the splitter action.

### Test plan

- Manual testing; please provide instructions so we can reproduce: @jbedard has the reproducer handy. I'll defer to him for confirmation.


js/private/js_image_layer.bzl
```diff
@@ -399,7 +400,6 @@ def _run_splitter(ctx, runfiles_dir, files, entries_json, layer_groups):
         executable = nodeinfo.node,
         progress_message = "Computing Layer Groups %{label}",
         mnemonic = "JsImageLayerGroups",
-        toolchain = "@rules_nodejs//nodejs:toolchain_type",
```
thesayyn, line 402: i am so mad at bazel for half-assing everything about toolchains.

jbedard (approved):

PR #2149, opened 2025-03-26
fix: bzlmod invalid lockfile entry error

jbedard:
There have been discussions about ["invalid lockfile entry error"](https://bazelbuild.slack.com/archives/C014RARENH0/p1742987773306169) that this may resolve.

---

### Test plan


alexeagle (approved):
The reporter verified the fix, let's land it even though we don't have a repro.

Issue #1126, opened 2023-06-28
[Bug]: next.js js_image_layer targets fails symlink identification for `.next`


archen:
### What happened?

This comment about accurately resolving symlinked directories still appears to bite at least `next.js` builds with js_image_layer https://github.com/aspect-build/rules_js/blob/e8278a09a8aabe4c170d85ba39245586cf2643c2/js/private/js_image_layer.mjs#L9590

truncated error output

```bash
Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: EISDIR: illegal operation on a directory, read
Emitted 'error' event on ReadStream instance at:
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -21,
  code: 'EISDIR',
  syscall: 'read'
}

Node.js v18.12.0
```

I can see in my `bin/image_arm64_layers_entries.json` that the `.next` path is still not being recognized as a directory. 
```json
...

  "/app/server/nextjs_server.runfiles/myworkspace/.next": {
    "dest": "bazel-out/darwin_arm64-fastbuild-ST-c7efbb1f8678/bin/.next",
    "is_directory": false,
    "is_source": false,
    "root": "bazel-out/darwin_arm64-fastbuild-ST-c7efbb1f8678/bin"
  }
...
```

I'm working around the issue by using a patched version of the "legacy" `js_image_layer` target that used the python tar generator, here: https://gist.github.com/archen/a8c02583861d138206051bb7f4298ff4

### Version

Development (host) and target OS/architectures: `MacOS 13.4.1 Apple Silicon`

Output of `bazel --version`: `bazel 6.2.0`

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
```
  http_archive(
      name = "aspect_rules_js",
      sha256 = "2a88d837f8fb7bfe46b1d9f413df9a777ec2973e1f812929b597c1971a3a1da5",
      strip_prefix = "rules_js-1.28.0",
      url = "https://github.com/aspect-build/rules_js/archive/refs/tags/v1.28.0.tar.gz",
  )
```

Language(s) and/or frameworks involved: `next.js 13.0.5`

### How to reproduce

[...]

Issue #1292, opened 2023-09-26
[FR]: Consider hosting framework-specific "contrib"


alexeagle:
### What is the current behavior?

Users are expected to do a lot of "last-mile" work to adapt the instructions from their framework to be viable under Bazel.

### Describe the feature

We could provide simple macros, imagine `contrib/vue.bzl`, `contrib/react.bzl`, etc. Each would be taking something like 100-300 lines of code from one of our examples and making it shared with users.

alexeagle:
A bunch of requests for help with Playwright recently, which would be another case for this

stevenpollack:
FYI came here from https://github.com/bazelbuild/examples/blob/main/frontend/next.js/defs.bzl -- I would love to see that functionality up-streamed.

stevenpollack:
Sadly, I'm a contractor and I don't have the budget to donate to OSS. I
understand where you're coming from, though.
-----

jbedard:
We've decided to add a contrib folder within rules_js to put things like framework specific macros with [nextjs](https://github.com/aspect-build/rules_js/commit/48622239739561b1634dc226013d2254da3faeee) being the first.

Additional frameworks can have separate github issues/discussions this one will be closed now.

PR #2150, opened 2025-04-02
feat: add contrib/nextjs rules for building and running nextjs applications

jbedard:
Ref #1292
Close https://github.com/aspect-build/rules_js/issues/1126

---

Provide `@aspect_rules_js//contrib/nextjs:defs.bzl` for convenience macros for NextJs applications built using `next`.

Primary API:
* `next()` wrapper of the 3 primary `build`, `start` and `dev` macros - aligning with package.json bin entries created by `create-next-app`
* `nextjs_build()` - the npm `next build`
* `nextjs_start()` - the npm `next start` (where the `build` is a dependency)
* `nextjs_dev()` - the npm `next dev`

Standalone server API, where custom pnpm+bazel logic is required:
* `nextjs_standalone_build()` is the standard `next build` but with some workarounds or `output: standalone`
* `nextjs_standalone_server()` follows the [standalone mode](https://nextjs.org/docs/app/api-reference/config/next-config-js/output#automatically-copying-traced-files) instructions for rearranging the directory structure, then defines a `js_binary` to run the `server.js` within that directory

### Test plan

- New test cases added


.bazelignore
```diff
@@ -7,6 +7,7 @@ examples/linked_empty_node_modules/node_modules
 examples/linked_lib/node_modules
 examples/linked_pkg/node_modules
 examples/macro/node_modules/
+examples/nextjs/node_modules/
```
alexeagle, line 10: we could get rid of this file and use `REPO.bazel` instead, it finally has wildcards for ignores

contrib/nextjs/defs.bzl
```diff
@@ -0,0 +1,317 @@
...
+            "{}/static".format(_next_build_out): "standalone/{}/{}/static".format(pkg, _next_build_out),
+            "public": "standalone/{}/public".format(pkg),
+        },
+        out = standalone_outdir,
+        visibility = ["//visibility:private"],
+        tags = ["manual"],
+    )
+
+# ---------------------------------------------------------------------------------------------
+# Next.js image layer - an optimal image layer for Next.js standalone applications.
```
alexeagle, line 248: nit: needs to mention container images - many readers won't connect what a "layer" is without some context

jbedard: I'm dropping the image layer stuff from here for now. This is more generic "exclude packages from my image" then it is nextjs specific imo.

.bazelignore
jbedard: We'll need to think about the bazelignore check and if we will remove that for v3 or support REPO.bazel instead etc.

Izzy-CodersPlanet (approved):
Looks good! ✅

jbedard:
@Izzy-CodersPlanet want to try it out? Have any feedback?

jbedard:
FYI @domenic-donato

alexeagle (approved):
Nice!

domenic-donato:
Awesome 🚀

PR #2153, opened 2025-04-03
fix: accept empty, relative or absolute js_image_layer(root) paths

jbedard:
### Test plan

- New test cases added


js/private/js_image_layer.bzl
```diff
@@ -607,6 +607,7 @@ js_image_layer_lib = struct(
         ),
         "root": attr.string(
             doc = "Path where the files from js_binary will reside in. eg: /apps/app1 or /app",
+            default = "/",
```
thesayyn, line 610:
```suggestion
```

we should probably revert this line even though result will be the same. just to adhere to semver

thesayyn (approved):

PR #2157, opened 2025-04-06
perf: make better use of macros in generated defs.bzl files

dzbarsky:


e2e/gyp_no_install_script/snapshots/bzlmod/segfault-handler_defs.bzl
```diff
@@ -4,441 +4,280 @@
...
-        output_group = "package_directory",
-        visibility = ["//visibility:public"],
-        tags = ["manual"],
+    return _npm_link_imported_package_store(
+        name,
+        package = PACKAGE,
+        version = VERSION,
+        root_package = _ROOT_PACKAGE,
+        link_packages = {
+            "": ["segfault-handler"],
```
jbedard, line 255: This string is in a constant up top, is that easy to use here?

e2e/gyp_no_install_script/snapshots/bzlmod/segfault-handler_defs.bzl
```diff
@@ -4,441 +4,280 @@
...
+        },
+        link_visibility = ["//visibility:public"],
+        bins = {},
+        link = None,
+        package_store_name = "segfault-handler@1.3.0",
```
jbedard, line 260: This string is probably also repeated a few places? Easy to use a constant here?

e2e/gyp_no_install_script/snapshots/bzlmod/segfault-handler_defs.bzl
dzbarsky: I can try, I was a bit intimidated by how the data flows through here until the very end (though I also think it's simpler now 😛). I'll poke at it

e2e/gyp_no_install_script/snapshots/bzlmod/segfault-handler_defs.bzl
```diff
@@ -4,441 +4,280 @@
...
-        tags = ["manual"],
+    return _npm_link_imported_package_store(
+        name,
+        package = PACKAGE,
+        version = VERSION,
+        root_package = _ROOT_PACKAGE,
+        link_packages = {
+            "": ["segfault-handler"],
+        },
+        link_visibility = ["//visibility:public"],
```
jbedard, line 257: I wonder about things like this as well. Are we constructing both an array and string here 100% of the time?

jbedard: Can't some things like this be put into `_npm_link_imported_package_store` defaults and then just drop this line? Same with other params maybe?

npm/private/npm_import.bzl
```diff
@@ -39,208 +42,286 @@ _LINK_JS_PACKAGE_LOADS_TMPL = """\
...
+        msg = "name must end with '/{package}' when linking the store in package '{package}'; recommended value is 'node_modules/{package}'".format(
+            package = package,
+        )
         fail(msg)
-    link_root_name = name[:-len("/{package}")]
+    link_root_name = name[:-len(suffix)]
 
-    deps = {deps}
-    ref_deps = {ref_deps}
+    deps = {k.format(link_root_name = link_root_name): v for k, v in deps.items()}
```
jbedard, line 114: It would probably be good to avoid `.items()` create giant new arrays?

e2e/gyp_no_install_script/snapshots/bzlmod/segfault-handler_defs.bzl
dzbarsky: ok i did it and pushed it as a separate commit on top of this, it was ok-ish, except the serialization code is kinda annoying to hack around. wdyt?

e2e/gyp_no_install_script/snapshots/bzlmod/segfault-handler_defs.bzl
dzbarsky: mmm for some reason I thought it wasn't repeated anywhere, but looks like it is, lets fix

e2e/gyp_no_install_script/snapshots/bzlmod/segfault-handler_defs.bzl
dzbarsky: yeah i was considering that, it would require a bit more changes to how we instantiate the templates but I agree it would be nice to be able to use better defaults. want me to do it here or a followup?

npm/private/npm_import.bzl
dzbarsky: i'd love your opinion here. the problem is that `link_root_name` is not known at the time these dicts are constructed, only later. is it possible to compute `link_root_name` in the module extension and fully materialize these?

alternately, we could defer this expansion until the point where the `deps` are actually used, but i wasn't sure if there was a nice single point for that or not. 

I figured you know this better, wdyt?

dzbarsky: my preference would be to fully materialize at extension time if that's possible

e2e/gyp_no_install_script/snapshots/bzlmod/segfault-handler_defs.bzl
jbedard: Probably a followup?

npm/private/npm_import.bzl
jbedard: Ideally it could all be avoided until later.

Is this where dropping the `link_root_name` param would help?

dzbarsky: Ok I will try to trace it and see if we can defer, but yeah if we can drop lonk_root_name this becomes a non-issue

e2e/gyp_no_install_script/snapshots/bzlmod/segfault-handler_defs.bzl
dzbarsky: ya lets do followup

npm/private/npm_import.bzl
dzbarsky: @jbedard I think it's not possible - the `deps` are passed to `_npm_package_store_impl` at which point they need to be a `label_keyed_string_dict` so we must construct the label names before instantiating the rule. But dropping `link_root_name` would certainly make this smoother

jbedard (approved):

jbedard (approved):

PR #2168, opened 2025-04-14
devserver: retry all file I/O with exponential backoff

dmiller-figma:
We were observing errors such as the following pretty regularly:

```
Error details: ENOENT: no such file or directory, open '/home/ubuntu/.cache/bazel/_bazel_ubuntu/d2c52710e031f37347cde534e5ba90f7/execroot/_main/bazel-out/k8-fastbuild/bin/web/bundler/server_/server.runfiles/_main/web/js/...
```

Bazel unlinks files from the output base at any time, and so the devserver can't rely on files that existed in the output base when the build initially finished to continue existing while it performs its sync operation.

This patch adds retries with exponential backoff to all file I/O operations. Prior to this patch if I edited a file in a loop it would take only about 1-2 minues to encounter an `ENOENT` error. With this change I have edited a file in a loop for 1h30m without any errors.

---

### Test plan

- Existing unit test coverage
- Manual stress testing


js/private/js_run_devserver.mjs
```diff
@@ -198,7 +223,13 @@ async function syncRecursive(src, dst, sandbox, writePerm) {
             ).reduce((s, t) => s + t, 0)
         } else {
             const lastChecksum = syncedChecksum.get(src)
-            const checksum = await generateChecksum(src)
+            let checksum
+            try {
+                checksum = await generateChecksum(src)
+            } catch (e) {
+                console.error(`Error generating checksum for ${src}: ${e.message}`)
```
jbedard, line 230: Didn't `generateChecksum` already output an error?

js/private/js_run_devserver.mjs
```diff
@@ -65,19 +65,41 @@ export function is1pPackageStoreDep(p) {
     return p.match(unscoped1p) || p.match(scoped1p)
 }
 
+// Utility function to retry an async operation with exponential backoff
```
jbedard, line 68: It doesn't look exponential. Should it be? I worry exponential across multiple `withRetry` calls (on all files etc) could get quite delayed?

dmiller-figma: Oops, yes sorry. I'm so used to typing "exponential" before "backoff"

js/private/js_run_devserver.mjs
```diff
@@ -226,7 +251,16 @@ async function syncRecursive(src, dst, sandbox, writePerm) {
                 mkdirpSync(path.dirname(dst))
             }
 
-            await fs.promises.copyFile(src, dst)
+            try {
+                await withRetry(
+                    () => fs.promises.copyFile(src, dst),
+                    `copyFile from ${src} to ${dst}`
+                )
+            } catch (e) {
```
dmiller-figma, line 259: Should I remove these logs as well?

jbedard: I think the `console.log` in `withRetry` is enough, especially with the `description` in there.

js/private/js_run_devserver.mjs
```diff
@@ -243,6 +272,7 @@ async function syncRecursive(src, dst, sandbox, writePerm) {
         }
     } catch (e) {
         console.error(e)
+        console.trace()
```
jbedard, line 275: What does this output?

jbedard (approved):

js/private/js_run_devserver.mjs
dmiller-figma: Ah yes, without this we were having a hard time telling where errors were coming from because the exception was swallowed, logged and then program exited. So in our logs we would just see 

```
...
ENOENT, open /whatever
Process died
```

[`console.trace()`](https://developer.mozilla.org/en-US/docs/Web/API/console/trace_static) logs a stack trace alongside this error so you can see which line of code originally threw the exception

PR #2172, opened 2025-04-17
refactor: simplify npm_*_package_store macro params

jbedard:
### Test plan


npm/private/npm_translate_lock_generate.bzl
```diff
@@ -271,10 +271,7 @@ def npm_link_all_packages(name = "node_modules", imported_links = []):
                 ),
             )
 
-        stores_bzl.append("""        store_{i}(name = "{{}}/{pkg}".format(name))""".format(
```
jbedard, line 274: The store macro only used this `name` to extract the `/{pkg}` and determine the package name.

npm/private/npm_import.bzl
```diff
@@ -186,25 +182,7 @@ _LINK_JS_PACKAGE_LIFECYCLE_TMPL = """\
 _LINK_JS_PACKAGE_LINK_IMPORTED_STORE_TMPL = """\
 # Generated npm_package_store and npm_link_package_store targets for npm package {package}@{version}
 # buildifier: disable=function-docstring
-def npm_link_imported_package_store(name):
```
jbedard, line 189: This macro is generated for each package, then invoked for each instance of that package. Then each invocation inspected the shared data to find which invocation it was.

Instead each invocation can just pass the data required.

alexeagle (approved):
I'd really like @gregmagolan to take a quick look here, but I can also trust your reasoning that this isn't a behavior change

dzbarsky (approved):
yeah this seemed overcomplicated when I was porting it to use the less memory-intensive macros, glad I'm not crazy :)

PR #2176, opened 2025-04-20
refactor: order generated packages in npm_translate_lock

jbedard:
To reduce make the generated bzl more deterministic.

### Test plan


npm/private/npm_translate_lock_generate.bzl
```diff
@@ -98,7 +98,7 @@ def generate_repository_files(rctx, pnpm_lock_label, importers, packages, patche
 
     npm_imports = helpers.get_npm_imports(importers, packages, patched_dependencies, only_built_dependencies, root_package, rctx.name, rctx.attr, rctx.attr.lifecycle_hooks, rctx.attr.lifecycle_hooks_execution_requirements, rctx.attr.lifecycle_hooks_use_default_shell_env, npm_registries, default_registry, npm_auth)
 
-    link_packages = [helpers.link_package(root_package, import_path) for import_path in importers.keys()]
+    link_packages = sorted([helpers.link_package(root_package, import_path) for import_path in importers.keys()])
```
dzbarsky, line 101: starlark spec says "Iteration yields the dictionary's keys in the order in which they were inserted; updating the value associated with an existing key does not affect the iteration order." so not sure this is necessary?

dzbarsky (approved):
i'm not sure of the full flow so its possible the non-determinism (was it observed or theoretical?) came from earlier. see comment below, `.keys()` and iteration should be deterministic

npm/private/npm_translate_lock_generate.bzl
jbedard: This did change the generated code, however maybe the `importers` should be sorted elsewhere? Or maybe it is deterministic (lockfile order) and there is no reason for this one. I might revert this line...

jbedard:
I'm mainly trying to ensure that refactoring all the pnpm logic doesn't unnecessary change the order of these fields. Maybe it's my other branch breaking the deterministic order and this isn't the root cause of it so I'm going to convert this to a draft for now...

jbedard:
I think you're right that we can depend on the insertion order. Most of the things that changed order in https://github.com/aspect-build/rules_js/pull/2177 are due to other reasons so I'll close this for now.

PR #2177, opened 2025-04-21
refactor: align pnpm package store with lockfile

jbedard:
---

### Changes are visible to end-users: yes/no

- Searched for relevant documentation and updated as needed: yes/no
- Breaking change (forces users to change their own code or config): yes/no
- Suggested release notes appear below: yes/no

### Test plan

- Covered by existing test cases
- New test cases added
- Manual testing; please provide instructions so we can reproduce:

PR #2183, opened 2025-04-22
test: aliases to local link: packages

jbedard:
### Test plan

- New test cases added

### DIFF
--- a/npm/private/test/snapshots/npm_defs.bzl
+++ b/npm/private/test/snapshots/npm_defs.bzl
@@ -2656,6 +2656,37 @@ def npm_link_all_packages(name = "node_modules", imported_links = []):
 
     if is_root:
         _npm_package_store(
+            name = ".aspect_rules_js/{}/js_lib_pkg_a-alias@0.0.0".format(name),
+            src = "//examples/js_lib_pkg/a:pkg",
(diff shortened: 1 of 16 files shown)


npm/private/test/snapshots/npm_defs.bzl
```diff
@@ -2654,6 +2654,37 @@ def npm_link_all_packages(name = "node_modules", imported_links = []):
         )
         link_targets.append(":{}/js_lib_pkg_a".format(name))
 
+    if is_root:
+        _npm_package_store(
+            name = ".aspect_rules_js/{}/js_lib_pkg_a-alias@0.0.0".format(name),
```
jbedard, line 2659: Notice how it is adding an additional `npm_package_store` even though we only want a link to a package that already has a store entry.

This is something fixed in https://github.com/aspect-build/rules_js/pull/2177 but I want to add now and only see the diff in https://github.com/aspect-build/rules_js/pull/2177.

jbedard: That functionality might actually be https://pnpm.io/package_json#dependenciesmetainjected, and rules_js just has a different default then pnpm where we always create a store entry for `workspace:`/`link:` deps...

PR #2185, opened 2025-04-22
refactor: normalize npm file/link: paths at parse time

jbedard:
To move more of the path/data normalization into the lockfile parsing instead of being spread throughout the repo.

---

### Test plan

### DIFF
--- a/npm/private/pnpm.bzl
+++ b/npm/private/pnpm.bzl
@@ -1,5 +1,6 @@
 "Pnpm lockfile parsing and conversion to rules_js format."
 
+load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_skylib//lib:types.bzl", "types")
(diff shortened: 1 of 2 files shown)


npm/private/pnpm.bzl
```diff
@@ -145,6 +146,10 @@ def _convert_v5_importer_dependency_map(specifiers, deps):
             alias, version = _strip_v5_v6_default_registry(version).lstrip("/").rsplit("/", 1)
             version = _convert_pnpm_v5_version_peer_dep(version)
             version = "npm:{}@{}".format(alias, version)
+        elif version.startswith("link:"):
+            version = version[:5] + paths.normalize(paths.join(import_path, version[5:]))
```
dzbarsky, line 150: bit inconsistent that we hardcode `5` here but `len("link:")` in the file above :)

jbedard: It is inconsistent all over the place atm, maybe that can be cleaned up another time.

PR #2189, opened 2025-04-23
refactor: reduce use of pnpm <9 lockfile properties

jbedard:
The lockfile `id` property is removed with v9. Our only use of it is to resolve `file:` paths in some scenarios however the `resolution.directory` always has the same information available across all lockfile versions.

The `resolution.directory` property also already has the `file:` prefix removed and `resolution.type == "directory"` is a more reliable way of checking for references to directories.

---

### Test plan

### DIFF
--- a/npm/private/npm_import.bzl
+++ b/npm/private/npm_import.bzl
@@ -667,12 +667,7 @@ def _npm_import_links_rule_impl(rctx):
                 deps[dep_store_target].append(dep_name)
     else:
         for (dep_name, dep_version) in rctx.attr.deps.items():
-            if dep_version.startswith("link:") or dep_version.startswith("file:"):
-                dep_store_target = """"//{root_package}:{package_store_root}/{{}}/{package_store_name}".format(link_root_name)"""
-            else:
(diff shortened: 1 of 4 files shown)


npm/private/npm_import.bzl
```diff
@@ -667,10 +667,7 @@ def _npm_import_links_rule_impl(rctx):
                 deps[dep_store_target].append(dep_name)
     else:
         for (dep_name, dep_version) in rctx.attr.deps.items():
-            if dep_version.startswith("link:") or dep_version.startswith("file:"):
-                dep_store_target = """"//{root_package}:{package_store_root}/{{}}/{package_store_name}".format(link_root_name)"""
-            else:
-                dep_store_target = """":{package_store_root}/{{}}/{package_store_name}".format(link_root_name)"""
+            dep_store_target = """":{package_store_root}/{{}}/{package_store_name}".format(link_root_name)"""
```
dzbarsky, line 670: Merge with the next line and drop the roor_package substitution unless I'm missing something?

jbedard: You're correct, unless I'm missing something 👍

PR #2191, opened 2025-04-24
refactor: create /ref and /pkg targets for local package store entries

jbedard:
Reduce the inspecting of packages and simply assuming all packages (imported and local) have store implementations for the regular + `/ref` + `/pkg` targets.

Create the `npm_local_package_store_internal` wrapper macro instead of expanding the template further.

---

### Test plan


dzbarsky (approved):
Cool makes sense

PR #2207, opened 2025-04-29
test: separate optional + lifecycle lockfile test packages

jbedard:
To ensure there is a test with the lockfile `option: true` property indicating it is **only** used as an optional dep.

### Test plan

- New test cases added


e2e/pnpm_lockfiles/v90/pnpm-lock.yaml
```diff
@@ -501,6 +508,9 @@ snapshots:
 
   '@aspect-test/e@1.0.0': {}
 
+  '@aspect-test/h@1.0.0':
+    optional: true
```
jbedard, line 512: This was the main goal, just a package within `e2e/pnpm_lockfiles` that has this flag. Where `@aspect-test/c` was also a non-optional dep so this flag wasn't set.

Issue #2212, opened 2025-05-01
[Bug]: js_image_layer, too aggressive splitter cache


ValeraS:
### What happened?

Splitter creates mtree files for layers, but adds all entries that are not symlinks to unused inputs. Because of this, if an entry is a folder and files in it have changed, bazel will not run splitter and will get stale results (mtree files) from the cache.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: 7.5.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: Module.bazel

Language(s) and/or frameworks involved: js

### How to reproduce

```shell

```

misty-fungus:
Splitter adds entries to mtree and instantly reports that it's output does not depend on these files:

https://github.com/aspect-build/rules_js/blob/ba38f3afb2a6f858b6b364dda05d5209bc905118/js/private/js_image_layer.mjs#L221-L237

This makes bazel consider that mtree files do not depend on src files:

https://github.com/aspect-build/rules_js/blob/8987d08adf24046e2d4fd9171b0c3ac3b7b2e6e6/js/private/js_image_layer.bzl#L415-L423

That's why upon source updates bazel won't re-run splitter and consequential `JsImageLayer` action will fail on invalid mtrees 


This dirty patch fixes build:

```diff

--- a/js/private/js_image_layer.bzl
+++ b/js/private/js_image_layer.bzl
@@ -415,7 +415,7 @@ else {
     ctx.actions.run(
         inputs = inputs,
         arguments = [splitter.path],
-        unused_inputs_list = unused_inputs,
+        #unused_inputs_list = unused_inputs,
         outputs = splitter_outputs,
         executable = node_exec,
         progress_message = "Computing Layer Groups %{label}",
```

----

In general I do not understand the logic of this functionality. 

1. Relying on inode type (symlink or dir) is questionable. AFAIK bazel does not guarantee where a dep will be brought to sandbox via copy/symlink or hardlink. Not to mention remote execution.
2. Splitter should rely on all user-provided it's inputs. It's up to user to provide correct depsets rather and trying to guess inside action.

thesayyn:
> In general I do not understand the logic of this functionality.

> Relying on inode type (symlink or dir) is questionable. AFAIK bazel does not guarantee where a dep will be brought to sandbox via copy/symlink or hardlink. Not to mention remote execution.

[...]

Issue #2215, opened 2025-05-12
[Bug]: `js_image_layer` error reading aspect_rules_js_metadata.json


tbarrella:
### What happened?

Hello,

I'm encountering an issue with `js_image_layer` + pnpm v8/v9 related to `pnpm.onlyBuiltDependencies`:

> ERROR: .../app/BUILD.bazel:85:16: JsImageLayer package_store_3p //app:layer failed: (Exit 1): tar failed: error executing JsImageLayer command (from target //app:layer) external/aspect_bazel_lib++toolchains+bsd_tar_darwin_arm64/tar --create --file bazel-out/darwin_arm64-fastbuild/bin/app/layer_package_store_3p.tar.gz --gzip '--options=gzip:!timestamp' ... (remaining 1 argument skipped)
> ...
> tar: Error reading archive bazel-out/darwin_arm64-fastbuild/bin/app/layer_package_store_3p.mtree: Can't open bazel-out/darwin_arm64-fastbuild/bin/node_modules/.aspect_rules_js/agent-base@7.1.3/node_modules/agent-base/aspect_rules_js_metadata.json

### Version

Development (host) and target OS/architectures: macos arm64

Output of `bazel --version`: 8.2.1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: 2.3.5

Language(s) and/or frameworks involved: JS

### How to reproduce

```shell
Roughly,
1. Build a `js_image_layer` target
2. Modify package.json to remove one of the layer's dependencies from `onlyBuiltDependencies`
3. Rebuild the `js_image_layer` target
```

### Any other information?

Step 1 produces a file declared [here](https://github.com/aspect-build/rules_js/blob/v2.3.7/js/private/js_image_layer.bzl#L334)
```
$(bazel info bazel-bin)/app/layer_package_store_3p.mtree
```
including the line
> ./app/app.runfiles/_main/node_modules/.aspect_rules_js/agent-base@7.1.3/node_modules/agent-base/aspect_rules_js_metadata.json uid=0 gid=0 time=0 mode=0555 type=file content=bazel-out/darwin_arm64-fastbuild/bin/node_modules/.aspect_rules_js/agent-base@7.1.3/node_modules/agent-base/aspect_rules_js_metadata.json

[...]

PR #2217, opened 2025-05-13
fix: correctly track treeartifact contents

thesayyn:
Fixes https://github.com/aspect-build/rules_js/issues/2212, https://github.com/aspect-build/rules_js/issues/2215

Fixes js_image_layer to correctly track treeartifact contents.

### Test plan

- Manual testing; please provide instructions so we can reproduce:  Try the repro in https://github.com/aspect-build/rules_js/issues/2212

Unfortunately its not trivial to add test for this case since it requires two step `bazel build` reproduce.

### DIFF
--- a/js/private/js_image_layer.bzl
+++ b/js/private/js_image_layer.bzl
@@ -26,7 +26,7 @@ _DEFAULT_LAYER_GROUPS = {
 
 _DOC = """Create container image layers from js_binary targets.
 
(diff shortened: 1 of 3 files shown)


alexeagle (approved):
Could you add a test?

thesayyn:
> Could you add a test?

> Unfortunately its not trivial to add test for this case since it requires two step bazel build reproduce.

I'd like to but it requires two step bazel build to reproduce.

PR #2220, opened 2025-05-14
chore: upgrade default pnpm to v9

alexeagle:
When setting up a new Angular 20 project, @alan-agius4 and I discovered that pnpm8 has a bug. New projects should be using the right default.

BREAKING CHANGE:
Users need to add pnpm: {onlyBuiltDependencies: []} to their package.json
or downgrade back to version 8


jbedard:
Replaced with https://github.com/aspect-build/rules_js/pull/2456

PR #2252, opened 2025-06-26
perf: reduce memory allocation in fs patches

jbedard:
While working in a large repo these made minor differences reducing memory usage + GC.

The first commit is just giving fs-patch callbacks better names to make profiling easier.

### Test plan

- Manual testing; running a `js_run_devserver` in a large repo


js/private/node-patches/fs.cjs
```diff
@@ -494,7 +494,9 @@ function patcher(fs = _fs, roots) {
                 return cb(undefined);
             }
             if (link !== HOP_NON_LINK) {
-                link = path.join(link, nested);
+                if (nested) {
+                    link = link + path.sep + nested;
```
thesayyn, line 498: could this break? does path.join() normalize `..` or `.` in the path segments?

js/private/node-patches/src/fs.cts
```diff
@@ -575,7 +577,7 @@ export function patcher(fs: any = _fs, roots: string[]) {
                 // not a link
                 return cb(escapedHop)
             }
-            nested = path.join(path.basename(maybe), nested)
+            nested = path.basename(maybe) + (nested ? path.sep + nested : '')
```
thesayyn, line 580: maybe an if else condition to make it more readable?

js/private/node-patches/src/fs.cts
```diff
@@ -683,7 +687,7 @@ export function patcher(fs: any = _fs, roots: string[]) {
                 return escapedHop
             }
 
-            nested = path.join(path.basename(maybe), nested)
+            nested = path.basename(maybe) + (nested ? path.sep + nested : '')
```
thesayyn, line 690: same as above

thesayyn (approved):

js/private/node-patches/fs.cjs
jbedard: `path.join()` does invoke `path.normalize()` at the end, but these paths are already normalized so when simply appending a single directory to the end we really don't have to do it again...

PR #2255, opened 2025-06-26
perf: avoid use of .to_list() in js_run_devserver data files list

jbedard:
### Changes are visible to end-users: no

Avoid use of `depset().to_list()` when computing data files for `js_run_devserver`.

### Test plan

- Manual testing; a diff of the data files list before+after in various repos


js/private/js_run_devserver.bzl
```diff
@@ -21,14 +21,33 @@ _attrs = dicts.add(js_binary_lib.attrs, {
...
+        package_name_segment = path_segments.index(".aspect_rules_js") + 1
+
+        # TODO: @0.0.0 is by default the version of all 1p linked packages, however, it can be overridden by users
+        # if they are manually linking a 1p package and not using workspace. A more robust solution would be to
+        # split handling of 1p and 3p package in the JsInfo provider itself. Other optimizations in the rule set
+        # could also be made if that was the case.
+        if not (len(path_segments) > package_name_segment and "@0.0.0" in path_segments[package_name_segment]):
+            return None
+
+    return json.encode(short_path)
```
dzbarsky: Curious if this is faster than '"%s"' % short_path or the simplest concat?

jbedard: I'm not sure, I was copying `js_image_layer`...

js/private/js_run_devserver.bzl
```diff
@@ -51,30 +70,20 @@ def _js_run_devserver_impl(ctx):
...
-            # could also be made if that was the case.
-            if len(path_segments) > package_name_segment and "@0.0.0" in path_segments[package_name_segment]:
-                # include this first party linked dependency
-                data_files.append(f)
-        else:
-            data_files.append(f)
+    # Build the list of data files to copy to the custom sandbox using ctx.actions.args()
+    entries = ctx.actions.args()
+    entries.set_param_file_format("multiline")
+    entries.add("[")
```
dzbarsky, line 74: Why not just do a newline-separats list? Do we really worry about newlines in filenames?

dzbarsky (approved):

js/private/js_run_devserver.bzl
jbedard: Good question. Mainly because json is easy though, and I'm copying `js_image_layer`.

However I'd like to explore streaming this file content instead of doing a full read + `JSON.parse` so we can start the ugly file-syncing without having to wait for this potentially-giant json file to reparse...

jbedard: I might try this later, for now this change requires hardly any change to the .mjs file so I'd like to stick with this for now...

PR #2256, opened 2025-06-27
perf: determine js_run_devserver file vs directory in rule

jbedard:
This removes a `lstat` call on almost all files being synced. Only those within directory artifacts will still be done.

### Test plan

- Manual testing; profile while using `aspect run --watch`

### DIFF
--- a/js/private/js_run_devserver.mjs
+++ b/js/private/js_run_devserver.mjs
@@ -176,17 +176,13 @@ async function syncSymlink(src, dst, sandbox, exists) {
     return 1
 }
 
-async function syncDirectory(src, dst, sandbox, exists, writePerm) {
+async function syncDirectory(src, dst, sandbox, writePerm) {
     if (process.env.JS_BINARY__LOG_DEBUG) {
         console.error(
             `Syncing directory ${src.slice(RUNFILES_ROOT.length + 1)}...`
         )
     }
     const contents = await fs.promises.readdir(src)
-    if (!exists) {
-        // Intentionally synchronous; see comment on mkdirpSync
-        mkdirpSync(dst)
-    }
     return (
         await Promise.all(
             contents.map(
@@ -264,7 +260,7 @@ async function syncRecursive(src, dst, sandbox, writePerm) {
         if (lstat.isSymbolicLink()) {
             return syncSymlink(src, dst, sandbox, exists)
         } else if (lstat.isDirectory()) {
-            return syncDirectory(src, dst, sandbox, exists, writePerm)
+            return syncDirectory(src, dst, sandbox, writePerm)
         } else {
             const lastChecksum = syncedChecksum.get(src)
             const checksum = await generateChecksum(src)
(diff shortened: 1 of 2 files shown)


js/private/js_run_devserver.mjs
```diff
@@ -382,10 +383,8 @@ async function syncFiles(files, sandbox, writePerm, doSync) {
 
     let totalSynced = (
         await Promise.all(
-            otherFiles.map(async (file) => {
-                const src = path.join(RUNFILES_ROOT, file)
-                const dst = path.join(sandbox, file)
-                return await doSync(src, dst, sandbox, writePerm)
+            otherFiles.map(async ([file, isDirectory]) => {
```
thesayyn, line 386:
```suggestion
            otherFiles.map(([file, isDirectory]) => {
```

js/private/js_run_devserver.mjs
```diff
@@ -382,10 +383,8 @@ async function syncFiles(files, sandbox, writePerm, doSync) {
...
+                return await doSync(file, isDirectory, sandbox, writePerm)
```
thesayyn, line 387:
```suggestion
                return doSync(file, isDirectory, sandbox, writePerm)
```

js/private/js_run_devserver.mjs
```diff
@@ -400,10 +399,8 @@ async function syncFiles(files, sandbox, writePerm, doSync) {
 
     totalSynced += (
         await Promise.all(
-            packageStore1pDeps.map(async (file) => {
-                const src = path.join(RUNFILES_ROOT, file)
-                const dst = path.join(sandbox, file)
-                return await doSync(src, dst, sandbox, writePerm)
+            packageStore1pDeps.map(async ([file, isDirectory]) => {
+                return await doSync(file, isDirectory, sandbox, writePerm)
```
thesayyn, line 403: same as above?

js/private/js_run_devserver.mjs
```diff
@@ -241,7 +235,10 @@ async function syncFile(src, dst, exists, lstat, writePerm) {
 // synced it is only re-copied if the file's last modified time has changed since the last time that
 // file was copied. Symlinks are not copied but instead a symlink is created under the destination
 // pointing to the source symlink.
-async function syncRecursive(src, dst, sandbox, writePerm) {
+async function syncRecursive(file, _, sandbox, writePerm) {
+    const src = path.join(RUNFILES_ROOT, file)
+    const dst = path.join(sandbox, file)
+
     try {
         const lstat = await withRetry(
```
thesayyn, line 243: do we still need this?

thesayyn (approved):

js/private/js_run_devserver.mjs
jbedard: Is there a reason you prefer that?

I think doing the async+await keeps stack traces nicer. I'm uncertain how it technically works though, and if adding the async+await causes additional promise instances and maybe extra overhead?

This PR didn't add these though so I'd be tempted to debate and maybe change that in a separate PR...

js/private/js_run_devserver.mjs
jbedard: We still need it for the `mtime` and `isSymbolicLink`. Note this code path is only for (a) ibazel protocol or (b) directory artifacts when using the new protocol

js/private/js_run_devserver.mjs
thesayyn: if it didn't add then lets keep it as is. its just going to be an extra promise.

PR #2257, opened 2025-06-27
perf: js_run_devserver

jbedard:
A lot of minor changes to the methods invoked extremely frequently, especially in the initial sync for targets with very large runfiles.

### Test plan

- Manual testing; run a devserver on a large target

### DIFF
--- a/js/private/js_run_devserver.mjs
+++ b/js/private/js_run_devserver.mjs
@@ -138,9 +138,7 @@ async function syncSymlink(src, dst, sandbox, exists) {
             linkPath = path.relative(src, linkPath)
         }
         // Special case for 1p node_modules symlinks
-        const maybe1pSync = path.normalize(
-            path.join(sandbox, srcWorkspacePath, linkPath)
-        )
+        const maybe1pSync = path.join(sandbox, srcWorkspacePath, linkPath)
         if (fs.existsSync(maybe1pSync)) {
             src = maybe1pSync
             symlinkMeta = '1p'
--- a/js/private/js_run_devserver.mjs
+++ b/js/private/js_run_devserver.mjs
@@ -188,7 +188,7 @@ async function syncDirectory(file, src, sandbox, writePerm) {
             contents.map(
                 async (entry) =>
                     await syncRecursive(
-                        path.join(file, entry),
+                        file + path.sep + entry,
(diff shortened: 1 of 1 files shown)


js/private/js_run_devserver.mjs
```diff
@@ -6,11 +6,18 @@ import * as child_process from 'node:child_process'
 import * as crypto from 'node:crypto'
 import * as readline from 'node:readline'
 
+// Environment constants
+const {
+    JS_BINARY__EXECROOT,
```
jbedard, line 11: Some of the `process.env.JS_BINARY__LOG_DEBUG` lines were actually showing up in the profiling, so may as well make them `const` 🤷

js/private/js_run_devserver.mjs
```diff
@@ -129,18 +136,15 @@ export function friendlyFileSize(bytes) {
     )
 }
 
-async function syncSymlink(src, dst, sandbox, exists) {
-    const srcWorkspacePath = src.slice(RUNFILES_ROOT.length + 1)
```
jbedard, line 133: Adding `file` (the runfiles-relative file path) avoids things like this `slice`, and we already had it on a few of the `sync*()` methods.

This specific line was invoked extremely often, other lines are rare/debug but make the code shorter.

js/private/js_run_devserver.mjs
```diff
@@ -236,49 +238,45 @@ async function syncFile(src, dst, exists, lstat, writePerm) {
...
+    const src = RUNFILES_ROOT + path.sep + file
+    const dst = sandbox + path.sep + file
 
     try {
         const lstat = await withRetry(
             () => fs.promises.lstat(src),
             `lstat for ${src}`
         )
-        const last = syncedTime.get(src)
+        const last = syncedTime.get(file)
```
jbedard, line 249: All the `Map`s (except the mkdir one) now have the runfiles-relative `file` instead of the absolute paths.

js/private/js_run_devserver.mjs
dzbarsky: Is it just the property access or were those actual getenv syscalls?

jbedard: Just property access

PR #2263, opened 2025-07-08
Feat ignore not compatible packages

Mivr:
Ignoring packages that are not compatible to the platform of the build

### Test plan

- New test cases added


jbedard:
Replaced with https://github.com/aspect-build/rules_js/pull/2538

PR #2270, opened 2025-07-10
fix: correct early exit from _parse_lockfile

lukaszwawrzyk:
`_parse_lockfile` is expected to return 5 values. If pnpm lockfile is invalid user gets an error about getting 4 instead of 5 expected values and nothing about actual error. Happy path worked ofc, but error path was broken.


jbedard:
It looks like some unit tests are asserting these error cases with the old incorrect number of args. Can you update the those tests?

Issue #2226, opened 2025-05-31
[FR]: Peer dependency resolution of first party packages


devversion:
### What is the current behavior?

(somewhat related to #1092)

Consider a layout like:

```diff
| package.json
| packages/
+    | compiler
      |   package.json
+    | compiler-cli
      |    package.json
```

The `compiler` and `compiler-cli` packages are part of the pnpm workspace!

`compiler-cli` has a peer dependency on a specific compiler version (set via packaging substitutions). e.g.

```js
// compiler-cli/package.json
{
  "peerDependencies": {
    "@angular/compiler": "0.0.0-PLACEHOLDER"
  }
}
```

To satisfy this peer dependency for local development, in some other folder of the project, where we rely on the `:pkg`-linked first party packages, we have a few options:

1. Rely on pnpm's auto-install peer dependency feature (enabled by default).
2. Specifically add a `dependency` to the `compiler-cli` `package.json`.
3. Leverage `.pnpmfile.cjs` to only insert a `dependency` (like 2) for Aspect/rules_js
4. Don't satisfy _officially_, but use a `js_library` that has a `data` dependency on the linked `compiler:pkg` then.

There are quite a few options, and that's good, emphasizing that there are tricks to achieve what is needed, but I believe it would make sense to also support `devDependencies` here (and was also highlighted in #1092).

Currently when we'd try to satisfy the peer dependency via `devDependency` in the compiler-cli package.json, it works with pnpm itself, but with `rules_js`, the compiler package is never picked up as **prod** dependency, and will never be propagated via the `npm_link_package_store` dependency chain (if I understand that correctly from a high-level analysis).

https://github.com/aspect-build/rules_js/blob/157f7553543036a72a318ec6147b11f8f09abd88/npm/private/npm_link_package_store.bzl#L131-L132

### Describe the feature

It would be great if there would be a way to control this behavior https://github.com/aspect-build/rules_js/blob/157f7553543036a72a318ec6147b11f8f09abd88/npm/private/npm_link_package_store.bzl#L131-L132, or if `rules_js` would pick up the "satisfied" compiler package as **prod** then.

PR #2273, opened 2025-07-11
fix: do not require `_lcov_merger` under coverage

### LINKED ISSUE #2229: [Bug]: _lcov_merger attribute check makes testonly js_binary targets fail under coverage
### What happened?

`js_binary` targets fail to build under coverage if they are marked `testonly`.

The culprit seems to be this piece of code:

https://github.com/aspect-build/rules_js/blob/157f7553543036a72a318ec6147b11f8f09abd88/js/private/js_binary.bzl#L577-L580

It seems to use `testonly` as a means to detect a test. In my understanding, this is not what testonly is for:

https://bazel.build/reference/be/common-definitions#common.testonly

### Version

Output of `bazel --version`: "bazel 8.2.1"

Language(s) and/or frameworks involved: None.

### How to reproduce

Take the following vanilla setup:

```bzl
# MODULE.bazel
bazel_dep(name = "aspect_rules_js", version = "2.3.7")

# BUILD.bazel
load("@aspect_rules_js//js:defs.bzl", "js_binary")

js_binary(
    name = "test",
    entry_point = "test.js",
    testonly = 1,
)

When running coverage, the `js_binary` target fails:

```sh
$ bazel coverage ...
WARNING: Build options --collect_code_coverage and --instrumentation_filter have changed, discarding analysis cache (this can be expensive, see https://bazel.build/advanced/performance/iteration-speed).
ERROR: /home/tos/dh/personal/hack/rules_js/BUILD.bazel:3:10: in js_binary rule //:test: 
	File "/home/tos/.cache/bazel/_bazel_tos/cefae97e122df1b427ef8483286d6423/external/aspect_rules_js+/js/private/js_binary.bzl", line 580, column 17, in _js_binary_impl
		fail("_lcov_merger attribute is missing and coverage was requested")
Error in fail: _lcov_merger attribute is missing and coverage was requested
ERROR: /home/tos/dh/personal/hack/rules_js/BUILD.bazel:3:10: Analysis of target '//:test' failed
ERROR: Analysis of target '//:test' failed; build aborted
INFO: Elapsed time: 0.856s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
ERROR: Build did NOT complete successfully
ERROR: No test targets were found, yet testing was requested
```

This is unexpected. I'd expect

```sh
$ bazel coverage ...
INFO: Analyzed target //:test (1 packages loaded, 2 targets configured).
INFO: Found 1 target and 0 test targets...
Target //:test up-to-date:
  bazel-bin/test_/test
INFO: Elapsed time: 0.137s, Critical Path: 0.00s
INFO: 3 processes: 5 action cache hit, 3 internal.
INFO: Build completed successfully, 3 total actions
ERROR: No test targets were found, yet testing was requested
```

(which is what happens if the `testonly = 1` flag is commented out).

### Any other information?

_No response_

gzm0:
Newer versions of bazel do not require it anymore: https://github.com/bazelbuild/bazel/issues/4033#issuecomment-2507162290

Removing the explicit failure:
- Fixes the false-positive failures when building `js_binary` `testonly = 1` targets under `coverage` (fixes #2229).
- Avoids that test rule authors targeting newer bazel versions only have to put a workaround attribute in place.

This change reduces discoverability of the _lcov_merger issue for new test rule authors. However, given that it is hopefully an issue of the past soon, I think this is acceptable.

---

### Test plan

Run `bazel coverage //a/testonly/js_binary:target` (see #2229 for a minimal example.

Before this PR: `Error in fail: _lcov_merger attribute is missing and coverage was requested`
After this PR: builds

### DIFF
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -577,6 +577,12 @@ def _js_binary_impl(ctx):
     if ctx.attr.testonly and ctx.configuration.coverage_enabled:
         # We have to propagate _lcov_merger runfiles since bazel does not treat _lcov_merger as a proper tool.
         # See: https://github.com/bazelbuild/bazel/issues/4033
+        # This is optional because:
+        # - We do not want to require it for js_binary targets
+        #   (but we cannot distinguish js_binary from js_test here, see #2229).
+        # - It is not required anymore on bazel 8
+        #   (https://github.com/bazelbuild/bazel/issues/4033#issuecomment-2507162290)
+        # TODO: Remove once bazel<8 support is dropped.
         if hasattr(ctx.attr, "_lcov_merger"):
             runfiles = runfiles.merge(ctx.attr._lcov_merger[DefaultInfo].default_runfiles)
         providers = [
@@ -649,6 +655,8 @@ See the Bazel [Test encyclopedia](https://bazel.build/reference/test-encyclopedi
 the contract between Bazel and a test runner.""",
     implementation = js_binary_lib.implementation,
     attrs = dict(js_binary_lib.attrs, **{
+        # TODO: Remove once bazel<8 support is dropped.
+        # See comment at usage site in the rule impl for more.
         "_lcov_merger": attr.label(
             executable = True,
             default = Label("//js/private/coverage:merger"),


js/private/js_binary.bzl
```diff
@@ -575,13 +575,10 @@ def _js_binary_impl(ctx):
...
 
     if ctx.attr.testonly and ctx.configuration.coverage_enabled:
-        # We have to instruct rule implementers to have this attribute present.
-        if not hasattr(ctx.attr, "_lcov_merger"):
-            fail("_lcov_merger attribute is missing and coverage was requested")
-
         # We have to propagate _lcov_merger runfiles since bazel does not treat _lcov_merger as a proper tool.
         # See: https://github.com/bazelbuild/bazel/issues/4033
-        runfiles = runfiles.merge(ctx.attr._lcov_merger[DefaultInfo].default_runfiles)
+        if hasattr(ctx.attr, "_lcov_merger"):
```
jbedard, line 586: Lets add a comment explaining why this is now optional, maybe with a link to whatever commit/issue explains it?

Will there be any reason for this in bazel8+? Should we add `# TODO: remove once bazel<8 support dropped`?

js/private/js_binary.bzl
```diff
@@ -575,13 +575,10 @@ def _js_binary_impl(ctx):
     providers = []
 
     if ctx.attr.testonly and ctx.configuration.coverage_enabled:
-        # We have to instruct rule implementers to have this attribute present.
-        if not hasattr(ctx.attr, "_lcov_merger"):
-            fail("_lcov_merger attribute is missing and coverage was requested")
```
jbedard, line 580: This issue will still exist in bazel <8, yet now we have no `fail()` informing users? I assume the fix is worth the lost error message but 🤔

js/private/js_binary.bzl
gzm0: Good point. I've added a commit that adds comments (also at the definition site of the attribute).

js/private/js_binary.bzl
gzm0: An alternative is that we only make this change on bazel >= 8 (I think I'd have to upstream a feature in bazel_features, but that seems not too hard).

The big downside is that we'd only fix #2226 on bazel >= 8.

My thinking was that at this point, fixing the bug is more important than providing guidance for old bazel version to rule implementers.

js/private/js_binary.bzl
jbedard: Great 👍

js/private/js_binary.bzl
jbedard: I don't think it's worth any extra code, especially if it prevents the actual fix for bazel <8

jbedard (approved):

Issue #1313, opened 2023-10-12
[FR]: support env_inherit attr on js_test


yaroslavvoloshchuk-codaio:
### What is the current behavior?

So far, we have only one way to pass the subset of environment variables to the js_test target - by using [--test_env=ENV_VAR](https://bazel.build/reference/command-line-reference#flag--test_env). However, those variables are defined globally for all test targets, so when they change, they invalidate the cache of all tests.

### Describe the feature

It would be great to add support for `env_inherit` described here: https://bazel.build/reference/be/common-definitions#common-attributes-tests. Then it would allow us to specify which env variables should be passed into tests per target level.

Implementation example: https://github.com/bazelbuild/rules_go/pull/3256/files#diff-2f4977ebad0d8fd012c0d532023ebe9202e795bf502029f7df454d5fc4f96656

WillMorrisonEnspi:
https://docs.aspect.build/rulesets/aspect_rules_js/docs/js_binary#js_test currently states "All [common test attributes](https://bazel.build/reference/be/common-definitions#common-attributes-tests) are supported", but without support for `env_inherit` that is not the case.

PR #2277, opened 2025-07-15
Add env_inherit to js_test

keith:
Fixes https://github.com/aspect-build/rules_js/issues/1313


js/private/js_binary.bzl
```diff
@@ -572,7 +576,11 @@ def _js_binary_impl(ctx):
     )
     runfiles = launcher.runfiles
 
-    providers = []
+    providers = [
+        RunEnvironmentInfo(
+            inherited_environment = getattr(ctx.attr, "env_inherit", []),
```
jbedard: Any reason not to just do `ctx.attr.env_inherit`?

keith: afaict this implementation is shared with js_binary and binary targets don't normally have env_inherit because they automatically inherit all available env with bazel run

jbedard: Ha, I should read which file I'm reviewing when I comment 😅 

I think that means we shouldn't append the `RunEnvironmentInfo` at all if the attribute doesn't exist. Can we change this to something such as:
```
if hasattr(ctx.attr, "env_inherit"):
    providers.append( ... )
```
?

keith: ah yea fair enough, done

jbedard (approved):

js/private/js_binary.bzl
```diff
@@ -129,6 +129,10 @@ _ATTRS = {
         and ["Make variable"](https://bazel.build/reference/be/make-variables) substitution if `expand_env` is set to True.
         """,
     ),
+    "env_inherit": attr.string_list(
```
jbedard: I think this should actually only be set for the `_test` version, while `env` seems to apply to both `*_binary` and `*_test` rules (see https://bazel.build/reference/be/common-definitions#common-attributes-binaries).

Can we add it down below in the `js_test = rule(...)` statement? Or if you want you can add a `_TEST_ATTRS = ` that merges `_ATTRS` plus a new set of additional `js_test` attributes.

keith: you're so right, fixed

jbedard:
One more thing... checkout the docs failure, or just do `cd docs && bazel test //...` and it will give you a `bazel run` snippet to update the docs.

keith:
done

jbedard:
Awesome, thanks! 👍

keith:
Thanks!

PR #2283, opened 2025-07-24
chore: switch to bazelrc-preset.bzl

jbedard:
### Test plan


alexeagle (approved):
Yay thanks for picking this one up

Issue #2077, opened 2025-01-25
[FR]: `js_binary` should provide `RunEnvironmentInfo`


jfirebaugh:
### Describe the feature

The implementation function of `js_binary` should return a [`RunEnvironmentInfo` provider](https://bazel.build/rules/lib/providers/RunEnvironmentInfo) instance. This allows downstream rule implementations to introspect the environment variables specified in the `env` attribute (and [eventually `args` as well](https://github.com/bazelbuild/bazel/pull/16430)).

PR #2303, opened 2025-08-11
feat: add action environment variables to RunEnvironmentInfo

Mivr:
### Test plan

- new tests added

Fixes: #2077

### DIFF
--- a/js/private/test/run_environment_info_test.bzl
+++ b/js/private/test/run_environment_info_test.bzl
@@ -93,7 +93,6 @@ def run_environment_info_test_suite(name):
             "LOCATION_VAR": "$(location :data.json)",
         },
         data = [":data.json"],
-        tags = ["manual"],
     )
 
     run_environment_info_test(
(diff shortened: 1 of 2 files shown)


js/private/test/run_environment_info_test.bzl
```diff
@@ -0,0 +1,169 @@
...
+        name = name + "_binary_no_env_subject",
+        entry_point = "test_env.js",
+        tags = ["manual"],
+    )
+
+    run_environment_info_test(
+        name = name + "_binary_no_env_test",
+        target_under_test = ":" + name + "_binary_no_env_subject",
+        expect_no_provider = True,
+        tags = ["manual"],
```
jbedard: Why make everything manual?

Mivr: Because otherwise it fails on the bazel coverage //... command

jbedard: Why? Why disable it instead of fixing it?

Mivr: Oh, just wasn't sure if it can be fixed in coverage. As the problem was that it could not find the provider type at all. I will look more into it and see if it can be fixed.

Mivr: I was not understanding the error ok, fixed

PR #2321, opened 2025-08-20
fix: indeterministic node-gyp makefile that is only used during install hook

### LINKED ISSUE #2302: [Bug]: Indeterministic npm_import packages with node-gyp
### What happened?

NPM Packages using node-gyp will always yield an indeterministic file `build/Makefile` in the resulting package after lifecycle hooks are run, with node versions that do not include this recent change https://github.com/nodejs/gyp-next/pull/293. 

Whereas node-gyp isn't straightforward in making all outputs deterministic, this common failure mode could be fixed by simply deleting this Makefile after the package install is completed. 

I'd be happy to submit a PR for this case, but wanted wanted to first clarify if this change would be accepted, and if the proper location would be in the node-gyp-bin shims or perhaps the bundle lifecycle-hooks script.

### Version

Development (host) and target OS/architectures: linux

Output of `bazel --version`: bazel 8.3.1

`WORKSPACE` or `MODULE.bazel` file: 2.3.8

Language(s) and/or frameworks involved: node 22.10.0

### How to reproduce

```shell
Add a npm_import for unix-dgram and see build/Makefile contain different results
```

### Any other information?

_No response_

joeljeske:
Fixes #2302 

Likely no, this file is an implementation of node-gyp that is highly unlikely to be used.

### Test plan

- Manual testing; please provide instructions so we can reproduce: https://github.com/aspect-build/rules_js/pull/2322


npm/private/lifecycle/lifecycle-hooks.js
```diff
@@ -124,6 +129,19 @@ async function runLifecycleHooks(opts, hooks) {
     }
 }
 
+// Remove the indeterministic Makefile from the resulting directory.
+// Can be removed when https://github.com/nodejs/gyp-next/pull/293
+// has sufficiently propagated.
```
jbedard: When will it be "sufficiently propagated" and how can we tell?

jbedard:
Awesome, thanks! 👍 

Can you sign the CLA? Otherwise LGTM 👍

npm/private/lifecycle/lifecycle-hooks.js
joeljeske: The relevant change is included in NPM package >=node-gyp@11.3.0. 

I would say that to be sufficiently propagated, we should see a very high percentage of downloads of this package be for this version and up: https://www.npmjs.com/package/node-gyp?activeTab=versions

By the looks of it, it may be a very long time if ever. For example, version 8.4.1 from ~4yrs ago still takes ~15% of total downloads.

Hopefully this isn't a particularly large burden to carry forward indefinitely 😅

joeljeske:
Done! Thanks, @jbedard

npm/private/lifecycle/lifecycle-hooks.js
jbedard: Can you add that version info to the comment? Just so some day when scanning this if someone notices node-gyp@11.3.0 and knows it's no longer supported then we can drop it...

PR #2322, opened 2025-08-20
repro: repro issue 2302

joeljeske:
Repro example case for # 2321

See README for instructions


joeljeske:
Closing, as it served only as repro steps.

PR #2325, opened 2025-08-21
Windows test fixes & enable tests in CI

peakschris:
Changes:
* Fix js_binary to run on windows without bash; works both in js_image_layer and via bazel
* Disable failing tests due to windows or --noenable_runfiles
* Enable windows CI
* Bump aspect_bazel_lib and tar.bzl to fix windows tar toolchain issues


peakschris:
@alexeagle could you take a look at / rerun the one failing job? Looks like an environment issue

peakschris:
Replacing with one that targets bazel-lib 3

PR #2328, opened 2025-08-23
Fix js_test nodejs toolchain exec_group

keith:
If you cross compile the js_test target, the nodejs toolchain needs to
be configured for the test exec group.

Depends on https://github.com/bazel-contrib/bazel_features/pull/112


keith:
can be replaced by https://github.com/aspect-build/rules_js/pull/2330 if it works

keith:
https://github.com/aspect-build/rules_js/pull/2330

PR #2336, opened 2025-08-26
refactor: ensure patched fs callbacks always invoked on error

jbedard:
### Test plan


jbedard:
I shouldn't merge such a thing until I can actually reproduce it in a test...

Issue #2275, opened 2025-07-13
[Bug]: Guidance for calling js_run_binary(chdir = package_name()) is incorrect in case the repo is external


EdSchouten:
### What happened?

There are many places in the tree where `js_run_binary()` is invoked with `chdir = package_name()`. It should be noticed that this is not correct in case the repo in which `js_run_binary()` is not the main binary. For example, if I create a Bazel project that depends on the Bonanza repo as follows:

```python
git_override(
    module_name = "build_bonanza",
    commit = "c45401ceea0071815df1d696afa26fb80eaeb0f5",
    remote = "https://github.com/buildbarn/bonanza.git",
)
```

Then building `@build_bonanza//cmd/bonanza_browser:generate_stylesheet` fails as follows:

```
bazel-out/darwin_arm64-opt-exec-ST-d57f47055a04/bin/external/build_bonanza+/cmd/bonanza_browser/generate_stylesheet_/generate_stylesheet: line 449: cd: cmd/bonanza_browser: No such file or directory
```

If I patch up the Bonanza repo as follows, the target does build:

```diff
diff --git a/cmd/bonanza_browser/BUILD.bazel b/cmd/bonanza_browser/BUILD.bazel
index 7c5d9fc..ce3b4bd 100644
--- a/cmd/bonanza_browser/BUILD.bazel
+++ b/cmd/bonanza_browser/BUILD.bazel
@@ -71,7 +71,10 @@ js_run_binary(
     name = "stylesheet",
     outs = ["stylesheet.css"],
     args = ["stylesheet.css"],
-    chdir = package_name(),
+    chdir = "external/%s/%s" % (
+        repo_name(),
+        package_name(),
+    ),
     tool = ":generate_stylesheet",
 )
```

I think what is needed is proper guidance on how to use the `chdir` option properly.

### Version

Bazel 8.3.1 running on a Mac, using rules_js v2.3.8

### How to reproduce

```shell

```

EdSchouten:
Alternatively, the `chdir` should be prefixed with the source path of the current repo implicitly?

jbedard:
I believe this is fixed. There's an e2e/js_binary_workspace test covering the cross-workspace case as an example

Issue #2343, opened 2025-09-06
[Bug]: `nextjs_*` rules do not resolve paths correctly when invocated across Bazel workspaces (i.e. `bazel build @workspace//ui:nextjs_target`)


f-luo:
### Issue

When building nextjs_* rules across Bazel workspaces, existing ruleset fails to resolve the paths correctly.

### Debug Info

`bazel build @workspace//ui --subcommands --sandbox_debug`

```
(20:30:49) SUBCOMMAND: # @@workspace+//ui:standalone_build [action 'Compile Next.js standalone app @@workspace+//ui:standalone_build', configuration: 801f737abe3f0e4b31b022204ec3da5310838309d8175bcfd6c726cf4920e12c, execution platform: @@platforms//host:host, mnemonic: NextJs]
(cd /private/var/tmp/_bazel_fernandol/d99cfb174deb8ddb3d1f2445a615959f/execroot/_main && \
  exec env - \
    BAZEL_BINDIR=bazel-out/darwin_arm64-fastbuild/bin \
    BAZEL_BUILD_FILE_PATH=ui/BUILD \
    BAZEL_COMPILATION_MODE=fastbuild \
    BAZEL_PACKAGE=ui \
    BAZEL_TARGET=@workspace+//ui:standalone_build \
    BAZEL_TARGET_CPU=darwin_arm64 \
    BAZEL_TARGET_NAME=standalone_build \
    BAZEL_WORKSPACE=_main \
    JS_BINARY__CHDIR=ui \   <------------- THIS IS POINTING TO THE PACKAGE NAME
    JS_BINARY__PATCH_NODE_FS=1 \
    JS_BINARY__USE_EXECROOT_ENTRY_POINT=1 \
    NEXTJS_STANDALONE_CONFIG=external/workspace+/ui/next.standalone.js \
    NEXT_PUBLIC_APP_NAME=APP \
    NODE_ENV=production \
    SKIP_ENV_VALIDATION=true \
  bazel-out/darwin_arm64-opt-exec-ST-d57f47055a04/bin/external/workspace+/ui/next_js_binary_/next_js_binary build)
# Configuration: 801f737abe3f0e4b31b022204ec3da5310838309d8175bcfd6c726cf4920e12c
# Execution platform: @@platforms//host:host
# Runner: local
```

```
[...]

PR #2344, opened 2025-09-06
js_binary and js_run_binary to support invocation across bazel workspaces

f-luo:
Addresses issue: https://github.com/aspect-build/rules_js/issues/2343

### Test plan

### DIFF
--- a/e2e/js_binary_workspace/.bazelignore
+++ b/e2e/js_binary_workspace/.bazelignore
@@ -1,4 +1,0 @@
-workspace/bazel-bin
-workspace/bazel-out
-workspace/bazel-testlogs
-workspace/bazel-workspace
--- a/e2e/js_binary_workspace/workspace/MODULE.bazel
+++ b/e2e/js_binary_workspace/workspace/MODULE.bazel
@@ -1,9 +1,6 @@
 module(name = "workspace")
 
 bazel_dep(name = "aspect_rules_js", version = "0.0.0")
-bazel_dep(name = "aspect_bazel_lib", version = "2.17.1")
-bazel_dep(name = "bazel_skylib", version = "1.5.0")
-
 local_path_override(
     module_name = "aspect_rules_js",
     path = "../../..",
--- a/e2e/js_binary_workspace/MODULE.bazel
+++ b/e2e/js_binary_workspace/MODULE.bazel
@@ -2,7 +2,6 @@ module(name = "js_binary_workspace")
 
 bazel_dep(name = "aspect_rules_js", version = "0.0.0", dev_dependency = True)
 bazel_dep(name = "bazel_skylib", version = "1.8.1", dev_dependency = True)
-bazel_dep(name = "aspect_bazel_lib", version = "2.17.1", dev_dependency = True)
 
 local_path_override(
     module_name = "aspect_rules_js",
--- a/e2e/js_binary_workspace/WORKSPACE.bazel
+++ b/e2e/js_binary_workspace/WORKSPACE.bazel
@@ -1,0 +1,20 @@
+# WORKSPACE mode setup for Bazel 7 when bzlmod is disabled.
+
+local_repository(
+    name = "aspect_rules_js",
+    path = "../..",
+)
+
+local_repository(
+    name = "workspace",
+    path = "workspace",
+)
+
+load("@aspect_rules_js//js:repositories.bzl", "rules_js_dependencies")
+
+rules_js_dependencies()
+
+load("@aspect_rules_js//js:toolchains.bzl", "rules_js_register_toolchains")
+
+# Use default Node toolchain registration suitable for CI.
+rules_js_register_toolchains()
--- a/e2e/js_binary_workspace/WORKSPACE.bzlmod
+++ b/e2e/js_binary_workspace/WORKSPACE.bzlmod
@@ -1,0 +1,2 @@
+# The presence of this file causes WORKSPACE to be ignored when bzlmod is enabled.
+# See https://docs.google.com/document/d/1JtXIVnXyFZ4bmbiBCr5gsTH4-opZAFf5DMMb-54kES0/edit#heading=h.y054fjub9max
--- a/e2e/js_binary_workspace/workspace/.bazelignore
+++ b/e2e/js_binary_workspace/workspace/.bazelignore
@@ -1,5 +1,0 @@
-bazel-bin
-bazel-out
-bazel-testlogs
-bazel-workspace
-external
--- a/e2e/js_binary_workspace/workspace/WORKSPACE
+++ b/e2e/js_binary_workspace/workspace/WORKSPACE
@@ -1,0 +1,2 @@
+# Stub WORKSPACE required for Bazel 6 when a module is referenced via a local_repository override.
+workspace(name = "workspace")
--- a/e2e/js_binary_workspace/workspace/WORKSPACE.bzlmod
+++ b/e2e/js_binary_workspace/workspace/WORKSPACE.bzlmod
@@ -1,0 +1,2 @@
+# The presence of this file causes WORKSPACE to be ignored when bzlmod is enabled.
+# See https://docs.google.com/document/d/1JtXIVnXyFZ4bmbiBCr5gsTH4-opZAFf5DMMb-54kES0/edit#heading=h.y054fjub9max


matthewjh:
@f-luo  could you try prefixing the chdir value with this in js_binary itself?
https://github.com/aspect-build/rules_js/blob/main/js/private/js_binary.bzl

That would then close https://github.com/aspect-build/rules_js/issues/2275.

f-luo:
> @f-luo could you try prefixing the chdir value with this in js_binary itself? https://github.com/aspect-build/rules_js/blob/main/js/private/js_binary.bzl
> 
> That would then close #2275.

I do think that would address it more broadly. My only concern is that doing it at the `js_binary` level would risk breaking existing usages, more specifically: if someone is already passing in `external/workspace+/pkg`.

Unless we also detect the presence of `external/...+/` to deduplicate.

Perhaps someone from `rules_js` (@alexeagle) and the original creator of `nextjs_*` rulesets (@jbedard) can chime in here?

alexeagle:
I agree this should be changed in js_binary. I don't think there are any existing functional use cases that would be broken by stripping `external/` from paths.

f-luo:
Okay, that sounds good. I can take a stab at it. But as an FYI I will be traveling for the next 2 weeks, so will come back to this afterwards... unless anyone else would like to drive it in the meantime :)

jbedard:
@f-luo since you're updating this... can you try reproducing the issue to demonstrate a) what's broken and b) that your fix works?

Most likely we'd need to add an `e2e` test? Or maybe move `examples/nextjs` into `e2e/` so it is its own MODULE?

f-luo:
@jbedard , thanks for checking in. Just repro'ed it on the latest `2.8.1`. The repro steps and outcome is consistent with the report at https://github.com/aspect-build/rules_js/issues/2343.

The fix in this PR does fix it. More specifically I tried:

From same workspace as //ui:
- Ran `bazel build <nextjs_standalone_server>` and `bazel run <nextjs_standalone_server>`

From another bazel workspace:
- Ran `bazel build @workspace//ui:<nextjs_standalone_server>` and `bazel run @workspace//ui:<nextjs_standalone_server>`

The behavior is now consistent.

I like the idea of have a test here. Let me move some stuff around and add coverage here, thank you.

P.S. seems like one of the [CI workflows failed](https://github.com/aspect-build/rules_js/actions/runs/19400485297/job/55507608627?pr=2344) due to disk issue:

For example:

```
Error in download_and_extract: java.io.IOException: Error extracting /home/runner/.cache/bazel/_bazel_runner/8d53428c79e05e77d1e55522fa21a0fd/external/rules_java~~toolchains~remotejdk11_linux/temp695009450509042695/zulu11.72.19-ca-jdk11.0.23-linux_x64.tar.gz to /home/runner/.cache/bazel/_bazel_runner/8d53428c79e05e77d1e55522fa21a0fd/external/rules_java~~toolchains~remotejdk11_linux/temp695009450509042695: write (No space left on device)
```

jbedard:
It would be great if we could get a super simple test demonstrating it. If not simple then maybe just converting the nextjs example to an e2e test?

> P.S. seems like one of the [CI workflows failed](https://github.com/aspect-build/rules_js/actions/runs/19400485297/job/55507608627?pr=2344) due to disk issue:

Yeah, sorry the GH runners seem overloaded atm and people have been busy. I've been spamming the re-run button but you might not have permission to do that. I hope just testing locally is enough for you while working on this PR?

f-luo:
Yup, np! Just added a simple e2e test. I wasn't sure if there's a hook somewhere to automate running e2e. I've testing via:

```
bash e2e/js_binary_workspace/test.sh
```

Let me know if there's anything I need to do in addition, happy to iterate on this.

Thanks again!

e2e/js_binary_workspace/test.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
jbedard: Can we try to reproduce it with just `bazel test //...` in this directory instead of the shell script? Or does the bug only occur with `bazel run`?

My initial idea would be this `js_binary_workspace` BUILD does something such as:
```
js_run_binary(
   name = "run_workspace_bin",
   tool = @workspace//:bin,
   stdout = "out.txt"
)
build_test(
   name = "test",
   targets = [":run_workspace_bin", "@workspace//:run_bin"]
)
```

`workspace/BUILD` could have the exact same `js_run_binar(name="run_bin")` but `tool = "//:bin"` without the `@workspace`.

js/private/test/image/custom_owner_test_app.listing
```diff
@@ -2,7 +2,7 @@ drwxr-xr-x  0 100    0           0 Jan  1  1970 ./js/
 drwxr-xr-x  0 100    0           0 Jan  1  1970 ./js/private/
 drwxr-xr-x  0 100    0           0 Jan  1  1970 ./js/private/test/
 drwxr-xr-x  0 100    0           0 Jan  1  1970 ./js/private/test/image/
--r-xr-xr-x  0 100    0       xxxxx Jan  1  1970 ./js/private/test/image/bin
```
jbedard: Search for `xxxxx` and you'll see a `sed` command that was doing this to make the test work on multiple platforms (where this file has different sizes for some reason 🤷).

I assume you need to update that `sed` and this diff should go away. Or if your change somehow made this file size consistent across platforms then lets delete the `sed` instead of modify it. I'll kickoff CI now and see if linux has a different value then what you have locally...

contrib/nextjs/next.bazel.mjs
```diff
@@ -12,6 +12,10 @@ if (nextjsStandaloneConfig.startsWith(process.env.BAZEL_BINDIR)) {
         process.env.BAZEL_BINDIR.length + 1
     )
 }
+
+// Support invocation from another workspace: strip "external/<repo>/" if present
+nextjsStandaloneConfig = nextjsStandaloneConfig.replace(/^external\/[^/]+\//, '')
```
jbedard, line 17: This is needed despite your fix? Or does your main fix just allow this one to be done?

jbedard:
Great, that is nice and simple (a lot simpler then involving nextjs in a test) 👍

We might be able to avoid the test.sh script, but otherwise this LGTM, thanks!

e2e/js_binary_workspace/test.sh
f-luo: Yup, great suggestion, updated this to be runnable via `bazel test //...` (both inside and outside the `workspace` dir).

js/private/test/image/custom_owner_test_app.listing
f-luo: Nice, thanks for the pointers. I don't think this PR changes the cross-platform behavior but updated the `sed` regex to handle the new changes.

contrib/nextjs/next.bazel.mjs
f-luo: This is still needed in addition to the chdir fix.

jbedard:
I think we just need to add the e2e test to `.github/workflows/ci.yaml` and `.aspect/workflows/config.yaml`

e2e/js_binary_workspace/workspace/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
...
+)
+
+js_run_binary(
+    name = "run_bin",
+    stdout = "out_workspace.txt",
+    tool = ":bin",
+    visibility = ["//visibility:public"],
+)
+
+build_test(
```
jbedard: I assume no one will ever actually run bazel in this directory so this will never be used? Then we can drop `bazel_skylib` 🤷

f-luo: Sounds good, thanks

e2e/js_binary_workspace/workspace/MODULE.bazel
```diff
@@ -0,0 +1,10 @@
+module(name = "workspace")
+
+bazel_dep(name = "aspect_rules_js", version = "0.0.0")
+bazel_dep(name = "aspect_bazel_lib", version = "2.17.1")
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
```
jbedard: Delete? 🤷

e2e/js_binary_workspace/.bazelignore
```diff
@@ -0,0 +1,4 @@
+workspace/bazel-bin
```
jbedard: why is `.bazelignore` necessary? Normally e2e tests only need to list the `node_modules` directories and I don't think we have any?

f-luo: You are correct, it is not necessary. I think I previously ran `bazel test //...` inside the `workspace` which left behind a bunch of bazel artifact and prevented me from running `bazel test //...` from the parent workspace.

But to your early comment, we won't be going into `e2e/js_binary_workspace/workspace` to run tests so we can get rid of this. Thanks

e2e/js_binary_workspace/workspace/MODULE.bazel
f-luo: Yup, thanks for catching.

jbedard:
@alexeagle @dzbarsky can you take a look at the js_binary changes being done here? WDYT?

matthewjh:
@DavidZbarsky-at  @jbedard can we get this in?

jbedard (approved):

PR #2349, opened 2025-09-15
feat: add pnpm_version_from

alexeagle:
Ideally pnpm version should be kept in-sync with non-Bazel codepaths, and the easiest way to do this is the packageManager field in package.json

See https://github.com/nodejs/corepack#when-authoring-packages

### DIFF
--- a/npm/private/pnpm_extension.bzl
+++ b/npm/private/pnpm_extension.bzl
@@ -1,7 +1,7 @@
 """pnpm extension logic (the extension itself is in npm/extensions.bzl)."""
 
 load("@aspect_bazel_lib//lib:lists.bzl", "unique")
-load(":pnpm_repository.bzl", "LATEST_PNPM_VERSION")
+load(":pnpm_repository.bzl", "DEFAULT_PNPM_VERSION", "LATEST_PNPM_VERSION")
 
 DEFAULT_PNPM_REPO_NAME = "pnpm"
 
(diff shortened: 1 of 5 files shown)


kormide (approved):

npm/private/pnpm_extension.bzl
```diff
@@ -40,11 +40,33 @@ def resolve_pnpm_repositories(modules):
...
+            elif attr.pnpm_version == "latest":
                 v = LATEST_PNPM_VERSION
+            else:
+                v = attr.pnpm_version
 
-            registrations[attr.name].append(v)
+            # Avoid inserting the default version from a non-root module
+            # (likely rules_js itself) if the root module already has a version.
+            if mod.is_root or len(registrations[attr.name]) == 0:
+                registrations[attr.name].append(v)
```
gzm0, line 69: @alexeagle it seems to me that this code block essentially breaks all non-root module registrations. Note that:

- There is always a registration for `attr.name == "pnpm"` (the one from rules_js itself)
- It is not allowed to register a different name (see above).

Do you recall why you added this (it is not clear to me how it relates to the PR).

Sorry for the comment on the old PR, I'll file a bug report once I get a chance, but wanted to share the finding anyways.

jbedard: I think you are correct, and I have a fix WIP: https://github.com/aspect-build/rules_js/pull/2883

PR #2365, opened 2025-09-25
feat(npm): make repo rules reproducible

tyler-french:
In order to use the `--repo_contents_cache`, we need to mark the repository rule as reproducible.

See https://github.com/bazel-contrib/bazel-gazelle/pull/2119 as an example

The contents of their `repo_contents_cache` might be different.


jbedard:
@tyler-french have time to fix the lint errors? Maybe switch to:
```
if not hasattr(rctx, "repo_metadata"):
    return None
return rctx.repo_metadata(reproducible = reproducible)
```
which will allow simply deleting 2 lines if/when we remove older bazel versions that need the `if`

jbedard:
Done in https://github.com/aspect-build/rules_js/pull/2416

PR #2366, opened 2025-09-25
fix: sandboxing and esm imports

jbedard:
---

### Test plan

- New test cases added


matthewjh:
@jbedard are you still working on this?

jbedard:
I still hope to do it but haven't been able to put any time/testing into it recently. It might be nice to enable by default in a v3, so merge for opt-in sooner?

jbedard:
Update as of https://github.com/aspect-build/rules_js/pull/2366/commits/8bf78b8a4c9fa89ce58d7d5162c9d9ea7ad4d24c

I think there are currently issues with the node internals [invoking `fs.*`](https://github.com/nodejs/node/blob/v24.12.0/lib/fs.js#L2951-L2954) and crossing between the internals vs patched vs non-patched.

Something such as https://github.com/aspect-build/rules_js/pull/2366/commits/dbf4482432a08021004feff80a938a881c082b98 fixes some of those scenarios, but just breaks others.

EDIT:

Maybe the [conditional wrapping](https://github.com/aspect-build/rules_js/commit/8bf78b8a4c9fa89ce58d7d5162c9d9ea7ad4d24c#diff-a79bad39cac7eb4caf6ee12fd544e34fbbf1145341d33e3e6c57b294c985984bR135-R136) needs to be dropped and we still need to wrap everything even when using esm patches? [Doing that](https://github.com/aspect-build/rules_js/pull/2366/commits/6276ff0bffb3f99dc371148b69d5838e5bb78ff5) may have made tests green...

MODULE.bazel
```diff
@@ -14,7 +14,7 @@ bazel_dep(name = "bazel_features", version = "1.9.0")
 bazel_dep(name = "bazel_lib", version = "3.0.0")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "platforms", version = "0.0.5")
-bazel_dep(name = "rules_nodejs", version = "6.3.0")
+bazel_dep(name = "rules_nodejs", version = "6.4.0")
```
jbedard: TODO: why was this done?

jbedard:
@devversion and @tenzinhl you both contributed to this PR so I hope you have some experience with both the issue and this fix. If either of you have time would you be able to try this PR out and see if it fixes the underlying issue?

I still want to add a proper test of actual esm import statements, in addition to all the existing `fs.*` tests that I believe are all passing 🤞

devversion:
Sorry @jbedard, just returning back from vacation. Unfortunately I don't recall at all where we've been seeing this, but I can try ask and search on Slack to figure out a concrete case where we've been seeing the ESM escaping. Would that be helpful?

devversion:
Also cc. @josephperrott and @clydin in case you remember a case where `rules_js` escaped with ESM.

matthewjh:
@jbedard can you check out these 2 comments I left couple of years ago? They give context on other sources of sandbox escape that may be addressable with your changes:

https://github.com/aspect-build/rules_js/issues/362#issuecomment-2079392107
https://github.com/aspect-build/rules_js/issues/362#issuecomment-2079410172

PR #2374, opened 2025-10-03
refactor: move AspectWatchProtocol to standalone library

jbedard:
For some experimental importing+usage of this class in binaries supporting the watch protocol. The goal is minimizing actual changes within rules_js so for now the js_run_devserver.mjs is bundled and is almost identical to before.

### Test plan


js/private/src/rollup.config.mjs
```diff
@@ -0,0 +1,16 @@
...
+
+/** @type {import("rollup").RollupOptions} */
+export default {
+    plugins: [
+        resolve(),
+        ts({
+            compilerOptions: {
+                target: 'ES2020',
+                module: 'ES2020',
+                moduleResolution: 'Node',
```
nickschaap: Can we get the declarations and source maps emitted?

js/private/js_run_devserver.mjs
```diff
@@ -1,27 +1,163 @@
-import * as path from 'node:path'
-import * as perf_hooks from 'node:perf_hooks'
-import * as fs from 'node:fs'
-import * as os from 'node:os'
-import * as child_process from 'node:child_process'
-import * as crypto from 'node:crypto'
-import * as readline from 'node:readline'
-
+import * as path from 'node:path';
```
nickschaap, line 1: Maybe a note that this is now a generated file

js/private/devserver/src/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "private": true,
```
nickschaap, line 2:
```suggestion
    "private": true,
    "type": "module",
```

nickschaap: You shouldn't need mjs/mts everywhere if you add this and make sure the package json is available at runtime for node

js/private/src/aspect_watch_protocol.d.mts
```diff
@@ -0,0 +1,74 @@
...
+    | 'NEGOTIATE_RESPONSE'
+    | 'EXIT'
+
+export interface Message {
+    readonly kind: MessageType
+}
+
+export interface SourceInfo {
+    readonly is_symlink: boolean
+    readonly is_source: boolean
```
nickschaap: I believe these are optional

jbedard: Good point 👍

js/private/devserver/src/package.json
jbedard: This file isn't used at runtime since we bundle everything into a single .js file, but I will add that just to be clear/correct 👍

js/private/js_run_devserver.mjs
jbedard: The `write_source_file(s)` rules directly copy from the tsc/rollup-generated code to src so we can't edit it, at least not without some extra `genrule` logic which I don't think is worth it atm. Tests will fail if you edit this file is not a direct clone of the generated file so I'm not too worried about it being bad or out of date...

dzbarsky:
@codex please review this

PR #2378, opened 2025-10-04
refactor: ensure fs patch tests are run on real fs library

jbedard:
To move toward the refactoring in https://github.com/aspect-build/rules_js/pull/2366 where this seems required.

### Test plan

- New test cases added


js/private/node-patches/src/fs.cts
```diff
@@ -283,18 +306,8 @@ export function patcher(fs: any = _fs, roots: string[]) {
                     ) {
                         return cb(null, next)
                     }
-                    // The escape from the root is not mappable back into the root; we must make
```
jbedard, line 286: Without this change some of the non-`Sync` `js/private/test/node-patches/readlink.js` tests fail. I did this by copy+pasting the `Sync` version, then changing it back callbacks and non-`Sync`. This block of code did not exist in the `Sync` version, removing it makes tests pass in this PR, and I don't know why...

js/private/node-patches/fs.cjs
```diff
@@ -45,18 +45,38 @@ const util = require("util");
...
+    'lstat',
+    'lstatSync',
+    'realpath',
+    'realpathSync',
+    'readlink',
+    'readlinkSync',
+    'readdir',
+    'readdirSync',
+    'opendir',
+];
```
dzbarsky, line 62: opendirSync?

js/private/node-patches/src/fs.cts
```diff
@@ -28,22 +28,45 @@ type Dirent = any
...
+    'lstat',
+    'lstatSync',
+    'realpath',
+    'realpathSync',
+    'readlink',
+    'readlinkSync',
+    'readdir',
+    'readdirSync',
+    'opendir',
+]
```
dzbarsky, line 49: same here

js/private/node-patches/src/fs.cts
```diff
@@ -448,15 +464,39 @@ export function patcher(fs: any = _fs, roots: string[]) {
...
                 const _promises = oldGetter()
                 Object.assign(cachedPromises, _promises, promises)
                 return cachedPromises
             }
-            Object.defineProperty(fs, 'promises', promisePropertyDescriptor)
+            Object.defineProperty(
+                fs,
+                'promises',
+                Object.assign(
+                    {
```
dzbarsky: are these arguments reversed? Shouldn't the getter be assigned onto the object created with the descriptor?

js/private/node-patches/src/fs.cts
```diff
@@ -448,15 +464,39 @@ export function patcher(fs: any = _fs, roots: string[]) {
...
+                    },
+                    Object.create(promisePropertyDescriptor)
+                )
+            )
+
+            unpatchPromises = function unpatchFsPromises() {
+                Object.defineProperty(fs, 'promises', promisePropertyDescriptor)
+            }
         } else {
+            const unpatched_promises = Object.keys(promises).reduce(
```
dzbarsky: nit: unpatchedPromises?

js/private/test/node-patches/readlink.js
```diff
@@ -152,36 +160,45 @@ describe('testing readlink', async () => {
...
+                    'CB: should read the symlink in the sandbox'
+                )
+
+                assert.deepStrictEqual(
+                    await fs.promises.readlink(linkPath),
                     filePath,
                     'Promise: should read the symlink in the sandbox'
                 )
+
+                revertPatches()
```
dzbarsky, line 201: i wonder if you can use the `using` RAII-type thing to automatically call revert at the end?

js/private/node-patches/fs.cjs
jbedard: this is the transpiled .cts

js/private/node-patches/src/fs.cts
jbedard: I think `opendirSync` doesn't exist, or at least didn't when we wrote this file, because we don't patch it... 🤷

js/private/node-patches/src/fs.cts
jbedard: I'll look at that again. Note that I only changed this because the tests were failing (the `fs.promises` were not being reverted properly after tests run) so I'm fairly confident it's working

js/private/test/node-patches/readlink.js
jbedard: Only in recent versions of node (24?), and we're running these on v16-24 atm.

js/private/node-patches/src/fs.cts
jbedard: You're correct it was backwards 👍

js/private/node-patches/src/fs.cts
dzbarsky: Exists since 2021... https://www.geeksforgeeks.org/node-js/node-js-fs-opendirsync-method/

js/private/test/node-patches/readlink.js
dzbarsky: Ok cool maybe leave a TODO to use it eventually next time you touch the file :)

PR #2380, opened 2025-10-07
test: test fs patches via cjs and esm

jbedard:
Converting the source files to mjs, then testing on both mjs and babel-transpiled cjs.

### Test plan

- New test cases added


js/private/test/node-patches/BUILD.bazel
```diff
@@ -1,13 +1,19 @@
 load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "copy_to_bin")
+load("@npm//:@babel/cli/package_json.bzl", babel_bin = "bin")
 load("//js:defs.bzl", "js_test")
 
 TESTS = [
```
dzbarsky, line 5: maybe a bit cleaner to define MJS_TESTS and put `escape.js` in its own list so you don't need to filter below? also as written you will run escape.js in CJS_TESTS as well, is that intended?

jbedard: The idea was `TESTS` get run directly, and `CJS_TESTS` are a subset that get transpiled+run.

You're correct escape.js is run twice which I'll fix...

dzbarsky: yeah it might be cleaner to combine the mjs + js test once at the js_test expansion site than combining here and then filtering multiple times in the babel targets

jbedard: Yeah that reduces the code a bit, updated 👍

PR #2382, opened 2025-10-07
fix: add fs patch for `fs.opendirSync`

jbedard:
---

### Test plan

- New test cases added


js/private/test/node-patches/opendir.mjs
```diff
@@ -52,7 +52,15 @@ describe('testing opendir', async () => {
                     path.join(fixturesDir, 'a')
                 )
                 const entry1 = await dir.read()
-                const entry2 = await util.promisify(dir.read.bind(dir))()
```
jbedard, line 55: This was causing a warning to stdout since `dir.read()` is already a promise-returning method, while `dir.read(callback)` is not.

js/private/test/node-patches/opendir.mjs
```diff
@@ -69,6 +77,23 @@ describe('testing opendir', async () => {
...
                 assert.equal(entry1.name, (await dir.read()).name)
+                assert.equal(entry2.name, (await dir.read()).name)
+
+                // Assert reading via opendirSync() produces the same result
+                dir = fs.opendirSync(path.join(fixturesDir, 'a'))
+                const entry1Sync = dir.readSync()
+                const entry2Sync = dir.readSync()
+                assert.equal(entry1.name, entry1Sync.name)
+                assert.equal(
+                    entry1.isSymbolicLink(),
```
jbedard, line 88: One of these `isSymbolicLink` calls is what failed before the "fix" commit which overrides `opendirSync` to be like the non-Sync

PR #2384, opened 2025-10-08
Stop reading pnpm_lock from label_store

dzbarsky:
We should stop using label_store completely; it served a useful purpose to precompute labels when restarts were expensive, but that's no longer true in modern bazels. We can use the labels direcly from `rctx.attr` and simplify some indirection.

### Test plan
existing tests


npm/private/npm_translate_lock.bzl
```diff
@@ -139,7 +139,7 @@ def _npm_translate_lock_impl(rctx):
 INFO: {} file updated. Please run your build again.
 
 See https://github.com/aspect-build/rules_js/issues/1445
-""".format(state.label_store.relative_path("pnpm_lock"))
```
dzbarsky: i changed some error messages to use the label instead of the path for simplicity, figured it doesn't matter too much. But can change it back if you prefer.

npm/private/npm_translate_lock_state.bzl
```diff
@@ -489,7 +489,7 @@ def _load_lockfile(priv, rctx, _, label_store):
 
     yq_args = [
         str(label_store.path("host_yq")),
```
jbedard, line 491: Can we drop them all?

dzbarsky: yes but i ran out of time to do a mega-PR

jbedard:
will need a rebase and to open against `3.x` branch

PR #2396, opened 2025-10-15
Remove usage of dicts.add

dzbarsky:
### Test plan
Covered by existing test cases

### DIFF
--- a/js/private/js_run_devserver.bzl
+++ b/js/private/js_run_devserver.bzl
@@ -1,10 +1,9 @@
 "Implementation details for js_run_devserver rule"
 
-load("@bazel_skylib//lib:dicts.bzl", "dicts")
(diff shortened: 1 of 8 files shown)


jbedard:
> Error: trying to mutate a frozen dict value

Which dictionaries are frozen now?

Does this mean we need to use `|` instead of modifying dictionaries in a bunch of places now? Will that always be better then what we have?

dzbarsky:
> > Error: trying to mutate a frozen dict value
> 
> Which dictionaries are frozen now?
> 
> Does this mean we need to use `|` instead of modifying dictionaries in a bunch of places now? Will that always be better then what we have?

Only the ones that came directly from an attribute. I added a copy to the place that was mutating it and we should be good now. (Previously _dict.add would always copy it anyway). Outside of that, no additional change should be needed

PR #2397, opened 2025-10-15
feat: add dev script to update snapshots

Mivr:
### Test plan
N/A

### DIFF
--- a/tools/update-snapshots.sh
+++ b/tools/update-snapshots.sh
@@ -13,37 +13,37 @@ YELLOW='\033[1;33m'
 NC='\033[0m' # No Color
 
 print_section() {
-  echo -e "${BLUE}==> $1${NC}"
+    echo -e "${BLUE}==> $1${NC}"
 }
 
 print_success() {
-  echo -e "${GREEN}✓ $1${NC}"
+    echo -e "${GREEN}✓ $1${NC}"
 }
 
 print_category() {
-  echo ""
-  echo -e "${YELLOW}========================================${NC}"
-  echo -e "${YELLOW}$1${NC}"
-  echo -e "${YELLOW}========================================${NC}"
+    echo ""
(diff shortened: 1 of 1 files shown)


tools/update-snapshots.sh
```diff
@@ -0,0 +1,136 @@
...
+echo "Updating all snapshots and sources..."
+echo "========================================"
+
+##############################################################################
+# E2E TEST SNAPSHOTS
+##############################################################################
+
+print_category "E2E TEST SNAPSHOTS"
+
+# pnpm_lockfiles - Multiple versions (bzlmod only, no workspace mode in this repo)
```
jbedard, line 59: There's also `e2e/pnpm_lockfiles/update-snapshots.sh`, I'm not sure if we should delete that or maybe invoke it from here?

tools/update-snapshots.sh
```diff
@@ -0,0 +1,136 @@
...
+run_target "$REPO_ROOT" "//npm/private/test:write_npm_translate_lock" "npm/private/test"
+
+# js/private/test - js_binary launcher snapshot
+run_target "$REPO_ROOT" "//js/private/test:write_launcher" "js/private/test"
+
+##############################################################################
+# GENERATED SOURCES
+##############################################################################
+
+print_category "GENERATED SOURCES"
```
jbedard, line 104: Can we `bazel query` to find these instead of hardcoding them? Or write a comment in this file saying how to regenerate this file...

tools/update-snapshots.sh
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env bash
```
jbedard, line 1: Lets move this to `tools/` maybe? Somewhere other then the root imo 🤷

jbedard (approved):

jbedard:
Does `//:format` need to run?

PR #2403, opened 2025-10-22
Fix fetching npm packages from git repositories

thomasdraebing:
Aspect_rules_js supports fetching a git repository containing an npm package. However, to follow that path, it is required that the URL of the repository is prefixed with `git+`. However, this URL is then also configured as the remote for the git repository. The `git+` prefix is not supported by git and thus fetching fails with:

git: 'remote-git+https' ist kein Git-Befehl. Siehe 'git --help'.

To prevent this, the `git+`-prefix is now being removed, if it is present before setting it as remote url.formation ends up in the git log.
-->

---

### Test plan

Install an NPM package from a git repository. The pnpm-lock.yaml may look like this:

```
lockfileVersion: '6.0'

settings:
  autoInstallPeers: true
  excludeLinksFromLockfile: false

dependencies:
  highlight.js:
    specifier: ^11.11.1
    version: 11.11.1
  highlightjs-ttcn3:
    specifier: https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git#6daccff309fca1e7561a43984d42fa4f829ce06d
    version: gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git#6daccff309fca1e7561a43984d42fa4f829ce06d

packages:
  /highlight.js@11.11.1:
    resolution: {integrity: sha512-Xwwo44whKBVCYoliBQwaPvtd/2tYFkRQtXDWj1nackaV2JPXx3L0+Jvd8/qCJ2p+ML0/XVkJ2q+Mr+UVdpJK5w==}
    engines: {node: '>=12.0.0'}
    dev: false

  gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git#6daccff309fca1e7561a43984d42fa4f829ce06d:
    resolution:
      type: git
      repo: git+https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git
      commit: 6daccff309fca1e7561a43984d42fa4f829ce06d
    name: highlightjs-ttcn3
    version: 0.0.1
    dependencies:
      highlight.js: 11.11.1
    dev: false
```

### DIFF
--- a/npm/private/npm_import.bzl
+++ b/npm/private/npm_import.bzl
@@ -493,6 +493,10 @@ def _fetch_git_repository(rctx):
     if not rctx.attr.commit:
         fail("commit required if url is a git repository")
 
+    remote_url = str(rctx.attr.url)
+    if remote_url.startswith("git+"):
+        remote_url = remote_url[4:]
+
     # Adapted from git_repo helper function used by git_repository in @bazel_tools//tools/build_defs/repo:git_worker.bzl:
     # https://github.com/bazelbuild/bazel/blob/5bdd2b2ff8d6be4ecbffe82d975983129d459782/tools/build_defs/repo/git_worker.bzl#L34
     git_repo = struct(
@@ -500,11 +504,11 @@ def _fetch_git_repository(rctx):
         shallow = "--depth=1",
         reset_ref = rctx.attr.commit,
         fetch_ref = rctx.attr.commit,
-        remote = str(rctx.attr.url),
+        remote = remote_url,
     )
     rctx.report_progress("Cloning %s of %s" % (git_repo.reset_ref, git_repo.remote))
     _git_init(rctx, git_repo)
-    _git_add_origin(rctx, git_repo, rctx.attr.url)
+    _git_add_origin(rctx, git_repo, remote_url)
     _git_fetch(rctx, git_repo)
     _git_reset(rctx, git_repo)
     _git_clean(rctx, git_repo)


jbedard:
Can you add a test which fails without your fix?

thomasdraebing:
> Can you add a test which fails without your fix?

I will do that. Currently I am waiting for my company to approve the CLA. As soon as this has been done, I will continue on this PR.

thomasdraebing:
@jbedard I finally got the approval to sign the CLA and added an E2E test. If you just check out the first commit that contains only the test, it fails, and it succeeds with the second commit containing the fix.

jbedard:
Would it be possible to test it by adding a package to `e2e/pnpm_lockfiles/base/package.json`? Or are you only able to reproduce when manually invoking `npm_import()`?

jbedard:
Otherwise I think you need to add the e2e test to the list in `.github/workflows/ci.yaml` and `.aspect/workflows/config.yaml`

thomasdraebing:
> Otherwise I think you need to add the e2e test to the list in `.github/workflows/ci.yaml` and `.aspect/workflows/config.yaml`

Done



> Would it be possible to test it by adding a package to `e2e/pnpm_lockfiles/base/package.json`? Or are you only able to reproduce when manually invoking `npm_import()`?

I have updated the test.

jbedard:
@thomasdraebing I'm just looking to port this to the 3.x branch and also add more testing and I keep getting errors when trying to install this `https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git#6daccff309fca1e7561a43984d42fa4f829ce06d` dep. But if I run pnpm (even a newer version that changes the lockfile format) in your new e2e test it seems to work, maybe because it maintains the metadata it already has and doesn't try to refetch it.

Is there something special with this `https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git#6daccff309fca1e7561a43984d42fa4f829ce06d` URL that may require auth or restrict it in some other way?

jbedard:
Even if I just `rm pnpm-lock.yaml` from `e2e/git_dep_no_tar` it then fails to install.

@thomasdraebing you must have some local config making this work work?

thomasdraebing:
@jbedard just using pnpm or npm also fails for me for that repository. Actually for the same reason, it tries to download a tar-archive, which isn't available on that git server.

```
❯ pnpm add https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git\#6daccff309fca1e7561a43984d42fa4f829ce06d
 WARN  GET https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git#6daccff309fca1e7561a43984d42fa4f829ce06d error (ERR_PNPM_TARBALL_EXTRACT). Will retry in 10 seconds. 2 retries left.
```

 It works however using yarn from which we are in the progress of migrating from due to the removal of support of `yarn_install` in `rules_nodejs`.

```
❯ yarn add https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git\#6daccff309fca1e7561a43984d42fa4f829ce06d
yarn add v1.22.22
info No lockfile found.
(node:13400) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
(Use `node --trace-deprecation ...` to show where the warning was created)
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 🔨  Building fresh packages...

success Saved lockfile.
success Saved 2 new dependencies.
info Direct dependencies
└─ highlightjs-ttcn3@0.0.1
info All dependencies
├─ highlight.js@11.11.1
└─ highlightjs-ttcn3@0.0.1
✨  Done in 4.84s.
```

jbedard:
How did you generate the pnpm-lock in your new e2e test then?

thomasdraebing:
The case used in the e2e test is actually the case, I wanted to solve for the Gerrit project. There I added the dependency manually to the lock-file based on the previously existing yarn.lock file. I guess, the same bug should be fixed in pnpm as well.

jbedard:
You manually created the lockfile? Does that mean it might be invalid and your fix is for a scenario that pnpm would never reproduce?

thomasdraebing:
I don't believe that it is invalid. The resolution format is implemented here: https://github.com/pnpm/pnpm/blob/v8.15.9/resolving/git-resolver/src/index.ts#L36.

This works:

```
lockfileVersion: '6.0'

settings:
  autoInstallPeers: true
  excludeLinksFromLockfile: false

dependencies:
  highlightjs-vue:
    specifier: github:paladox/highlightjs-vue#44eed074ea0110d1ad03d2cbd77d27027cf7bb04
    version: github.com/paladox/highlightjs-vue/44eed074ea0110d1ad03d2cbd77d27027cf7bb04

packages:

  github.com/paladox/highlightjs-vue/44eed074ea0110d1ad03d2cbd77d27027cf7bb04:
    resolution:
      type: git
      repo: git+https://codeload.github.com/paladox/highlightjs-vue.git
      commit: 44eed074ea0110d1ad03d2cbd77d27027cf7bb04
    name: highlightjs-vue
    version: 1.1.0
    dev: false
```

However, pnpm again uses this data to compute the tar-ball download URL.

This is not the case for rules_js. It will actually try to clone. Just see the diff of my fix and that is clearly broken.
I would be open to have a look at pnpm itself and to add the support of git clone to fetching dependencies as well, because I believe that this is a valid usecase.

jbedard:
If it is impossible to generate that lockfile via pnpm then I would not consider that to be a valid lockfile.

See https://github.com/aspect-build/rules_js/commit/e75843134afffffda92e2789d7319a32d034a367 where I was trying to add tests for this. Anywhere I use `git+https` specifiers it results in `resolution: {... repo: https://}` without any `git+` in the resolution.

How did you end up with `git+https` in the `resolution.repo`?

jbedard:
@thomasdraebing do you know of any way to produce the lockfile that had this bug? If not then this fix won't make it into rules_js v3...

thomasdraebing:
@jbedard Sorry for the delayed response.
Except for manual creation of the lock file, I don't. I came across that by switching from yarn (`yarn_install`) to pnpm. With yarn this worked perfectly well, with pnpm it did not although the spec indicates it should. I guess, a fix in pnpm has to come first though.

thomasdraebing:
I have created a bug in the pnpm project: https://github.com/pnpm/pnpm/issues/10468

thomasdraebing:
According to the discussion in https://github.com/pnpm/pnpm/issues/10468, this scenario is supported by pnpm. It was just a mistake of mine not to use the `git+` prefix while adding that dependency:

```
root@ab2a4dc830ac:/# pnpm add git+https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git\#6daccff309fca1e7561a43984d42fa4f829ce06d
Packages: +2
++
Progress: resolved 2, reused 0, downloaded 2, added 2, done

dependencies:
+ highlightjs-ttcn3 0.0.1

Done in 1.5s using pnpm v10.28.0
root@ab2a4dc830ac:/# cat p
package.json    pnpm-lock.yaml  proc/
root@ab2a4dc830ac:/# cat pnpm-lock.yaml
lockfileVersion: '9.0'

settings:
  autoInstallPeers: true
  excludeLinksFromLockfile: false

importers:

  .:
    dependencies:
      highlightjs-ttcn3:
        specifier: git+https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git#6daccff309fca1e7561a43984d42fa4f829ce06d
        version: git+https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git#6daccff309fca1e7561a43984d42fa4f829ce06d

packages:

  highlight.js@11.11.1:
    resolution: {integrity: sha512-Xwwo44whKBVCYoliBQwaPvtd/2tYFkRQtXDWj1nackaV2JPXx3L0+Jvd8/qCJ2p+ML0/XVkJ2q+Mr+UVdpJK5w==}
    engines: {node: '>=12.0.0'}

  highlightjs-ttcn3@git+https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git#6daccff309fca1e7561a43984d42fa4f829ce06d:
    resolution: {commit: 6daccff309fca1e7561a43984d42fa4f829ce06d, repo: https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git, type: git}
    version: 0.0.1

snapshots:

  highlight.js@11.11.1: {}

  highlightjs-ttcn3@git+https://gitea.osmocom.org/ttcn3/highlightjs-ttcn3.git#6daccff309fca1e7561a43984d42fa4f829ce06d:
    dependencies:
      highlight.js: 11.11.1
```

jbedard:
It looks like they also enhanced it a bit which will be released in the next v10 release: https://github.com/zys-contrib/pnpm/commit/ec7c5d7d1a9b30be23d3e2459fe78be7e2d05287

PR #2411, opened 2025-11-01
feat: support watch protocol v1

jbedard:
Update the `AspectWatchProtocol` class to support v1 of the protocol.

### Test plan

### DIFF
--- a/js/private/watch/src/aspect_watch_protocol.mts
+++ b/js/private/watch/src/aspect_watch_protocol.mts
@@ -11,11 +11,44 @@ export enum MessageType {
     CYCLE_COMPLETED = 'CYCLE_COMPLETED',
     NEGOTIATE = 'NEGOTIATE',
     NEGOTIATE_RESPONSE = 'NEGOTIATE_RESPONSE',
+    CAPS = 'CAPS',
+    CAPS_RESPONSE = 'CAPS_RESPONSE',
     EXIT = 'EXIT',
 }
 
 export interface Message {
     readonly kind: MessageType
+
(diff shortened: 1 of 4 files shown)


js/private/watch/src/aspect_watch_protocol.mts
```diff
@@ -124,6 +173,15 @@ export class AspectWatchProtocol {
      * build has completed before proceeding.
      */
     async awaitFirstCycle() {
+        if (this._version > 0) {
+            await this._send<SubscribeMessage>(MessageType.SUBSCRIBE, {
```
jbedard: Maybe this is just re-inventing the `CAP` message from the original proposal?

@thesayyn were you thinking that a socket has a **single** subscription, and the `CAP` message is essentially declaring the options for that socket/subscription? This PR does similar but "subscribe" seems to imply that you can have more then one subscription per socket - which is probably more complicated and not worth it?

thesayyn: my worry with SUBSCRIBE is that it blocks cycle messages from being sent until the implementor tells us what it needs. the name implies that it can be sent N many times and every subscription needs an id to distinguish the CYCLE events from each other.

jbedard: I'm starting to think we should keep the existing behaviour where 1 socket = 1 subscription, and this PR (and the associated protocol spec changes) should instead implement the original `CAP` message - which I think aligns with your original idea?

thesayyn: Yes, and revisit the `scope` idea where as part of the CAP, we can decide what the user is asking for.

PR #2412, opened 2025-11-01
test: update tests to pnpm v9+

jbedard:
Upgrade tests to use pnpm v9+ in preparation for dropping <v9 in the next major release.

### Test plan

- New test cases added


examples/nextjs/package.json
```diff
@@ -9,12 +9,12 @@
         "lint": "next lint"
     },
     "dependencies": {
-        "react": "^19.0.0",
```
jbedard, line 12: This makes the versions in the lockfile align better between pnpm 8+9. It seems pnpm v9 changed the ^ to only incremental the last semver specified (which sounds right to me 🤔). Without this some transitive deps change and actually caused this nextjs example to fail.

package.json
```diff
@@ -1,16 +1,19 @@
 {
     "private": true,
     "devDependencies": {
-        "@babel/cli": "^7.23.0",
-        "@babel/core": "^7.23.0",
-        "@babel/plugin-transform-modules-commonjs": "^7.23.0",
+        "@babel/cli": "^7.28.3",
```
jbedard, line 4: Also required to get the versions in the lockfile to align better between pnpm 8+9

examples/linked_consumer/test_pkg_deps_linked.js
```diff
@@ -21,7 +21,17 @@ try {
     e = ex
 }
 if (e == null) {
-    throw new Error('devDependency should NOT be available')
+    // TODO: see https://github.com/aspect-build/rules_js/issues/2013 and similar issues.
```
jbedard, line 24: <v9 the lockfile `dev` flag was used (incorrectly) for this feature, >=v9 that flag no longer exists so the (buggy) behaviour is gone but we have nothing replacing it yet...

dzbarsky (approved):

package.json
```diff
@@ -1,16 +1,27 @@
...
         "@types/node": "18.19.54",
         "chalk": "5.1.1",
         "inline-fixtures": "1.1.0",
         "jsonpath-plus": "7.2.0",
         "typescript": "^5.8.0"
     },
     "pnpm": {
+        "onlyBuiltDependencies": [
+            "@aspect-test/c",
+            "@figma/nodegit",
```
jbedard, line 16: @dzbarsky I added a few things here since you approved the PR.

Adding `fsevents` reduced the diff on the snapshot for that file. Adding `@kubernetes/client-node` (which I think I did before you approved) is what caused the `@kubernetes/client-node` `packageExtensions` entry down below due to pnpm9 deciding transitive deps differently and causing a tsc error with a newer version of `@types/node`.

PR #2421, opened 2025-11-05
chore: allow DirectoryPathInfo to vary bazel_lib vs aspect_bazel_lib

alexeagle:
During the migration users will have a mix of both.

For example rules_ts uses this provider, and can't switch to bazel_lib unless there's a release of its dependencies that permits either.


jbedard:
We can't do this without an explicit dep on `bazel-lib`, are we ready to do that? Or can it wait for rules_js v3? Or we might want to do it now and drop the `aspect_bazel_lib` support in v3?

alexeagle:
Yes we should add the bazel-lib 3 dep now, it's non-breaking. We can leave the back-compat to aspect_bazel_lib indefinitely, or sure you could drop it in the next major

js/private/js_binary.bzl
```diff
@@ -106,11 +107,11 @@ _ATTRS = {
 
         This is the module referenced by the `require.main` property in the runtime.
 
-        This must be a target that provides a single file or a `DirectoryPathInfo`
-        from `@aspect_bazel_lib//lib::directory_path.bzl`.
+        This must be a target that provides a single file or a `Info`
```
jbedard: typo

alexeagle: thanks

PR #2428, opened 2025-11-06
refactor: simplify generated npm_link_all_packages() code

jbedard:
Simplify the generated code so the initial `link_targets` and `scope_targets` values are assigned generated constants, then append the dynamic `npm_link_all_packages(imported_links)` at the end.

This allows both the generating and generated code to be simplified. The only side-effect _should be_ that the `imported_links` targets are now appended to the end of the `link_targets` array, however this should only effect the `js_library(name = "node_modules", srcs = link_targets)` where order should not matter.

### Test plan

### DIFF
--- a/e2e/npm_translate_lock_disable_hooks/snapshots/defs.bzl
+++ b/e2e/npm_translate_lock_disable_hooks/snapshots/defs.bzl
@@ -30,13 +30,9 @@ def npm_link_all_packages(name = "node_modules", imported_links = [], prod = Tru
     if link:
         if bazel_package == "":
             link_0("{}/@aspect-test/c".format(name), False, name, "@aspect-test/c")
-            link_targets = [
-                ":{}/@aspect-test/c".format(name),
-            ]
+            link_targets = [":{}/@aspect-test/c".format(name)]
             scope_targets = {
-                "@aspect-test": [
-                    ":{}/@aspect-test/c".format(name),
-                ],
+                "@aspect-test": [":{}/@aspect-test/c".format(name)],
             }
     
     for link_fn in imported_links:
@@ -79,7 +75,5 @@ def npm_link_targets(name = "node_modules", package = None, prod = True, dev = T
 
     if bazel_package == "":
         if prod:
-            link_targets.extend([
-                ":{}/@aspect-test/c".format(name),
-            ])
+            link_targets.extend([":{}/@aspect-test/c".format(name)])
     return link_targets
--- a/npm/private/starlark_codegen_utils.bzl
+++ b/npm/private/starlark_codegen_utils.bzl
@@ -3,6 +3,9 @@
 def _to_list_attr(list, indent_count = 0, indent_size = 4, quote_value = True):
     if not list:
         return "[]"
+    if len(list) == 1:
+        val = "\"{}\"".format(list[0]) if quote_value else list[0]
+        return "[%s]" % val
     tab = " " * indent_size
     indent = tab * indent_count
     result = "["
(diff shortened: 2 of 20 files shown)


jbedard:
FWIW I think reviewing the diff of `npm/private/test/snapshots/npm_defs.bzl` is the best summary of the actual change and easier to understand then reading the generating code.

npm/private/starlark_codegen_utils.bzl
```diff
@@ -3,6 +3,9 @@
 def _to_list_attr(list, indent_count = 0, indent_size = 4, quote_value = True):
     if not list:
         return "[]"
+    if len(list) == 1:
```
jbedard, line 6: To partially reduce side-effects from the other change in this file

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -668,6 +656,41 @@ def _gen_npm_import(rctx, _import, link_workspace):
         maybe_exclude_package_contents = maybe_exclude_package_contents,
     )
 
+def _generate_npm_link_targets(links_targets_bzl):
```
jbedard: Now that `links_targets_bzl` is built up in a nicer way (it is metadata about which links belong to which packages) it is easier to put the code-generation for `npm_link_targets` into an isolated util.

See the `npm/private/test/snapshots/npm_defs.bzl` diff to notice the generated code has hardly changed.

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -394,37 +374,20 @@ def npm_link_all_packages(name = "node_modules", imported_links = [], prod = Tru
                     if fp_link_package not in links_targets_bzl:
                         links_targets_bzl[fp_link_package] = {"prod": [], "dev": []}
 
-                    fp_append_stmt = """link_targets.append(":{{}}/{pkg}".format(name))""".format(pkg = fp_package)
+                    link_target = '":{{}}/{alias}".format(name)'.format(alias = fp_package)
```
jbedard: We are generating a list of targets now, not list of code snippets invoking `link_targets.append()`.

The "list of targets" still contains the ugly `.format(name)` but that will go away in v3 when `name == "node_modules"` 100% of the time.

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -383,9 +356,16 @@ def npm_link_all_packages(name = "node_modules", imported_links = [], prod = Tru
                         ),
                     )
 
-    # Add first-party packages to npm_link_targets before generating the function
-    for fp_link in fp_links.values():
+    # Generate the first-party package stores and linking of first-party packages
+    for i, fp_link in enumerate(fp_links.values()):
         fp_package = fp_link.get("package")
+        fp_path = fp_link.get("path")
```
jbedard, line 352: The two `fp_links` loops were combined, none of these vars actually changed in value

e2e/npm_translate_lock_disable_hooks/snapshots/defs.bzl
```diff
@@ -19,39 +19,47 @@ def npm_link_all_packages(name = "node_modules", imported_links = [], prod = Tru
...
+                name = "{}/{}".format(name, scope),
+                srcs = scoped_targets,
+                tags = ["manual"],
+                visibility = ["//visibility:public"],
+            )
 
     _js_library(
         name = name,
-        srcs = link_targets,
+        srcs = link_targets if link_targets else [],
```
dzbarsky, line 62: does `srcs = link_targets` not work (when link_targets is None)?

npm/private/npm_translate_lock_generate.bzl
```diff
@@ -256,15 +228,23 @@ def npm_link_all_packages(name = "node_modules", imported_links = [], prod = Tru
         ),
     ]
 
-    # The 3 code sections that will make up the body of npm_link_all_packages()
+    # Imports and header comments at the top of the generatedfile
```
dzbarsky, line 231: nit: "generatedfile" -> "generated file"

PR #2431, opened 2025-11-07
chore: test on bazel 9

alexeagle:


js/private/test/image/asserts.bzl
```diff
@@ -25,7 +25,7 @@ def assert_tar_listing(name, actual, expected):
         in_file = actual_listing,
         out_file = expected,
         testonly = True,
-        tags = ["skip-on-bazel6", "skip-on-bazel8"],
+        tags = ["skip-on-bazel6", "skip-on-bazel7", "skip-on-bazel8"],
```
jbedard, line 28: Why switch this to 9? Any version specific tests should align with .bazelversion by default imo

alexeagle: I'm planning to update .bazelversion to 9 in the next PR. No reason we're developing on 7

jbedard: Please don't do that right now 🤦

jbedard:
The aspect_rules_lint upgrade breaks bazel 6? I think we need to revert that until we start the 3.x branch and drop bazel 6/7, that is also where we can switch to bazel8 by default (although 8.4 is buggy and I think we're still waiting on 8.5?).

PR #2432, opened 2025-11-07
chore: bump bazel-lib to pick up bazel 9 fix

alexeagle:
need https://github.com/bazel-contrib/bazel-lib/commit/772d605a6ff3660c21b3476bb065e9433a584e7a

Might green up Bazel 9 downstream tests - see https://github.com/bazelbuild/bazel-central-registry/issues/6402

However #2431 adds Bazel 9 testing to our own presubmit.

### DIFF
--- a/e2e/stamped_package_json/WORKSPACE
+++ b/e2e/stamped_package_json/WORKSPACE
@@ -3,6 +3,15 @@ local_repository(
     path = "../..",
 )
 
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "package_metadata",
+    sha256 = "7446b71263039c7a5e90c4a069fd4c090e12a1a6c69a20eb4cd76034d610beb6",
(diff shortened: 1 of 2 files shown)


jbedard:
```
(02:52:31) ERROR: Failed to load Starlark extension '@@tar.bzl//tar:tar.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @@tar.bzl
This could either mean you have to add the '@@tar.bzl' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
```
Does this bazel-lib upgrade change its deps and something isn't propagating right?

alexeagle:
Yup there's a missing call to bazel_lib_install_dependencies here.

I got it green on the bigger PR.

jbedard (approved):

PR #2434, opened 2025-11-07
refactor: remove old aspect CLI

alexeagle:
Also reduce duplication in AW config

Prep for adding testing on Bazel 9

### DIFF
--- a/tools/BUILD.bazel
+++ b/tools/BUILD.bazel
@@ -17,6 +17,7 @@ format_multirun(
 
 bazelrc_preset(
     name = "preset",
+    doc_link_template = "https://registry.build/flag/bazel?filter={flag}",
     # The output is specific to the version in .bazelversion
     tags = [
         "skip-on-bazel6",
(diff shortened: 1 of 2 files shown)


.aspect/workflows/config.yaml
```diff
@@ -5,312 +5,122 @@ workspaces:
...
         icon: bazel
-        tasks:
+        tasks: &e2e_tasks
             - test:
                   queue: aspect-medium
             - format:
                   without: true
             - buildifier:
                   without: true
+            - bazel-8:
```
jbedard, line 15: @alexeagle doesn't this have a side-effect on every test?

alexeagle: What side-effect do you mean?

jbedard: I don't know, but removing this diff implies bazel-8 is no longer tested? I don't understand how that is related to "remove old aspect CLI"?

alexeagle: Bazel 8 was introduced to the test matrix in this commit, so this is just keeping e2e tests the same

jbedard: Oh I see, thanks 👍 

So we'll start enabling multiple bazel versions in workflows now?

PR #2435, opened 2025-11-07
fix: do not override all snapshots of a package due to one tarball override

### LINKED ISSUE #2300: [Bug]: npm_translate_lock erroneously resolves importer link references to tarball if a tarball packages entry exists
### What happened?

Hi there, it looks like when a common workspace project dependency is dependend upon both via `link` and via tarball `file` in the same workspace, the logic in `npm_translate_lock` will result in all references to the common dependency pointing to the tarball file version, which is surprising behavior.

My expectation is that only projects that reference the common dependency with `"@my-scope/project-dep": "file:my_registry/@my-scope/project-dep/0.2.0/my-scope-project-dep-0.2.0.tgz"` in their package.json files will refer to the tarbal, while packages that refer to the common dependency as `"@my-scope/project-dep": "workspace:*"` in their package.json will point directly to the source in the repo.

I believe this is caused by https://github.com/aspect-build/rules_js/blob/03bde54fd8e3de4c57e84247dffef9c450d956d5/npm/private/transitive_closure.bzl#L107-L112 re-writing all deps to use the tarball version if it exists.  Does it make sense to add `and deps[info["name"]].startswith("file:")` to the if statements here?  

### Version

Development (host) and target OS/architectures: macos

Output of `bazel --version`:

```
% bazel version
Bazelisk version: v1.21.0
Bazelisk version: v1.21.0
INFO: Invocation ID: 96afa5f7-7ca2-4d59-b339-18ad507430c3
Build label: 8.3.1
Build target: @@//src/main/java/com/google/devtools/build/lib/bazel:BazelServer
Build time: Mon Jun 30 16:26:17 2025 (1751300777)
Build timestamp: 1751300777
Build timestamp as int: 1751300777
```

```
bazel_dep(name = "aspect_rules_js", version = "2.4.0")
```

```
JavaScript
```

### How to reproduce

Setup a pnpm-workspace with 3 projects, `project-dep`, `project-a` and `project-b`.

In `project-a`'s package.json:

```
"dependencies": {
    "@my-scope/project-dep": "file:my_registry/@my-scope/project-dep/0.2.0/my-scope-project-dep-0.2.0.tgz",
}
```

In `project-b`'s package.json:

```
"dependencies": {
    "@my-scope/project-dep": "workspace:*",
}
```

regenerate the pnpm-lock.yaml file. Then re-build a `npm_translate_lock`/`npm_link_all_packages`-based `js_library` target for `project-b` and notice that it erroneously is referring to the 0.2.0 tarball version of `project-dep`, as opposed to the version from the source tree.

### Any other information?

_No response_

jbedard:
Fix #2300

### Test plan

- New test cases added

### DIFF
--- a/e2e/pnpm_lockfiles/v90/snapshots/defs.bzl
+++ b/e2e/pnpm_lockfiles/v90/snapshots/defs.bzl
@@ -47,7 +47,7 @@ load("@lock-<LOCKVERSION>__is-odd__3.0.1__links//:defs.bzl", link_43 = "npm_link
 load("@lock-<LOCKVERSION>__jquery__https___codeload.github.com_jquery_jquery_tar.gz_399b201bb3143a3952894cf3489b4848fc003967__links//:defs.bzl", link_44 = "npm_link_imported_package_store_internal", store_44 = "npm_imported_package_store_internal")
 load("@lock-<LOCKVERSION>__jquery__https___codeload.github.com_jquery_jquery_tar.gz_e61fccb9d736235b4b011f89cba6866bc0b8997d__links//:defs.bzl", link_45 = "npm_link_imported_package_store_internal", store_45 = "npm_imported_package_store_internal")
 load("@lock-<LOCKVERSION>__kleur__4.1.5__links//:defs.bzl", store_46 = "npm_imported_package_store_internal")
-load("@lock-<LOCKVERSION>__lodash__4.0.1__links//:defs.bzl", store_47 = "npm_imported_package_store_internal")
+load("@lock-<LOCKVERSION>__lodash__4.0.1__links//:defs.bzl", link_47 = "npm_link_imported_package_store_internal", store_47 = "npm_imported_package_store_internal")
 load("@lock-<LOCKVERSION>__lodash__4.17.20__links//:defs.bzl", link_48 = "npm_link_imported_package_store_internal", store_48 = "npm_imported_package_store_internal")
 load("@lock-<LOCKVERSION>__lodash__4.17.21__links//:defs.bzl", link_49 = "npm_link_imported_package_store_internal", store_49 = "npm_imported_package_store_internal")
 load("@lock-<LOCKVERSION>__lodash__file_.._vendored_lodash-4.17.21.tgz__links//:defs.bzl", link_50 = "npm_link_imported_package_store_internal", store_50 = "npm_imported_package_store_internal")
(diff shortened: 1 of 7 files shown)


jbedard:
Depends on https://github.com/aspect-build/rules_js/pull/2445 so lets merge that first. Only review the second commit if you want to review this PR on its own.

jbedard:
It looks like this actually only fixed it for pnpm v9+. I think a proper solution would be to normalize the lockfile data better, where today it comes out slightly different depending on which pnpm version was used. I'll play with it a bit but I plan to drop pnpm <9 support so I don't want to refactor anything that we plan to throw out anyway...

dzbarsky:
> It looks like this actually only fixed it for pnpm v9+. I think a proper solution would be to normalize the lockfile data better, where today it comes out slightly different depending on which pnpm version was used. I'll play with it a bit but I plan to drop pnpm <9 support so I don't want to refactor anything that we plan to throw out anyway...

Sounds like more carrots to migrate

jbedard:
If I can improve the pnpm v9+ code path without too much <v9 hassle then I might do this now. Otherwise this existing PR fixes pnpm 9+ and we can just wait for rules_js v3, which I hope to start soon...

jbedard:
FWIW this will require a bit more cleanup, which is now possible in 3.x with pnpm <9 dropped.

e2e/pnpm_lockfiles/v90/snapshots/defs.bzl
```diff
@@ -47,7 +47,7 @@ load("@lock-<LOCKVERSION>__is-odd__3.0.1__links//:defs.bzl", link_43 = "npm_link
 load("@lock-<LOCKVERSION>__jquery__https___codeload.github.com_jquery_jquery_tar.gz_399b201bb3143a3952894cf3489b4848fc003967__links//:defs.bzl", link_44 = "npm_link_imported_package_store_internal", store_44 = "npm_imported_package_store_internal")
 load("@lock-<LOCKVERSION>__jquery__https___codeload.github.com_jquery_jquery_tar.gz_e61fccb9d736235b4b011f89cba6866bc0b8997d__links//:defs.bzl", link_45 = "npm_link_imported_package_store_internal", store_45 = "npm_imported_package_store_internal")
 load("@lock-<LOCKVERSION>__kleur__4.1.5__links//:defs.bzl", store_46 = "npm_imported_package_store_internal")
-load("@lock-<LOCKVERSION>__lodash__4.0.1__links//:defs.bzl", store_47 = "npm_imported_package_store_internal")
+load("@lock-<LOCKVERSION>__lodash__4.0.1__links//:defs.bzl", link_47 = "npm_link_imported_package_store_internal", store_47 = "npm_imported_package_store_internal")
```
jbedard, line 50: Notice how this lodash version is actually linked now, where previously anything named "lodash" was blindly overwritten with the `file:` version

PR #2437, opened 2025-11-07
chore: sanitize some file sizes in golden files

alexeagle:
Prefactor for Bazel 9 which has a minor change to the binary launcher making these a few bytes different in size.


js/private/test/image/asserts.bzl
```diff
@@ -23,7 +25,7 @@ def assert_tar_listing(name, actual, expected):
         in_file = actual_listing,
         out_file = expected,
         testonly = True,
-        tags = ["skip-on-bazel6", "skip-on-bazel8"],
+        tags = ["skip-on-bazel6", "skip-on-bazel7"],
```
jbedard: Version 7 is currently our default version tested on CI. I don't think we should be changing this until we change our default version?

PR #2440, opened 2025-11-07
chore: downgrade rules_lint to 1.1.0 for bazel6 compat

jbedard:
### Test plan


jbedard:
rules_lint needs an upgrade to fix a `sh_binary` issue for bazel9, but upgrading rules_lint breaks bazel6 - we need to disable testing of the root module on bazel9 for now...

PR #2441, opened 2025-11-07
chore(deps): bump bazel to 9

alexeagle:


jbedard:
We shouldn't be depending on an RC version by default like this, please no. The default version we use should be the primary version we support which is 7-8 atm.

Lets switch to v9 once a) it is released and b) we have a 3.x branch where we can drop v6-7 at the same time.

PR #2443, opened 2025-11-07
test: run tests on same version as .bazelversion

jbedard:
By default tests must pass without modifying `.bazelversion` files

### Test plan


js/private/test/image/asserts.bzl
```diff
@@ -52,7 +52,7 @@ def assert_js_image_layer_listings(name, js_image_layer, additional_layers = [])
             "assert_{}_{}".format(name, layer)
             for layer in all_layers
         ],
-        tags = ["skip-on-bazel6", "skip-on-bazel7", "skip-on-bazel8"],
+        tags = ["skip-on-bazel6", "skip-on-bazel8", "skip-on-bazel9"],
```
alexeagle, line 55: if we only tested the root module against the "devDependency" version of Bazel, we wouldn't need any skips at all - that's the desired right?

jbedard: If I make no changes I think `bazel test` should pass. That is my initial goal and what this pr accomplishes.

If we want to run alternate versions of bazel then additional changes might need to be made such as excluding tags etc

PR #2445, opened 2025-11-08
fix: treat file: references in package store the same as pnpm

jbedard:
This aligns rules_js with pnpm when it comes to `file:` references in the package store, as well as bazel target throughout the various store/link targets.

Now only `workspace:*` deps (which resolve to `link:path` versions) will continue to use `@0.0.0` in the package store. This is different then pnpm where local workspace deps are symlinks into source and do not exist in the package store at all.

Ref #2300 which will depend on this change.

### Test plan

- New test cases added

### DIFF
--- a/e2e/pnpm_lockfiles/v101/snapshots/lodash-4.17.21_tgz_defs.bzl
+++ b/e2e/pnpm_lockfiles/v101/snapshots/lodash-4.17.21_tgz_defs.bzl
@@ -12,7 +12,7 @@ load("@aspect_rules_js//npm/private:npm_import.bzl",
 PACKAGE = "lodash"
 VERSION = "file:../vendored/lodash-4.17.21.tgz"
 _ROOT_PACKAGE = "<LOCKVERSION>"
-_PACKAGE_STORE_NAME = "lodash@0.0.0"
+_PACKAGE_STORE_NAME = "lodash@file+..+vendored+lodash-4.17.21.tgz"
 
 # Generated npm_package_store targets for npm package lodash@file:../vendored/lodash-4.17.21.tgz
 # buildifier: disable=function-docstring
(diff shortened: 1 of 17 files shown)


npm/private/npm_translate_lock_generate.bzl
```diff
@@ -121,6 +121,7 @@ js_binary(name = "sync", entry_point = "noop.js")
                 transitive_deps[dep_store_target] = ",".join(transitive_deps[dep_store_target])
             fp_links[dep_key] = {
                 "package": name,
+                "version": version,
```
jbedard, line 124: This causes a `.get("version", "0.0.0")` to now use the raw `file:` version instead of dropping it. Then the change in `utils.bzl` is the other spot that was previously dropping `file:` versions.

e2e/pnpm_lockfiles/v101/snapshots/lodash-4.17.21_tgz_defs.bzl
```diff
@@ -12,7 +12,7 @@ load("@aspect_rules_js//npm/private:npm_import.bzl",
 PACKAGE = "lodash"
 VERSION = "file:../vendored/lodash-4.17.21.tgz"
 _ROOT_PACKAGE = "<LOCKVERSION>"
-_PACKAGE_STORE_NAME = "lodash@0.0.0"
```
jbedard, line 15: Notice how the package store name here and the `VERSION` above were different. Now they align, and the store name is the exact same as the pnpm store name (including `file+..+` chars).

jbedard:
FWIW the `js_image_layer` makes use of the `@0.0.0` string to separate first-party vs third-party packages into different layers.  This fix will cause vendored third-party packages to go into the 3p layer, while pnpm workspace projects remain in the 1p layer. IMO this aligns with pnpm better and is a positive change.

FYI @thesayyn

PR #2449, opened 2025-11-09
refactor: use pnpm file: id field

jbedard:
... and add a `fail` in the `if` where we extract all info from the key and are assuming there is no explicit `name/version/id`. If this `fail` is ever hit we should fix it.

### Test plan

### DIFF
--- a/e2e/pnpm_lockfiles/v54/pnpm-lock.yaml
+++ b/e2e/pnpm_lockfiles/v54/pnpm-lock.yaml
@@ -142,8 +142,11 @@ importers:
   ../projects/c:
     specifiers:
       '@scoped/a': link:../a
+      a-types: file:../a-types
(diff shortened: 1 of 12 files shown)


e2e/pnpm_lockfiles/v54/pnpm-lock.yaml
```diff
@@ -578,6 +581,13 @@ packages:
     name: hello
     version: 1.0.0
 
+  file:../projects/a-types:
```
jbedard, line 584: In v54-61 lockfiles this is a case where there is a `file:` reference to a project with no version in it. So we have a `name` field but no `version` or `id` and fallback to just using the key (this line) as the `package_key` / "version"

dzbarsky (approved):

PR #2456, opened 2025-11-10
refactor: remove support for pnpm <v9

jbedard:

Remove support for PNPM versions 8 and lower. Only modern pnpm 9 and 10+ are supported.

Update the default pnpm version to `^10.0`.

### Test plan


npm/extensions.bzl
```diff
@@ -231,7 +231,6 @@ WARNING: Cannot determine home directory in order to load home `.npmrc` file in
             commit = i.commit,
             custom_postinstall = i.custom_postinstall,
             deps = i.deps,
-            dev = i.dev_only,
```
jbedard, line 234: This is also key. The `npm_import(dev)` and `NpmPackageStoreInfo.dev` flags are gone along with the `npm_imported_package_store_internal(dev_only)` and the `struct(dev_only)`.

PR #2457, opened 2025-11-10
refactor: upgrade aspect_bazel_lib to bazel_lib

jbedard:
### Test plan

### DIFF
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -15,11 +15,11 @@ js_binary(
 ```
 """
 
-load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "COPY_FILE_TO_BIN_TOOLCHAINS")
 load("@aspect_bazel_lib//lib:directory_path.bzl", _LegacyDirectoryPathInfo = "DirectoryPathInfo")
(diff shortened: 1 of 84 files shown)


js/private/js_binary.bzl
```diff
@@ -15,11 +15,10 @@ js_binary(
 ```
 """
 
-load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "COPY_FILE_TO_BIN_TOOLCHAINS")
-load("@aspect_bazel_lib//lib:directory_path.bzl", _LegacyDirectoryPathInfo = "DirectoryPathInfo")
```
jbedard, line 19: Should we keep this?

I assume this will make upgrading to 3.x a lot harder for people, it might require users upgrade a lot of rulesets all at once etc.

I propose we keep this, but set the `aspect_bazel_lib` version in MODULE to 2.0.0? However that means we are stuck with that dependency declaration until rules_js v4?

jbedard: WDYT @alexeagle?

jbedard: I've reverted to keep this 2.0 for now...

PR #2467, opened 2025-11-11
chore: upgrade rules_nodejs to 6.6

jbedard:
### Test plan

### DIFF
--- a/docs/js_binary.md
+++ b/docs/js_binary.md
@@ -21,6 +21,8 @@ js_binary(
 ## js_binary
 
 <pre>
+load("@aspect_rules_js//js/private:js_binary.bzl", "js_binary")
+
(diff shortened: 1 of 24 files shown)


docs/js_binary.md
```diff
@@ -21,6 +21,8 @@ js_binary(
 ## js_binary
 
 <pre>
+load("@aspect_rules_js//js/private:js_binary.bzl", "js_binary")
```
jbedard, line 24: FYI @alexeagle upgrading rules_nodejs must have caused a transitive dep to be upgraded and now stardoc is outputting these.

I guess the doc comments are within the `/private` implementation so it outputs the "wrong" path here. Have you seen this elsewhere?

Issue #2379, opened 2025-10-06
[Bug]: `js_run_devserver` does not execute with updated `data` values starting with Bazel 8.4.0


scottpledger:
### What happened?

Changes to the `data` attribute of `js_run_devserver` can result in I/O errors like:

```
[Error: ENOENT: no such file or directory, lstat '/private/var/tmp/_bazel_scott.pledger/90d2640d2b39ca9b655f207fb9f07c7b/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/apps/testApp/dev_/dev.runfiles/_main/apps/testApp/index.html'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'lstat',
  path: '/private/var/tmp/_bazel_scott.pledger/90d2640d2b39ca9b655f207fb9f07c7b/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/apps/testApp/dev_/dev.runfiles/_main/apps/testApp/index.html'
}
Trace
    at syncRecursive (file:///private/var/tmp/_bazel_scott.pledger/90d2640d2b39ca9b655f207fb9f07c7b/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/apps/testApp/dev_/dev.runfiles/aspect_rules_js+/js/private/js_run_devserver.mjs:284:17)
```

Specifically, this error is triggered by having a `js_run_devserver` rule where I run it with an `index.html` `data` entry first, then remove that entry and run it again.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: N/A

By bisecting https://github.com/bazelbuild/bazel/compare/8.3.1...8.4.0, I've narrowed this down to one of two Bazel commits:

- [8.4.0] Set working directory in interactive --sandbox_debug command - https://github.com/bazelbuild/bazel/commit/b8abd841730e7aee9abe8fb0da614315f8e5d485
- [8.4.0] Fix stale action key with expand_directories=False and manual expansion https://github.com/bazelbuild/bazel/commit/84cd58ca51fd32bb9f18433fee870f6cfad9f807

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: 2.6.0

Language(s) and/or frameworks involved:
JS + Vite, though this seems to be an issue independent of the underlying binary and has something to do with the js_run_devserver.mjs binary itself and/or the rule that manages it.

### How to reproduce

```shell
[...]

PR #2470, opened 2025-11-12
fix: Expand directories in js_run_devserver

purkhusid:
This fixes a bug I encountered where it seems like the args file that is being constructed is not invalidated when files that are passed into `js_run_devserver` are renamed.

The exact steps I had to do to reproduce this were:
1. Have a NextJS project which takes in a `ts_project` as input in srcs
2. Run `nextjs_dev` and everything works fine
3. Rename a .ts file
4. Run `nextjs_dev` again
5. `js_run_devserver` complains that it can't find a file with the same name as BEFORE the rename.
6. Run `bazel clean`
7. Run `nextjs_dev` again
8. Everything works again

I narrowed this bug down to the changes in this PR: https://github.com/aspect-build/rules_js/pull/2255

### DIFF
--- a/js/private/js_run_devserver.bzl
+++ b/js/private/js_run_devserver.bzl
@@ -73,7 +73,7 @@ def _js_run_devserver_impl(ctx):
     entries.add("[")
     entries.add_joined(
         depset(transitive = transitive_runfiles + [dep.files for dep in ctx.attr.data] + default_data_runfiles),
-        expand_directories = False,
+        expand_directories = True,
         map_each = _file_to_entry_json,
         join_with = ",",
     )


purkhusid:
cc @jbedard since you seem to be the expert on this

jbedard:
What version of bazel are you using? This sounds like https://github.com/aspect-build/rules_js/issues/2379?

purkhusid:
I'm on 8.4.2

I downgraded to 8.3.1 and could not reproduce this there. I guess I'll just have to downgrade until 8.5.0 is released.
Thanks for taking a look, I'll close this issue for now.

jbedard:
Thanks for confirming that :+1:

PR #2478, opened 2025-11-13
refactor: extricate lockfile verification from npm_translate_lock repo rule

dzbarsky:
After this we should be able to do the work in module extension pretty easily. I'm not sure about the bootstrap step, may need a hack. let's see how it goes!

---

### Test plan

### DIFF
--- a/npm/extensions.bzl
+++ b/npm/extensions.bzl
@@ -69,13 +69,23 @@ def _npm_extension_impl(module_ctx):
     # Process npm_translate_lock and npm_import tags
     for mod in module_ctx.modules:
         for attr in mod.tags.npm_translate_lock:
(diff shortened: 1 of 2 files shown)


npm/extensions.bzl
```diff
@@ -132,6 +142,10 @@ The 'replace_packages' attribute will be removed in rules_js version 3.0.
                 fail("Package replacement conflict: {} specified in both replace_packages attribute and npm_replace_package tag".format(package))
             replace_packages[package] = replacement
 
+    # TODO(zbarsky): Use these
```
jbedard, line 145: Is there any reason to merge this PR before doing this?

dzbarsky: Just wanted to keep this no-op to make actual change smaller

jbedard (approved):

PR #2482, opened 2025-11-14
test: importing of dev alias of a non-dev dep

jbedard:
### Test plan

- New test cases added


npm/private/test/snapshots/npm_defs.bzl
```diff
@@ -2568,35 +2568,41 @@ def npm_link_all_packages(name = "node_modules", imported_links = [], prod = Tru
         elif bazel_package == "examples/linked_lib":
             link_45("node_modules/@aspect-test/e", False, "@aspect-test/e")
             link_45("node_modules/alias-e", False, "alias-e")
+            link_45("node_modules/@aspect-test/e-dev", True, "@aspect-test/e-dev")
```
jbedard, line 2571: Notice `link_45` is called 3 times here with the default name, then with 2 aliases, the second alias being the new one that also has `dev=True`...

PR #2486, opened 2025-11-14
refactor: remove npm_import(link_packages)

jbedard:

Remove `npm_import(link_packages)` which caused some npm repo contents to be clone for each pnpm-workspace location it is imported.

### Test plan


npm/private/npm_import.bzl
```diff
@@ -272,16 +272,11 @@ def npm_link_imported_package_internal(
         dev,
         root_package,
         link,
-        link_packages,
         public_visibility,
         npm_link_imported_package_store_macro,
         npm_imported_package_store_macro):
     bazel_package = native.package_name()
```
jbedard, line 279: This is the (unwrapped) method that is documented as `npm_link_imported_package` from each npm package. For example in the root BUILD we have:
```
load("@acorn__8.4.0__links//:defs.bzl", npm_link_acorn = "npm_link_imported_package")
```

which we pass to:

```
npm_link_all_packages(
    imported_links = [
        npm_link_acorn,
    ],
)
```

However I don't think the `link_packages` param deleted from this method is ever used, so that might not even be a breaking change (where removing `npm_import(link_packages)` is breaking).

npm/private/npm_import.bzl
```diff
@@ -1115,13 +1082,6 @@ def npm_import(
 
             Can be left unspecified if the link workspace is the user workspace.
 
-        link_packages: Dict of paths where links may be created at for this package to
```
jbedard, line 1118: Breaking change 1:  I have never seen this used

npm/private/npm_import.bzl
```diff
@@ -673,20 +660,16 @@ def _npm_import_rule_impl(rctx):
                 bin_fields = "\n".join(bin_fields),
             ))
 
-            # Duplicate the bzl+BUILD into each linking package
-            for link_package in rctx.attr.link_packages.keys():
```
jbedard, line 677: Breaking change 2: each `npm__{pkg}__{version}` repo now has one `package_json.bzl` in the root. It is no longer duplicated into magic directories parallel to the pnpm workspace project directories.

If you want to import this package it is recommended to do `@npm//your/workspace/project:{pkg}/package_json.bzl` instead.

PR #2510, opened 2025-11-19
refactor: enable strict type checking of fs patches src

jbedard:
### Test plan


js/private/node-patches/fs.cjs
```diff
@@ -282,19 +284,14 @@ function patcher(roots) {
...
-            if (!next) {
-                if (next == undefined) {
-                    // The escape from the root is not mappable back into the root; throw EINVAL
-                    throw enoent('readlink', args[0]);
-                }
-                else {
-                    // The escape from the root is not mappable back into the root; throw EINVAL
-                    throw einval('readlink', args[0]);
-                }
+            if (next === false) {
```
jbedard: This is the only functional difference AFAICT. Previously an empty string would return an error, now... idk if/when that is applicable 🤷

dzbarsky (approved):

js/private/node-patches/fs.cjs
jbedard: I've reverted this and instead changed the method signature to align with how it was actually implemented. I'll look at changing the it functionally later maybe 🤷

jbedard:
FWIW I did this because the incorrect types were confusing me in https://github.com/aspect-build/rules_js/pull/2366

PR #2525, opened 2025-11-25
refactor: do not create empty name="node_modules" targets

jbedard:
### Test plan

### DIFF
--- a/npm/private/npm_translate_lock_generate.bzl
+++ b/npm/private/npm_translate_lock_generate.bzl
@@ -216,8 +216,8 @@ def npm_link_all_packages(name = "node_modules", imported_links = [], prod = Tru
     bazel_package = native.package_name()
     root_package = "{root_package}"
     is_root = bazel_package == root_package
-    link = bazel_package in _IMPORTER_PACKAGES
-    if not is_root and not link:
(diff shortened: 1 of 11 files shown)


npm/private/npm_translate_lock_generate.bzl
```diff
@@ -456,23 +456,23 @@ def npm_link_all_packages(name = "node_modules", imported_links = [], prod = Tru
...
-    _js_library(
-        name = name,
-        srcs = link_targets if link_targets else [],
-        tags = ["manual"],
-        visibility = ["//visibility:public"],
-    )""")
+    if is_importer:
+        _js_library(
+            name = "node_modules",
+            srcs = link_targets if link_targets else [],
```
dzbarsky, line 472: It's still empty if not lonk_targets, no? Should we condition on both of those?

dzbarsky (approved):

npm/private/npm_translate_lock_generate.bzl
jbedard: Correct, I'm not sure if we should or should not create the target though.

We have `e2e/npm_translate_lock_empty` today that verifies `:node_modules` exists even though we `npm_link_all_packages()` produces no targets because `no_dev = True` 🤷

PR #2526, opened 2025-11-26
test: remove use of gitpkg.vercel.app tarball in tests

jbedard:
This vercel tarball URL now returns a [`402 Payment Required` response](https://github.com/aspect-build/rules_js/actions/runs/19521588508/job/56394858916#step:6:114).

This test case seems to be for pnpm <9 where for some reason the key in the yaml file startswith("@") which is an odd edge case. With rules_js v3 removing pnpm <9 this test will no longer be necessary in the future so I have not bothered finding a replacement.

### Test plan


e2e/pnpm_lockfiles/v54/pnpm-lock.yaml
```diff
@@ -576,11 +572,6 @@ packages:
     engines: {node: '>=0.3.1'}
     dev: false
 
-  '@gitpkg.vercel.app/EqualMa/gitpkg-hello/packages/hello':
```
jbedard, line 579: Here you can see the odd yaml key

PR #2535, opened 2025-11-27
refactor: do not resort npm packages multiple times

jbedard:
These are already deterministic based on the pnpm-lock.yaml order.

### Test plan


npm/private/npm_translate_lock_generate.bzl
```diff
@@ -613,9 +613,6 @@ def _generate_npm_package_locations(fp_links, npm_imports):
                     if alias not in location_to_packages[link_package]:
                         location_to_packages[link_package].append(alias)
 
-    # TODO(zbarsky): `sorted_map` could just return the dict's items which is what `to_dict_attr` wants anyway.
-    location_to_packages = utils.sorted_map(location_to_packages)
```
jbedard, line 617: I guess I'm not certain about this one since it is based on the order of `npm_imports` (which should be the order it was in the lockfile) but then also the `.link_packages.items()` that I would not trust as much? 🤷

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -465,13 +465,7 @@ ERROR: can not apply both `pnpm.patchedDependencies` and `npm_translate_lock(pat
             replace_package = replace_package,
         )
 
-        if repo_name in result:
```
jbedard, line 474: This is just (an escaped version of) `{name}__{package_key}` and the key should already be unique. It's not worth the extra code here...

e2e/pnpm_lockfiles/v101/snapshots/defs.bzl
```diff
@@ -1,29 +1,29 @@
...
-load("@lock-<LOCKVERSION>__at_types_archiver__5.3.1__links//:defs.bzl", link_18 = "npm_link_imported_package_store_internal", store_18 = "npm_imported_package_store_internal")
-load("@lock-<LOCKVERSION>__at_types_estree__0.0.39__links//:defs.bzl", store_19 = "npm_imported_package_store_internal")
-load("@lock-<LOCKVERSION>__at_types_glob__8.1.0__links//:defs.bzl", store_20 = "npm_imported_package_store_internal")
-load("@lock-<LOCKVERSION>__at_types_minimatch__5.1.2__links//:defs.bzl", store_21 = "npm_imported_package_store_internal")
-load("@lock-<LOCKVERSION>__at_types_node__18.19.54__links//:defs.bzl", link_22 = "npm_link_imported_package_store_internal", store_22 = "npm_imported_package_store_internal")
-load("@lock-<LOCKVERSION>__at_types_sizzle__2.3.10__links//:defs.bzl", link_23 = "npm_link_imported_package_store_internal", store_23 = "npm_imported_package_store_internal")
+load("@lock-<LOCKVERSION>__at_aspect-test_a__5.0.2__links//:defs.bzl", link_0 = "npm_link_imported_package_store_internal", store_0 = "npm_imported_package_store_internal")
```
jbedard, line 3: This no longer resorts after escaping the names so `@` is now before `a` just like in the lockfile...

npm/private/npm_translate_lock_generate.bzl
dzbarsky: could leave it for now, since there's still the opportunity to optimize as the comment suggests?

dzbarsky (approved):

npm/private/npm_translate_lock_generate.bzl
jbedard: I'd rather delete it and we can restore if we notice the order randomly changing on us 🤷

PR #2537, opened 2025-11-28
refactor: npm_translate_lock simplifying and cleanup

jbedard:
### Test plan

### DIFF
--- a/npm/extensions.bzl
+++ b/npm/extensions.bzl
@@ -71,9 +71,7 @@ def _npm_extension_impl(module_ctx):
         for attr in mod.tags.npm_translate_lock:
             state = parse_and_verify_lock(module_ctx, attr.name, attr)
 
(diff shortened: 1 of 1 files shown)


npm/extensions.bzl
```diff
@@ -134,22 +124,9 @@ _hub_repo = repository_rule(
     },
 )
 
-def _npm_translate_lock_bzlmod(mctx, attr, state, npm_imports):
-    mctx.report_progress("Generating starlark for npm dependencies")
+def _npm_translate_lock_bzlmod(module_ctx, attr, state, exclude_package_contents_config, replace_packages):
```
dzbarsky, line 127: personally i prefer `mctx` cause its less noisy but we can keep things consistent and maybe swap in a followup

jbedard: That would require renaming the var throughout the method so I'm going to keep the diff minimal for now

PR #2538, opened 2025-11-29
feat: platform specific optional npm deps

### LINKED ISSUE #714: [FR]: provide select constraints for optional dependencies that define specific platforms

Currently when an NPM package contains optional dependencies, and `npm_translate_lock` has `no_optional = False`, all of the optional dependencies of a package are added without constraints deps attribute of the rerespective `npm_package_store_internal` for the desired package. 

In the case of nextjs, for example, it means that all of these [optional dependencies](https://github.com/aspect-build/bazel-examples/blob/ec19825e711ef2d1db68352ed8b152cad65cd257/next.js/pnpm-lock.yaml#L2298-L2311) will be added to the deps. This results in all of the various platform binaries for @next/swc being installed on the user's machine.

### Describe the feature

I am looking for something similar to this pnpm PR for rules_js: https://github.com/pnpm/pnpm/pull/3672

In short, I'd expect that all optional dependencies that provide a platform constraint, have the constraint mapped into a proper bazel select when adding the dependencies to the store target. We should be able to achieve this using the pnpm-lock file, as seen on this example swc-linux package: https://github.com/aspect-build/bazel-examples/blob/ec19825e711ef2d1db68352ed8b152cad65cd257/next.js/pnpm-lock.yaml#L268-L275

The end goal is that the build stage would only depend on dependencies that match the allowed host, which would result in fewer unneeded binaries from being installed when using packages like nextjs.

I can provide more information if necessary. 

### Fund our work

### LINKED ISSUE #2140: [Bug]: npm_translate_lock should respect [os] and [cpu] flags
### What happened?

See https://github.com/pnpm/spec/blob/master/lockfile/9.0.md#packagesdependencyidos

### Version

Output of `bazel --version`:

### How to reproduce

```shell

```

### Any other information?

_No response_

jbedard:
Initial attempt at `select()` constraints based on pnpm `os` and `cpu` flags.

Current issues/WIP:
* transitive dep constraints don't get carried through when the `transitive_closure` attribute is used

Fix #714
Close #2140

### Test plan

- New test cases added


npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -495,6 +499,27 @@ Either remove this patch file if it is no longer needed or change its key to mat
...
+################################################################################
+def _dep_constraints(packages, package_info):
+    constraints = {}
+
+    for name, key in package_info["optional_dependencies"].items():
+        # Only set constraints on optional deps
+        if name in package_info["dependencies"]:
+            continue
+
+        if packages[key]["os"] and packages[key]["cpu"]:
```
dzbarsky: nit: store `packages[key]` in a local? maybe os and cpu as well?

npm/private/pnpm.bzl
```diff
@@ -44,8 +47,19 @@ def _new_package_info(name, dependencies, optional_dependencies, has_bin, option
...
         "resolution": resolution,
+        "cpu": cpu,
+        "os": os,
     }
 
+def _to_bazel_os_cpu_constraint(os, cpu):
+    return "@aspect_rules_js//platforms/pnpm:{}_{}".format(os, cpu)
+
+def _to_bazel_os_constraint(os):
+    return "@aspect_rules_js//platforms/pnpm:{}".format(os)
```
dzbarsky, line 71: should we just map these directly to `@platforms//os:_` and same with cpu? That feels clearer than mixing these with "fuller" constraints in same package

platforms/pnpm/BUILD.bazel
```diff
@@ -0,0 +1,60 @@
...
+        visibility = ["//visibility:public"],
+    )
+    for p in PNPM_PLATFORMS.keys()
+]
+
+# Available pnpm architectures, may map to incompatible
+[
+    alias(
+        name = a,
+        actual = PNPM_ARCHS[a] if PNPM_ARCHS[a] else "@platforms//:incompatible",
```
dzbarsky: for ones that are incompatible, should we just not generate them? They can't work anyway and it will lead to a smaller build graph

platforms/pnpm/index.bzl
```diff
@@ -0,0 +1,39 @@
...
+    "darwin": "@platforms//os:macos",
+    "freebsd": "@platforms//os:freebsd",
+    "linux": "@platforms//os:linux",
+    "openbsd": "@platforms//os:openbsd",
+    "win32": "@platforms//os:windows",
+    "sunos": None,  # TODO: confirm
+
+    # TODO: confirm
+    # Additional platforms found in pnpm [os] fields not on Node.js docs
+    "android": None,
```
dzbarsky: https://github.com/bazelbuild/platforms/blob/5cf94563e35494b0dab15435868dd7f9e3cab2c8/os/BUILD#L37

platforms/pnpm/index.bzl
```diff
@@ -0,0 +1,39 @@
...
+    "netbsd": None,
```
dzbarsky: https://github.com/bazelbuild/platforms/blob/5cf94563e35494b0dab15435868dd7f9e3cab2c8/os/BUILD#L22

platforms/pnpm/BUILD.bazel
jbedard: If they truly are incompatible then ideally we wouldn't create this `alias` and they would be filtered out far sooner?

dzbarsky: right, I think we're getting at the same thing. We can drop them in the module extension :)

platforms/pnpm/index.bzl
jbedard: Why didn't I see that? 🤔 lol

platforms/pnpm/BUILD.bazel
jbedard: I've refactored it so we no longer have any `@platforms//:incompatible` references/aliases. Instead we essentially have `constraints: []` meaning it is compatible with nothing, where `constraints: None` means it has no constraints and is compatible with everything.

I know `[]` vs `None` is not ideal but that's how it is atm. Can debate refactoring that...

e2e/pnpm_lockfiles/v101/snapshots/esbuild_links_defs.bzl
```diff
@@ -24,91 +24,216 @@ def npm_imported_package_store_internal():
...
-            ":.aspect_rules_js/node_modules/@esbuild+linux-mips64el@0.27.0/pkg": "@esbuild/linux-mips64el",
-            ":.aspect_rules_js/node_modules/@esbuild+linux-ppc64@0.27.0/pkg": "@esbuild/linux-ppc64",
-            ":.aspect_rules_js/node_modules/@esbuild+linux-riscv64@0.27.0/pkg": "@esbuild/linux-riscv64",
-            ":.aspect_rules_js/node_modules/@esbuild+linux-s390x@0.27.0/pkg": "@esbuild/linux-s390x",
-            ":.aspect_rules_js/node_modules/@esbuild+linux-x64@0.27.0/pkg": "@esbuild/linux-x64",
-            ":.aspect_rules_js/node_modules/@esbuild+netbsd-arm64@0.27.0/pkg": "@esbuild/netbsd-arm64",
-            ":.aspect_rules_js/node_modules/@esbuild+netbsd-x64@0.27.0/pkg": "@esbuild/netbsd-x64",
-            ":.aspect_rules_js/node_modules/@esbuild+openbsd-arm64@0.27.0/pkg": "@esbuild/openbsd-arm64",
-            ":.aspect_rules_js/node_modules/@esbuild+openbsd-x64@0.27.0/pkg": "@esbuild/openbsd-x64",
-            ":.aspect_rules_js/node_modules/@esbuild+openharmony-arm64@0.27.0/pkg": "@esbuild/openharmony-arm64",
```
jbedard, line 48: Note that this one is gone because `platforms/pnpm/index.bzl` has `"openharmony": None,  # TODO: confirm` atm.

`None` means "known but unsupported" and it gets filtered out

npm/private/pnpm.bzl
jbedard: One issue was that means each `@npm___` repo now depends on `@platforms` which I think was causing issues. I think that was one reason I went with these `@aspect_rules_js//platforms/pnpm:X` targets.

dzbarsky: you probably had repo visibility issues because you generated it as `"@platforms//os:linux"`. You should instead do `Label("@platforms//os:linux")` and then let it get str'ed into the template, that way it goes through name canonicalization and turns into `@@platforms+//os:linux` or whatever :)

e2e/pnpm_lockfiles/v101/snapshots/esbuild_links_defs.bzl
dzbarsky: great! :)

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -495,6 +497,44 @@ Either remove this patch file if it is no longer needed or change its key to mat
...
+            continue
+
+        dep = packages[dep_key]
+
+        c = None
+        if dep["os"] and dep["cpu"]:
+            c = []
+            for os in dep["os"]:
+                for cpu in dep["cpu"]:
+                    c.append(pnpm.to_bazel_os_cpu_constraint(os, cpu))
```
dzbarsky, line 524: maybe this would be a bit cleaner if you checked for `None` here and avoided appending it, then you wouldn't need to filter it out below?

platforms/pnpm/BUILD.bazel
```diff
@@ -0,0 +1,49 @@
...
+    alias(
+        name = p,
+        actual = target,
+        visibility = ["//visibility:public"],
+    )
+    for p, target in PNPM_PLATFORMS.items()
+    if target
+]
+
+# Available pnpm architectures, may map to incompatible
```
dzbarsky: "may map to incompatible" -> snip. But I still hope we can just remove these arch and cpu aliases

dzbarsky (approved):

npm/private/npm_translate_lock_helpers.bzl
jbedard: I found I then needed to repeat the `to_bazel_*_constraint` calls instead doing that and it was more code. I'm not sure which is cleaner 🤷

PR #2552, opened 2025-12-04
refactor: remove npm_translate_lock(root_package)

jbedard:

Remove `npm_translate_lock(root_package)` and always use the lockfile package as the root package.

### Test plan

### DIFF
--- a/npm/extensions.bzl
+++ b/npm/extensions.bzl
@@ -12,7 +12,6 @@ load("//npm/private:npm_translate_lock.bzl", "npm_translate_lock_lib", "parse_an
 load("//npm/private:npm_translate_lock_generate.bzl", "generate_repository_files")
 load("//npm/private:npm_translate_lock_helpers.bzl", npm_translate_lock_helpers = "helpers")
 load("//npm/private:npm_translate_lock_macro_helpers.bzl", macro_helpers = "helpers")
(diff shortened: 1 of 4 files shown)


jbedard:
Note that the public `npm_import(root_package)` still exists. While `npm_translate_lock` sets this to `pnpm_lock.package` it must still be manually set when invoking `npm_import()` manually, for now at least...

e2e/npm_translate_lock_link_workspace/MODULE.bazel
```diff
@@ -15,7 +15,7 @@ npm = use_extension(
 npm.npm_translate_lock(
     name = "foo",
     additional_file_contents = {
-        "BUILD.bazel": [
+        "foo/BUILD.bazel": [
```
jbedard: You can no longer override the `root_package` where the `.aspect_rules_js` store is...

e2e/npm_translate_lock_link_workspace/foo/BUILD.bazel
```diff
@@ -1,6 +1,6 @@
 load("@bazel_lib//lib:testing.bzl", "assert_contains")
-load("@foo//:@aspect-test/a/package_json.bzl", "bin")
 load("@foo//:defs.bzl", "js_run_binary")
+load("@foo//foo:@aspect-test/a/package_json.bzl", "bin")
```
jbedard: This is where the `package.json` is, you can't change it anymore. I think it was super confusing or maybe buggy that workspace paths (directories containing `package.json`) were being modified based on `root_package`... 🤔

PR #2554, opened 2025-12-04
fix: change use of rctx.environ to rctx.getenv

jbedard:
### Test plan


npm/private/test/npm_auth_test.bzl
```diff
@@ -173,6 +173,11 @@ def _plain_username_password_test_impl(ctx):
 def _env_var_token_test_impl(ctx):
     env = unittest.begin(ctx)
 
+    renv = {}
+    rctx = struct(
```
dzbarsky, line 177: Getenv = renv.get :)

PR #2557, opened 2025-12-05
refactor: add simplified package store definition for no-hook non-circular packages

jbedard:
Add a simplified codepath (`npm_imported_simple_package_store_internal`) for packages with no circular deps and no lifecycle hooks. While it adds a bit of code it makes the most common scenario much simpler to follow.

### Test plan

### DIFF
--- a/npm/private/npm_import.bzl
+++ b/npm/private/npm_import.bzl
@@ -31,19 +31,46 @@ _LINK_JS_PACKAGE_LOADS_TMPL = """\
 # buildifier: disable=bzl-visibility
 load(
     "@aspect_rules_js//npm/private:npm_import.bzl",
-    _npm_imported_package_store_internal = "npm_imported_package_store_internal",
     _npm_link_imported_package_internal = "npm_link_imported_package_internal",
     _npm_link_imported_package_store_internal = "npm_link_imported_package_store_internal",
 )
 """
 
-_LINK_JS_PACKAGE_TMPL = """\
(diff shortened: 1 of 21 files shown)


npm/private/npm_import.bzl
```diff
@@ -67,14 +93,72 @@ def npm_imported_package_store_internal():
...
+        package = package,
+        version = version,
+        deps = deps,
+        visibility = ["//visibility:public"],
+        tags = ["manual"],
+        exclude_package_contents = exclude_package_contents,
+    )
+
+    # reference target used when referenced by a package with cycles
+    _npm_package_store(
```
jbedard: One reason to create `npm_imported_simple_package_store_internal`... if we can determine that this package is never part of a circular dep then `/ref` and `/pkg` can either be omitted or maybe `alias()`s to the main target (we do aliases for local package store entries already - or we can have a breaking change to no longer guarantee these targets exist).

That will be a followup...

npm/private/npm_import.bzl
```diff
@@ -86,16 +170,6 @@ def npm_imported_package_store_internal(
         lifecycle_hooks_execution_requirements,
         use_default_shell_env,
         exclude_package_contents):
-    bazel_package = native.package_name()
-    is_root = bazel_package == root_package
-    if not is_root:
```
jbedard, line 90: This macro is only invoked from generated code where we already checked this, so I'm deleting it instead of duplicating it again into the new macro.

jbedard:
Note worth the complexity for little/no gain

PR #2559, opened 2025-12-05
refactor: do not use transitive_closure for lifecycle hooks

jbedard:
Do not use `transitive_closure` for lifecycle hooks but only for scenarios with circular deps.

### Test plan


npm/private/npm_import.bzl
```diff
@@ -213,8 +213,8 @@ def npm_imported_package_store_internal(
         npm_package_internal(
             name = "{}/pkg_lc".format(store_target_name),
             src = ":{}/lc".format(store_target_name),
-            package = "{package}",
```
jbedard, line 216: This bug even exists in `2.x` but I think this info always gets overridden. @dzbarsky this again makes me think we can delete at least 1 of these lc packages and maybe completely delete this `npm_package_internal` rule that is only used in this one case.

npm/private/transitive_closure.bzl
```diff
@@ -86,6 +86,7 @@ def calculate_transitive_closures(packages):
             cache,
         )
 
-        packages[package]["is_circular"] = is_circular
-        packages[package]["transitive_closure"] = transitive_closure
+        if is_circular:
+            packages[package]["transitive_closure"] = utils.sorted_map(transitive_closure)
```
jbedard, line 90: Now we only sort when we actually persist the info to the package struct, so that will remove a `sorted_map` call for 99% of packages.

dzbarsky: Cool!

dzbarsky (approved):

PR #2560, opened 2025-12-05
perf: optimize package store symlink creation more

jbedard:
The same as https://github.com/aspect-build/rules_js/pull/2401

### Test plan


jbedard:
Your code path was in the `elif not ctx.attr.src:`, now that I've removed the use of `transitive_closure` for non-circular deps these other 2 paths should get hit significantly more I think.

I'll try putting this in a helper before merging.

PR #2567, opened 2025-12-09
refactor: remove direct dep on aspect_bazel_lib

jbedard:
### Test plan

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -8,7 +8,6 @@ module(
 
 # Lower-bounds (minimum) versions for direct runtime dependencies.
 # Do not bump these unless rules_js requires a newer version to function.
-bazel_dep(name = "aspect_bazel_lib", version = "2.17.1")  # TODO(alexeagle): remove
 bazel_dep(name = "tar.bzl", version = "0.6.0")
 bazel_dep(name = "yq.bzl", version = "0.3.2")
 bazel_dep(name = "jq.bzl", version = "0.4.0")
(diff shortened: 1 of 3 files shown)


MODULE.bazel
```diff
@@ -19,6 +18,10 @@ bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "platforms", version = "0.0.5")
 bazel_dep(name = "rules_nodejs", version = "6.6.2")
 
+# Ensure any version of aspect_bazel_lib used is at least 2.22.0 to include
+# https://github.com/bazel-contrib/bazel-lib/commit/371362199e5e5bde58aeb15b54ea6cfb164af337
+bazel_dep(name = "aspect_bazel_lib", version = "2.22.0", repo_name = None)
```
dzbarsky, line 23: neat

dzbarsky (approved):

malt3:
This breaks on some Bazel versions (Bazel 7.4.0 and older):


```
ERROR: https://bcr.bazel.build/modules/aspect_rules_js/3.0.3/MODULE.bazel:34:10: Traceback (most recent call last):
	File "https://bcr.bazel.build/modules/aspect_rules_js/3.0.3/MODULE.bazel", line 34, column 10, in <toplevel>
Error in bazel_dep: in call to bazel_dep(), parameter 'repo_name' got value of type 'NoneType', want 'string'
```

EDIT: noticed that the minimal Bazel version you require is now higher, so that's probably fine.

jbedard:
Yes, if you are getting that error then you're on an older unsupported bazel version.

I know it's not the greatest error message, but it's better then odd errors that we'd get if we did not ensure a minimum `aspect_bazel_lib` version.

Issue #2584, opened 2025-12-16
[Bug]: js_image_layer omit .tar extension when compression is set to none


fa93hws:
### What happened?

`ext` is set to "" when compression is none by the following code:
```
ext = tar_lib.common.compression_to_extension[compress] if compress else ""
output = ctx.actions.declare_file("%s_%s%s" % (ctx.label.name, typ, ext))
```
https://github.com/aspect-build/rules_js/blob/v2.8.3/js/private/js_image_layer.bzl#L571

Is that expected? I think output `*.tar` for uncompressed file makes more sense?

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: `bazel 8.5.0`

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: 2.8.3

Language(s) and/or frameworks involved: `js_image_layer`

### How to reproduce

```shell

```

PR #2585, opened 2025-12-16
js image layer to output with .tar extension when uncompressed

fa93hws:
Fixes https://github.com/aspect-build/rules_js/issues/2584
I think adding `*.tar` makes more sense? WDYT?

### Test plan

`bazel test //...` passes


jbedard:
That makes sense to me. @thesayyn?

jbedard:
For 3.x: https://github.com/aspect-build/rules_js/pull/2586

PR #2592, opened 2025-12-19
fix: npm deps on git+https urls

jbedard:
Found while adding more tests for https://github.com/aspect-build/rules_js/pull/2403

### Test plan

- New test cases added


npm/private/utils.bzl
```diff
@@ -102,7 +102,7 @@ def _npm_registry_download_url(package, version, registries, default_registry):
     )
 
 def _is_git_repository_url(url):
-    return url.startswith("git+ssh://") or url.startswith("git+https://") or url.startswith("git@")
+    return url.startswith("git+ssh://") or url.startswith("git+https://") or url.startswith("git@") or url.startswith("https://git@")
```
jbedard: The pnpm <9 (lockfile v54, 60, 61) tests reproduce this where the lockfile itself has the git repo specified as:
```
resolution: {repo: https://git@github.com/jquery/jquery.git}
```

I assume this is possible for pnpm 9+ as well but 9+ must fallback to non-git `tarball` https resolution instead of git.

npm/private/utils.bzl
```diff
@@ -101,9 +101,6 @@ def _npm_registry_download_url(package, version, registries, default_registry):
         version[:version.find("_")] if version.find("_") != -1 else version,
     )
 
-def _is_git_repository_url(url):
-    return url.startswith("git+ssh://") or url.startswith("git+https://") or url.startswith("git@")
```
jbedard, line 105: There is also `https://git@` which I had originally added to this. But there is also plain `https://` which might be a git repo (and it probably ends with `.git` but might end with `.git#commithash`).

I think it's better to not do things based on URL and instead based on if a `commit` attr is specified, which we required anyway (see removed `fail()`).

npm/private/npm_import.bzl
```diff
@@ -621,10 +618,10 @@ def _npm_import_rule_impl(rctx):
 
     reproducible = False
     package_src = _EXTRACT_TO_DIRNAME
-    if utils.is_git_repository_url(rctx.attr.url):
```
jbedard, line 624: We explicitly set `url` to the lockfile `repo` field which for `git+https` version specifiers will be the plain `https://github.com/foo/bar.git` URL and not appear to be a git protocol.

We then assert that when fetching from git a `commit` is required. So instead lets just do things based on the `commit` attribute...

npm/private/utils.bzl
jbedard: See newer comment regarding deleting this `is_git_repository_url`

e2e/pnpm_lockfiles/base/package.json
```diff
@@ -11,9 +11,13 @@
         "@isaacs/cliui": "8.0.2",
         "debug": "ngokevin/debug#9742c5f383a6f8046241920156236ade8ec30d53",
         "esbuild": "0.27.0",
+        "highlightjs-git-https-notar": "git+https://gitea.osmocom.org/vyanitskiy/highlight.js.git#58dc5961f6f2bb8bc8bb1e7ce39f268a2fdd874f",
```
jbedard, line 14: Github seems to provide tarballs for all commits so the pnpm 9+ lockfile references the tarball instead of using the git protocol.

Using this non-github public git repo ensures the git protocol is used on all pnpm version

e2e/pnpm_lockfiles/v90/pnpm-lock.yaml
```diff
@@ -552,6 +564,11 @@ packages:
     resolution: {integrity: sha512-0hJU9SCPvmMzIBdZFqNPXWa6dqh7WdH0cII9y+CyS8rG3nL48Bclra9HmKhVVUHyPWNH5Y7xDwAB7bfgSjkUMQ==}
     engines: {node: '>= 0.4'}
 
+  highlight.js@git+https://gitea.osmocom.org/vyanitskiy/highlight.js.git#58dc5961f6f2bb8bc8bb1e7ce39f268a2fdd874f:
+    resolution: {commit: 58dc5961f6f2bb8bc8bb1e7ce39f268a2fdd874f, repo: https://gitea.osmocom.org/vyanitskiy/highlight.js.git, type: git}
```
jbedard, line 568: Notice the `type: git` where all the jquery ones  are `tarball`

PR #2610, opened 2026-01-02
chore: checkin MODULE.bazel.lock files

jbedard:
I think the lockfiles are stable enough now, and `bazelrc-presets.bzl` seems to suggest checking them in now?

### Test plan

### DIFF
--- a/tools/preset.bazelrc
+++ b/tools/preset.bazelrc
@@ -67,6 +67,11 @@ common --experimental_fetch_all_coverage_outputs
 common --experimental_remote_cache_eviction_retries=5
 # Docs: https://registry.build/flag/bazel?filter=experimental_remote_cache_eviction_retries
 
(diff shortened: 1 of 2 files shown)


tools/preset.bazelrc
```diff
@@ -135,6 +144,11 @@ common --incompatible_strict_action_env
 common --nolegacy_external_runfiles
 # Docs: https://registry.build/flag/bazel?filter=legacy_external_runfiles
 
+# Fail the build if the MODULE.bazel.lock file is out of date.
+# Using this mode in ci prevents the lockfile from being out of date.
+common:ci --lockfile_mode="error"
```
jbedard, line 149: So this doesn't work great on things like rulesets where multiple bazel versions are tested, and different bazel versions may change the lockfile format.

Have we seen this used in rulesets anywhere else?

alexeagle (approved):

PR #2615, opened 2026-01-06
refactor: do not duplicate js_library result JsInfo in runfiles

jbedard:
The rules_js providers should not be destructed into generic runfiles/`DefaultInfo` until necessary such as `js_binary` preparing for executing node.

See https://github.com/aspect-build/rules_js/pull/1781 for more context.

The `@aspect_rules_js//js:libs.bzl` `js_lib_helpers.gather_runfiles` only gathers runfiles and no longer gathers files from `JsInfo`. To gather files from `JsInfo` use `js_lib_helpers.gather_files_from_js_infos`.

As a result `js_library`, the `aspect_rules_ts` `ts_project` any similar rules using `gather_runfiles` will no longer collect `JsInfo` provider files as runfiles. It is the responsibility of a `*_binary`-like target to convert the rules_js specific `JsInfo` into generic bazel runfiles.

### Test plan

### DIFF
--- a/js/private/js_helpers.bzl
+++ b/js/private/js_helpers.bzl
@@ -117,12 +117,7 @@ def gather_runfiles(
         deps = [],
         data_files = [],
         copy_data_files_to_bin = False,
-        no_copy_to_bin = [],
-        include_sources = True,
-        include_types = False,
-        include_transitive_sources = True,
-        include_transitive_types = False,
-        include_npm_sources = True):
+        no_copy_to_bin = []):
     """Creates a runfiles object containing files in `sources`, default outputs from `data` and transitive runfiles from `data` & `deps`.
 
     As a defense in depth against `data` & `deps` targets not supplying all required runfiles, also
@@ -158,16 +153,6 @@ def gather_runfiles(
             file such as a file in an external repository. In most cases, this option is not needed.
             See `copy_data_files_to_bin` docstring for more info.
 
-        include_sources: see js_info_files documentation
-
-        include_types: see js_info_files documentation
-
-        include_transitive_sources: see js_info_files documentation
-
-        include_transitive_types: see js_info_files documentation
-
-        include_npm_sources: see js_info_files documentation
-
     Returns:
         A [runfiles](https://bazel.build/rules/lib/runfiles) object created with [ctx.runfiles](https://bazel.build/rules/lib/ctx#runfiles).
     """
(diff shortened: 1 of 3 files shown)


js/private/js_helpers.bzl
```diff
@@ -112,17 +112,10 @@ this option is not needed.
 
 def gather_runfiles(
```
jbedard, line 113: ~~I think this is considered public-API so maybe I need to change this to merge to `3.x` instead of `main`...~~
EDIT: I've switched this to 3.x

jbedard: Yes [ts_project uses this](https://github.com/aspect-build/rules_ts/blob/6e5ab380f972642b7c76e522c5b6b8f26f54ae7c/ts/private/ts_project.bzl#L377) which IMO shows another reason for this PR... removing more duplication. Although the fact that this logic is duplicated into other rulesets makes this change a bit harder to do...

jbedard:
I believe this is the same as what @gregmagolan was suggesting as an opt-in? https://github.com/aspect-build/rules_js/pull/1781

jbedard:
As a general question... what is the correct/standard behaviour of defining "runfiles for a library"? Should we even be doing that or is it the job of things like `js_binary` to collect provider information and convert it to the generic runfiles instead of every "library" doing the same thing?

gregmagolan (approved):

js/private/js_library.bzl
```diff
@@ -240,17 +240,11 @@ def _js_library_impl(ctx):
 
     runfiles = gather_runfiles(
         ctx = ctx,
-        sources = transitive_sources,
```
jbedard, line 243: These are passed along via `js_info(transitive_sources = transitive_sources)` and targets consuming `JsInfo` should get them from there (which they should already be doing anyway...).

js/private/js_library.bzl
```diff
@@ -240,17 +240,11 @@ def _js_library_impl(ctx):
...
         data = ctx.attr.data,
         deps = srcs_types_deps,
         data_files = ctx.files.data,
         copy_data_files_to_bin = ctx.attr.copy_data_to_bin,
         no_copy_to_bin = ctx.files.no_copy_to_bin,
-        include_sources = True,
```
jbedard, line 249: The `include_*` decisions should be made by `*_binary`-like targets and those can be collected from the `JsInfo`

jbedard:
I've updated this to remove the params from `gather_runfiles` and will make rules_ts compatible with rules_js v3 while keeping existing behaviour when used with rules_js v2: https://github.com/aspect-build/rules_ts/pull/878

js/private/js_helpers.bzl
jbedard: See https://github.com/aspect-build/rules_ts/pull/878 which makes rules_ts forward compatible with rules_js v3 without breaking changes.

Other rulesets seem to not use the `include_*` params and will inherit the breaking change from rules_js when people opt-in to that while remaining compatible with rules_js v2.

PR #2632, opened 2026-01-10
refactor: remove support for undeclared .npmrc and pnpm-workspace files

jbedard:

Remove support for undeclared `.npmrc` and `pnpm-workspace.yaml` inputs. All inputs that will effect package fetching or `pnpm install | import` must be declared.

### Test plan

### DIFF
--- a/npm/private/npm_translate_lock_state.bzl
+++ b/npm/private/npm_translate_lock_state.bzl
@@ -189,27 +189,19 @@ def _init_workspace(priv, rctx, label_store, is_windows):
 
 ################################################################################
 def _init_npmrc(priv, rctx, attr, label_store):
-    if not label_store.has("npmrc"):
-        # check for a .npmrc next to the pnpm-lock.yaml file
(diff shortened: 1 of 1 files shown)


npm/private/npm_translate_lock_state.bzl
```diff
@@ -189,28 +189,20 @@ def _init_workspace(priv, rctx, label_store, is_windows):
 
 ################################################################################
 def _init_npmrc(priv, rctx, attr, label_store):
-    if not label_store.has("npmrc"):
-        # check for a .npmrc next to the pnpm-lock.yaml file
-        _maybe_npmrc(priv, rctx, attr, label_store, "sibling_npmrc")
-
     if label_store.has("npmrc"):
         _load_npmrc(priv, rctx, label_store.path("npmrc"))
+    else:
```
jbedard, line 194: We could potentially only do this assertion `elif _should_update_pnpm_lock(priv)` but I figured we should start with simply always `fail`ing and can reduce it later if necessary

Issue #2138, opened 2025-03-19
[Bug]: memory pressure in Bazel due to non-internable strings


alexeagle:
### What happened?

A Java Profiler run against Bazel's server reports lots of duplicate string data like

> link attribute cannot be specified when link_packages are set

coming from https://github.com/aspect-build/rules_js/blob/2679d3670885c9222848fb86c95d0e6c187cd2b2/npm/private/npm_import.bzl#L244

Fabian on Bazel slack:
https://bazelbuild.slack.com/archives/CDCMRLS23/p1742380269194209?thread_ts=1742216248.839529&cid=CDCMRLS23
explains this is because .bzl files generated into external repos are not de-duplicated by Bazel.

### Version

HEAD

### How to reproduce

```shell

```

PR #2636, opened 2026-01-12
feat: enable default npm_exclude_package_contents

jbedard:

The default option for `exclude_package_contents` now excludes unnecessary files and folders at package extraction time. This default aligns with [yarn autoclean](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/commands/autoclean.js#L16) and may cause some packages to be misconfigured.

### Test plan


dzbarsky (approved):

npm/private/test/snapshots/circular_links_defs.bzl
```diff
@@ -47,7 +47,40 @@ def npm_imported_package_store_internal():
         lifecycle_hooks_env = {},
         lifecycle_hooks_execution_requirements = {},
         use_default_shell_env = False,
-        exclude_package_contents = [],
+        exclude_package_contents = [
```
jbedard, line 50: Notice how since `npm_import` is a repo rule the default on there causes this BUILD content to be generated, so these strings will be duplicated A LOT.

Is it worth the effort to make the attribute unspecified when defaulted and rely on the `.attr(default)`? Note this will most likely be reproducing https://github.com/aspect-build/rules_js/issues/2138...

Also note the `next_defs.bzl` is not in the diff because of the exclusion in MODULE.

js/private/test/image/custom_layers_nomatch_test_package_store_3p.listing
```diff
@@ -10,9 +10,7 @@ drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/acorn@8.12.1/
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/acorn@8.12.1/node_modules/
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/acorn@8.12.1/node_modules/acorn/
--r-xr-xr-x  0 0      0       22267 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/acorn@8.12.1/node_modules/acorn/CHANGELOG.md
```
jbedard, line 13: These are nice examples of the intention of this change...

docs/troubleshooting.md
```diff
@@ -177,17 +177,6 @@ can effect performance and are sometimes worth explicitly excluding content.
 
 In these cases you can add such packages and the respective files/folders you want to exclude.
 
-**For WORKSPACE builds:**
```
jbedard: Cleanup related to the attribute while we're touching it 🤷

jbedard:
Doing https://github.com/aspect-build/rules_js/pull/2652 instead

PR #2642, opened 2026-01-13
Make node 22 the default node version

hofbi:
https://nodejs.org/en/about/previous-releases says that Node 18 is already EOL and 20 will become EOL in April. So I am proposing to use Node 22 by default to not run on old and unsupported node versions in case someone does not configure their own node toolchain

---

- Searched for relevant documentation and updated as needed: No relevant docs found
- Breaking change (forces users to change their own code or config): Could happen if they don't declare a node toolchain.
- Suggested release notes appear below: Switch default Node toolchain from Node 18 to Node 22

### Test plan

- New test cases added

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -135,7 +135,7 @@ use_repo(node_dev, "node24_linux_arm64")
 use_repo(node_dev, "node24_linux_s390x")
 use_repo(node_dev, "node24_linux_ppc64le")
 use_repo(node_dev, "node24_windows_amd64")
-node_dev.toolchain(node_version = "18.20.4")
+node_dev.toolchain(node_version = "22.21.1")
 node_dev.toolchain(
     name = "node16",
     node_version = "16.20.0",
@@ -150,7 +150,7 @@ node_dev.toolchain(
 )
 node_dev.toolchain(
     name = "node22",
-    node_version = "22.20.0",
+    node_version = "22.21.1",
 )
 node_dev.toolchain(
     name = "node24",


jbedard:
Note rules_js v3 has switched to simply [inheriting the default from rules_nodejs](https://github.com/aspect-build/rules_js/commit/3a20f40ec5669c023acd965f4ece8bf86cd7ef7c#diff-6136fc12446089c3db7360e923203dd114b6a1466252e71667c6791c20fe6bdcR122) and rules_nodejs now [defaults to 20](https://github.com/bazel-contrib/rules_nodejs/commit/8f3a65701d795b932ce3369d97e7891074731f24).

However 20 will also be EOL soon so I wonder if we should do something more aggressive, at least in rules_js v3. Maybe changing the default to always be "current LTS" and explicitly stating that it can change when updating rules_js minor releases etc. @alexeagle wdyt of that?

jbedard:
I'd like to propose https://github.com/bazel-contrib/rules_nodejs/pull/3891 and making the rules_js 3.x include that by default, while keeping 2.x unchanged...

alexeagle:
Another option is to stop having a default at all, make users provide the major version

jbedard:
@alexeagle I'll try that out but I think we should include https://github.com/bazel-contrib/rules_nodejs/pull/3891 either way

jbedard:
See https://github.com/aspect-build/rules_js/pull/2649 for the change defaulting to 22 in rules_js v3

PR #2645, opened 2026-01-13
refactor: remove support and references to bazel-lib <3.x, bazel<7.6

jbedard:
### Test plan

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -2,7 +2,7 @@
 
 module(
     name = "aspect_rules_js",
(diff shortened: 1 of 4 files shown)


jbedard:
Note the [MODULE still contains a 2.22.5 minimum](https://github.com/aspect-build/rules_js/blob/d55294ced9d58420310940d6262042b3135d5494/MODULE.bazel#L21-L27) to ensure any v2 is compatible with the v3 we use, but the version we use from rules_js will always be 3+.

PR #2646, opened 2026-01-13
refactor: move npm/private/lifecycle to pnpm workspace, bazel build

jbedard:
I want to move the build+bundling of `lifecycle/min/index.min.js` into bazel in preparation for upgrading the `@pnpm/*` packages for 3.x

### Test plan


npm/private/lifecycle/package.json
```diff
@@ -2,13 +2,17 @@
     "name": "lifecycle-deps",
     "private": true,
     "dependencies": {
+        "path-exists": "^4.0.0",
```
jbedard, line 5: The use of npm and hoisting allowed rollup to bundle these without declaring the direct dependencies on them

PR #2650, opened 2026-01-14
refactor: remove use of bazel_features, old docs/tests, deprecate verify_node_modules_ignored

jbedard:
### Test plan


jbedard:
Things robots said I could do...

PR #2651, opened 2026-01-14
chore: update release notes snippets

jbedard:
### Test plan

### DIFF
--- a/.github/workflows/release_prep.sh
+++ b/.github/workflows/release_prep.sh
@@ -34,34 +34,30 @@ Many companies are successfully building with rules_js.
 If you're getting value from the project, please let us know!
 Just comment on our [Adoption Discussion](https://github.com/aspect-build/rules_js/discussions/1000).
 
-## Using Bzlmod:
-
 Add to your \`MODULE.bazel\` file:
 \`\`\`starlark
 bazel_dep(name = "aspect_rules_js", version = "${TAG:1}")
 
-####### Node.js version #########
-# By default you get the node version from DEFAULT_NODE_VERSION in @rules_nodejs//nodejs:repositories.bzl
-# Optionally you can pin a different node version:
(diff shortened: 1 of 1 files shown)


.github/workflows/release_prep.sh
```diff
@@ -34,31 +34,33 @@ Many companies are successfully building with rules_js.
...
-# By default you get the node version from DEFAULT_NODE_VERSION in @rules_nodejs//nodejs:repositories.bzl
-# Optionally you can pin a different node version:
-bazel_dep(name = "rules_nodejs", version = "6.6.0")
-node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node", dev_dependency = True)
-node.toolchain(node_version = "18.14.2")
-#################################
-
-npm = use_extension("@aspect_rules_js//npm:extensions.bzl", "npm", dev_dependency = True)
-
+Normally a \`pnpm-lock.yaml\` file is is used to manage dependencies:
```
jbedard: `pnpm-lock.yaml` is almost always used, so maybe it can just be part of the main snippet?

But the `rules_nodejs` logic and exposing of `pnpm` are not as common and imo should not be part of the main snippet users will be copy and pasting. WDYT?

alexeagle: I think everyone needs the bazel-managed pnpm too.

whatever bazel-starters/js does is what I prefer.

jbedard: Ok. So I'll only move the `rules_nodejs` out to its own snippet? The bazel-managed pnpm will be set to `dev_dependency = True` while the other's wont...

alexeagle: I don't think we have to be prescriptive about dev_dependency - very few users publish a module for others to consume

jbedard: IMO that means we should NOT have `dev_dependency = True` like we do today and it should be removed...

PR #2652, opened 2026-01-15
feat: add package exclusion presets, enable default exclusion list

jbedard:
This is an alternative to #2636. This allows us to define multiple "exclude presets" such as the existing `yarn_autoclean` list, but also a less aggressive one that we can enable by default and possibly others for different use cases or strictness.

The default package content exclusion via `npm_exclude_package_contents(use_defaults=True)` is now done using `npm_exclude_package_contents(presets = ["yarn_autoclean"])`. A simpler `"default"` preset is available and enabled by default for all packages unless overridden by a custom `npm_exclude_package_contents(patterns|presets)`.

### Test plan

- New test cases added


MODULE.bazel
```diff
@@ -287,12 +287,6 @@ npm.npm_translate_lock(
     npmrc = "//:.npmrc",
     pnpm_lock = "//:pnpm-lock.yaml",
 )
-
-# Configure package exclusions using tag class
-npm.npm_exclude_package_contents(
-    package = "chalk",
-    patterns = ["**/README*"],
```
jbedard, line 294: This is covered by the new defaults

js/private/test/image/custom_layers_nomatch_test_package_store_3p.listing
```diff
@@ -10,9 +10,7 @@ drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/acorn@8.12.1/
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/acorn@8.12.1/node_modules/
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/acorn@8.12.1/node_modules/acorn/
--r-xr-xr-x  0 0      0       22267 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/acorn@8.12.1/node_modules/acorn/CHANGELOG.md
```
jbedard, line 13: The diff of this file is the best example of the goal+effects of this PR, although this file only really shows the `*.md` glob being applied and has nothing such as the CI configs being excluded

npm/private/npm_import.bzl
```diff
@@ -95,6 +110,16 @@ def npm_imported_package_store_internal(
...
 
+    # Expand presets into patterns
+    expanded_exclude_patterns = list(exclude_package_contents) if exclude_package_contents else []
+    for preset in exclude_package_contents_presets:
+        if preset not in EXCLUDE_PACKAGE_CONTENTS_PRESETS:
+            fail("Unknown exclude_package_contents preset '{}'. Valid presets are: {}".format(
+                preset,
+                ", ".join(EXCLUDE_PACKAGE_CONTENTS_PRESETS.keys()),
+            ))
+        expanded_exclude_patterns = expanded_exclude_patterns + EXCLUDE_PACKAGE_CONTENTS_PRESETS[preset]
```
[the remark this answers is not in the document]
jbedard: Good bot, correcting another bot 👍

kormide:
I like the new API for this and that the "default" exclusions are on by default, so long as it's clear as a breaking release note.

MODULE.bazel
dzbarsky: was this here because necessary or just as an example? I guess either way examlpe doesn't belong here :)

js/private/test/image/regex_edge_cases_test_package_store_3p.listing
```diff
@@ -59,10 +54,6 @@ drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/uuid@8.3.2/
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/uuid@8.3.2/node_modules/
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/uuid@8.3.2/node_modules/uuid/
--r-xr-xr-x  0 0      0       12680 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/uuid@8.3.2/node_modules/uuid/CHANGELOG.md
--r-xr-xr-x  0 0      0         513 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/uuid@8.3.2/node_modules/uuid/CONTRIBUTING.md
--r-xr-xr-x  0 0      0        1109 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/uuid@8.3.2/node_modules/uuid/LICENSE.md
```
dzbarsky, line 64: i wonder how come sometimes the LICENSE gets excluded and sometimes not. Is it just excluding markdowns? Technically we should keep licenses around (they may require shipping with the package) but if we're ok stripping we should strip all ;)

npm/extensions.bzl
```diff
@@ -267,18 +266,24 @@ _EXCLUDE_PACKAGE_CONTENT_ATTRS = {
...
 
 _EXCLUDE_PACKAGE_CONTENT_DOCS = """Configuration for excluding package contents from npm packages.
 
 This tag can be used multiple times to specify different exclusion patterns for different package specifiers.
-Multiple exclusion patterns for the same package are combined.
+More specific package matches override less specific ones (the wildcard "*" is only used if no specific
+package match is found).
 
-The `yarn autoclean` default exclusion patterns can be included by setting `use_defaults` to `True`.
+By default, `presets` is set to `["basic"]` which excludes common files such as `*.md` and development-related
```
dzbarsky, line 285: this sounds at odds with the above (`- "basic": Minimal defaults (README.md)`). Is it just README.md or *.mod and others?

dzbarsky: reading code below it's not just README.md, lets amend the above to be cleaer?

npm/private/exclude_package_contents_presets.bzl
```diff
@@ -0,0 +1,88 @@
...
+    "**/powered-test/**",
+
+    # asset directories
+    "**/docs/**",
+    "**/doc/**",
+    "**/website/**",
+    "**/images/**",
+    "**/assets/**",
+
+    # examples
```
dzbarsky, line 50: should this extend off _DEFAULT? I.e. make it clearer that _DEFAULT is a safer subset? are there cases where we'd add something extra to _DEFAULT? likely not?

dzbarsky (approved):
this is cool, because we could extend it to allow specifying your own reusable preset via a tag class. Ship it!

js/private/test/image/regex_edge_cases_test_package_store_3p.listing
jbedard: The "default exclusions" is a subset of the yarn one atm and has `*.md` in it. I assume a file named `LICENSE` is standard and intentionally not excluded

npm/extensions.bzl
jbedard: You're correct this doc is bad. Originally I had only README.md but expanded it to include a few more things

npm/private/exclude_package_contents_presets.bzl
jbedard: I wanted to keep the yarn one copy+pasted from the linked above and not do any concat/subset magic, even if they do overlap.

We might want to make the exclusions unique before we pass them as cli args to the tar binary but that can be a different discussion if necessary...

MODULE.bazel
jbedard: I think just an example, and it was reflected in one of the `js_image_layer` tests - the new defaults still exclude that readme for the test

PR #2653, opened 2026-01-15
fix: ensure COVERAGE_DIR is writable for split coverage postprocessing

anonrig:
When Bazel's --experimental_split_coverage_postprocessing flag is enabled,
   coverage post-processing runs as a separate action from the test execution.
   The COVERAGE_DIR directory may be created by a previous action with
   restrictive permissions (read-only), causing the coverage merger to fail
   with EACCES (permission denied) when attempting to write lcov.info.

---

   This fix checks if COVERAGE_DIR exists and ensures it has write permissions
   before the coverage report attempts to write to it.

```
Error: EACCES: permission denied, open '/home/runner/.cache/bazel/_bazel_runner/be54bd8ac890d1817c772d91eb8f6cb4/execroot/_main/bazel-out/k8-fastbuild/testlogs/types/test/index.spec/_coverage/lcov.info'
    at Object.openSync (node:fs:561:18)
    at FileWriter.writeFile (/home/runner/.cache/bazel/_bazel_runner/be54bd8ac890d1817c772d91eb8f6cb4/external/aspect_rules_js+/js/private/coverage/coverage.js:11043:42)
    at LcovOnlyReport.onStart (/home/runner/.cache/bazel/_bazel_runner/be54bd8ac890d1817c772d91eb8f6cb4/external/aspect_rules_js+/js/private/coverage/coverage.js:12029:46)
    at Visitor.value (/home/runner/.cache/bazel/_bazel_runner/be54bd8ac890d1817c772d91eb8f6cb4/external/aspect_rules_js+/js/private/coverage/coverage.js:11192:39)
    at ReportTree.visit (/home/runner/.cache/bazel/_bazel_runner/be54bd8ac890d1817c772d91eb8f6cb4/external/aspect_rules_js+/js/private/coverage/coverage.js:11280:18)
    at LcovOnlyReport.execute (/home/runner/.cache/bazel/_bazel_runner/be54bd8ac890d1817c772d91eb8f6cb4/external/aspect_rules_js+/js/private/coverage/coverage.js:11945:45)
    at Report.run (/home/runner/.cache/bazel/_bazel_runner/be54bd8ac890d1817c772d91eb8f6cb4/external/aspect_rules_js+/js/private/coverage/coverage.js:15606:11) {
  errno: -13,
  code: 'EACCES',
  syscall: 'open',
  path: '/home/runner/.cache/bazel/_bazel_runner/be54bd8ac890d1817c772d91eb8f6cb4/execroot/_main/bazel-out/k8-fastbuild/testlogs/types/test/index.spec/_coverage/lcov.info'
}
```

### Test plan

1. Enable on `--experimental_split_coverage_postprocessing`
2. Build using `bazel build --config=coverage`
3. Run a js_test with `bazel coverage //...`

### Related

Had to disable the flag to use js_test... https://github.com/cloudflare/workerd/pull/5881/changes/16fa7f2fcfe07166b1d44071c6229fdbec8d15dd

### DIFF
--- a/js/private/coverage/bundle/c8.js
+++ b/js/private/coverage/bundle/c8.js
@@ -14,6 +14,24 @@ try {
     // in case file doesn't exist or some other error is thrown, just ignore it.
 } catch {}
 
+// When --experimental_split_coverage_postprocessing is enabled, Bazel runs
+// coverage post-processing as a separate action. The COVERAGE_DIR may have been
+// created by a previous action with restrictive permissions. Ensure we can write
+// to it by making the directory writable.
+try {
+    const coverageDir = process.env.COVERAGE_DIR
+    if (coverageDir) {
+        const stats = fs.statSync(coverageDir, { throwIfNoEntry: false })
+        // Check if directory exists and is not writable by owner
+        if (stats && (stats.mode & 0o200) === 0) {
+            fs.chmodSync(coverageDir, stats.mode | 0o200)
+        }
+    }
+} catch {
+    // Ignore errors - if we can't fix permissions, the write will fail with
+    // a more descriptive error message below
+}
+
 const include = fs
     .readFileSync(process.env.COVERAGE_MANIFEST)
     .toString('utf8')
--- a/js/private/coverage/coverage.js
+++ b/js/private/coverage/coverage.js
@@ -1,19 +1,19 @@
 'use strict';
 
-var require$$0$2 = require('path');
+var require$$0$1 = require('path');
 var require$$2$1 = require('node:url');
 var require$$1$1 = require('node:path');
-var require$$0$1 = require('fs');
+var fs = require('fs');
 var require$$4 = require('node:fs');
 var require$$5 = require('node:fs/promises');
 var require$$0 = require('node:events');
 var require$$1 = require('node:stream');
 var require$$2 = require('node:string_decoder');
 var require$$2$2 = require('util');
-var require$$0$3 = require('os');
+var require$$0$2 = require('os');
 var require$$1$2 = require('tty');
 var require$$1$3 = require('url');
-var require$$0$4 = require('assert');
+var require$$0$3 = require('assert');
 var require$$12 = require('module');
 
 var c8 = {};
@@ -4808,7 +4808,7 @@ function requireCommonjs$1 () {
 	const lru_cache_1 = /*@__PURE__*/ requireCommonjs$3();
 	const node_path_1 = require$$1$1;
 	const node_url_1 = require$$2$1;
-	const fs_1 = require$$0$1;
+	const fs_1 = fs;
 	const actualFS = __importStar(require$$4);
 	const realpathSync = fs_1.realpathSync.native;
 	// TODO: test perf of fs/promises realpath vs realpathCB,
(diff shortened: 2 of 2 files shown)


js/private/coverage/bundle/c8.js
```diff
@@ -14,6 +14,24 @@ try {
...
+    const coverageDir = process.env.COVERAGE_DIR
+    if (coverageDir) {
+        const stats = fs.statSync(coverageDir, { throwIfNoEntry: false })
+        // Check if directory exists and is not writable by owner
+        if (stats && (stats.mode & 0o200) === 0) {
+            fs.chmodSync(coverageDir, stats.mode | 0o200)
+        }
+    }
+} catch {
+    // Ignore errors - if we can't fix permissions, the write will fail with
```
jbedard, line 31: Should we output a WARNING to stderr just so we have more info when that write failure does occur? Or are there scenarios where we hit this path but no error will occur?

jbedard:
Do you have coverage working with `--experimental_split_coverage_postprocessing` after this patch? From what I am reading it sounds broken, see https://github.com/aspect-build/rules_js/pull/991 and the issues it links to.

anonrig:
> Do you have coverage working with `--experimental_split_coverage_postprocessing` after this patch? From what I am reading it sounds broken, see #991 and the issues it links to.

I haven't tested it yet. I'll open as many pull-requests as possible in order to make sure it works.

jbedard:
> The COVERAGE_DIR directory may be created by a previous action with restrictive permissions (read-only),

Can you provide an example of what previous action that might be? Does the fix belong in that previous action?

js/private/coverage/bundle/c8.js
anonrig: We will already receive a good error just after this line (if permission is required). I'm sure this might bloat the build output.

anonrig:
> > The COVERAGE_DIR directory may be created by a previous action with restrictive permissions (read-only),
> 
> Can you provide an example of what previous action that might be? Does the fix belong in that previous action?

This is the previous action I've encountered this issue: https://github.com/cloudflare/workerd/pull/5881/changes#diff-763c106cd93999d6533f6849ce26ce51598d42215459051ded837b4b255b1c14R136

jbedard:
Do you know which previous action is creating the bad `COVERAGE_DIR`? Ideally we could avoid creating the bad permissions instead of having to override them like this, won't every coverage tool for every language need this same fix otherwise?

anonrig:
> Do you know which previous action is creating the bad `COVERAGE_DIR`? Ideally we could avoid creating the bad permissions instead of having to override them like this, won't every coverage tool for every language need this same fix otherwise?

I'm not sure what the root cause is. When `--experimental_split_coverage_postprocessing` is enabled the coverage is processed in another task and does not have write access to the coverage directory. 

I've also validated that this fixes js_tests (in: https://github.com/cloudflare/workerd/commit/9c70d739834b204ddf954a65a59766e852f60d9f)

I've submitted a bug report to Bazel as well: https://github.com/bazelbuild/bazel/issues/28310

jbedard:
Ok to close this if the proper solution is in bazel?

anonrig:
> Ok to close this if the proper solution is in bazel?

I haven't received any feedback from Bazel team yet. But let's close it for the time being https://github.com/bazelbuild/bazel/pull/28311

PR #2655, opened 2026-01-16
fix: return full label paths from npm_link_targets

### LINKED ISSUE #2417: [FR]: Return full path labels from npm_link_targets

`npm_link_targets` now returns labels with relative paths to targets (e.g. `:pkg_name`). This makes it difficult to get correct targets for different packages in one list. For example, if I want to get targets for current package and root package, I have to do next:
```
npm_link_targets() + ["//{}".format(label) for label in npm_link_targets(package = "")]
```

### Describe the feature

If `npm_link_targets` returns full path labels (e.g. `//bzl_pkg:pkg_name`), it will simplify the code to this
```
npm_link_targets() + npm_link_targets(package = "")
```

jbedard:
Fix #2417

### Test plan

### DIFF
--- a/e2e/npm_translate_lock_empty/snapshots/npm_defs.bzl
+++ b/e2e/npm_translate_lock_empty/snapshots/npm_defs.bzl
@@ -70,4 +70,4 @@ def npm_link_targets(name = "node_modules", package = None, prod = True, dev = T
 
     link_targets = []
 
-    return link_targets
+    return ["//%s%s" % (bazel_package, target) for target in link_targets]
--- a/npm/private/test/snapshots/npm_defs.bzl
+++ b/npm/private/test/snapshots/npm_defs.bzl
@@ -3509,7 +3509,7 @@ def npm_link_targets(name = "node_modules", package = None, prod = True, dev = T
                 ":node_modules/@mycorp/pkg-d",
                 ":node_modules/@mycorp/pkg-b",
             ])
-    return link_targets
+    return ["//%s%s" % (bazel_package, target) for target in link_targets]
 
 # Generated npm_link_package_store for linking of first-party "@mycorp/pkg-a" package
 # buildifier: disable=function-docstring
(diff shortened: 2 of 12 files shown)


npm/private/test/snapshots/npm_defs.bzl
```diff
@@ -2736,34 +2736,34 @@ def npm_link_all_packages(name = "node_modules", imported_links = [], prod = Tru
...
-                ":node_modules/mobx-react",
-                ":node_modules/mobx",
-                ":node_modules/ms",
-                ":node_modules/react",
-                ":node_modules/rollup",
-                ":node_modules/uvu",
-                ":node_modules/@mycorp/pkg-a",
-                ":node_modules/@mycorp/pkg-d",
-                ":node_modules/@mycorp/pkg-e",
+                "//examples/npm_deps:node_modules/@aspect-test/a",
```
jbedard: Is there any harm in these strings increasing in size?

This PR also creates N instances of each `:node_modules/{pkg}` where N = number of pnpm projects that import `{pkg}`.

The alternative is only changing `npm_link_targets()` to prepend the package before returning the array. Then we don't have N versions of each package, but instead prepend `//{pkg}` to every string, but that is the rarely used `npm_link_targets()` API... 🤔

jbedard: Actually I think that was a stupid idea and we should just prepend the path in `npm_link_targets()`. Updated...

e2e/npm_translate_lock_empty/snapshots/npm_defs.bzl
```diff
@@ -70,4 +70,4 @@ def npm_link_targets(name = "node_modules", package = None, prod = True, dev = T
 
     link_targets = []
 
-    return link_targets
+    return ["//%s%s" % (bazel_package, target) for target in link_targets]
```
dzbarsky, line 73: do we care that these are not fully-qualified? Is someone gonna want that next? :)

jbedard: That's a good question, but I will leave that for someone to complain about next 😅

PR #2657, opened 2026-01-16
fix: quotes in js_binary(env)

### LINKED ISSUE #2590: [Bug]: js_binary env values should be quote-escaped
### What happened?

Currently when setting env vars via the `js_binary` `env` attribute to a JSON-encoded struct or list, the string value will contain quotes which break when inlined in the launcher bash script.

e.g.

```python
js_binary(
...
env = {
            "LINT_ANGULAR_SELECTORS_ALLOWLIST": json.encode(kwargs.pop("angular_selectors_allowlist", None))
}
)
```

with 
`angular_selectors_allowlist = ["twx"]` above, `LINT_ANGULAR_SELECTORS_ALLOWLIST` is bound to the following value in the bash wrapper script:

```shell
export LINT_ANGULAR_SELECTORS_ALLOWLIST="["twx"]"
```

which produces unintuitive errors downstream.

rules_js (or bazel-lib) should pass through values (esp json arrays or objects) that contain embedded double-quotes seamlessly, by escaping the embedded quotes. 

### Version

Output of `bazel --version`:
8.5.0

bazel_dep(name = "aspect_rules_js", version = "2.8.3")

### How to reproduce

```shell

```

### Any other information?

_No response_

jbedard:
Fix #2590

`js_binary(env)` values are now bash-escaped before being inserted into the generated bash shell script. If escaping was previously done manually before passing variables to `js_binary(env)` that escaping may have to be removed to avoid double-escaping.

### Test plan

- New test cases added

### DIFF
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -324,14 +324,17 @@ _ATTRS = {
     ),
 }
 
-_ENV_SET = """export {var}=\"{value}\""""
-_ENV_SET_IFF_NOT_SET = """if [[ -z "${{{var}:-}}" ]]; then export {var}=\"{value}\"; fi"""
+_ENV_SET = """export {var}={quoted_value}"""
+_ENV_SET_IFF_NOT_SET = """if [[ -z "${{{var}:-}}" ]]; then export {var}={quoted_value}; fi"""
 _NODE_OPTION = """JS_BINARY__NODE_OPTIONS+=(\"{value}\")"""
 
 def _expand_env_if_needed(ctx, value):
     if ctx.attr.expand_env:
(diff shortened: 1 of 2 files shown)


jbedard:
FYI @matthewjh does this look like it will solve #2590 for your scenario?

jbedard:
Note that [js/private/test/snapshots/launcher.sh](https://github.com/aspect-build/rules_js/blob/7c85b56117bd3abadaa709f75a45618be2391e9a/js/private/test/snapshots/launcher.sh#L15-L26) does not change which is an example of all the quoted `"1"`s etc

matthewjh:
@jbedard  I think this should do it - thanks!

kormide (approved):

PR #2697, opened 2026-01-24
chore: Fix up a few broken links

MatrixFrog:
It looks like there are a few others as well but I wanted to at least fix a few that I came across.

### DIFF
--- a/tools/preset.bazelrc
+++ b/tools/preset.bazelrc
@@ -103,7 +103,7 @@ common --heap_dump_on_oom
 
 # Allow the Bazel server to check directory sources for changes. Ensures that the Bazel server
 # notices when a directory changes, if you have a directory listed in the srcs of some target.
-# Recommended when using [copy_directory](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/copy_directory_doc.md)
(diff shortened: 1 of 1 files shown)


tools/preset.bazelrc
```diff
@@ -103,7 +103,7 @@ common --heap_dump_on_oom
 
 # Allow the Bazel server to check directory sources for changes. Ensures that the Bazel server
 # notices when a directory changes, if you have a directory listed in the srcs of some target.
-# Recommended when using [copy_directory](https://github.com/bazel-contrib/bazel-lib/blob/main/docs/copy_directory.md)
+# Recommended when using [copy_directory](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/copy_directory_doc.md)
```
jbedard: I think this one is wrong?

jbedard: Actually this file shouldn't be updated at all, this is generated and that is the reason the build is failing...

jbedard: Can you just revert this one? Other LGTM and thanks for the updates 👍

alexeagle:
Thanks!!

alexeagle:
@MatrixFrog note https://github.com/aspect-build/rules_js/pull/2703#issuecomment-3820595004 is the follow-up to make all broken links a merge-blocker

MatrixFrog:
That's awesome ✨

PR #2698, opened 2026-01-26
feat: support reproducible pnpm downloads using bazel facts api

### LINKED ISSUE #2695: [FR]: use bzlmod facts instead of declaring all pnpm versions+hashes
... or use `reproduceable` for bazel <9?

jbedard:
Close #2695

---

PNPM versions not known to rules_js may now be used without manually specifying integrity hashes by using the Bazel 9 [facts API](https://bazel.build/rules/lib/builtins/module_ctx#facts) to fetch+persist pnpm integrity hashes in the MODULE lock file.

### Test plan

- Manual testing; please provide instructions so we can reproduce: when using bazel9, update MODULE.bazel to use a `pnpm_version` not specified in `npm/private/versions.bzl`

### DIFF
--- a/npm/extensions.bzl
+++ b/npm/extensions.bzl
@@ -358,8 +358,15 @@ def _pnpm_extension_impl(module_ctx):
         pnpm_repository(
             name = name,
             pnpm_version = pnpm["version"],
+            integrity = pnpm["integrity"],
             include_npm = pnpm["include_npm"],
         )
+
+    kwargs = {}
+    if resolved.facts:
+        kwargs["facts"] = resolved.facts
+
+    return module_ctx.extension_metadata(reproducible = True, **kwargs)
 
 pnpm = module_extension(
     implementation = _pnpm_extension_impl,
--- a/npm/private/pnpm_extension.bzl
+++ b/npm/private/pnpm_extension.bzl
@@ -2,6 +2,7 @@
 
 load("@bazel_lib//lib:lists.bzl", "unique")
 load(":pnpm_repository.bzl", "DEFAULT_PNPM_VERSION", "LATEST_PNPM_VERSION")
+load(":versions.bzl", "PNPM_VERSIONS")
 
 DEFAULT_PNPM_REPO_NAME = "pnpm"
 
@@ -20,16 +21,12 @@ def resolve_pnpm_repositories(mctx):
       A struct with the following fields:
       - `repositories`: dict (name -> {version, include_npm, integrity}) to invoke `pnpm_repository` with.
       - `notes`: list of notes to print to the user.
+      - `facts`: Bazel module Facts to persist for future use (may be None).
     """
 
-    registrations = {}
+    # Collect all the module tags and associated versions/integrities/options
     integrity = {}
-
-    result = struct(
-        notes = [],
-        repositories = {},
-    )
-
+    registrations = {}
     for mod in mctx.modules:
         for attr in mod.tags.pnpm:
             if attr.name != DEFAULT_PNPM_REPO_NAME and not mod.is_root:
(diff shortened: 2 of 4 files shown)


npm/private/pnpm_repository.bzl
```diff
@@ -27,38 +27,35 @@ def pnpm_repository(name, pnpm_version = DEFAULT_PNPM_VERSION, include_npm = Fal
...
-                """js_binary(
+    if native.existing_rule(name):
+        fail("Repository with name '{}' already exists".format(name))
+
+    key = "{}@{}".format("pnpm", pnpm_version)
+
+    _npm_import_rule(
+        name = name,
+        key = key,
+        integrity = integrity,
```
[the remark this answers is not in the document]
jbedard: This is an internal API. The logic to automatically determine the sha512 is in the public bzlmod API now.

npm/private/pnpm_extension.bzl
```diff
@@ -102,6 +107,64 @@ def resolve_pnpm_repositories(mctx):
...
+                integrity = mctx.facts.get(pnpm["version"], None)
+                if not integrity:
+                    # Only fetch the list of pnpm versions once, and only if the version is not found in existing facts
+                    if not fetched_facts:
+                        fetched_facts = _fetch_pnpm_versions(mctx)
+                    integrity = fetched_facts.get(pnpm["version"], None)
+
+                if integrity:
+                    pnpm["integrity"] = integrity
+                    used_facts[pnpm["version"]] = integrity
```
jbedard, line 130: Note the facts are a plain `version=>integrity` map resulting in things like:
```
  "facts": {
    "//npm:extensions.bzl%pnpm": {
      "10.28.2": "sha512-QYcvA3rSL3NI47Heu69+hnz9RI8nJtnPdMCPGVB8MdLI56EVJbmD/rwt9kC1Q43uYCPrsfhO1DzC1lTSvDJiZA==",
      "11.0.0-alpha.1": "sha512-SODM4kjibEEtzW4S+H57MNR6xr6Bwi1oOvqFgDT/dH0Lp7rZAGCyA1kSBXQG6Si9WxNErRmcSMqQ11E/uY5cZg=="
    }
  }
```
(that was my test trying 10.28.2 and a v11 alpha which are not in our `versions.bzl`)

Note the `//npm:extensions.bzl%pnpm` key because that is the extension this is in, and the data is ONLY this `version=>integrity` map and it gets overwritten every time the `//npm:extensions.bzl%pnpm` extension runs.

We could potentially wrap this map into a parent `pnpm_versions` map or something like that but currently this extension is fairly minimal and since it gets overwritten each time I think it's ok if it has breaking changes in the future and it just gets overwritten 🤷

npm/private/pnpm_extension.bzl
```diff
@@ -102,6 +107,64 @@ def resolve_pnpm_repositories(mctx):
...
+    )
+
+def _fetch_pnpm_versions(module_ctx):
+    """Fetches pnpm versions and their integrity hashes from the npm registry.
 
-    return result
+    Returns:
+        A dict mapping version strings to dicts with 'integrity' keys.
+    """
+    result = module_ctx.download(url = ["https://registry.npmjs.org/pnpm"], output = "pnpm_versions.json")
```
dzbarsky, line 144: url is a list? is that right?

npm/extensions.bzl
```diff
@@ -354,13 +354,22 @@ def _pnpm_extension_impl(module_ctx):
         # buildifier: disable=print
         print(note)
 
+    extension_metadata = {
```
dzbarsky: nit: i usually name it `kwargs` and put it right before the call (so flip with the loop below?)

npm/private/pnpm_extension.bzl
jbedard: I think I copied it from the rules_go change to use facts which had multiple URLs listed. I assume it accepts either? 🤷

npm/private/pnpm_extension.bzl
fmeum: That's fine as long as a future extension is prepared to deal with any namespace collisions. A pnpm version called `pnpm_versions` seems unlikely though :⁠-⁠)

fmeum (approved):

Issue #2261, opened 2025-07-06
[Bug]: bazel test //examples/npm_deps:test10 fails on windows with bazel 8+


peakschris:
### What happened?

We have an issue in our repo that prevents any of our tests from running. I have found the same issue in rules_js itself. It appears to have regressed between bazel 7.6.1 and bazel 8.0.0 and is still failing with bazel 8.3.1 and coincidentally perhaps I have noticed that the node symlinks were previously displayed as 'SYMLINKD' but in bazel 8.0.0+ are displayed as 'JUNCTION'

Setup:
git clone rules_js

%workspace%/windows.bazelrc:
```
# Enable platform specific config
common --enable_platform_specific_config

# Enable symlinks on windows
startup --windows_enable_symlinks

# Runfiles are disabled by default on Windows but rules_js requires them.
common:windows --enable_runfiles
```

Add to .bazelrc:
```
import %workspace%/windows.bazelrc
```

Update .bazelversion:
```
8.0.0
```

>bazel test //examples/npm_deps:test10
```
INFO: From Testing //examples/npm_deps:test10:
DEBUG: D:/workdir/github/rules_js/npm/private/npm_translate_lock_state.bzl:27:14:
WARNING: `update_pnpm_lock` attribute in `npm_translate_lock(name = "npm")` is not yet supported on Windows. This feature will be disabled for this build.
==================== Test output for //examples/npm_deps:test10:
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module '@mycorp/pkg-e'
Require stack:
- D:\udu\b\6fsffodd\execroot\_main\bazel-out\x64_windows-fastbuild\bin\examples\npm_deps\test10_\test10.bat.runfiles\_main\examples\npm_deps\case10.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (D:\udu\b\6fsffodd\execroot\_main\bazel-out\x64_windows-fastbuild\bin\examples\npm_deps\test10_\test10.bat.runfiles\_main\examples\npm_deps\case10.js:1:1)
    at Module._compile (node:internal/modules/cjs/loader:1103:14)
[...]

PR #2700, opened 2026-01-27
fix: npm_deps tests on Windows, add a fix for Bazel 9.0.0 #2261 

Mivr:
Bazel 8.x is not compatible with Windows at the moment due to: https://github.com/bazelbuild/bazel/issues/26701.
This branch fixes some common Windows issues so that the bug formentioned can be observed at all.
Then it introduces a fix for Bazel 9.0.0+ where the needed attribute is present.

Bazel 8.x will not work with rules_js until the fix is backported.

For now this seems unlikely: https://github.com/bazelbuild/bazel/issues/27607

Fixes: #2261 

### Test plan

### DIFF
--- a/npm/private/npm_package_store.bzl
+++ b/npm/private/npm_package_store.bzl
@@ -1,5 +1,6 @@
 "npm_package_store rule"
 
+load("@bazel_features//:features.bzl", "bazel_features")
 load("@bazel_lib//lib:copy_directory.bzl", "copy_directory_bin_action")
 load("@tar.bzl//tar:tar.bzl", "tar_lib")
 
@@ -8,6 +9,10 @@ load("//js/private:js_info.bzl", "JsInfo", "js_info")
 load(":npm_package_info.bzl", "NpmPackageInfo")
 load(":npm_package_store_info.bzl", "NpmPackageStoreInfo")
 load(":utils.bzl", "utils")
+
+_SUPPORTS_SYMLINK_TARGET_TYPE = bazel_features.rules.symlink_action_has_target_type
+
+_PACKAGE_STORE_PREFIX_LEN = len("node_modules/" + utils.package_store_root)
 
 _DOC = """Defines a npm package that is linked into a node_modules tree.
 
@@ -182,6 +187,10 @@ def _npm_package_store_impl(ctx):
     if not version:
         fail("No package version specified to link to. Package version must either be specified explicitly via 'version' attribute or come from the 'src' 'JsInfo|NpmPackageInfo', typically a 'js_library|npm_package' target")
 
+    package_store_prefix_len = _PACKAGE_STORE_PREFIX_LEN
+    if ctx.label.package:
+        package_store_prefix_len += len(ctx.label.package) + 1
+
     package_key = "{}@{}".format(package, version)
     package_store_name = utils.package_store_name(package_key)
     package_store_directory = None
@@ -282,7 +291,8 @@ deps of npm_package_store must be in the same package.""" % (ctx.label.package, 
             if dep_package_store_directory:
                 linked_package_store_directories.append(dep_package_store_directory)
                 for dep_alias in dep_aliases:
-                    files.append(_symlink_package_store(ctx, package_store_name, dep_package_store_directory, dep_alias))
+                    target = dep_package_store_directory.short_path[package_store_prefix_len:]
+                    files.append(_symlink_package_store(ctx, package_store_name, target, dep_alias))
             else:
                 # this is a ref npm_link_package, a downstream terminal npm_link_package
                 # for this npm dependency will create the dep symlinks for this dep;
@@ -296,7 +306,8 @@ deps of npm_package_store must be in the same package.""" % (ctx.label.package, 
             # only link npm package store deps from NpmPackageInfo if they have _not_ already been linked directly
             # from deps; fixes https://github.com/aspect-build/rules_js/issues/1110.
             if dep_package_store_directory not in linked_package_store_directories:
-                files.append(_symlink_package_store(ctx, package_store_name, dep_package_store_directory, dep_info.package))
+                target = dep_package_store_directory.short_path[package_store_prefix_len:]
+                files.append(_symlink_package_store(ctx, package_store_name, target, dep_info.package))
 
                 # Include the store info of all linked dependencies
                 npm_package_store_infos.append(dep_info)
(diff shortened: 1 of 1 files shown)


examples/npm_package/packages/pkg_a/index.js
```diff
@@ -14,32 +14,34 @@ function getAcornVersion() {
 }
 
 function sandboxAssert() {
-    if (!/-sandbox\/\d+\/execroot\//.test(__filename)) {
+    if (!/[/\\]execroot[/\\]/.test(__filename)) {
```
jbedard: If these tests are broken on Windows let's exclude them on Windows, not break the test

npm/private/utils.bzl
```diff
@@ -1,9 +1,19 @@
...
 load("@bazel_skylib//lib:paths.bzl", "paths")
 
+# Bazel 9+ supports target_type="directory" on ctx.actions.symlink which creates
+# junctions instead of file symlinks on Windows. Without this, Node.js gets EPERM
+# when traversing symlinks created by declare_symlink on Bazel 8+/Windows.
+# See https://github.com/bazelbuild/bazel/issues/26701
+# Seems like it won't be cherry picked: https://github.com/bazelbuild/bazel/issues/27607
+# Added in commit https://github.com/bazelbuild/bazel/commit/b9bbda939cddab807e34559cb7ee798febfa3861
+# If a backport happens this version just needs to be lowered to the 8.x version that has the backport
+_SUPPORTS_SYMLINK_TARGET_TYPE = ge("9.0.0")
```
cameron-martin: It would be good to add this to bazel_features instead of depending on internals of it.

examples/npm_package/packages/pkg_a/index.js
Mivr: Sorry, added an if statement that uses the proper regex per OS

npm/private/utils.bzl
Mivr: opened a PR for the addition: https://github.com/bazel-contrib/bazel_features/pull/124

examples/npm_package/packages/pkg_a/index.js
```diff
@@ -14,32 +14,37 @@ function getAcornVersion() {
 }
 
 function sandboxAssert() {
-    if (!/-sandbox\/\d+\/execroot\//.test(__filename)) {
+    const sandboxRe = process.platform === 'win32'
```
jbedard, line 17: The issue is that windows doesn't even have sandboxing, correct?

npm/private/utils.bzl
Mivr: Merged and released in https://github.com/bazel-contrib/bazel_features/releases/tag/v1.41.0

Updating this PR to use it

examples/npm_package/packages/pkg_a/index.js
Mivr: Yes Windows doesn't have a sandox, thus we need to validate different paths.

npm/private/utils.bzl
Mivr: Made also a PR to cherry pick the needed changes in to the 8.5.1 branch of Bazel: https://github.com/bazelbuild/bazel/pull/28477

examples/npm_package/packages/pkg_a/index.js
```diff
@@ -14,32 +14,37 @@ function getAcornVersion() {
...
     // Use of npm_package() copies files into the npm package store.
-    if (!__filename.includes('/node_modules/.aspect_rules_js/')) {
+    if (!/[/\\]node_modules[/\\]\.aspect_rules_js[/\\]/.test(__filename)) {
         throw new Error(`Not in package store: ${__filename}`)
     }
 
     // When running under test, files should be in runfiles.
     // This package may also be used as a run_binary(tool) and not in a test.
-    if (process.env.TEST_WORKSPACE) {
+    // On Windows, Node.js resolves junctions to their real path so __filename
```
jbedard, line 31: Is that a fs-patch bug? Why does that only happen on windows? 🤔

npm/private/utils.bzl
```diff
@@ -133,10 +140,13 @@ def _friendly_name(name, version):
 
 def _make_symlink(ctx, symlink_path, target_path):
```
jbedard: Does this need a `target_type` param? Or it needs to be renamed to `_make_directory_symlink`...

jbedard:
@Mivr can we fix the failing tests as well?

npm/private/utils.bzl
Mivr: Renamed to `_make_directory_symlink` since it's only used for directories.

examples/npm_package/packages/pkg_a/index.js
Mivr: Not an fs-patch bug. Node.js on Windows resolves junctions (directory symlinks) to their real path when returning `__filename`. This is expected Windows junction behavior.

jbedard: I mean... that is a bug. I'm just not sure if we can ever fix it?

npm/private/npm_package_store.bzl
```diff
@@ -493,8 +493,11 @@ def _symlink_package_store(ctx, package_store_name, target, name):
         name,
     )
     symlink = ctx.actions.declare_symlink(store_symlink_path)
-    ctx.actions.symlink(
```
jbedard, line 496: I wonder if there's a reason this doesn't use the util method? Should we change this to `make_directory_symlink`?

npm/private/utils.bzl
```diff
@@ -131,12 +138,15 @@ def _friendly_name(name, version):
     "Make a name@version developer-friendly name for a package name and version"
     return "%s@%s" % (name, version)
 
-def _make_symlink(ctx, symlink_path, target_path):
+def _make_directory_symlink(ctx, symlink_path, target_path):
     symlink = ctx.actions.declare_symlink(symlink_path)
-    ctx.actions.symlink(
```
jbedard, line 139: Can we keep the 2 params like before and only add `**_symlinksTypeArgs` to the end? Also make `_symlinksTypeArgs` a global (replace `_SUPPORTS_SYMLINK_TARGET_TYPE`) so we aren't constructing it 100% of the time?

npm/private/npm_package_store.bzl
jbedard: If we use the util method then utils won't have to expose `supports_symlink_target_type`?

npm/private/utils.bzl
jbedard: Or just `if/else` and each can hardcode it's args without any `**`

alexeagle:
Replaces #2325?

examples/npm_package/packages/pkg_a/index.js
Mivr: I think this is just how node.js is with junctions and we are kind of working around it.


What do you have in mind as alternative?

npm/private/npm_package_store.bzl
```diff
@@ -485,16 +476,11 @@ def npm_local_package_store_internal(package_store_name, package, version, src,
...
     # "node_modules/{package_store_root}/{package_store_name}/node_modules/{package}"
     store_symlink_path = "node_modules/{}/{}/node_modules/{}".format(
         utils.package_store_root,
         package_store_name,
         name,
     )
-    symlink = ctx.actions.declare_symlink(store_symlink_path)
-    ctx.actions.symlink(
-        output = symlink,
-        target_path = ("../../.." if "/" in name else "../..") + target,
```
jbedard, line 498: This logic got replaced with a `relative_file()` invocation? Is that expensive?

Mivr: It should be just string concats I think

jbedard (approved):

PR #2703, opened 2026-01-29
fix: produce starlark_doc_extract rules for public API

alexeagle:
Just switch to bazel-lib which has a macro for bzl_library that produces starlark_doc_extract if `native.package_name()` isn't /private/

Also introduce Lychee broken link checker, it comments on PRs, when broken links are fixed we can honor its exit code

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -110,22 +110,6 @@ jobs:
             folder: ${{ steps.folder.outputs.res }}
             os: ${{ steps.os.outputs.res }}
 
-    check-links:
-        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -110,6 +110,22 @@ jobs:
             folder: ${{ steps.folder.outputs.res }}
             os: ${{ steps.os.outputs.res }}
 
+    check-links:
```
jbedard: Drop?

alexeagle: sure, I'll just have to put back in the next PR

jbedard (approved):

PR #2707, opened 2026-01-30
chore: remove tests from gha which also run on aspect workflows

jbedard:
### Test plan

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,9 +4,9 @@ name: CI
 on:
     # Triggers the workflow on push or pull request events but only for the main branch
     push:
-        branches: [main, 2.x, 3.x]
+        branches: [main, 2.x]
     pull_request:
-        branches: [main, 2.x, 3.x]
+        branches: [main, 2.x]
 
     # Allows you to run this workflow manually from the Actions tab
     workflow_dispatch:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -210,19 +111,49 @@ jobs:
...
               env:
                   ASPECT_GH_PACKAGES_AUTH_TOKEN: ${{ secrets.ASPECT_GH_PACKAGES_AUTH_TOKEN }}
                   ASPECT_NPM_AUTH_TOKEN: ${{ secrets.ASPECT_NPM_AUTH_TOKEN }}
                   ASPECT_RULES_JS_FROZEN_PNPM_LOCK: 1
 
+    # Mac/Windows tests - only e2e/bzlmod
+    # Only run on main branch (not PRs) to minimize minutes (billed at 10X and 2X respectively)
+    # https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes
+    smoke:
+        if: github.ref_name == 'main' || contains(github.head_ref, 'macos') || contains(github.head_ref, 'windows')
```
jbedard, line 121: I'm keeping this logic, and wondering if I should have done the same with rules_ts?

.github/workflows/ci.yaml
```diff
@@ -24,52 +24,25 @@ jobs:
...
+    # Prepares dynamic test matrix values for ubuntu tests
+    e2e-tests-list:
         runs-on: ubuntu-latest
         steps:
             - uses: actions/checkout@v6
             - id: folder
               name: Prepare 'folder' matrix axis
+              # Include tests not run on Aspect Workflows: tests with test.sh, tests depending on secrets
               run: |
                   a=(
```
alexeagle, line 36: wdyt about radically simplifying not to have this bash at all?

alexeagle (approved):

.github/workflows/ci.yaml
jbedard: and just assuming the `secrets.*` always exist?

PR #2709, opened 2026-01-30
fix: convert hex from packageManager integrity to base64 format

sbarfurth:
The hash on the `packageManager` field in `package.json` is specified as a hex string for the sha512 checksum. This representation is what corepack [expects](https://github.com/nodejs/corepack?tab=readme-ov-file#when-authoring-packages) and what something like `pnpm install` can understand. However, Bazel needs an integrity hash in SRI format with a base64 representation of the checksum. The repository code will now convert from hex to base64 on the fly. This makes the `pnpm_version_from` field actually useful with a hash. Before it was only useful without it since the hash format was mismatched between corepack and Bazel and thus caused errors on one side or the other.

---

### Test plan

- New test cases added


jbedard:
Does this mean no one has ever used this feature before? lol
I wonder if we should instead just drop the feature? Especially with bazel9 facts [allowing the hash to be unspecified](https://github.com/aspect-build/rules_js/commit/37eb63ff72a677203aec58f57720dd97a4f788a8)...

e2e/pnpm_package_manager_integrity/package.json
```diff
@@ -0,0 +1,6 @@
+{
+    "packageManager": "pnpm@10.28.2+sha512.41872f037ad22f7348e3b1debbaf7e867cfd448f2726d9cf74c08f19507c31d2c8e7a11525b983febc2df640b5438dee6023ebb1f84ed43cc2d654d2bc326264",
```
jbedard: Can we maybe change this from an e2e test to either just be a unit test or maybe merge it into an existing e2e test? Every e2e test adds load to CI and unless it really needs to be an isolated module I'd like to avoid them...

sbarfurth:
It seems like it's somewhat new. People use it, but no usage on GitHub uses an integrity hash. I would like to, so selfishly I want to keep it. We probably won't upgrade to Bazel 9 right away, so this would be great to have for a while at least. Perhaps deprecate it for a while?

e2e/pnpm_package_manager_integrity/package.json
sbarfurth: Sure, I do have a unit test already in this PR. I'll revert the e2e.

sbarfurth: Done.

jbedard:
Maybe we can update at least 1 existing e2e test to use this though? `e2e/bzlmod` uses `pnpm_version_from` that has no hash in the package.json, so lets keep that test case, but maybe switch a random one like `e2e/update_pnpm_lock` to use `pnpm_version_from` with the version+hash in package.json?

sbarfurth:
> Maybe we can update at least 1 existing e2e test to use this though? `e2e/bzlmod` uses `pnpm_version_from` that has no hash in the package.json, so lets keep that test case, but maybe switch a random one like `e2e/update_pnpm_lock` to use `pnpm_version_from` with the version+hash in package.json?

Done.

e2e/update_pnpm_lock/package.json
```diff
@@ -1,5 +1,6 @@
 {
     "private": true,
+    "packageManager": "pnpm@10.28.2+sha512.41872f037ad22f7348e3b1debbaf7e867cfd448f2726d9cf74c08f19507c31d2c8e7a11525b983febc2df640b5438dee6023ebb1f84ed43cc2d654d2bc326264",
```
jbedard, line 3: This is different then the `9.15.9` version previously listed in MODULE. Does that mean this test might now be upgrading pnpm while it updates the lockfile? Maybe that's fine or a good thing 👍

sbarfurth: I would assume so. My thought here was that I figured I might as well bump this when I touch it.

sbarfurth:
Yeah, CI is unhappy since the lockfile needs updating. Would you prefer I update it or change the ref to the old version?

e2e/update_pnpm_lock/package.json
jbedard: That's probably fine. See the test failure though... I think that's just because the package.json changed so the hashes are different, so it updates the lockfile in the initial `bazel test //...` which the test doesn't want.

I think you just need to run that initial `bazel test //...` so it updates the lockfiel to v10 and updates the hashes file? Then include the updated lockfile and hashes in your change...

jbedard:
I think just run `bazel test //...` in that directory, it will update to the v10 lockfile format and you'll commit that updated file. Then the test (run via `test.sh`) should function again after that, where it does an initial `bazel test //....` and assert no changes, then it updates `@types/node` and runs `bazel test //...` again and asserts it does update.

sbarfurth:
I think the lockfile is actually unchanged, but the hash of the `package.json` obviously needed an update. I ran the update now. Test passes locally at least.

jbedard:
Do you need to pull or rebase for the new "org-checks" CI task to pass?

sbarfurth:
Merged upstream

sbarfurth:
@jbedard what do you think about backporting this to `2.x`?

jbedard:
Is something stopping you from using `3.x`?

sbarfurth:
> Is something stopping you from using `3.x`?

Mostly that it's not stable released yet. If that's imminent I can try bumping us to an RC with the expectation that it won't be for very long. Is there a timeline?

jbedard:
I hope to release the final 3.0 soon, and would prefer avoiding further work on 2.x unless there are bazel or ruleset compatibility issues.

sbarfurth:
> I hope to release the final 3.0 soon, and would prefer avoiding further work on 2.x unless there are bazel or ruleset compatibility issues.

I'll give 3.x a try on our repo tomorrow. Release notes seems relatively harmless, though it's hard to get a good overview across RCs.

jbedard:
It's mainly removing support for outdated things (bazel6, WORKSPACE, pnpm 8) which allows moving everything else forward. See the [announcement on slack](https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1768870647487119) for a bit more info.

So far I've had no one comment on the removal of old things and just a few minor bugs fixed in followup RCs, so I'm hoping that means everything is stable and ready for a final release 🤞

sbarfurth:
Turns out we cannot use `3.x`. We also depend on [flatbuffers](https://github.com/google/flatbuffers) which uses the `root_package` on `npm_translate_lock`. Seems that attribute is gone in v3.

```
ERROR: in tag at https://bcr.bazel.build/modules/flatbuffers/25.9.23/MODULE.bazel:61:23, unknown attribute root_package provided
```

So I guess it would be helpful to have this in v2 still.

jbedard:
Interesting. Can you poke them to remove the use of `root_package`? IIUC it isn't even necessary there.

You can also cherry-pick your fix into the 2.x branch if you'd like to land it there for now.

sbarfurth:
Did both now. Proposed removal of `root_package` on `flatbuffers`, which is probably not gonna release fast. Also created #2717 for cherry-pick to `2.x`.

jbedard:
Awesome, thanks! 👍

PR #2711, opened 2026-01-30
chore(docs): document public APIs, not private ones

alexeagle:


avm99963:
Hey @alexeagle, with this change we no longer have documentation for the attributes of private rules such as `js_binary`. Although the rule is private, it is exposed via a macro that doesn't have any sort of documentation.

Do you think we should add docs to the macros? Or move the docs from the rules to the macros? What would be the best practice here?

I could work on this, it's the least I can do to contribute back!

alexeagle:
@avm99963 oops! Thanks for the report. What we should do is re-export the underlying rule in the public API alongside the macro that wraps it, so both get documented. I appreciate the offer of help, but it's probably faster for me to do myself this time.

alexeagle:
https://docs.aspect.build/bazel/javascript/aspect_rules_js/js_defs shows the output of the starlark_doc_extract rules in this repo at HEAD. Seems to me that we do have js_library rule documented - can you help me understand where you're looking and what is missing?

avm99963:
@alexeagle Thanks! I was referring to the BCR docs: https://registry.bazel.build/docs/aspect_rules_js#function-js_binary.

BTW, since I always saw a login screen when accessing docs.aspect.build, I thought it was something internal to the Aspect organization (akin to *.corp.google.com). From your comment, I see that I can just sign up to view them :P

avm99963:
Some more "off-topic" stuff:

- To be more specific: I've been testing a little bit, and I only see the login screen when accessing some URLs like https://docs.aspect.build/bazel/javascript/aspect_rules_js/js_defs, but not at others like the homepage https://docs.aspect.build/.
- Also, in case it isn't under your radar, I saw there are many [outdated links](https://github.com/search?q=repo%3Aaspect-build%2Frules_js+docs.aspect.build%2Frules%2F&type=code) to docs.aspect.build in the codebase which end up redirecting to https://docs.aspect.build/rules-redirect.

alexeagle:
I fixed https://docs.aspect.build/bazel/javascript to load without requiring a login, I'm a bit confused on Mintlify json syntax for `public: true` not applying on parent elements :)

alexeagle:
https://bcr.stack.build/modules/aspect_rules_js/3.0.0-rc4/docs/js/defs.bzl/js_library is going to become the official BCR UI soon, I think that's showing everything it's supposed to?

Yes, I'd like to unwind all the links going to our useless rules/ and rulesets/ redirects. FYI @kapunahelewong

PR #2718, opened 2026-02-07
fix: native FS sandbox for ESM imports (#362)

Mivr:
## Summary

- Adds a native C shared library (`fs_patch_linux.so` / `fs_patch_macos.dylib`) loaded via `LD_PRELOAD` / `DYLD_INSERT_LIBRARIES` that intercepts libc `realpath()` to prevent ESM imports from escaping the Bazel sandbox
- The core `guarded_realpath` algorithm resolves paths component-by-component, stopping at the first symlink hop that would escape the configured sandbox roots (execroot + runfiles)
- Integrated into `js_binary` / `js_test` via the launcher template — automatically active when `patch_node_fs = True` (the default)

## Problem

Issue https://github.com/aspect-build/rules_js/issues/362: Node.js ESM resolver captures `realpathSync.native()` via destructuring **before** `--require` patches run:

```js
// Node.js internal ESM loader (runs at module evaluation time)
const { realpathSync } = require('fs');  // captures BEFORE patches
```

This means the JS-level fs patches applied by `register.cjs` never intercept the ESM resolver's `realpath` calls, causing `.mjs` entry points and dynamic `import()` to resolve through Bazel sandbox symlinks to the real source tree.

## Solution

Intercept `realpath()` at the C level via `LD_PRELOAD` — this runs before Node.js even starts, so the ESM resolver's captured reference is already patched.

### Files

| File | Lines | Purpose |
|------|-------|---------|
| `fs_patch.h` | 118 | Config struct, function declarations |
| `fs_patch_init.c` | 152 | `__attribute__((constructor))` — env var parsing, dlsym |
| `fs_patch_common.c` | 344 | Core: `guarded_realpath`, `is_sub_path`, `check_escape`, `normalize_path` |
| `fs_patch_linux.c` | 50 | Linux interpositions: `realpath`, `__realpath_chk`, `canonicalize_file_name` |
| `fs_patch_macos.c` | 35 | macOS interpositions via `__DATA,__interpose` |
| `js_binary.bzl` | +7 | `_fs_patch_native` attr |
| `js_binary.sh.tpl` | +33 | `LD_PRELOAD` / `DYLD_INSERT_LIBRARIES` setup |

### Algorithm

1. **Fast path**: Call real `realpath()`, check if the result escapes any sandbox root → if not, return it
2. **Slow path** (escape detected): Walk path component-by-component with `lstat` + `readlink`, stop at the first symlink hop that escapes → return the guarded (non-escaped) path

## Test plan

- [x] 2 C unit test suites (is_sub_path, check_escape, normalize_path, make_absolute, guarded_realpath) — `bazel test //js/private/fs_patches_native/test:all`
- [x] E2E test at `e2e/esm_sandbox/` — `.mjs` entry point verifying `realpathSync.native()` stays within sandbox roots
- [x] All 122 existing `//js/...` tests pass with zero regressions
- [x] Golden launcher file updated

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -16,6 +16,7 @@ bazel_dep(name = "bazel_lib", version = "3.0.0")
 bazel_dep(name = "bazel_features", version = "1.41.0")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "rules_cc", version = "0.1.0")
 bazel_dep(name = "rules_nodejs", version = "6.7.3")
 
 # Ensure any version of aspect_bazel_lib used includes:
--- a/e2e/esm_sandbox/.bazelignore
+++ b/e2e/esm_sandbox/.bazelignore
@@ -1,0 +1,1 @@
+node_modules
--- a/e2e/esm_sandbox/.bazelrc
+++ b/e2e/esm_sandbox/.bazelrc
@@ -1,0 +1,2 @@
+import %workspace%/../../tools/preset.bazelrc
+import %workspace%/../e2e.bazelrc
--- a/e2e/esm_sandbox/.bazelversion
+++ b/e2e/esm_sandbox/.bazelversion
@@ -1,0 +1,1 @@
+../../.bazelversion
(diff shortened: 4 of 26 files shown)


jbedard:
this is not a viable solution

PR #2721, opened 2026-02-11
feat: allow proto_library as js_library#dep

alexeagle:
Design:
- user expected to register a toolchain providing their protoc plugin for gen-es
- use an aspect to adapt `ProtoInfo` to `JsInfo`, so a proto_library target can appear anywhere a js_library does
- no need for `js_proto_library` targets to pollute the package where the proto_library appears
- similar to https://github.com/bazel-contrib/rules_ruby/pull/292

Example of use currently in https://github.com/aspect-build/rules_ts/pull/894

TODO:
- for typescript, place .d.ts files in the source tree, with @kormide using diff.bzl as a validation action. Maybe this belongs in rules_ts
- support connect_query - we didn't have any test/example usage of it

### DIFF
--- a/js/private/BUILD.bazel
+++ b/js/private/BUILD.bazel
@@ -32,6 +32,7 @@ bzl_library(
     deps = [
         ":js_helpers",
         "//js:providers",
+        "@protobuf//bazel/common:proto_common_bzl",
         "@protobuf//bazel/common:proto_info_bzl",
     ],
 )


js/toolchains/BUILD
```diff
@@ -0,0 +1,3 @@
+package(default_visibility = ["//visibility:public"])
+
+toolchain_type(name = "protoc_plugin")
```
alexeagle, line 3: @dzbarsky should we drop this in `contrib_toolchains` ?

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+load(
+    "//js/private:js_helpers.bzl",
+    "gather_npm_package_store_infos",
+    "gather_npm_sources",
+    "gather_transitive_sources",
+    "gather_transitive_types",
+)
+
+GEN_ES_PLUGIN_TOOLCHAIN = Label("//js/toolchains:protoc_plugin")
+PROTOC_TOOLCHAIN = "@protobuf//bazel/private:proto_toolchain_type"
```
jbedard: Why is this one not a `Label()` but the one above is?

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+GEN_ES_OUTPUT_FILE = "{proto_file_basename}_pb.js"
+GEN_ES_TYPINGS_OUTPUT_FILE = "{proto_file_basename}_pb.d.ts"
+
+# FIXME: support stripping import prefixes
+def _js_proto_aspect_impl(target, ctx):
+    if not ProtoInfo in target:
+        # Should not be possible, since our aspect declaration has required_providers = [ProtoInfo]
+        fail("ts_proto_aspect_impl: target %s does not provide ProtoInfo" % target.label)
+
+    # Skip generation for well-known-types (is this the right way??)
```
jbedard: Where do those "well-known-types" live? In a core `@protobuf//something` target? Or are they magic globals that no one needs to depend on?

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+    protoc_info = ctx.toolchains[PROTOC_TOOLCHAIN].proto
+    gen_es_info = ctx.toolchains[GEN_ES_PLUGIN_TOOLCHAIN].proto
+    deps = [gen_es_info.runtime]
+    js_outputs = []
+    dts_outputs = []
+
+    for src in target[ProtoInfo].direct_sources:
+        proto_file_basename = src.basename.replace(".proto", "")
+        js_output = ctx.actions.declare_file(GEN_ES_OUTPUT_FILE.format(proto_file_basename = proto_file_basename))
+        dts_output = ctx.actions.declare_file(GEN_ES_TYPINGS_OUTPUT_FILE.format(proto_file_basename = proto_file_basename))
```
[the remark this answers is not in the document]
jbedard: ☝️

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+        #       '--es_out=bazel-out/k8-fastbuild/bin' \
+        #       '--descriptor_set_in=bazel-out/k8-fastbuild/bin/external/com_google_protobuf/timestamp_proto-descriptor-set.proto.bin:bazel-out/k8-fastbuild/bin/example/thing/thing_proto-descriptor-set.proto.bin:bazel-out/k8-fastbuild/bin/example/place/place_proto-descriptor-set.proto.bin:bazel-out/k8-fastbuild/bin/example/person/person_proto-descriptor-set.proto.bin' \
+        #       example/person/person.proto
+        args = ctx.actions.args()
+        args.add_joined(["--plugin", "protoc-gen-es", gen_es_info.plugin.executable], join_with = "=")
+        args.add_joined(["--es_out", ctx.bin_dir.path], join_with = "=")
+        args.add_all(gen_es_info.out_replacement_format_flag.split(" "))
+        args.add("--descriptor_set_in")
+        args.add_joined(target[ProtoInfo].transitive_descriptor_sets, join_with = ctx.configuration.host_path_separator)
+        args.add_all(target[ProtoInfo].direct_sources)
```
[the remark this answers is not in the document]
jbedard: Shouldn't this just be `src`?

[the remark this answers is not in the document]
jbedard: Or should we be invoking `protoc_info.proto_compiler` only once? That would be even better I assume...

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+        )
+        js_outputs.append(js_output)
+        dts_outputs.append(dts_output)
+
+    return [
+        js_info(
+            target = ctx.label,
+            sources = depset(js_outputs),
+            types = depset(dts_outputs),
+            transitive_sources = gather_transitive_sources(js_outputs, ctx.rule.attr.deps),
```
jbedard, line 73: What is `ctx.rule.attr.deps`? I assume we don't want to be following `js_library(deps)` or `ts_project(deps)` here, but we would want to include the `JsInfo` outputed by the aspect for `proto_library(deps)`...? I'm not sure what this line is actually doing :/

jbedard:
If this is the direction other rulesets such as ruby then I definitely think we should do the same.

It might be nice to land this for the v3 release, which I thought was basically ready to go until seeing this...

js/private/proto.bzl
dzbarsky: huh? I haven't followed what's going on with proto but surely this can't be the public toolchain API lolz?

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+)
+
+GEN_ES_PLUGIN_TOOLCHAIN = Label("//js/toolchains:protoc_plugin")
+PROTOC_TOOLCHAIN = "@protobuf//bazel/private:proto_toolchain_type"
+GEN_ES_OUTPUT_FILE = "{proto_file_basename}_pb.js"
+GEN_ES_TYPINGS_OUTPUT_FILE = "{proto_file_basename}_pb.d.ts"
+
+# FIXME: support stripping import prefixes
+def _js_proto_aspect_impl(target, ctx):
+    if not ProtoInfo in target:
```
dzbarsky: dunno if you need this, i feel like the error on line 42 will be clear enough anyway, this is just noise?

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+        # Should not be possible, since our aspect declaration has required_providers = [ProtoInfo]
+        fail("ts_proto_aspect_impl: target %s does not provide ProtoInfo" % target.label)
+
+    # Skip generation for well-known-types (is this the right way??)
+    if target.label.workspace_root.startswith("external/protobuf"):
```
dzbarsky: wait, why? they do have runtime code that you need to import, no? They should not really be that special

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+        # Follow https://www.npmjs.com/package/@bufbuild/protoc-gen-es
+        # Create an action like
+        #     bazel-out/k8-opt-exec-2B5CBBC6/bin/external/com_google_protobuf/protoc $@' '' \
+        #       '--plugin=protoc-gen-es=bazel-out/k8-opt-exec-2B5CBBC6/bin/plugin/bufbuild/protoc-gen-es.sh' \
+        #       '--es_opt=keep_empty_files=true' '--es_opt=target=ts' \
+        #       '--es_out=bazel-out/k8-fastbuild/bin' \
+        #       '--descriptor_set_in=bazel-out/k8-fastbuild/bin/external/com_google_protobuf/timestamp_proto-descriptor-set.proto.bin:bazel-out/k8-fastbuild/bin/example/thing/thing_proto-descriptor-set.proto.bin:bazel-out/k8-fastbuild/bin/example/place/place_proto-descriptor-set.proto.bin:bazel-out/k8-fastbuild/bin/example/person/person_proto-descriptor-set.proto.bin' \
+        #       example/person/person.proto
+        args = ctx.actions.args()
+        args.add_joined(["--plugin", "protoc-gen-es", gen_es_info.plugin.executable], join_with = "=")
```
dzbarsky: i think this is shorter as `args.add(gen_es_info.plugin.executable, format="--plugin=protoc-gen-es=%s")` and generatess less garbage

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+        args.add_joined(["--es_out", ctx.bin_dir.path], join_with = "=")
```
dzbarsky: ctx.bin_dir.path is a bit evil, it prevents path mapping. Can we not avoid it?

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+        args.add_all(gen_es_info.out_replacement_format_flag.split(" "))
+        args.add("--descriptor_set_in")
+        args.add_joined(target[ProtoInfo].transitive_descriptor_sets, join_with = ctx.configuration.host_path_separator)
+        args.add_all(target[ProtoInfo].direct_sources)
+        ctx.actions.run(
+            executable = protoc_info.proto_compiler.executable,
+            arguments = [args],
+            progress_message = "Generating .js/.d.ts from %{label}",
+            mnemonic = "ProtocGenEs",
+            env = {"BAZEL_BINDIR": ctx.bin_dir.path},
```
dzbarsky: ugh, this also breaks path mapping. I really hate this rules_js needs this, it's a huge efficiency drag when using transitions. We should figure out a way to fix that (and honestly maybe this plugin could be rewritten in not-JS so it's lighter-weight? I love writing protoc plugins in Go ;) )

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+    required_providers = [ProtoInfo],
+    # Be a valid dependency of a ts_project rule
+    provides = [JsInfo],
+    toolchains = [
+        GEN_ES_PLUGIN_TOOLCHAIN,
+        PROTOC_TOOLCHAIN,
+    ],
+)
+
+# def _windows_path_normalize(path):
```
dzbarsky: snip?

js/private/proto.bzl
alexeagle: Lolz indeed my friend

alexeagle: Because this one is already fully qualified

js/private/proto.bzl
alexeagle: The @bufbuild/protoc-gen-es package has special handling for it. If you choose a different plugin then the answer might be different

js/private/proto.bzl
alexeagle: We only visit rules that produce ProtoInfo, so indeed this is just propagating the JsInfo from transitive proto_library

js/private/proto.bzl
dzbarsky: it's not fully qualified, that's an apparent name. If someone imports this symbol and they've aliased the repo (hi `com_google_protobuf`) it won't work. Best to just wrap for consistency IMO

js/private/proto.bzl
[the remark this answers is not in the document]
alexeagle: yeah just call it once. nice review Codex

js/private/proto.bzl
alexeagle: removed

js/private/proto.bzl
alexeagle:
```
Could not find file in descriptor database: bazel-out/darwin_arm64-fastbuild/bin/external/protobuf+/src/google/protobuf/_virtual_imports/any_proto/google/protobuf/any.proto: No such file or directory
```

which I think means I'm missing some include dir like `-I that_external/...`

js/private/proto.bzl
alexeagle: hm, maybe we can use a representative one of the output files we produce and derive the path protoc-gen-es wants to see here

js/private/proto.bzl
alexeagle: haha as the user you are free to choose whatever plugin you like, nothing here forces you to use a nodejs_binary

I filed an issue on Bazel that actions should be able to know the root of the output directory, that's the principled fix

js/private/proto.bzl
thesayyn: Yes, well-known-types are usually Message like types that the language specific runtimes implement as a standard library and have their generated code depend on for wkt. 

You never need to generate anything for them, you still need a dep on the runtime label.

js/private/proto.bzl
```diff
@@ -0,0 +1,187 @@
...
+    dts_outputs = []
+
+    for src in target[ProtoInfo].direct_sources:
+        proto_file_basename = src.basename.replace(".proto", "")
+        js_output = ctx.actions.declare_file(GEN_ES_OUTPUT_FILE.format(proto_file_basename = proto_file_basename))
+        dts_output = ctx.actions.declare_file(GEN_ES_TYPINGS_OUTPUT_FILE.format(proto_file_basename = proto_file_basename))
+
+        # Follow https://www.npmjs.com/package/@bufbuild/protoc-gen-es
+        # Create an action like
+        #     bazel-out/k8-opt-exec-2B5CBBC6/bin/external/com_google_protobuf/protoc $@' '' \
```
thesayyn: Hmm, this will make this code incompatible with any other ts/proto implementation. Generally i am all for making an opinionated choice but proto ecosystem passed beyond that long time ago.

thesayyn: i am referring to: `--es_opt`, `--es_out`

js/private/proto.bzl
jbedard: So this is `JsInfo` from `proto_library(deps)`? Does that mean most likely it will just be the same aspect applied to `proto_library(deps)`? Or will it be nothing?

js/private/proto.bzl
alexeagle: https://github.com/bazelbuild/bazel/issues/15470

js/private/proto.bzl
alexeagle: yes, the `--es_opt` is already under user control, I should be able to do the same with `--es_out` using https://github.com/protocolbuffers/protobuf/blob/95dc8a02ad7d64527f0850c17edb1b86d3996ad8/bazel/private/proto_lang_toolchain_rule.bzl#L97-L104

js/private/proto.bzl
alexeagle: yup, turns out it works fine to just descend and let protoc generate what it wants, the buf codegen doesn't reference the well-known-types from that output

js/private/proto.bzl
alexeagle: Yes it should be this same aspect that walked the deps of the immediate proto_library and returned JsInfo

js/private/proto.bzl
alexeagle: fixed with the `-I` flags from proto_common

js/private/proto.bzl
alexeagle: removed the evil, by using the `root` of the first js output

js/private/proto.bzl
alexeagle: fixed with `outputs[0].root` - I think :)

js/private/proto.bzl
alexeagle: got it now, no more plugin-specifics appear in this file. The user has complete control of the command line via their `lang_proto_toolchain`

js/private/proto_common.bzl
```diff
@@ -0,0 +1,58 @@
...
+    if proto_source_root.startswith(root.path):
+        #TODO: remove this branch when bin_dir is removed from proto_source_root
+        proto_source_root = proto_source_root.removeprefix(root.path).removeprefix("/")
+
+    if proto_source_root == "" or proto_source_root == ".":
+        return root.path
+
+    return root.path + "/" + proto_source_root
+
+proto_common = struct(
```
thesayyn, line 52: our copy differs from the upstream i assume?

js/private/proto.bzl
```diff
@@ -0,0 +1,92 @@
...
+    "gather_npm_sources",
+    "gather_transitive_sources",
+    "gather_transitive_types",
+)
+load(":proto_common.bzl", "proto_common")
+
+LANG_PROTO_TOOLCHAIN = Label("//js/toolchains:protoc_plugin")
+PROTOC_TOOLCHAIN = Label("@protobuf//bazel/private:proto_toolchain_type")
+
+def _js_proto_aspect_impl(target, ctx):
```
thesayyn, line 25: Much better now.

thesayyn (approved):

js/private/proto_common.bzl
alexeagle: yes, only in making the three helpers into public api. Their implementations are identical.

In fact, I think we could switch to `proto_common.compile` in the future. But rules_js requires a `BAZEL_BINDIR` env var which they don't pass, so the plugin won't execute.

PR #2722, opened 2026-02-12
chore(docs): add API docs for proto

alexeagle:
Includes some minor syntax sugaring to better match how protoc plugins are documented

### DIFF
--- a/js/proto.bzl
+++ b/js/proto.bzl
@@ -23,7 +23,7 @@ gen_es.protoc_gen_es_binary(
 ### Usage
 
 Just write `proto_library` targets as usual, or have Gazelle generate them.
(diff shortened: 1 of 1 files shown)


js/proto.bzl
```diff
@@ -1,16 +1,85 @@
...
+    name = "protoc_gen_es",
+)
+```
+3. Define a js_proto_toolchain that uses the plugin. See the rule documentation below.
+4. Update MODULE.bazel to register it, typically with a simple statement like `register_toolchains("//tools/toolchains:all")`
+
+### Usage
+
+Just write `proto_library` targets as usual, or have Gazelle generate them.
+Then reference them anywhere a js_library could appear, for example:
```
jbedard: backticks for `js_library`?

PR #2723, opened 2026-02-13
chore: trigger docs update after publish

alexeagle:
For https://github.com/aspect-build/docs/issues/140

Based on how bazel-central-registry triggers bcr-ui

Note, a PR is also coming on the docs repo to declare the trigger event type referenced here.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -11,6 +11,8 @@ on:
         secrets:
             BCR_PUBLISH_TOKEN:
                 required: true
+            ASPECT_DOCS_REPO_ACCESS_TOKEN:
+                required: true
     # In case of problems, let release engineers retry by manually dispatching
     # the workflow from the GitHub UI
     workflow_dispatch:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -10,6 +10,8 @@ on:
         secrets:
             BCR_PUBLISH_TOKEN:
                 required: true
+            ASPECT_DOCS_REPO_ACCESS_TOKEN:
+                required: true
     # Or, developers can manually push a tag from their clone
     push:
         tags:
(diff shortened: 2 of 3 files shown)


jbedard:
We can't have the docs website scrape each ruleset? So we don't have to have publish logic like this on every ruleset...?

alexeagle:
> We can't have the docs website scrape each ruleset? 

Sure, we could, but how would it trigger? I did this on a cron before, but then you may have to wait to find out the new release doesn't look good on the docsite.
We can tuck this new thing into the .github/ repo so it's a reusable job, would you prefer that?

jbedard:
> Sure, we could, but how would it trigger?

I have no idea. How will we trigger it if we start scraping docs for repos we don't own? Maybe we do both? Start with this PR for now?

>  We can tuck this new thing into the .github/ repo so it's a reusable job, would you prefer that?

Does that at least prevent us from duplicating it in all our rulesets?

alexeagle:
Sharing it doesn't really make it any shorter, it's already calling one action:

If it's inline:
```
    trigger_docsite:
        runs-on: ubuntu-latest
        needs: ['publish']
        steps:
            - name: Trigger stardoc pipeline for docs.aspect.build
              uses: peter-evans/repository-dispatch@v4
              with:
                  repository: aspect-build/docs
                  token: ${{ secrets.ASPECT_DOCS_REPO_ACCESS_TOKEN }}
                  event-type: on-docs-trigger
                  client-payload: '{"release_tag": "${{ inputs.tag_name }}", "ruleset_repo": "${{ github.repository }}"}'

```

If it's in aspect-build/.github:
```
    publish-docs:
        runs-on: ubuntu-latest
        needs: publish
        steps:
          - name: Publish stardocs to docs.aspect.build
            uses: aspect-build/.github/actions/publish-docs@main
            with:
                tag_name: ${{ inputs.tag_name }}
                token: ${{ secrets.ASPECT_DOCS_REPO_ACCESS_TOKEN }}
```

jbedard:
Then we don't have to repeat the secrets in every repo? Whatever you think is best. But as soon as we copy+paste things into N repos they will get out of sync and some repos will forget to do it completely...

alexeagle:
Dropping in favor of a GitHub App that can be installed to repos where we want to act on release publish events

Issue #2725, opened 2026-02-13
[Bug]: Broken symlinks when rules_js is used in a module in the current module


snakethatlovesstaticlibs:
### What happened?

Reproduction repository https://github.com/snakethatlovesstaticlibs/rules_js-1218 (the repo name is not longer accurate since we broke this out into a new bug)

Start of conversation https://github.com/aspect-build/rules_js/issues/1218#issuecomment-3887279263

```
openapi-typescript failed: node:internal/modules/package_json_reader:316
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@redocly/openapi-core' imported from .../765cb3b671779a00aa5f54fa3538df00/sandbox/linux-sandbox/42/execroot/_main/bazel-out/aarch64-opt-exec/bin/external/rules_openapi_typescript+/openapi_typescript_cli_/openapi_typescript_cli.runfiles/rules_openapi_typescript+/node_modules/.aspect_rules_js/openapi-typescript@7.10.1_typescript@5.9.3/node_modules/openapi-typescript/bin/cli.js
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:316:9)
    at packageResolve (node:internal/modules/esm/resolve:768:81)
    at moduleResolve (node:internal/modules/esm/resolve:858:18)
    at defaultResolve (node:internal/modules/esm/resolve:990:11)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:718:20)
    at #resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:735:38)
    at ModuleLoader.resolveSync (node:internal/modules/esm/loader:764:52)
    at #resolve (node:internal/modules/esm/loader:700:17)
    at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:620:35)
    at ModuleJob.syncLink (node:internal/modules/esm/module_job:143:33) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

```
[...]

PR #2726, opened 2026-02-13
fix: run npm package target in other repo

jbedard:
Ref #2725

### Test plan

- New test cases added


jbedard:
@snakethatlovesstaticlibs are you able to try this fix out? It does resolve the primary issue with the bad symlink for me, but there may be other bugs regarding maybe the cwd when launching binaries in other repos (see the test I left commented out).

PR #2737, opened 2026-02-27
test: upgrade version of node-gyp used in tests

acozzette:
`//npm/private/test:node_modules_test` and
`//npm/private/test:pkg_versions_test` have been failing to build for me with an error like this:
```
  File "/home/adam/.cache/bazel/_bazel_adam/6ed6476489dfd4242e2dc3f0487ba25a/execroot/_main/bazel-out/k8-fastbuild/bin/node_modules/.aspect_rules_js/node-gyp@9.3.0/node_modules/node-gyp/gyp/pylib/gyp/input.py", line 19, in <module>
    from distutils.version import StrictVersion
ModuleNotFoundError: No module named 'distutils'
```
I suspect this has something to do with distutils being removed from Python 3.12. This commit fixes the tests by upgrading node-gyp to its most recent version.

### Changes are visible to end-users: no


jbedard:
That's a little concerning if the tests are depending on a system python? IIUC?

That said, upgrading the version used in tests SGTM either way 👍

PR #2742, opened 2026-03-03
Added Rust Wasm-Bindgen rules

UebelAndre:
This change adds rules for interfacing outputs from [wasm-bindgen](https://github.com/wasm-bindgen/wasm-bindgen) into these javascript rules. This was inspired on a convo with @alexeagle 

---

Added `wasm-bindgen` rules at `@aspect_rules_js//contrib/wasm_bindgen:defs.bzl`.

### Test plan

- New test cases added

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -17,8 +17,9 @@ bazel_dep(name = "bazel_features", version = "1.41.0")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "platforms", version = "1.0.0")
 bazel_dep(name = "rules_nodejs", version = "6.7.3")
-bazel_dep(name = "rules_rust", version = "0.69.0")
 bazel_dep(name = "rules_rust_wasm_bindgen", version = "0.69.0")
+
+bazel_dep(name = "rules_rust", version = "0.69.0", dev_dependency = True)
 
 # NB: LOWER BOUND on earliest BCR release of protobuf module, to avoid upgrading the root module by accident
 bazel_dep(name = "protobuf", version = "3.19.6")
--- a/contrib/wasm_bindgen/defs.bzl
+++ b/contrib/wasm_bindgen/defs.bzl
@@ -1,7 +1,5 @@
 """Rust WASM-bindgen rules for interfacing with aspect-build/rules_js"""
 
-# buildifier: disable=bzl-visibility
-load("@rules_rust//rust/private:providers.bzl", "ClippyInfo", "RustAnalyzerGroupInfo", "RustAnalyzerInfo")
 load("@rules_rust_wasm_bindgen//private:wasm_bindgen.bzl", "WASM_BINDGEN_ATTR", "rust_wasm_bindgen_action")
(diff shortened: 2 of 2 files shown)


MODULE.bazel
```diff
@@ -17,6 +17,8 @@ bazel_dep(name = "bazel_features", version = "1.41.0")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "platforms", version = "1.0.0")
 bazel_dep(name = "rules_nodejs", version = "6.7.3")
+bazel_dep(name = "rules_rust", version = "0.69.0")
```
jbedard: We can't add such deps to rules_js, that is too much of a mess.

Should this just be an example in the bazel-examples repo instead?

UebelAndre: My opinion is that it's a pretty unrelated dependency and that these rules should just live in a separate repo. My original plan was just to see https://github.com/UebelAndre/aspect_rules_js_rust_wasm_bindgen could be moved to this org but @alexeagle suggested this go in the contrib directory.

alexeagle: This is only to get three symbols

```
load("@rules_rust//rust/private:providers.bzl", "ClippyInfo", "RustAnalyzerGroupInfo", "RustAnalyzerInfo")
```

and I don't even know why it needs those. @UebelAndre is it possible to just skip forwarding those providers through this rule? Does it need to appear in a place a crate could?

UebelAndre: Great catch! I deleted that but the examples still use `@rules_rust` so I moved it to `dev_dependency = True`. Is that good enough isolation?

alexeagle: yup it's fine to be a dev_dep

contrib/wasm_bindgen/defs.bzl
```diff
@@ -0,0 +1,58 @@
...
+        js_info(
+            target = ctx.label,
+            sources = info.js,
+            transitive_sources = info.js,
+            types = info.ts,
+            transitive_types = info.ts,
+        ),
+    ]
+
+    crate = ctx.attr.wasm_file[0]
```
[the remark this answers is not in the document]
UebelAndre: This is due to the transition and is an expected part of Bazel's interface.

MODULE.bazel
```diff
@@ -17,6 +17,9 @@ bazel_dep(name = "bazel_features", version = "1.41.0")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "platforms", version = "1.0.0")
 bazel_dep(name = "rules_nodejs", version = "6.7.3")
+bazel_dep(name = "rules_rust_wasm_bindgen", version = "0.69.0")
```
jbedard, line 20: We can't have this either...

UebelAndre: That would suggest the rules need to live elsewhere. @alexeagle

UebelAndre:
Closing per https://github.com/aspect-build/rules_js/pull/2742#discussion_r2881195501

PR #2743, opened 2026-03-04
feat: generalize js_proto_toolchain() to work with other implementations

acozzette:
Currently we are not set up to support any JavaScript or TypeScript protobuf implementations other than protobuf-es from Buf.build. This is because we have a hard-coded assumption that the protobuf plugin will produce files matching the .proto file name except with "_pb.js" and "_pb.d.ts" in place of ".proto". This assumption holds for protobuf-es but not for other implementations such as Google's protobuf-javascript implementation.

This change addresses that problem by adding a new `output_file_extensions` parameter on `js_proto_toolchain()` to allow users to specify the generated files they expect the plugin to produce.

To make this work, I had to copy most of the logic from the `proto_lang_toolchain()` rule into a new `js_proto_toolchain()` rule. This was necessary because we need a rule that continues to produce the `ProtoLangToolchainInfo` provider but additionally produces a new `JsProtoToolchainInfo` provider with the new file extension information we need.

---

### Test plan

- New test cases added


js/private/js_proto_toolchain.bzl
```diff
@@ -0,0 +1,106 @@
...
+            providers = [ProtoInfo],
+            doc = """
+No code will be generated for files in the <code>srcs</code> attribute of
+<code>denylisted_protos</code>.
+This is used for .proto files that are already linked into proto runtimes, such as
+<code>any.proto</code>.""",
+        ),
+        "toolchain_type": attr.label(),
+        "output_file_extensions": attr.string_list(),
+    } | ({} if proto_common.INCOMPATIBLE_ENABLE_PROTO_TOOLCHAIN_RESOLUTION else {
```
alexeagle: nit: I think we could assume the flag is flipped, either the user is on Bazel 9 or has flipped it themselves

alexeagle (commented):
Did you confirm that other plugins actually make use of this new ability?

acozzette:
> Did you confirm that other plugins actually make use of this new ability?

So far I have only confirmed that it works with protobuf-es and Google's protobuf-javascript implementation, so I still need to confirm that for stephenh/ts-proto and timostamm/protobuf-ts.

js/private/js_proto_toolchain.bzl
acozzette: Great, that simplifies things a bit then. I removed all the conditionals on this flag so that we just assume it's enabled.

acozzette:
@alexeagle I'd like to clean up and improve my examples a bit, but otherwise I think this PR is about ready for a closer review when you have a chance. I was able to set up an example for each of the JS/TS implementations that Buf provides an SDK for.

Do you have any suggestions on the best way to make sure my examples run on CI?

js/proto.bzl
```diff
@@ -88,12 +99,18 @@ def js_proto_toolchain(name, plugin_name, plugin_options, plugin_bin, runtime, *
...
     command_line_flags = ["--{}_opt=%s".format(plugin_name) % o for o in plugin_options]
     command_line_flags.append("--{}_out=$(OUT)".format(plugin_name))
-    proto_lang_toolchain(
+    _js_proto_toolchain(
         name = name,
         command_line = " ".join(command_line_flags),
         plugin_format_flag = "--plugin=protoc-gen-{}=%s".format(plugin_name),
         toolchain_type = LANG_PROTO_TOOLCHAIN,
         plugin = plugin_bin,
+        output_file_extensions = output_file_extensions,
```
alexeagle, line 116: this needs to have a default value, otherwise it's a breaking change

alexeagle:
is there a good reason not to just make these examples/ and part of the root module? then they are easier for users to discover and don't require maintaining nested modules or CI changes

acozzette:
> is there a good reason not to just make these examples/ and part of the root module? then they are easier for users to discover and don't require maintaining nested modules or CI changes

I think that should work. Once we have all four toolchains registered in the same module, I suspect we'll need a config transition to be able to choose the one we want for each example without requiring a command-line flag. Let me give that a try.

js/proto.bzl
acozzette: That makes sense, I've set a default now.

acozzette:
I think I now have something mostly working that allows all the examples to live in the same module. It still needs a bit of cleaning up, though.

MODULE.bazel
```diff
@@ -24,6 +24,12 @@ bazel_dep(name = "bazel_lib", version = "3.2.2")
 # NB: LOWER BOUND on earliest BCR release of protobuf module, to avoid upgrading the root module by accident
 bazel_dep(name = "protobuf", version = "3.19.6")
 
+# Require a newer version for our own testing only.
```
jbedard: Can that move down into the "Dev dependencies" section?

MODULE.bazel
```diff
@@ -326,3 +332,26 @@ use_repo(
...
     "acorn__8.4.0__links",
 )
+
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+# The usual way to obtain this protoc plugin is to use the
+# @protocolbuffers/protoc-gen-js package. However, that package downloads
+# protoc using a post-install script, and this is not easy to accommodate in
+# Bazel. Instead, we just download the release binary from GitHub.
+http_archive(
```
jbedard: `dev_dependency = True,`?

examples/protobuf/stephenh-ts-proto/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
+load("//examples/protobuf:js_proto_test.bzl", "js_proto_test")
+load("//js:defs.bzl", "js_library")
+
+js_library(
+    name = "main_lib",
+    srcs = ["main.ts"],
```
jbedard: `.ts` in `js_library(srcs)`?

MODULE.bazel
acozzette: Good catch, thanks! Fixed

MODULE.bazel
acozzette: Good point, fixed

examples/protobuf/stephenh-ts-proto/BUILD.bazel
acozzette: Huh, I guess I just assumed that js_library() intentionally supported .ts files, but is that not the case? Some of these files are valid as JavaScript, but I don't think `timostamm-protobuf-ts/main.ts` is.

Do you have any suggestions on the best way to fix this? I don't think I can take a dependency on `rules_ts` since that would create a dependency cycle. I guess I could just use JS only, but maybe that's not ideal for the TypeScript-specific protobuf implementations.

jbedard: You can have an e2e test that depends on rules_ts for a full test, I assume you'll have to? Then for basic unit tests (such as anything in `examples/` or `*/test/`) just don't use .ts files.

acozzette: OK, that makes sense. I went ahead and replaced the examples to just use .js files for everything. I think that should be OK for now, just to get some working examples.

acozzette:
@jbedard I think this is ready for another look whenever you have a chance.

package.json
```diff
@@ -1,11 +1,19 @@
...
-        "google-closure-compiler": "^20251111.0.0",
+        "@bufbuild/protoc-gen-es": "2.11.0",
+        "@protobuf-ts/plugin": "2.11.1",
         "@types/node": "22.18.13",
         "chalk": "5.1.1",
         "circles": "npm:es5-ext@0.10.64",
+        "google-closure-compiler": "^20251111.0.0",
+        "google-protobuf": "4.0.2",
         "jsonpath-plus": "10.3.0",
+        "ts-proto": "2.11.4",
```
jbedard: This isn't used?

package.json
```diff
@@ -1,11 +1,19 @@
 {
     "private": true,
+    "dependencies": {
+        "@bufbuild/protobuf": "2.11.0",
```
jbedard: I don't see any of these being used?

js/proto.bzl
```diff
@@ -84,16 +90,35 @@ def js_proto_toolchain(name, plugin_name, plugin_options, plugin_bin, runtime, *
 
             Note that node module resolution requires the runtime to be in a parent folder of any package containing generated code.
 
+        output_file_extensions: The file extensions that the plugin is expected to produce.
+
+            These are interpreted as the suffix that should replace ".proto"
+            from the input file name. This parameter has a default value for
+            backward compatibility, but should be set explicitly.
+
+        target_settings: List of target config settings the toolchain is compatible with.
```
jbedard, line 99: I don't see these new params being tested anywhere?

package.json
acozzette: I'm not directly using it in code anywhere, but it is used in the toolchain definition in `examples/protobuf/toolchains/BUILD.bazel` as the runtime for gencode to depend on.

package.json
acozzette: I believe all of these are used too, sometimes not directly but just in the toolchain definition.

js/proto.bzl
acozzette: The examples under `examples/protobuf/` do rely on these parameters. The different plugins output different file names, so we exercise a few different values of `output_file_extensions`, and then `target_settings` we rely on for the config transition to be able to select the right implementation. (If we end up going with separate modules for the examples then we can just delete the `target_settings` param.)

package.json
jbedard: Why can't I find the the string "@bufbuild/protobuf" in this PR then? Or is the github UI hiding things from me?

acozzette: Hmm, I'm not sure, I can find it if I go to the "Files changed" tab and use Ctrl-F, as long as the affected file is expanded.

jbedard: Sorry it's just github un-loading things and breaking the native browser search...

examples/protobuf/tools/toolchains/BUILD
```diff
@@ -1,6 +1,8 @@
 load("@aspect_rules_js//js:proto.bzl", "js_proto_toolchain")
```
alexeagle: add a note in the moduledoc that we chose these four plugins because they're the ones Buf Schema registry picked

examples/proto/js_proto_test.bzl
```diff
@@ -0,0 +1,66 @@
...
+)
+
+def _js_proto_test_impl(ctx):
+    # Bazel does not allow one rule to forward an executable from another rule,
+    # so we need to create a new executable that just invokes the original one.
+    actual_exe = ctx.attr.actual[0][DefaultInfo].files_to_run.executable
+    out_exe = ctx.actions.declare_file(ctx.label.name)
+    ctx.actions.write(
+        output = out_exe,
+        content = "#!/bin/bash\nexec ./%s \"$@\"\n" % actual_exe.short_path,
```
alexeagle: I think you can fix this by having a different shape:

js_library() -> js_proto_transition() -> js_test()

in other words, don't try to expose a new rule that behaves like a js_test. Just create a new rule that does nothing but forward providers, and happens to have a transitioned config.

I think it's just

```
def _js_proto_transition_library_impl(ctx):
    # Just forward the providers
    return [
        DefaultInfo(
            files = ctx.attr.target[DefaultInfo].files,
            runfiles = ctx.attr.target[DefaultInfo].default_runfiles,
        )
    ]

js_proto_transition_library = rule(
    implementation = _js_proto_transition_library_impl,
    attrs = {
        "target": attr.label(cfg = _js_proto_toolchain_transition),
        "js_proto_implementation": attr.string(mandatory = True),
    },
)

```

see pattern in https://github.com/bazel-contrib/bazel-lib/blob/cac2d7855949d1b222fa26888892fbbe1d31015d/lib/transitions.bzl#L16 where we have other rules that just blindly apply a transition, independent of other things. Follows the Unix Philosophy: just do one thing and let the user compose

(There could still be a macro to compose them, but in an example I think it's better to optimize for readability than for maintenance)

examples/proto/BUILD.bazel
```diff
@@ -13,8 +14,51 @@ proto_library(
     deps = ["@protobuf//:timestamp_proto"],
 )
 
+# We define a flag here for determinining which JS protobuf implementation
```
alexeagle: the root BUILD file is a first place users will look, and the flag stuff is intimidating. Following the locality principle, it's also just referenced in tools/toolchains/BUILD - so I think it wants to move there

examples/protobuf/tools/toolchains/BUILD
acozzette: Sure, I added this to the README. (I'm not familiar with the term moduledoc, is that something else I should update?)

Actually though, the TypeScript examples are currently broken and it's not obvious how to make them work, because js_proto_aspect() is not set up to be able to produce TypeScript code. It's not clear to me if we can make this work unless we take some functionality from rules_ts and move it into rules_js. So depending on what we decide to do, I might need to remove the TypeScript examples for now.

examples/proto/js_proto_test.bzl
acozzette: Thanks, that does seem a lot simpler! I had to slightly tweak your code to additionally forward the JsInfo provider, but once I did that it worked like a charm.

examples/proto/BUILD.bazel
acozzette: That makes sense, I went and moved it there.

examples/protobuf/tools/toolchains/BUILD
alexeagle: moduledoc was just my shorthand for the bare string literal at the top of the starlark file

> TypeScript examples are currently broken

Ah, you mean the case where `.ts` files are produced instead of `.js/.d.ts` pairs. I think we should be able to make this work since `ts_project` can accept generated sources https://github.com/aspect-build/rules_ts/blob/main/examples/srcs/BUILD.bazel#L33

examples/proto/js_proto_test.bzl
alexeagle: great, yeah I thought about forwarding that one also, good job

examples/protobuf/tools/toolchains/BUILD
```diff
@@ -1,6 +1,51 @@
...
+    build_setting_default = "protobuf_es",
+    values = [
+        "protobuf_es",
+        "ts_proto",
+        "google_protobuf",
+        "protobuf_ts",
+    ],
+)
+
+config_setting(
```
alexeagle: nit: to make it a bit more readable I'd suggest organizing the file with a header for each plugin, then all the targets for that plugin

```
##############
# Protobuf-es
# https://github.com/bufbuild/protobuf-es
config_setting(
    name = "protobuf_es",
    flag_values = {
        ":js_proto_implementation": "protobuf_es",
    },
)

gen_es.protoc_gen_es_binary(
    name = "protoc_gen_es",
)
...

##############
# Ts-proto
# https://github.com/stephenh/ts-proto

...

```
etc

js/private/js_proto_toolchain_info.bzl
```diff
@@ -0,0 +1,6 @@
+"""The JsProtoToolchainInfo provider."""
```
alexeagle: nit: is there a rationale for having this in a separate file? seems like it's only referenced in js_proto_toolchain.bzl

examples/proto/js_proto_transition_library.bzl
```diff
@@ -0,0 +1,27 @@
...
+    return [
+        ctx.attr.target[0][DefaultInfo],
+        ctx.attr.target[0][JsInfo],
+    ]
+
+js_proto_transition_library = rule(
+    implementation = _js_proto_transition_library_impl,
+    attrs = {
+        "target": attr.label(cfg = _js_proto_toolchain_transition),
+        "js_proto_implementation": attr.string(mandatory = True),
```
alexeagle: nit: better error handling if we repeat the legal values here

alexeagle (commented):
looks approx ready to merge to me!

@jbedard for final approve

examples/protobuf/tools/toolchains/BUILD
acozzette: That makes sense, but it still seems not entirely clear how to fully get things working, because we need to produce .ts files all throughout the proto_library graph and not just at a single level where we can pass it to `ts_project`. Currently `js_proto_aspect` produces a `JsInfo` provider and my PR puts the .ts files straight into `JsInfo.sources`, but I don't think we're set up to handle .ts files there.

I did notice that the timostamm implementation has an `output_javascript` option that fixes this problem for us (though unfortunately it's still broken for an unrelated reason). But the stephenh implementation doesn't seem to have an option for this.

Maybe the ideal solution would be that in rules_ts we should have a `ts_proto_aspect` which does the same thing as `js_proto_aspect` but then additionally runs `tsc` if necessary to produce a `.js` file. Both aspects could use the same proto plugin toolchain type, but you would just have to use `ts_project()` if you need support for pure TypeScript protos. What do you think?

acozzette: As I think about it some more, I don't think my idea would work well, because then you could not have a `js_library()` depend on a pure-TypeScript proto. It seems to me that `ts_project()` somehow needs to kick off an aspect that can run `tsc` on the whole proto_library() graph if necessary.

examples/protobuf/tools/toolchains/BUILD
acozzette: Good idea, done

js/private/js_proto_toolchain_info.bzl
acozzette: There's no good reason, I went ahead and moved it into js_proto_toolchain.bzl.

examples/proto/js_proto_transition_library.bzl
acozzette: Good point, done.

examples/proto/tools/package.json
```diff
@@ -2,6 +2,9 @@
     "devDependencies": {
         "@bufbuild/protoc-gen-es": "~2.2.4",
         "@connectrpc/protoc-gen-connect-query": "^2.0.1",
-        "@types/node": "catalog:"
+        "@protobuf-ts/plugin": "2.11.1",
+        "@types/node": "catalog:",
+        "google-protobuf": "4.0.2",
```
jbedard: This one is never used?

acozzette: Good catch, thanks! I pushed another commit that removes it.

acozzette:
@jbedard Ok, I think this is ready for another look and maybe even ready to merge if it looks good to you.

examples/protobuf/tools/toolchains/BUILD
```diff
@@ -0,0 +1,87 @@
...
+support all the JS and TS implementations for which Buf provides an SDK:
+https://buf.build/bazel/bazel/sdks/main:protobuf
+"""
+
+load("@aspect_rules_js//js:proto.bzl", "js_proto_toolchain")
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
+load("@npm//tools:@bufbuild/protoc-gen-es/package_json.bzl", gen_es = "bin")
+load("@npm//tools:@connectrpc/protoc-gen-connect-query/package_json.bzl", gen_connect_query = "bin")
+
+# We define a flag here for determinining which JS protobuf implementation
```
jbedard: What is the reason for using this? Would a user reading this "example" ever do this or is this just to reduce the code in rules_js examples? Would 2 different example-modules be a more realistic example but just more code?

js/private/proto.bzl
```diff
@@ -26,9 +26,18 @@ def _js_proto_aspect_impl(target, ctx):
     proto_info = target[ProtoInfo]
     protoc_info = ctx.toolchains[PROTOC_TOOLCHAIN].proto
     proto_lang_toolchain_info = ctx.toolchains[LANG_PROTO_TOOLCHAIN].proto
-    js_outputs = proto_common.declare_generated_files(ctx.actions, proto_info, "_pb.js")
-    dts_outputs = proto_common.declare_generated_files(ctx.actions, proto_info, "_pb.d.ts")
-    output_root = js_outputs[0].root
+    js_proto_toolchain_info = ctx.toolchains[LANG_PROTO_TOOLCHAIN].js
```
jbedard, line 29: What if `.js` doesn't exist? When/why would that happen?

acozzette: I believe the only way this could happen would be if someone registered the toolchain without using our `js_proto_toolchain()` macro. It seems to me like it's safe to disregard that possibility, but if you disagree I can update this to check for that.

jbedard: If the `js_proto_toolchain()` macro is already the standard API then we can disregard this 👍

examples/protobuf/tools/toolchains/BUILD
acozzette: As we talked about yesterday, I went ahead and changed things back to just having a separate module for each protobuf implementation.

acozzette:
@jbedard I think this is now ready for another look. I'm not sure what the CI errors are about but maybe you have a better idea of what the problem is there?

jbedard:
You can ignore the `CI / org-check / *` CI for now

.aspect/workflows/config.yaml
```diff
@@ -147,7 +147,10 @@ workspaces:
     # e2e/update_pnpm_lock_with_import:
     # e2e/webpack_devserver:
     # e2e/webpack_devserver_esm:
-    examples/proto:
+    examples/protobuf/bufbuild-es:
```
jbedard, line 150: Do these 2 directories not need to be in `examples/.bazelignore`?

examples/protobuf/bufbuild-es/tools/toolchains/BUILD
```diff
@@ -1,18 +1,16 @@
 load("@aspect_rules_js//js:proto.bzl", "js_proto_toolchain")
 load("@npm//tools:@bufbuild/protoc-gen-es/package_json.bzl", gen_es = "bin")
-load("@npm//tools:@connectrpc/protoc-gen-connect-query/package_json.bzl", gen_connect_query = "bin")
 
 gen_es.protoc_gen_es_binary(
     name = "protoc_gen_es",
 )
 
-# FIXME: wire this up to another toolchain_type symbol?
```
jbedard, line 9: What happened to this? Is this one of the ones we declared we do not support?

.aspect/workflows/config.yaml
acozzette: I believe these are already OK since `examples/protobuf` is in the top-level `.bazelignore` file.

examples/protobuf/bufbuild-es/tools/toolchains/BUILD
acozzette: I just removed this one because it was not actually being used, though as far as I know there's nothing stopping us from supporting it.

jbedard (approved):

PR #2748, opened 2026-03-05
fix: declare public RUNFILES_DIR for use by @bazel/runfiles

### LINKED ISSUE #2747: [Bug]: Latest @bazel/runfiles does not work with rules_js@3.0.1
### What happened?

The latest @bazel/runfiles release no longer works correctly with rules_js v3 when running a js_binary with `bazel run`

Seems to be related to this commit: https://github.com/aspect-build/rules_js/commit/6b65708a04b239f8f9a104ed44707f7a0f849000

The RUNFILES environment variable is no longer set nor is RUNFILES_DIR

### Version

Output of `bazel --version`: 
```
Build label: 7.7.1
Build target: @@//src/main/java/com/google/devtools/build/lib/bazel:BazelServer
Build time: Wed Nov 12 17:35:45 2025 (1762968945)
Build timestamp: 1762968945
Build timestamp as int: 1762968945
```

### How to reproduce

```shell

```

### Any other information?

_No response_

jbedard:
Fix #2747

### Test plan

- New test cases added

### DIFF
--- a/e2e/runfiles/BUILD.bazel
+++ b/e2e/runfiles/BUILD.bazel
@@ -55,3 +55,47 @@ assert_contains(
     actual = ":out1.txt",
     expected = "test_fixture.md.generated_file_suffix",
 )
+
+# Test that a js_binary wrapping another js_binary correctly propagates RUNFILES_DIR
+# so the inner binary's @bazel/runfiles can locate its runfiles.
+js_binary(
+    name = "outer_binary",
+    data = [
+        ":node_modules/@bazel/runfiles",
+        ":test_binary",
+    ],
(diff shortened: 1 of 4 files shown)


dzbarsky:
Sure but that runfiles lib seems like it's doing the wrong thing, it should autodetect runfiles from binary location if env var is missing

jbedard:
> Sure but that runfiles lib seems like it's doing the wrong thing, it should autodetect runfiles from binary location if env var is missing

Even if it should do that... Don't you think we should still set `RUNFILES_DIR` so that it doesn't **have** to inspect the fs looking for the directory manually?

PR #2750, opened 2026-03-06
fix: retry rctx.execute() on SIGKILL (exit 137) for macOS 26

yetanotheralex:
## Problem

macOS 26 (Tahoe) has a [known bug](https://github.com/GoogleContainerTools/skaffold/issues/9925) in its networking framework's `pthread_atfork` handler that intermittently kills child processes with `SIGKILL` (exit code 137) when spawned via `fork+exec`. This affects Bazel's `rctx.execute()` calls in repository rules.

In `rules_js`, the `npm_translate_lock` repository rule calls `rctx.execute()` to run `yq` (for parsing `pnpm-lock.yaml`), `mkdir`, and `cp` (for copying input files). On macOS 26, these commands are intermittently killed, producing errors like:

```
ERROR: pnpm-lock.yaml parse error failed to parse pnpm lock file with yq.
'...yq .../pnpm-lock.yaml -o=json' exited with 137
```

or:

```
Failed to copy file. 'cp -f .npmrc ...' exited with 137
```

The failure is transient — retrying the same command succeeds.

## Fix

Add an `execute_with_retry()` helper in `utils.bzl` that catches exit code 137 (SIGKILL) and retries up to 3 times. Apply it to all `rctx.execute()` call sites in:

- `npm/private/npm_translate_lock_state.bzl` — `_yaml_to_json()` (yq), `_copy_input_file()` (mkdir, cp)
- `npm/private/utils.bzl` — `_reverse_force_copy()` (mkdir, cp)

The function is also exported via `utils.execute_with_retry` so downstream consumers can use it.

## References

- https://github.com/GoogleContainerTools/skaffold/issues/9925 — Go's `fork+exec` broken on macOS 26
- https://github.com/bazelbuild/bazel/issues/27026 — Bazel `wrapped_clang` broken on macOS 26
- https://github.com/basecamp/upright/issues/27 — macOS fork safety with Objective-C frameworks

## Testing

Verified on macOS 26.3 (Tahoe) with Bazel 7.7.1 — the patch resolves the intermittent SIGKILL failures during `npm_translate_lock` repository rule evaluation.

### DIFF
--- a/npm/private/npm_translate_lock_state.bzl
+++ b/npm/private/npm_translate_lock_state.bzl
@@ -335,7 +335,8 @@ def _copy_input_file(priv, rctx, attr, path, repository_path):
     if len(dst_segments) > 1:
         dirname = "/".join(dst_segments[:-1])
         mkdir_args = [coreutils, "mkdir", "-p", dirname]
-        result = rctx.execute(
+        result = utils.execute_with_retry(
+            rctx,
             mkdir_args,
             quiet = attr.quiet,
         )
@@ -344,7 +345,8 @@ def _copy_input_file(priv, rctx, attr, path, repository_path):
             fail(msg)
 
     cp_args = [coreutils, "cp", path, repository_path]
-    result = rctx.execute(
+    result = utils.execute_with_retry(
+        rctx,
         cp_args,
         quiet = attr.quiet,
     )
@@ -395,7 +397,7 @@ def _yaml_to_json(rctx, yaml_path, is_windows):
         yaml_path,
         "-o=json",
     ]
-    result = rctx.execute(yq_args)
+    result = utils.execute_with_retry(rctx, yq_args)
     if result.return_code:
         return None, "failed to parse {} with yq. '{}' exited with {}: \nSTDOUT:\n{}\nSTDERR:\n{}".format(yq_args, yaml_path, result.return_code, result.stdout, result.stderr)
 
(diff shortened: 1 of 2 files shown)


jbedard:
This is a bug with golang and macos 26? Can we fix this where that golang is located instead?

yetanotheralex:
> This is a bug with golang and macos 26? Can we fix this where that golang is located instead?

It's not a golang-specific bug -- it's a macOS 26 (Tahoe) bug in the OS networking framework's pthread_atfork handler that intermittently sends SIGKILL to any child process spawned via fork+exec.

During investigation we saw exit code 137 hitting not just yq (Go) but also cp, mkdir, openssl, and rg -- none of which are Go programs. The issue is on the parent side (Bazel's JVM calling ProcessBuilder/fork+exec), and the OS kills the child before it even gets a chance to run.

There's an open report on the skaffold project with the same root cause: https://github.com/GoogleContainerTools/skaffold/issues/9925

Until Apple ships a fix for macOS 26, the retry at the rctx.execute() call site is the pragmatic workaround -- there's nothing Go (or any of the affected binaries) can do differently since the OS is killing the process externally.

jbedard:
The [bazel issue](https://github.com/bazelbuild/bazel/issues/27026) seems to be resolved in 6.6 and no longer an issue in bazel 7.

What version of bazel are you using?

yetanotheralex:
> The [bazel issue](https://github.com/bazelbuild/bazel/issues/27026) seems to be resolved in 6.6 and no longer an issue in bazel 7.
> 
> What version of bazel are you using?

7.7.1

PR #2753, opened 2026-03-09
fix: handle missing packages in _collect_dep_constraints

alan-agius4:
When an optional dependency is a "link:" dependency which doesn't exist yet or otherwise not present in the `packages` dictionary, `npm_translate_lock` would fail with a "key not found" error such as: `Error: key "link:@angular/core|packages/core" not found in dictionary`

This occurs because `_collect_dep_constraints` iterates over optional dependencies but assumes they all exist in the `packages` dictionary. Link dependencies are often not represented in the lockfile as standard packages, leading to this missing key.


jbedard:
Can we just add a `link:./does-not-exist` to `optionalDependencies` somewhere such as in `e2e/pnpm_lockfiles/base/package.json`?

jbedard:
Done in https://github.com/aspect-build/rules_js/pull/2765

PR #2762, opened 2026-03-10
test: add tests of optionalDependencies link:s

jbedard:
### Test plan

- New test cases added


jbedard:
This isn't reproducing the error I was looking for and doesn't add enough value to be worth it atm

PR #2766, opened 2026-03-11
feat: copy proto .d.ts back to source tree

alexeagle:
This is a missing feature compared with the now-deprecated ts_proto_library

Thanks to @kormide for getting the new diff.bzl out the door!


js/private/proto.bzl
```diff
@@ -90,3 +99,20 @@ js_proto_aspect = aspect(
         PROTOC_TOOLCHAIN,
     ],
 )
+
+def _js_proto_library_impl(ctx):
+    return [
+        DefaultInfo(files = ctx.attr.proto[ProtoGeneratedTypesInfo].types),
```
jbedard: Normally we output the .js as the `DefaultInfo` and not types, I don't think `js_proto_library` should be any different?

js/proto.bzl
```diff
@@ -97,3 +100,22 @@ def js_proto_toolchain(name, plugin_name, plugin_options, plugin_bin, runtime, *
         runtime = runtime,
         **kwargs
     )
+
+def js_proto_library(name, proto, copy_types = None):
```
jbedard: Can't the `proto` target have many proto files? Doesn't that mean we need to output many "types" files?

kormide (commented):
LGTM. Will let @jbedard approve.

```
in deps attribute of starlark_doc_extract rule //js:proto.doc_extract: missing bzl_library targets for
Starlark module(s) @diff.bzl//:diff.bzl
```
Did I mess up the `bzl_library` targets in diff.bzl?

js/private/proto.bzl
```diff
@@ -22,6 +22,13 @@ load(":proto_common.bzl", "proto_common")
 LANG_PROTO_TOOLCHAIN = Label("//js/toolchains:protoc_plugin")
 PROTOC_TOOLCHAIN = Label("@protobuf//bazel/private:proto_toolchain_type")
 
+ProtoGeneratedTypesInfo = provider(
+    "provides the generated types for a proto target",
```
jbedard: Isn't that what `JsInfo` outputted by a proto target does already?

js/private/proto.bzl
alexeagle: yeah, but then the diff rule needs a label to use that accesses the types via an output_group

js/proto.bzl
alexeagle: yes, but I'm also hoping to avoid making the user repeat the list of output files like we did in rules_ts... working on it

js/private/proto.bzl
jbedard: Our macro wrapping the diff rule having an extra target to access the types sounds better then `js_proto_library` having completely different `DefaultInfo` then every other rule?

alexeagle: yup, and we already have output_group="types" convention, updated 👍🏻

alexeagle:
@kormide
> Did I mess up the bzl_library targets in diff.bzl

no, it's the same pattern as the other `*.bzl` modules - the syntax sugar is for end users, rules should use the `@diff.bzl//diff:defs.bzl` load statement for stardocs. Updated here.

js/private/proto.bzl
alexeagle: yup you're right we can just pick `types` off the JsInfo we get from the `proto` dep edge, no need for added provider kind

jbedard: Then we can completely delete this new provider?

js/private/proto.bzl
```diff
@@ -65,11 +65,12 @@ def _js_proto_aspect_impl(target, ctx):
         use_default_shell_env = True,
     )
 
+    types = depset(dts_outputs)
     return [
         js_info(
             target = ctx.label,
             sources = depset(js_outputs),
-            types = depset(dts_outputs),
```
jbedard, line 82: revert

js/proto.bzl
```diff
@@ -97,3 +101,30 @@ def js_proto_toolchain(name, plugin_name, plugin_options, plugin_bin, runtime, *
...
+
+    js_proto_library_rule(name = name, proto = proto)
+    if len(copy_types) > 0:
+        native.filegroup(name = "_{}.types".format(name), srcs = [name], output_group = "types")
+        diff_targets = []
+        for i, src_file in enumerate(copy_types):
+            gen_file = "_{}.gen_{}.d.ts".format(name, i)
+            diff_target = "{}.diff_{}".format(name, i)
+            select_file(name = gen_file, srcs = "_{}.types".format(name), subpath = src_file)
+            diff(
```
jbedard, line 120: @kormide did you guys think about any APIs to support multiple files?

js/private/proto.bzl
```diff
@@ -90,3 +90,24 @@ js_proto_aspect = aspect(
         PROTOC_TOOLCHAIN,
     ],
 )
+
+def _js_proto_library_impl(ctx):
+    return [
+        OutputGroupInfo(types = ctx.attr.proto[JsInfo].types),
```
jbedard, line 96: Is it ok this has no `DefaultInfo(files)` with the `JsInfo.sources` like `js_library` does?

I think having no `DefaultInfo(runfiles)` is probably correct though? Or that might be a discussion+test for another day...

jbedard:
You can rebase/pull to get my workaround for the bazel 9.0.1 issue (pinning to 9.0.0 for now).

js/proto.bzl
kormide: Not yet. Open to suggestions. @alexeagle and I are also discussing async.

js/proto.bzl
```diff
@@ -97,3 +101,30 @@ def js_proto_toolchain(name, plugin_name, plugin_options, plugin_bin, runtime, *
         runtime = runtime,
         **kwargs
     )
+
+def js_proto_library(name, proto, copy_types = []):
+    """Wrap a proto_library to invoke the js_proto_toolchain.
+
+    Can copy .d.ts type definitions back to the source folder.
```
acozzette, line 106: Will users want to create a `js_proto_library()` for every `proto_library()`, so that all .d.ts files get copied back to the source folder? This makes me wonder if `js_proto_library()` should use an aspect for this, so that you can get all the .d.ts files copied over without needing to replicate the whole proto_library dependency graph.

alexeagle: Users probably do want a lang_proto_library for each proto_library, that's already the common practice and it's Gazelle's job to create those targets in the BUILD file.

If we only wanted to have an aspect on a js_library -> proto_library edge, we already have one of those. I could experiment with the DX of doing it without any `js_proto_library` targets in the BUILD file.

I'll get an example landed first so we can base our PRs from that.

PR #2767, opened 2026-03-12
feat: add example of using proto_library with Node.js

alexeagle:
Also fix a bug where `js_library` deps ran the proto aspect, but js_binary/js_test did not.

/cc @acozzette


js/private/js_binary.bzl
```diff
@@ -99,6 +100,7 @@ _ATTRS = {
         NB: `data` files are copied to the Bazel output tree before being passed
         as inputs to runfiles. See `copy_data_to_bin` docstring for more info.
         """,
+        aspects = [js_proto_aspect],
```
jbedard, line 103: Putting this on a `data` instead of `deps` is confusing, but I guess that confusion already exists with `JsInfo` on those attributes 🤷

examples/proto/status_pb.d.ts
```diff
@@ -0,0 +1,29 @@
+// @generated by protoc-gen-es v2.2.5 with parameter "keep_empty_files=true,target=js+dts,import_extension=js"
```
jbedard: What wrote this to source?

jbedard: I think this file (and `user.d.ts`) shouldn't be in this PR?

js/private/js_binary.bzl
alexeagle: yeah js_binary doesn't have `deps` so this is where it has to go

examples/proto/status_pb.d.ts
alexeagle: They are manually copied to source at this point. Sure I can drop them for clarity.

js/private/js_binary.bzl
jbedard: Actually... can we drop this and just force a `js_library` to be used? Otherwise we will have to at this to 100 places...

PR #2791, opened 2026-04-03
test: reduce Bazel module count by consolidating e2e tests into examples

acozzette:
Remove 5 standalone e2e Bazel modules (44 → 39) by merging or moving 
tests that don't require their own module-level configuration:

- e2e/npm_link_package-esm: merged into e2e/npm_link_package by adding 
  .mjs ESM variants alongside existing .js files

- e2e/package_json_module: moved to examples/package_json_module; the
  "type: module" scoping is per-directory so it works as a sub-package

- e2e/runfiles: merged into examples/runfiles; added module-name path
  tests (module_name_test.js), nested binary runfiles propagation tests
  (outer.js), and js_run_binary tests (resolve-args.js)

- e2e/stamped_package_json: moved to examples/stamped_package_json;
  added jq.bzl as a dev dep to examples/MODULE.bazel; excluded from
  pnpm workspace since package.json contains a stamp template version

- e2e/worker: moved to examples/worker; no npm deps and bazel_lib +
  rules_nodejs are already available in examples

I also moved the protobuf examples from examples/protobuf into e2e/ for consistency.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

---

### Test plan


acozzette:
@jbedard Like we talked about yesterday, I had Claude try to consolidate the e2e tests to try to remove unnecessary extra Bazel modules. I'm not sure about the npm_link_package and runfiles changes, but the other 3 that were moved into examples/ look reasonable to me. What do you think?

e2e/npm_link_package/lib/package.json
```diff
@@ -1,4 +1,10 @@
 {
     "name": "@e2e/lib",
-    "private": true
+    "private": true,
+    "exports": {
```
jbedard: So this package now supports both cjs and esm imports? But we lost the `"type": "module"` test case?

acozzette: I believe the idea is that we can use `.mjs` files to override the default of CommonJS from the package.json file. But I do find this a bit confusing and would probably lean toward reverting the part of the change that combined these two test cases together.

jbedard: I would vote for merging the 2 e2e tests into 2 MODULE but keeping 2 sub-directories separate. I think the idea was you can diff the 2 directories and they are the exact same except a few key differences like `"type": "module"`...

acozzette: Did you mean to say merging them into one module with 2 sub-directories?

jbedard: Yeah. I don't think having them as separate MODULEs adds any value? But they are intentionally identical other then a few things like the `"type": "module"` and maybe js vs cjs vs mjs file extensions?

jbedard:
@codex

e2e/npm_link_package/lib/package.json
acozzette: OK, I went ahead kept them in one module but with separate subdirectories, and can confirm that they are mostly identical with only minor changes.

e2e/npm_link_package/MODULE.bazel
```diff
@@ -13,10 +13,17 @@ npm = use_extension(
     dev_dependency = True,
 )
 npm.npm_translate_lock(
-    name = "npm",
-    data = ["//:package.json"],
+    name = "npm_cjs",
+    data = ["//cjs:package.json"],
```
jbedard: I don't think we need separate `npm_translate_lock()` calls though? Part of the reason for merging the 2 directories is this code was duplicated with no differences?

acozzette: OK, I went back to a single call to `npm_translate_lock()`.

jbedard (approved):

jbedard (approved):

PR #2797, opened 2026-04-08
feat: add js_rpc_library and js_rpc_toolchain for ConnectRPC support

acozzette:
Adds a new `js_rpc_library` rule and `js_rpc_toolchain` macro to support protoc-based RPC code generators such as `protoc-gen-connect-query`.

 - `js_rpc_library`: a rule that runs a protoc plugin on its immediate proto_library deps and exposes the output as JsInfo.
 - `js_rpc_toolchain`: a macro wrapping `_js_proto_toolchain` but with a new `//js/toolchains:protoc_rpc_plugin` toolchain type.
 - Updates to `e2e/protobuf-es` showing how to use Connect and Connect-Query
 - Updates to `e2/protobuf-google` showing how to use grpc-js
    
Support for Connect-Query seems to be the only thing in the deprecated `ts_proto_library` rule that was lacking in `rules_js`. Hopefully now that we have support for it, we can start recommending that everyone migrate away from `ts_proto_library`.

---

A `js_rpc_library` rule is now available and can generate RPC bindings for any RPC system that is built on a protoc plugin. The plugin just needs to be registered with the `js_rpc_toolchain` macro.

### Test plan

- New test cases added


e2e/connect-query/tools/package.json
```diff
@@ -0,0 +1,6 @@
+{
+    "devDependencies": {
```
jbedard: Are these deps similar enough to be combined into an existing `e2e` test to avoid _another_ MODULE?

e2e/connect-query/.nvmrc
```diff
@@ -0,0 +1 @@
+24.11.0
```
jbedard: I think this is unused by bazel and the file should be deleted?

acozzette: This seems to be needed because it's referenced in MODULE.bazel, though I'm going probably going to remove this whole directory anyway and merge it into another module.

e2e/connect-query/tools/package.json
acozzette: Connect-Query is actually just a layer on top of protobuf-es and Connect, so I think they can all go in the same module in protobuf-es. I just pushed another commit to do that.

js/private/proto.bzl
```diff
@@ -53,47 +53,30 @@ def _js_proto_aspect_impl(target, ctx):
...
-    ctx.actions.run(
-        executable = protoc_info.proto_compiler.executable,
-        arguments = [args],
-        progress_message = "Generating .js/.d.ts from %{label}",
+        rewrite_args = [
+            "--es_opt=rewrite_imports=" + r
+            for r in _get_import_rewrites(ctx.rule.attr.deps, js_proto_toolchain_info.out_js_extension)
+        ]
+
+    proto_common.compile(
```
jbedard, line 66: Do you promise this is equivalent to before?

js/private/rpc.bzl
```diff
@@ -0,0 +1,143 @@
...
+    proto_lang_toolchain_info = ctx.toolchains[LANG_RPC_TOOLCHAIN].proto
+
+    if not ctx.attr.outs:
+        fail("The outs parameter must specify the expected output files.")
+
+    if len(ctx.attr.deps) != 1:
+        fail("There must be exactly one dependency in deps.")
+
+    output_files = [ctx.actions.declare_file(f) for f in ctx.attr.outs]
+    proto_info = ctx.attr.deps[0][ProtoInfo]
```
jbedard, line 30: What about `.deps[1...]`?

js/private/proto.bzl
acozzette: It's hard to say for certain but let me see if I can add some more tests so that we can be more confident this is equivalent.

js/private/rpc.bzl
acozzette: We know that there can be only one dep because we error out if len(deps) is not 1. This is a common convention at least for Google's protobuf and gRPC rules that even though deps is a list, it's supposed to have exactly one element. cc_proto_library() also does this for example: https://github.com/protocolbuffers/protobuf/blob/48ce31daef690ea584a6d7e0e9bc34928013f2e2/bazel/private/oss/cc_proto_library.bzl#L139 I'm not sure if this convention has caught on outside of Google, though, so I'm not set on it if you think it's too weird.

jbedard: What about other languages+rulesets other then cc?

If there is an odd convention we want to align with then sure, lets try to stay aligned. But if we were writing designing this ourselves I think we'd just do `proto = ":the-proto-target"` instead?

acozzette: The rules in the protobuf repo do this for upb, C++, hpb, and Lua. For other languages it looks like it's not enforced, but it is a convention that you only have one dep per `LANG_proto_library()` target. Either way is OK with me but I think I'd lean toward your idea since the `deps` convention is not universal.

jbedard: If it's not universal what do some other languages do?

acozzette: It looks like the name `deps` is pretty near universal among Google's rules, but a lot of them don't enforce that you have to have exactly one. (I was actually surprised to learn this just now, I assumed it was enforced).

jbedard: Ok, so `deps` and the `len() == 1` validation aligns with other `{lang}_proto_library()` rules the most then?

acozzette: Yes, I believe so. I asked Claude to check all the rules in the protobuf repo and it counted 8 with this validation and 3 without.

jbedard: Ok, well lets just follow that odd standard then 🤷

e2e/protobuf-es/BUILD.bazel
```diff
@@ -38,3 +39,56 @@ js_test(
...
+# Generates eliza_pb.js with message types and the service descriptor via js_proto_aspect.
+js_library(
+    name = "eliza_js",
+    deps = [":eliza_proto"],
+)
+
+# Generates TanStack Query hooks via protoc-gen-connect-query.
+js_rpc_library(
+    name = "eliza_rpc",
+    outs = [
```
jbedard, line 65: Didn't we discuss a scenario where the output files were predictable just given the proto inputs? I assume where all services are in one `eliza_connectquery.{js,d.ts}`?

If that is possible it would nice to have an example where that can be used and this `outs` doesn't have to be manually specified? 🤷

js/BUILD.bazel
```diff
@@ -39,6 +39,24 @@ starlark_doc_extract(
...
+        "//js/private:rpc",
+    ],
+)
+
+starlark_doc_extract(
+    name = "rpc.doc_extract",
+    src = "rpc.bzl",
+    deps = [
+        ":rpc",
+        "@protobuf//bazel/toolchains:proto_lang_toolchain_bzl",
```
jbedard: Shouldn't this be in the `:rpc` `deps`?

acozzette: Good catch, thanks! Though actually I've now changed things enough that this particular dep is no longer needed.

e2e/protobuf-es/BUILD.bazel
acozzette: Unfortunately we can't do this with Connect-Query since its output files are not predictable, but I added an example in e2e/protobuf-google of how to do this with grpc-js.

js/private/rpc.bzl
```diff
@@ -0,0 +1,143 @@
...
+    all_direct_sources = proto_info.direct_sources
+    all_descriptor_sets = [proto_info.transitive_descriptor_sets]
+    all_proto_paths = [proto_info.transitive_proto_path]
+
+    proto_common.compile(
+        actions = ctx.actions,
+        proto_lang_toolchain_info = proto_lang_toolchain_info,
+        protoc_info = protoc_info,
+        output_files = output_files,
+        proto_outdir = output_files[0].root.path,
```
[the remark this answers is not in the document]
acozzette: Done

js/private/proto.bzl
acozzette: I went ahead and some analysis tests and confirmed that they pass both before and after the other changes, so hopefully that along with the e2e tests can make us reasonably confident that this refactor is not breaking anything.

js/private/rpc.bzl
```diff
@@ -0,0 +1,158 @@
...
+            doc = """proto_library targets whose sources will be passed to the RPC plugin.
+
+There must be exactly one target in this list.""",
+            providers = [ProtoInfo],
+        ),
+        "outs": attr.string_list(
+            doc = """Explicit list of files the plugin will generate, relative to the package directory.
+
+This parameter must be specified if and only if the registered
+js_grpc_toolchain does not specify out_js_extension or out_dts_extension.
```
jbedard: `js_grpc_toolchain` => `js_rpc_toolchain`?

js/rpc.bzl
```diff
@@ -0,0 +1,137 @@
...
+
+        plugin_bin: The plugin binary. This should be the label of a binary target declared using the plugin's package_json.bzl entry point.
+        runtime: The npm package that the generated code imports from at runtime.
+
+            Note that node module resolution requires the runtime to be in a parent folder of any package containing generated code.
+
+        out_dts_extension: The suffix that should replace ".proto" in determining the .d.ts output file name, or None if the plugin does not produce a type declaration file.
+        out_js_extension: The suffix that should replace ".proto" in determining the .js output file name.
+
+            If neither of the two above parameters are set, then each js_grpc_library target must indicate its expected outputs via the outs parameter.
```
jbedard: `js_grpc_library` => `js_rpc_library`?

js/rpc.bzl
```diff
@@ -0,0 +1,137 @@
...
+
+    Args:
+        name: The name of the toolchain. A target named [name]_toolchain is also created, which is the one to use in register_toolchains.
+        plugin_name: The `NAME` of the plugin program, used in command-line flags to protoc, as follows:
+
+            > `protoc --plugin=protoc-gen-NAME=path/to/mybinary --NAME_out=OUT_DIR`
+
+            See https://protobuf.dev/reference/cpp/api-docs/google.protobuf.compiler.plugin
+
+        plugin_options: (List of strings) Command line flags used to invoke the plugin.
```
jbedard: A robot told me:

> These are plugin option values, not full flags. The macro assembles them into --NAME_opt=VALUE flags. The doc should say something like "Option strings passed to the plugin via --NAME_opt=".

Is the robot correct?

js/private/rpc.bzl
```diff
@@ -0,0 +1,158 @@
...
+js_rpc_library = rule(
+    implementation = _js_rpc_library_impl,
+    doc = """Generates RPC client stubs from proto_library targets using a configurable protoc plugin.
+
+Runs a protoc plugin (such as protoc-gen-connect-query) on the specified proto_library
+targets and provides the generated files as JsInfo, allowing them to be used as deps
+in any rule that accepts JsInfo (js_library, js_binary, js_test, etc.).
+
+Unlike js_library with proto deps (which uses js_proto_aspect to generate message
+serialization code), js_rpc_library operates only on its immediate deps and generates
```
jbedard: Should things like `js_rpc_library` have backticks? 🤔 
Same with `JsInfo`, `js_library` etc above and maybe more

js/private/rpc.bzl
```diff
@@ -0,0 +1,158 @@
...
+
+    output_files = []
+    if ctx.attr.outs:
+        output_files.extend([ctx.actions.declare_file(f) for f in ctx.attr.outs])
+    if js_proto_toolchain_info.out_js_extension:
+        output_files.extend(proto_common.declare_generated_files(ctx.actions, proto_info, js_proto_toolchain_info.out_js_extension))
+    if js_proto_toolchain_info.out_dts_extension:
+        output_files.extend(proto_common.declare_generated_files(ctx.actions, proto_info, js_proto_toolchain_info.out_dts_extension))
+
+    all_direct_sources = proto_info.direct_sources
```
jbedard: The `all_*` are just single entries.

Can:
```
proto_paths = depset(transitive = all_proto_paths),
```
just be
```
proto_paths = proto_info.transitive_proto_path,
```
?

js/rpc.bzl
```diff
@@ -0,0 +1,137 @@
...
+
+            If neither of the two above parameters are set, then each js_grpc_library target must indicate its expected outputs via the outs parameter.
+
+        target_settings: List of target config settings the toolchain is compatible with.
+        exec_compatible_with: List of constraint_values that the execution platform must be compatible with.
+        target_compatible_with: List of constraint values that the target platform must be compatible with.
+
+        **kwargs: Additional arguments to pass to the underlying rule.
+    """
+    command_line_flags = ["--{}_opt=%s".format(plugin_name) % o for o in plugin_options]
```
jbedard: Can the `.format() % x` just be replaced with a single `.format()` or single `%`?

js/private/rpc.bzl
acozzette: Good catch! Done.

js/rpc.bzl
acozzette: Done

js/rpc.bzl
acozzette: That does sound more correct, so I went with that phrasing instead.

js/private/rpc.bzl
acozzette: Done

js/private/rpc.bzl
acozzette: Done

js/rpc.bzl
acozzette: Done

jbedard (approved):

jbedard:
@acozzette can you just review the codex comments to see if those are applicable?

js/private/rpc.bzl
```diff
@@ -0,0 +1,158 @@
...
+    dts_extension = js_proto_toolchain_info.out_dts_extension
+
+    if not ctx.attr.outs and not js_extension and not dts_extension:
+        fail("No outputs were specified.")
+    if ctx.attr.outs and (js_extension or dts_extension):
+        fail("Outputs can be specified on the js_rpc_library or js_rpc_toolchain, but not both.")
+
+    output_files = []
+    if ctx.attr.outs:
+        output_files.extend([ctx.actions.declare_file(f) for f in ctx.attr.outs])
```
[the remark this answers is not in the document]
acozzette: Added an appropriate `sibling` argument to `declare_file()` to fix this.

js/private/rpc.bzl
```diff
@@ -0,0 +1,154 @@
...
+        output_files.extend(proto_common.declare_generated_files(ctx.actions, proto_info, js_proto_toolchain_info.out_js_extension))
+    if js_proto_toolchain_info.out_dts_extension:
+        output_files.extend(proto_common.declare_generated_files(ctx.actions, proto_info, js_proto_toolchain_info.out_dts_extension))
+
+    proto_common.compile(
+        actions = ctx.actions,
+        proto_lang_toolchain_info = proto_lang_toolchain_info,
+        protoc_info = protoc_info,
+        output_files = output_files,
+        proto_outdir = proto_common.output_directory(proto_info, output_files[0].root),
```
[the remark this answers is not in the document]
acozzette: I added logic to explicitly fail on cross-package deps to address this issue.

acozzette:
@codex review

acozzette:
@jbedard Codex made some good points about edge cases, and I added some fixes to address those. I think this should now be good to go unless Codex finds something else (I kicked it off one more time just to be safe).

js/private/rpc.bzl
```diff
@@ -0,0 +1,157 @@
...
+    )
+
+    js_outputs = [f for f in output_files if not f.path.endswith(".d.ts")]
+    dts_outputs = [f for f in output_files if f.path.endswith(".d.ts")]
+
+    return [
+        js_info(
+            target = ctx.label,
+            sources = depset(js_outputs),
+            types = depset(dts_outputs),
```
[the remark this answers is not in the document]
jbedard: Unless proto rules are special this might be a good idea? Normally `.js` (and `.js.map`? double check that...) are provided in `DefaultInfo` while things like dts and deps must be fetched from `JsInfo`.

Runfiles should probably be empty, as of v3 runfiles is "things not in `JsInfo`"...

jbedard:
Did any of the fixes have appropriate tests that could be added?

js/private/rpc.bzl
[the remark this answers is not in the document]
acozzette: That sounds good, I updated it to return a `DefaultInfo()` with the .js output files.

jbedard:
Ready to merge then?

acozzette:
> Did any of the fixes have appropriate tests that could be added?

I added an analysis test verifying that you get an error if you try to have a `js_rpc_library` depend on a `proto_library` in a different package.

The other fix only applies to the case where you try to use `js_rpc_library` on a `proto_library` that sets `import_prefix` or `strip_import_prefix`. I kind of hesitate to add a test for this, though, because I hope no one ever does this and I wouldn't want people to see an example of that and imitate it. I can add that test if you disagree, though.

acozzette:
> Ready to merge then?

I think it's ready to merge as long as it looks OK to you.

jbedard:
Worth adding release notes? Just 1-2 sentences is fine...

acozzette:
Sure, I added a couple sentences of suggested release notes to the PR description.

PR #2802, opened 2026-04-14
feat: add support for pnpm 11

### LINKED ISSUE #2798: [FR]: Support PNPM 11

Right now, when using PNPM 10, everything works well. However, when PNPM 11 is released (currently RC.0), it will break usage with this library.

This is because lockfiles are seemingly being updated to dual-document YAML files, which YQ doesn't seem to understand:

Image

This is made worse by this issue: https://github.com/pnpm/action-setup/issues/225

Where every CI using setup v6 will use PNPM 11 regardless of what's configured.

### Describe the feature

Support these new `pnpm-lock.yaml` files of PNPM 11

acozzette:
pnpm 11 introduced several breaking changes to the lockfile format that this commit addresses:
1. patchedDependencies path moved to pnpm-workspace.yaml: pnpm 11 no longer includes the patch file path in pnpm-lock.yaml (only the hash). Add _normalize_patched_dependencies() to look up the path from pnpm-workspace.yaml and restore the dict form expected by the rest of the code.
2. Multi-document lockfile format: in some situations, such as when configDependencies are present, pnpm 11 prepends an "env lockfile" YAML document to pnpm-lock.yaml.  Update the yq invocation to always return the last document regardless of whether the file has one or two YAML documents.
3. Changed peer dependency path encoding in snapshot keys: pnpm 11 encodes workspace-relative peer dep paths with a leading '+' (e.g.  `@scoped/b@+projects+b` instead of `@scoped/b@projects+b`), which changes the generated virtual store target names.

Also adds v110 test fixtures (lockfile, workspace config, snapshots) and reorganizes the e2e/pnpm_lockfiles setup so each version (v90, v101, v110) is self-contained. v90 needs its own package.json and pnpm-workspace.yaml, because pnpm 9 requires some things to be in package.json that are expected in pnpm-workspace.yaml in later versions.

Fixes #2798.

---

### Test plan

- New test cases added


docs/pnpm.md
```diff
@@ -271,9 +271,9 @@ npm packages have "lifecycle scripts" such as `postinstall` which are documented
 
 We refer to these as "lifecycle hooks".
 
-The lifecycle hooks of a package are determined by the `package.json` [`pnpm.onlyBuiltDependencies` attribute](https://pnpm.io/package_json#pnpmonlybuiltdependencies).
+The lifecycle hooks of a package are determined by the `package.json` [`pnpm.onlyBuiltDependencies` attribute](https://pnpm.io/package_json#pnpmonlybuiltdependencies), and by [allowBuilds](https://pnpm.io/next/settings#allowbuilds) for pnpm 10.26 and newer.
 
-If `pnpm.onlyBuiltDependencies` is unspecified `npm_translate_lock` will fallback to the legacy pnpm lockfile `requiresBuild` attribute.
+If `onlyBuiltDependencies` and `allowBuilds` are unspecified, `npm_translate_lock` will fallback to the legacy pnpm lockfile `requiresBuild` attribute.
```
jbedard: rules_js v3 no longer supports <v9 so I think this full paragraph should just be deleted?

e2e/pnpm_lockfiles/base/package.json
```diff
@@ -1,78 +0,0 @@
-{
```
jbedard, line 1: This file was intentionally shared across tests - what is the new setup?

npm/private/npm_translate_lock_state.bzl
```diff
@@ -392,6 +417,11 @@ def _yaml_to_json(rctx, yaml_path, is_windows):
     host_yq = Label("@yq_{}//:yq{}".format(repo_utils.platform(rctx), ".exe" if is_windows else ""))
     yq_args = [
         rctx.path(host_yq),
+        "eval-all",
+        # Do a cross-document reduce that simply returns the last document.
```
jbedard, line 399: Do we have anything testing this? Lets add some simple ones to `e2e/pnpm_lockfiles/cases`?

npm/private/npm_translate_lock_state.bzl
```diff
@@ -48,6 +48,10 @@ WARNING: `update_pnpm_lock` attribute in `npm_translate_lock(name = "{rctx_name}
     _init_root_package(priv)
     _init_workspace(priv, rctx, is_windows)
 
+    # pnpm 11 removed patch paths from pnpm-lock.yaml, so let's put them back
+    # in and restore the old format.
+    _normalize_patched_dependencies(priv)
```
jbedard: Can we move this into `_load_lockfile` and make it part of the lockfile parse/loading instead of a modification afterwards?

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -645,10 +645,13 @@ def _normalize_bazelignore(lines):
 def _verify_lifecycle_hooks_specified(state):
     if state.only_built_dependencies() == None:
         msg = """\
-ERROR: pnpm 'onlyBuiltDependencies' configuration required.
+ERROR: pnpm 'onlyBuiltDependencies' (or 'allowBuilds' in pnpm v10.26+) configuration required.
```
jbedard: Lets rephrase it to emphasize+encourage the modern one? Something such as:

> ERROR: pnpm 'allowBuilds' (or 'onlyBuiltDependencies' in pnpm <10.26) configuration required.

e2e/pnpm_lockfiles/lockfile-test.bzl
```diff
@@ -145,7 +145,7 @@ def lockfile_test(npm_link_all_packages, name = None):
             ":node_modules/scoped/bad",
             ":.aspect_rules_js/node_modules/@scoped+a@0.0.0",
             ":.aspect_rules_js/node_modules/@scoped+b@0.0.0",
-            ":.aspect_rules_js/node_modules/@scoped+c@file+..+projects+c_@scoped+b@projects+b",
+            ":.aspect_rules_js/node_modules/@scoped+c@file+..+projects+c_@scoped+b@%sprojects+b" % ("+" if lock_version == "v110" else ""),
```
jbedard, line 151: Did you discover why this changed in the lockfile?

docs/pnpm.md
acozzette: Ok, done.

e2e/pnpm_lockfiles/base/package.json
acozzette: I assumed that we needed to stop sharing it since parts of it are moving to pnpm-workspace.yaml in v10 and v11, but after trying just now I see I was wrong about that. It seems that we can keep sharing package.json and pnpm-workspace.yaml since pnpm doesn't mind if there is duplication between them. I went back to one common copy in base/ that each test symlinks to.

npm/private/npm_translate_lock_state.bzl
acozzette: `e2e/pnpm_lockfiles/v110` exercises the multi-document case, but I went ahead and added a smaller test case in `cases/` too.

npm/private/npm_translate_lock_helpers.bzl
acozzette: Good idea, done.

docs/pnpm.md
```diff
@@ -271,10 +271,7 @@ npm packages have "lifecycle scripts" such as `postinstall` which are documented
 
 We refer to these as "lifecycle hooks".
 
-The lifecycle hooks of a package are determined by the `package.json` [`pnpm.onlyBuiltDependencies` attribute](https://pnpm.io/package_json#pnpmonlybuiltdependencies).
-
-If `pnpm.onlyBuiltDependencies` is unspecified `npm_translate_lock` will fallback to the legacy pnpm lockfile `requiresBuild` attribute.
-This attribute is only available in pnpm _before_ v9, see [pnpm #7707](https://github.com/pnpm/pnpm/issues/7707) for reasons why this attribute was removed.
+The lifecycle hooks of a package are determined by the `package.json` [`pnpm.onlyBuiltDependencies` attribute](https://pnpm.io/package_json#pnpmonlybuiltdependencies), and by [allowBuilds](https://pnpm.io/next/settings#allowbuilds) for pnpm 10.26 and newer.
```
jbedard: Specify that `allowBuilds`s is in the workspace and not `package.json`?

Put `allowBuilds` first in the sentence? Maybe say "or `pnpm.onlyBuildDependencies` in package for older versions of pnpm (<10.26)" or something like that?

e2e/pnpm_lockfiles/cases/multi-document-v11/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
...
+multi_document_v11_link_all()
+
+# Verify that rules_js can parse a pnpm 11 multi-document lockfile (where
+# configDependencies cause an env lockfile to be prepended as a second YAML
+# document) and correctly resolves the main lockfile from the last document.
+build_test(
+    name = "multi_document_v11",
+    targets = [
+        ":node_modules",
+        ":node_modules/semver",
```
jbedard: In the lockfile `semver` is in both sections/subdocuments/whatever-they-are-called - how can we make this test assert we are loading the correct one?

npm/private/npm_translate_lock_state.bzl
acozzette: Sure, I went ahead and did that. I had to reorder things a bit so that we parse `pnpm-workspace.yaml` before the lockfile, but luckily that seemed to work without any problems.

docs/pnpm.md
acozzette: I reworded it to put `allowBuilds` first and de-emphasize `onlyBuiltDependencies`; let me know if this sounds OK.

npm/private/npm_translate_lock_state.bzl
```diff
@@ -447,6 +477,15 @@ def _has_workspaces(priv):
 def _should_update_pnpm_lock(priv):
     return priv["should_update_pnpm_lock"] and priv["src_root"] != None
 
+################################################################################
+def _only_built_dependencies(pnpm_settings):
+    # pnpm 11 replaced onlyBuiltDependencies (a list) with allowBuilds (a dict
+    # mapping package names to booleans). Support both.
+    allow_builds = pnpm_settings.get("allowBuilds", None)
+    if allow_builds != None:
+        return [name for name, allowed in allow_builds.items() if allowed]
```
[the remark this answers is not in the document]
jbedard: ☝️ does this make `allowBuilds` different then `onlyAllowWhatever`?

e2e/pnpm_lockfiles/cases/multi-document-v11/BUILD.bazel
acozzette: Yeah, good point. The existence of `:node_modules/semver` doesn't prove anything about which document we read, since it exists in both. I added a regular non-config dependency and updated this to check for that dep instead.

npm/private/npm_translate_lock_state.bzl
```diff
@@ -423,6 +427,24 @@ def _load_lockfile(priv, rctx, attr, pnpm_lock_path, is_windows):
 
     priv["importers"] = importers
     priv["packages"] = packages
+
+    # pnpm 11 changed the patchedDependencies format in pnpm-lock.yaml: values are now just
+    # a hash string rather than a dict with 'path' and 'hash' keys. The patch file path moved
+    # to pnpm-workspace.yaml. Normalize to always use dict form with 'path' and 'hash' keys.
+    if any([type(v) == "string" for v in pnpm_patched_dependencies.values()]):
```
jbedard: This still seems ugly imo. How can we clean this up? Is it any better if `parse_pnpm_lock_json` does it? I forget how much "normalization" `parse_pnpm_lock_json` does vs doing it after so scope it out before you do anything...

npm/private/npm_translate_lock_state.bzl
```diff
@@ -489,7 +520,7 @@ def _new(rctx, mod, attr):
         importers = lambda: priv["importers"],
         packages = lambda: priv["packages"],
         pnpm_patched_dependencies = lambda: priv["pnpm_patched_dependencies"],
-        only_built_dependencies = lambda: priv["pnpm_settings"].get("onlyBuiltDependencies", None),
+        only_built_dependencies = lambda: _only_built_dependencies(priv["pnpm_settings"]),
```
jbedard: Why not normalize the data once as we parse+load the files instead of everytime this getter is invoked?

e2e/pnpm_lockfiles/cases/multi-document-v11/BUILD.bazel
jbedard: What makes `ms` "non-config"? I don't understand this new stuff... maybe you can explain or have a link to what this is all about?

npm/private/npm_translate_lock_state.bzl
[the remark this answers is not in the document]
acozzette: Actually I just realized that `onlyBuiltDependencies` similarly supports a version specifier as of pnpm 10.19: https://pnpm.io/blog/releases/10.19#versions-in-onlybuiltdependencies I could not find any existing test coverage of this in rules_js, so I'm guessing we probably don't support it yet.

e2e/pnpm_lockfiles/cases/multi-document-v11/BUILD.bazel
acozzette: Config dependencies are just listed in the `configDependencies` field in `pnpm-workspace.yaml`: https://pnpm.io/config-dependencies I had not heard of these until today, but my understanding is that they're essentially special dependencies that are installed before all other deps and can do fancy things like dynamically change pnpm settings.

I'm guessing it would be a fair bit of work to support them properly, so I'm not trying to do that right now, but just using them as a way to trigger the multi-document lockfile behavior.

npm/private/npm_translate_lock_state.bzl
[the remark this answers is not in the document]
jbedard: Ok, if we don't support it maybe you can find or log an issue and we don't have to do anything here for now...

e2e/pnpm_lockfiles/cases/multi-document-v11/BUILD.bazel
jbedard: Oh I see, they are in the `pnpm-workspace.yaml` and not a regular `package.json` file. That's odd, and I do not want to even think about supporting them for now...

npm/private/npm_translate_lock_state.bzl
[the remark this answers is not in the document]
acozzette: Sounds good, I filed an issue for this [here](https://github.com/aspect-build/rules_js/issues/2803).

npm/private/npm_translate_lock_state.bzl
acozzette: Good idea, done.

npm/private/npm_translate_lock_state.bzl
acozzette: I kind of think pushing it down to `parse_pnpm_lock_json` would be potentially even worse, since that function doesn't know about the workspace file at all.

I tried going in a different direction and just removed the normalization entirely, in favor of having the accessors hide the internal representation and just expose a more friendly API. Let me know what you think and if you don't like this approach then I can try something else.

e2e/pnpm_lockfiles/lockfile-test.bzl
acozzette: No, I put this in as a temporary workaround but still need to figure out why this changed. I couldn't find anything in the pnpm release notes so I'll probably just have to ask the pnpm maintainers if this change was intentional.

acozzette: I went ahead and filed an issue with pnpm and included a simple example: https://github.com/pnpm/pnpm/issues/11272

e2e/pnpm_lockfiles/cases/multi-document-v11/package.json
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "config-deps-v11",
+    "version": "0.0.0",
+    "dependencies": {
+        "ms": "^2.1.3",
+        "semver": "^7.7.4"
```
jbedard: remove?

e2e/pnpm_lockfiles/v101/pnpm-lock.yaml
```diff
@@ -498,36 +498,42 @@ packages:
     engines: {node: '>= 10.0.0'}
     cpu: [arm]
     os: [linux]
+    libc: [glibc]
```
jbedard, line 501: Any idea why this changed? Did the .sh script doing `npx pnpm@~10` start to resolve a newer pnpm version that now adds this `libc` thing?

npm/private/npm_translate_lock_state.bzl
```diff
@@ -447,6 +455,33 @@ def _has_workspaces(priv):
...
+        return None
+    if type(patch_info) == "string":
+        patch_path = priv["pnpm_settings"].get("patchedDependencies", {}).get(name)
+        if not patch_path:
+            fail("ERROR: patchedDependencies entry '{}' in pnpm-lock.yaml has no corresponding path in pnpm-workspace.yaml".format(name))
+        return patch_path
+    return patch_info["path"]
+
+################################################################################
+def _only_built_dependencies(pnpm_settings):
```
jbedard: What is "pnpm settings"?

Why does this method accept `pnpm_settings` while the previous method accepts `priv` and then inspects `priv["pnpm_settings"]`?

Is there a more consistent way we can do this?

npm/private/npm_translate_lock_state.bzl
```diff
@@ -447,6 +455,33 @@ def _has_workspaces(priv):
...
+################################################################################
+def _patch_path_for(priv, name):
+    """Returns the patch file path for a patchedDependency entry by name, or None if not found.
+
+    Handles both lockfile formats:
+    - pnpm <= 10: pnpm-lock.yaml stores a dict with 'path' and 'hash' keys
+    - pnpm 11+: pnpm-lock.yaml stores only a hash string; the path is in pnpm-workspace.yaml
+    """
+    patch_info = priv["pnpm_patched_dependencies"].get(name)
+    if patch_info == None:
```
jbedard, line 467: When we drop support for pnpm 10, how will this function change? Can we make it so that change is only deleting code?

e2e/pnpm_lockfiles/v101/pnpm-lock.yaml
acozzette: Yes, I think this must have been a change in the pnpm 10.x series somewhere. I tried going back to 10.19 and found that the `libc` diffs disappeared, so whatever it was must have been made somewhere between 10.19 and 10.33.

npm/private/npm_translate_lock_state.bzl
acozzette: Sure, I tweaked the logic so that the pnpm<=10 code is just an if block that we can delete whenever we're ready to drop support for pnpm 10.

e2e/pnpm_lockfiles/cases/multi-document-v11/package.json
acozzette: Done.

npm/private/npm_translate_lock_state.bzl
acozzette: pnpm_settings is just `priv["pnpm_settings"]`, which comes from the `pnpm` field in package.json and/or the contents of `pnpm-workspace.yaml`. I went ahead and tweaked this function to take `priv` as a parameter instead so that it's consistent with the other helper functions around it.

e2e/pnpm_lockfiles/base/pnpm-workspace.yaml
```diff
@@ -2,3 +2,26 @@ packages:
...
+
+# The presence of configDependencies causes pnpm-lock.yaml to be a
+# multi-document YAML file with pnpm 11.
+configDependencies:
+    semver: 7.7.4+sha512-vFKC2IEtQnVhpT78h1Yp8wzwrf8CM+MzKMHGJZfBtzhZNycRFnXsHk6E5TxIkkMsgNS7mdX3AGB7x2QM2di4lA==
+
+patchedDependencies:
+    meaning-of-life@1.0.0: patches/meaning-of-life@1.0.0-pnpm.patch
+
+allowBuilds:
```
jbedard, line 18: Does base/package.json still duplicate this info as well as the patches + overrides? Maybe just add a comment in this file saying to ensure the info is duplicated and a TODO to drop it from package.json once pnpm9 is dropped?

jbedard:
@codex

e2e/pnpm_lockfiles/v101/pnpm-lock.yaml
jbedard: I guess we can ignore it then 🤷

e2e/pnpm_lockfiles/base/pnpm-workspace.yaml
acozzette: Yes, allowBuilds/onlyBuiltDependencies, patchedDependencies, and overrides are all duplicated in package.json. I added a comment here about this and noted that we should remove the duplication once we drop pnpm 9.

npm/private/npm_translate_lock_state.bzl
```diff
@@ -447,6 +455,38 @@ def _has_workspaces(priv):
...
+    return patch_path
+
+################################################################################
+def _only_built_dependencies(priv):
+    # pnpm 11 replaced onlyBuiltDependencies (a list) with allowBuilds (a dict
+    # mapping package names to booleans). Support both.
+    pnpm_settings = priv["pnpm_settings"]
+    allow_builds = pnpm_settings.get("allowBuilds", None)
+    if allow_builds != None:
+        return [name for name, allowed in allow_builds.items() if allowed]
```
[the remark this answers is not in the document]
acozzette: I filed an issue for this yesterday: https://github.com/aspect-build/rules_js/issues/2803 I think we should probably support this eventually, but it's not a new problem since we already lack support for it with onlyBuiltDependencies.

e2e/pnpm_lockfiles/setup.sh
```diff
@@ -2,9 +2,13 @@
 
 #   9.0 - pnpm v9.0.0 bumped the lockfile version to 9.0; this includes breaking changes regarding lifecycle hooks and patches
 #  10.0 - pnpm v10 kept the lockfile version at 9.0, but has minor differences such as length of hashes, yaml key order etc.
+#  11.0 - pnpm v11 kept the lockfile version at 9.0 and made several minor changes
 
 # pnpm v9.0.0 bumped the lockfile version to 9.0
-pushd base && npx -y pnpm@~9 install --lockfile-only && mv pnpm-lock.yaml ../v90 && popd
+pushd v90 && npx -y pnpm@~9 install --lockfile-only && popd
```
jbedard, line 8: Does this setup.sh still work despite having pnpm11 fields in the workspace file?

jbedard (approved):

e2e/pnpm_lockfiles/setup.sh
acozzette: Yeah, surprisingly it does still work. I believe none of the new fields are brand new in pnpm 11--they were added in the 10.x series and so that may partially explain it. And I guess pnpm 9 must be just ignoring fields it doesn't know about.

jbedard: Ok, that SGTM then 👍

jbedard:
All good to land?

acozzette:
Yes, I believe so! Feel free to just squash it, no need to keep all the later small commits.

PR #2804, opened 2026-04-15
Demonstration of odd behavior with deps

tyler-breisacher-zipline:
At https://github.com/tyler-breisacher-zipline/rules_js/commit/4e59a57d9c7e79caa6858016f7c67a1f5bc5130b (first commit in this branch) I added @sentry/nextjs as a dependency. The bazel target for that package does not depend on esbuild: `bazel query "deps(//examples/nextjs:node_modules/@sentry/nextjs)" | grep esbuild` returns no results.

But then in https://github.com/tyler-breisacher-zipline/rules_js/commit/98c514552b746b3b8ad5c10a83cf60f23fa992bd (second commit) I also added storybook, and then the same bazel query shows that we do depend on esbuild.


jbedard:
Can you create an isolated pnpm-lock that reproduces this with the absolute minimum deps? This seems like a pnpm issue (adding one package causes the deps of another package to change) really, but seeing it in isolation will make that easier to under.

tyler-breisacher-zipline:
Here's [what I have so far](https://github.com/tyler-breisacher-zipline/rules_js_deps_investigation/commit/a256c471b46e4a52ece958334097856554942c9c) but to understand what's happening you kind of have to trace through a bunch of packages (storybook, webpack, terser-webpack-plugin...) so I'm going to try to simplify it more. The crux of it seems to be that terser-webpack-plugin has an optional dep on esbuild, but pnpm decides to create two different snapshots of it, one without esbuild, and one with (lines 3488 and 3498 of the lock file)

jbedard:
That seems to confirm that this is mainly a npm/pnpm usability issue? When you reach this level of peer+optional deps it can get quite complicated, but I think rules_js is just respecting the lockfile?

tyler-breisacher-zipline:
I agree, I don't think this is a bug in rules_js, and maybe not really a bug in pnpm, but it's a surprising behavior in both that I would like to understand better, especially if it turns out to be the cause of some of our ts_project() targets being slow.

jbedard:
It might be worth putting into `troubleshooting.md` or something like that, if we can provide a simple example and solution of the problem(s). A section titled something such as "slow builds due to unnecessary (transitive) npm deps" with an explanation and possible workarounds...?

Issue #2754, opened 2026-03-09
[Bug]: js_run_binary is not configured for the execution platform when the use_execroot_entrypoint = True and the target platform != exec platform


nickschaap:
### What happened?

In rules_js v3 support was added for 3p packages which specify optional dependencies that are included based on platform. For example, esbuild specifies various optional dependencies which included pre-compiled binaries for a variety of supported platforms. `js_run_binary` uses `run_binary` from `bazel_lib` and specifically configures its tool label under the `exec` configuration. However, if you are using `use_execroot_entrypoint = True` (the default). Only target platform configured sources will be available. This causes issues when you are building a target that transitions to a target that does not match the host.

Ideally the default configuration for `js_run_binary` would support this use case.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:

```
Bazelisk version: 1.27.0
Aspect CLI version: 2026.7.4-75a8f73d
INFO: Invocation ID: 419c9392-c1cd-417b-8d0a-80a6f2e244e1
Build label: 7.7.1
Build target: @@//src/main/java/com/google/devtools/build/lib/bazel:BazelServer
Build time: Wed Nov 12 17:35:45 2025 (1762968945)
Build timestamp: 1762968945
Build timestamp as int: 1762968945
```

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

aspect_rules_js v3.0.2
rules_nodejs v6.7.3

Language(s) and/or frameworks involved:

### How to reproduce

```shell

```

sallustfire:
This is a pre-existing bug #2121, but v3 now makes it breaking because of logic in `npm/private/npm_import.bzl` that generates platform constraints for npm package deps like:

```starlark
deps = {} | select({
    "@aspect_rules_js//platforms/pnpm:linux_x64": {"@esbuild/linux-x64": "..."},
    "@aspect_rules_js//platforms/pnpm:darwin_arm64": {"@esbuild/darwin-arm64": "..."},
    "//conditions:default": {},
})
```

acozzette:
[...]

Issue #2121, opened 2025-03-10
[Bug]: js_run_binary incorrectly transitions hoisted tools runfiles


sallustfire:
### What happened?

The immediate symptom is that upgrading to Bazel 8.x and generating sources via `js_run_binary` consumed by a Windows binary causes the following build error:
```
external/rules_cc+/cc/BUILD:103:19: in cc_toolchain_alias rule @@rules_cc+//cc:current_cc_toolchain: 
Traceback (most recent call last):
	File "/virtual_builtins_bzl/common/cc/cc_toolchain_alias.bzl", line 26, column 48, in _impl
	File "/virtual_builtins_bzl/common/cc/cc_helper.bzl", line 163, column 13, in _find_cpp_toolchain
Error in fail: Unable to find a CC toolchain using toolchain resolution. Target: @@rules_cc+//cc:current_cc_toolchain, Platform: @@platforms//host:host, Exec platform: @@bazel_tools//tools:host_platform
```

The cause of this is that `js_run_binary` by default sets `use_execroot_entry_point` which hoists runfiles of the js_binary to execroot to better mimic how JS tools are run. However, `run_binary` forces a `cfg = "exec"` transition on `tool`. But, the same is not done for these runfile srcs which I believe by default use the target platform. 

Bazel 8 is forcing CC toolchain resolution for reasons unknown to me, which may suggest another bug about why a CC toolchain is required at all. 

### Version

Development (host) and target OS/architectures: Linux, targeting Windows x86_64

Output of `bazel --version`: 8.1.1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

Language(s) and/or frameworks involved: js_run_binary

### How to reproduce

```shell
Quick demo of how to reproduce this:

https://github.com/sallustfire/rules_js/tree/js-run-binary-platform-runfiles
```

### Any other information?

This is loosely related: https://github.com/aspect-build/rules_js/issues/2073

The most naive change here would be to force a transition on the hoisted runfiles.

acozzette:
I wrote up some documentation [here](https://github.com/aspect-build/rules_js/blob/main/docs/use_execroot_entry_point.md) on how to avoid this problem by setting `use_execroot_entry_point = False`. As of v3.2.1, you can also pass the flag `--@aspect_rules_js//js:use_execroot_entry_point=False` to make that option False by default (but still possible to override on specific targets). We may eventually flip the default, but that will have to wait for a major version bump.

Issue #2073, opened 2025-01-16
[Bug]: `js_run_binary` tool runfiles get built twice


jfirebaugh:
### What happened?

`js_run_binary` sets `use_execroot_entry_point = True` by default, and hoists runfiles of the tool to srcs to that they are included as execroot inputs to the action, while also preserving the runfiles input. The downside of this is that any actions required to build those runfiles will get executed twice: once under the default `fastbuild` configuration, for the execroot input, and once under a transition, for the runfiles input. For example, if there are npm packages in the runfiles dependencies, then their `NpmPackageExtract` (and `NpmLifecycleHook` if present) actions run twice.

I was able to work around this in a specific case where it had a significant performance impact by avoiding the use of data dependencies, but it would be nice if there was a generic solution.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: bazel 7.4.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: rules_js 2.1.2

Language(s) and/or frameworks involved: JS

### How to reproduce

```shell
In rules_js, run `bazel clean && bazel build --execution_log_compact_file=compact_exec.log //examples/js_binary:run2 && ../bazel/bazel-bin/src/tools/execlog/parser --log_path compact_exec.log`.

acorn/CHANGELOG.md output from first `NpmPackageExtract`, in `darwin_arm64-fastbuild`:

actual_outputs {
  path: "bazel-out/darwin_arm64-fastbuild/bin/node_modules/.aspect_rules_js/acorn@8.4.0/node_modules/acorn/CHANGELOG.md"
  digest {
    hash: "6767297d491d22ed8897c4df234d0f83ee0258091066ce344102effdba304c9f"
    size_bytes: 17478
    hash_function_name: "SHA-256"
  }
}


acorn/CHANGELOG.md output from second `NpmPackageExtract`, in `darwin_arm64-opt-exec-ST-d57f47055a04`:

actual_outputs {
  path: "bazel-out/darwin_arm64-opt-exec-ST-d57f47055a04/bin/node_modules/.aspect_rules_js/acorn@8.4.0/node_modules/acorn/CHANGELOG.md"
  digest {
    hash: "6767297d491d22ed8897c4df234d0f83ee0258091066ce344102effdba304c9f"
    size_bytes: 17478
    hash_function_name: "SHA-256"
  }
}


First acorn/CHANGELOG.md input:

inputs {
[...]

PR #2809, opened 2026-04-22
fix: allow js_run_binary to hoist runfiles in the exec configuration

acozzette:
The `js_run_binary` macro has a `use_execroot_entry_point` option which is enabled by default and which hoists out the runfiles into the execroot.  Currently this hoisting effectively rebuilds the runfiles in the target platform config, but they should really be in the exec config since they consist of code that is going to run during the build action.

This change fixes that problem by making sure we hoist the runfiles in the exec config. I also updated the launcher script, since it needs to be prepared to look for the entry point in a different directory. This is all guarded by the `hoist_runfiles_to_exec_cfg` option on `js_run_binary`, which effectively defaults to False for now. The default is actually `None` but only for technical reasons.

The default is controlled by a new flag `//js:_hoist_runfiles_to_exec_cfg`, but this is intended only for our own CI purposes, allowing us to run most tests against both behaviors. There are a handful of `js_run_binary` targets that require one behavior or the other, so I updated those ones to explicitly specify the flag they need.

This addresses #2754, #2121, and #2073, but does not automatically fix them yet since the fix is opt-in for now.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

---

The `js_run_binary` macro now supports a `hoist_runfiles_to_exec_cfg` option, which will cause runfiles to be hoisted in the exec configuration instead of the target configuration. This is currently disabled by default, but you can opt in to achieve a more correct build setup.

### Test plan

- New test cases added


.aspect/workflows/config.yaml
```diff
@@ -154,7 +162,7 @@ tasks:
           name: Bazel 7
           id: bazel-7
           bazel:
-              flags: ['--test_tag_filters=-skip-on-bazel7']
+              flags: ['--test_tag_filters=-skip-on-bazel7', '--@aspect_rules_js//js:_hoist_runfiles_to_exec_cfg=true']
```
acozzette: For now I just updated our Bazel 7 tests here to all opt into the new behavior by default. I was thinking this would provide some good coverage without adding a ton of new test runs, but I'm not sure if this is the best approach.

js/private/js_run_binary.bzl
```diff
@@ -380,14 +393,29 @@ See https://github.com/aspect-build/rules_js/tree/main/docs#using-binaries-publi
...
+                "//conditions:default": [runfiles_label],
+            })
 
     if allow_execroot_entry_point_with_no_copy_data_to_bin:
         fixed_env["JS_BINARY__ALLOW_EXECROOT_ENTRY_POINT_WITH_NO_COPY_DATA_TO_BIN"] = "1"
 
     _run_binary(
         name = name,
         tool = tool,
+        tools = tools,
```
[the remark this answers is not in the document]
acozzette: This is technically true, but I think we can safely disregard this possibility given that no released `run_binary` supported the `tools` parameter until yesterday. It seems fair to say that there has not been enough time for anyone to start relying on `tools` being successfully passed through.

js/private/js_run_binary.bzl
```diff
@@ -149,6 +150,17 @@ def js_run_binary(
...
             setting `allow_execroot_entry_point_with_no_copy_data_to_bin` to True.
 
+        hoist_runfiles_to_exec_cfg: Hoist runfiles to the execution platform configuration instead of the target platform
+            configuration.
+
+            This only has an effect when `use_execroot_entry_point` is True. Setting `hoist_runfiles_to_exec_cfg` to True
+            results in a more correct build by ensuring that code that is executed during a build action is built for the
+            execution platform, which may not be the same as the target platform.
+
+            When set to `None` (the default), the value of the internal `//js:_hoist_runfiles_to_exec_cfg` build flag
```
jbedard: Should we just remove this paragraph if it is internal?

js/private/js_run_binary.bzl
```diff
@@ -31,6 +31,7 @@ def js_run_binary(
         exit_code_out = None,
         silent_on_success = True,
         use_execroot_entry_point = True,
+        hoist_runfiles_to_exec_cfg = None,
```
jbedard: Why use the term "hoist"? Do we use that anywhere else in the public API/docs?

WDYT about `use_execroot_exec_cfg` or something like that? That implies it is related to `user_execroot_*`?

js/private/test/data/BUILD.bazel
```diff
@@ -117,12 +117,12 @@ js_test(
...
     name = "run-write-srcs",
+    srcs = ["data.json"],
     outs = ["rb0.json"],
     args = [
-        "rb0.json",
-        "data.json",
+        "$(rootpath rb0.json)",
+        "$(rootpath data.json)",
     ],
-    chdir = package_name(),
```
jbedard, line 125: Why this change?

.aspect/workflows/config.yaml
```diff
@@ -123,7 +123,15 @@ workspaces:
...
         icon: js
-        tasks: *e2e_tasks
+        tasks:
+            - test:
+                  queue: aspect-medium
+            - format:
+                  without: true
+            - buildifier:
+                  without: true
+            - bazel-7:
```
jbedard: This broke on bazel 7?

js/private/js_run_binary.bzl
jbedard: Or we name+document it better and make it public so users can easily opt-in to what we hope will be the default in the future...

.aspect/workflows/config.yaml
acozzette: I went ahead and added a separate test category for this, so we now have Bazel 7, 8, 9, and 9 + `use_execroot_exec_cfg`.

js/private/js_run_binary.bzl
acozzette: As far as I can tell we don't use it anywhere except in the description of `use_execroot_entry_point`. There are some references in the docs to hoisting related to PNPM, but I think that's a separate concept. `use_execroot_exec_cfg` sounds good to me, so I went ahead and switched to that name.

js/private/test/data/BUILD.bazel
acozzette: Good question, I think a bunch of these changes were necessary with an earlier version of this PR, but not anymore. I reverted all the changes to this file except for the ones adding dependencies to `srcs`, which is still necessary.

.aspect/workflows/config.yaml
acozzette: No, it was fine with Bazel 7, but I added the new flag to the Bazel 7 tests and it was the flag that caused the breakage. `e2e/repo_mapping` assigns an unusual name to `aspect_rules_js`, and so passing any flag beginning with `@aspect_rules_js//` doesn't work.

MODULE.bazel
```diff
@@ -19,7 +19,7 @@ bazel_dep(name = "rules_nodejs", version = "6.7.3")
 
 # Changes ensured by rules_js:
 # 3.2.2: https://github.com/bazel-contrib/bazel-lib/commit/cac2d7855949d1b222fa26888892fbbe1d31015d
```
jbedard, line 21: Add a comment just like this for the reason we need 3.3.0?

js/private/test/data/BUILD.bazel
```diff
@@ -117,6 +117,7 @@ js_test(
 # TEST: js_test(data = js_run_binary(srcs)) ---------------
 js_run_binary(
     name = "run-write-srcs",
+    srcs = ["data.json"],
```
jbedard, line 120: This was essentially an undeclared input before, but magically worked because the `tool` also declared the input? Now it correctly fails with the new option enabled?

js/private/js_binary.sh.tpl
```diff
@@ -97,10 +97,18 @@ function logf_debug {
 
 function resolve_execroot_bin_path {
     local short_path="$1"
+    local suffix
     if [[ "$short_path" == ../* ]]; then
-        echo "$JS_BINARY__EXECROOT/${BAZEL_BINDIR:-$JS_BINARY__BINDIR}/external/${short_path:3}"
+        suffix="external/${short_path:3}"
```
jbedard, line 102: Can we just modify `short_parth` instead of introducing the extra `suffix` var? Any disadvantage to that?

js/BUILD.bazel
```diff
@@ -1,6 +1,20 @@
 "Public API"
 
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+
+# This flag controls the default value of the use_execroot_exec_cfg flag on
+# js_run_binary.
+bool_flag(
+    name = "use_execroot_exec_cfg",
```
jbedard, line 9: WDYT about prefixing this with `js_run_binary_` since that is the one and only usage?

js/private/test/data/BUILD.bazel
acozzette: Yes, that's right. With the new option enabled and without this line adding the dep via `srcs`, this would fail since data.json would no longer be in the same place.

jbedard: That sounds correct to me 👍

PR #2812, opened 2026-04-24
fix: support for pnpm 11.0.0-rc.5

acozzette:
When we initially added support for 11.0.0-rc.0, we had to put in a workaround in lockfile-test.bzl due to the way peer suffixes changed unexpectedly: https://github.com/pnpm/pnpm/issues/11272 This has now been fixed in more recent release candidates, so we can drop that workaround.

---

### Test plan


e2e/pnpm_lockfiles/base/pnpm-workspace.yaml
```diff
@@ -17,6 +17,7 @@ patchedDependencies:
 
 allowBuilds:
     '@aspect-test/c': true
+    'jquery': true
```
jbedard: Why add this?

acozzette: It's mysterious to me why this is needed now when it wasn't before, but I had to add it to get past errors from e2e/pnpm_lockfiles/setup.sh.

acozzette: I removed them and now it works again somehow.

acozzette:
The Bazel 9 test failures seem to be an unrelated regression Bazel 9.1.0, so I filed a bug about that [here](https://github.com/bazelbuild/bazel/issues/29393).

e2e/pnpm_lockfiles/base/pnpm-workspace.yaml
jbedard: idk, but we want to discourage lifecycle hooks and there's no way jquery should need such a thing so I don't want it here...

jbedard:
If CI fails on main atm maybe we should change `9.x` on CI to `9.0.x` (or something like that, idk exact syntax)?

e2e/pnpm_lockfiles/base/pnpm-workspace.yaml
acozzette: Yeah, that makes sense. For whatever reason it's working now and so I removed jquery.

acozzette:
Unfortunately Bazelisk does not seem to support a syntax like `9.0.x` or `9.0.*`, so for now I just pinned the version to 9.0.2.

jbedard (approved):

PR #2813, opened 2026-04-25
refactor: simplify launcher script logic for `chdir`

acozzette:
The `js_binary()` and `js_run_binary()` rules provide a `chdir` option that you can use to indicate a directory to cd to before the main entry point is invoked. This turns into a `JS_BINARY__CHDIR` environment variable that the launcher script looks at to decide where to cd.

The logic has a special case for when `JS_BINARY__CHDIR` begins with `external/`, in which case we call `resolve_execroot_bin_path` to canonicalize the path. This special casing seems to be unnecessary, and `resolve_execroot_bin_path` also contains more logic than we need. I therefore replaced all this with a one-liner that cds to the correct absolute path.

The motivation for this is that in another PR I am going to change the behavior of `resolve_execroot_bin_path` so that it always refers to the exec platform bin directory, but this is not something we want to affect `chdir`.

---

### Test plan

### DIFF
--- a/js/private/js_binary.sh.tpl
+++ b/js/private/js_binary.sh.tpl
@@ -328,10 +328,7 @@ fi
 # Change directory to user specified package if set
 if [ "${JS_BINARY__CHDIR:-}" ]; then
     logf_debug "changing directory to user specified package %s" "$JS_BINARY__CHDIR"
-    case "$JS_BINARY__CHDIR" in
-    external/*) cd "$(resolve_execroot_bin_path "$JS_BINARY__CHDIR")" ;;
-    *) cd "$JS_BINARY__CHDIR" ;;
-    esac
+    cd "$JS_BINARY__EXECROOT/${BAZEL_BINDIR:-$JS_BINARY__BINDIR}/$JS_BINARY__CHDIR"
 fi
 
 # Gather node options
--- a/js/private/test/snapshots/launcher.sh
+++ b/js/private/test/snapshots/launcher.sh
@@ -448,10 +448,7 @@ fi
 # Change directory to user specified package if set
 if [ "${JS_BINARY__CHDIR:-}" ]; then
     logf_debug "changing directory to user specified package %s" "$JS_BINARY__CHDIR"
-    case "$JS_BINARY__CHDIR" in
-    external/*) cd "$(resolve_execroot_bin_path "$JS_BINARY__CHDIR")" ;;
-    *) cd "$JS_BINARY__CHDIR" ;;
-    esac
+    cd "$JS_BINARY__EXECROOT/${BAZEL_BINDIR:-$JS_BINARY__BINDIR}/$JS_BINARY__CHDIR"
 fi
 
 # Gather node options


js/private/js_binary.sh.tpl
```diff
@@ -328,10 +328,7 @@ fi
 # Change directory to user specified package if set
 if [ "${JS_BINARY__CHDIR:-}" ]; then
     logf_debug "changing directory to user specified package %s" "$JS_BINARY__CHDIR"
-    case "$JS_BINARY__CHDIR" in
-    external/*) cd "$(resolve_execroot_bin_path "$JS_BINARY__CHDIR")" ;;
-    *) cd "$JS_BINARY__CHDIR" ;;
-    esac
+    cd "$JS_BINARY__CHDIR"
```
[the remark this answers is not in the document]
jbedard: Is this right?

js/private/js_binary.sh.tpl
```diff
@@ -328,10 +328,7 @@ fi
 # Change directory to user specified package if set
 if [ "${JS_BINARY__CHDIR:-}" ]; then
     logf_debug "changing directory to user specified package %s" "$JS_BINARY__CHDIR"
-    case "$JS_BINARY__CHDIR" in
-    external/*) cd "$(resolve_execroot_bin_path "$JS_BINARY__CHDIR")" ;;
```
jbedard, line 332: I don't understand why this is no longer required. When could that have changed? When we dropped WORKSPACE support?

Does `--[no]legacy_external_runfiles` effect this which I think removes the `{target}.runfiles/_main/external` (previously `{target}.runfiles/{workspace_name}/external`

e2e/js_binary_workspace/BUILD.bazel
```diff
@@ -1,16 +1,33 @@
 load("@aspect_rules_js//js:defs.bzl", "js_run_binary")
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
 
 js_run_binary(
     name = "run_workspace_bin",
     stdout = "out.txt",
-    tool = "@workspace//:bin",
+    tool = "@workspace//pkg:bin",
```
jbedard: Was moving this into a subdir required for some reason?

acozzette: Yes, the problem is that when you have `chdir = package_name()` in the top-level directory, you end up with `chdir = ""`, so `chdir` is falsy and therefore this [block](https://github.com/aspect-build/rules_js/blob/bde0ec578139d5330f53dc8083fb1125ffd153eb/js/private/js_binary.bzl#L346) does not take effect. This is probably a bug, because even if when `chdir` is an empty string, it may still mean something when it's in a foreign repo.

js/private/js_binary.sh.tpl
acozzette: It looks to me like this logic was probably never needed. I see that it was added a few months ago in https://github.com/aspect-build/rules_js/pull/2344, but if I go back to that commit I find that the tests under `e2e/js_binary_workspace` all pass even without this logic.

acozzette: I don't think that `--legacy_external_runfiles` flag should have an effect since the script does not run from inside the runfiles directory.

js/private/js_binary.sh.tpl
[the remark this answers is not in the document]
acozzette: I think it's saying that a `js_binary` could be invoked from some other binary and could therefore start in a different working directory. E.g. let's say you're using a `js_binary` as a protoc plugin and protoc invokes the plugin from some random directory, in which case we can't assume that we're already in `$BAZEL_BINDIR` when this code runs. I think that is a valid point, though it's weird that we were only handling this case for external repos. I updated my PR so that it will always cd to the absolute path whether we are working with an external repo or not.

e2e/js_binary_workspace/BUILD.bazel
jbedard: The fact that a test had to change seems to imply this "refactor" is a breaking change. Can we avoid this?

acozzette: The thing is that the test does not currently seem to be exercising the code that it is supposed to, since the use of `chdir` I mentioned above is not having any effect. I could add a new test, but it seems like I might as well fix the existing one.

The other possibility would be to fix the code so that the test *does* exercise it. If I run `bazel build @workspace//:run_bin` on `main` and look at the resulting output file, I see:
```
Running within directory: /home/adam/.cache/bazel/_bazel_adam/8b4505806fe4ef0e1042d702014949ed/sandbox/processwrapper-sandbox/59/execroot/_main/bazel-out/k8-fastbuild/bin
```
This seems wrong to me, as I would expect it to end with `bazel-out/k8-fastbuild/bin/external/workspace+`. That would be fairly straightforward to fix and would remove the need for the `pkg` subdirectory. That could theoretically be interpreted as a breaking change, but I doubt it would break anything as it would be more of a bug fix.

What do you think would be the best path forward?

jbedard: Do you think this test was broken before? Then we can fix the test first in its own PR and then this "refactor" PR could avoid modifying tests?

Or just change this PR to a "fix" if there is a fix to be done? I don't understand the change here so it's hard for me to tell, which might mean a separate PR would explain it better 🤷

acozzette: OK, I just created another PR here with the fix: https://github.com/aspect-build/rules_js/pull/2815 That will make it so that `chdir = package_name()` has the intended effect even in the top-level directory of an external repo, and then there won't be a need to create a `pkg` subdirectory.

jbedard (approved):

PR #2815, opened 2026-04-27
fix: correctly handle empty chdir on js_binary

acozzette:
The `js_binary` rule has a `chdir` attribute, and it is common to set `chdir = package_name()` to indicate that the binary should run under the target bin directory in the subdirectory corresponding to the current package.

The logic for this checks `if ctx.attr.chdir`, which is usually fine but does not work correctly when the target is at the top level of an external repo. In that case `package_name()` is empty, making `chdir` falsy, but we do still need to change directories since the package is in an external repo.

This change fixes that bug and adds a test verifying that the binary runs in the expected working directory.

---

The js_binary rule now correctly handles an empty chdir in a target in an external repo.

### Test plan

- New test cases added

### DIFF
--- a/js/defs.bzl
+++ b/js/defs.bzl
@@ -83,6 +83,15 @@ def js_binary(**kwargs):
     Args:
         **kwargs: All attributes of the [js_binary](#js_binary) rule.
     """
+
+    # Often a js_binary target will set "chdir = package_name()", and if it is
+    # in the top-level directory then this will result in an empty string. That
+    # argument may still be significant, though, particularly if the target is
+    # in an external repo. We make sure to replace an empty string with "." so
+    # that the underlying rule can distinguish this from an unset chdir
+    # parameter.
+    if kwargs.get("chdir") == "":
+        kwargs["chdir"] = "."
     _js_binary(
         enable_runfiles = select({
             Label("@bazel_lib//lib:enable_runfiles"): True,
@@ -117,6 +126,8 @@ def js_test(**kwargs):
     Args:
         **kwargs: All attributes of the [js_test](#js_test) rule.
     """
+    if kwargs.get("chdir") == "":
+        kwargs["chdir"] = "."
     _js_test(
         enable_runfiles = select({
             Label("@bazel_lib//lib:enable_runfiles"): True,
--- a/e2e/js_binary_workspace/.bazelignore
+++ b/e2e/js_binary_workspace/.bazelignore
@@ -1,0 +1,1 @@
+workspace/
--- a/e2e/js_binary_workspace/BUILD.bazel
+++ b/e2e/js_binary_workspace/BUILD.bazel
@@ -14,3 +14,13 @@ build_test(
         "@workspace//:run_bin",
     ],
 )
+
+# Verify that js_binary and js_test in an external repo with chdir =
+# package_name() run in the expected working directory.
+test_suite(
+    name = "external_tests",
+    tests = [
+        "@workspace//:js_binary_chdir_test",
+        "@workspace//:js_test_chdir_test",
+    ],
+)
--- a/e2e/js_binary_workspace/MODULE.bazel
+++ b/e2e/js_binary_workspace/MODULE.bazel
@@ -2,6 +2,7 @@ module(name = "js_binary_workspace")
 
 bazel_dep(name = "aspect_rules_js", version = "0.0.0", dev_dependency = True)
 bazel_dep(name = "bazel_skylib", version = "1.8.1", dev_dependency = True)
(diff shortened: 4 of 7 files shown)


js/private/js_binary.bzl
```diff
@@ -7,8 +7,13 @@ load("@bazel_lib//lib:windows_utils.bzl", "create_windows_native_launcher_script
 load(":bash.bzl", "BASH_INITIALIZE_RUNFILES")
 load(":js_helpers.bzl", "LOG_LEVELS", "envs_for_log_level", "gather_files_from_js_infos", "gather_runfiles")
 
+# We need to be able to distinguish whether the chdir attribute was explicitly
+# set, so we use this sentinel as the default.
+_CHDIR_UNSET_SENTINEL = "<unset>"
```
jbedard: My only thought or alternative idea... can the rule assume `""` means unset and `"."` means root dir? Then the public API macro converts `None` to `""` and `""` to `"."`.

Is that any better/worse then having this magic `"<unset>"` value? In some ways this `"<unset>"` is simpler because it lives in this one file only?

acozzette: Hmm, good question. I forgot that we already had a wrapper macro where we could do this. I think that would work. Either way seems fine to me. Converting things in the macro seems maybe slightly better because then we don't have to prevent users from working with a directory called `<unset>` (although presumably that would be a rare occurrence). What do you think?

jbedard: I think I'd vote for the macro to avoid the magic string. Even if internal it would still show in bazel query etc. 

The macro also means we can put a breaking change in the macro in the future without touching the underlying rule...

acozzette: Sounds good, I moved the fix to the macro.

js/private/test/image/asserts.bzl
```diff
@@ -6,7 +6,7 @@ load("//js:defs.bzl", "js_image_layer")
 # buildifier: disable=function-docstring
 def assert_tar_listing(name, actual, expected):
     # Either of these two file sizes may be observed on a file like /js/private/test/image/bin
-    sanitize_cmd = "sed -E 's/239[0-9]{2}|24[0-9]{3}/xxxxx/g'"
+    sanitize_cmd = "sed -E 's/250[0-9]{2}|24[0-9]{3}/xxxxx/g'"
```
jbedard: Was this actually required? lol

js/defs.bzl
```diff
@@ -83,6 +83,15 @@ def js_binary(**kwargs):
...
         **kwargs: All attributes of the [js_binary](#js_binary) rule.
     """
+
+    # Often a js_binary target will set "chdir = package_name()", and if it is
+    # in the top-level directory then this will result in an empty string. That
+    # argument may still be significant, though, particularly if the target is
+    # in an external repo. We make sure to replace an empty string with "." so
+    # that the underlying rule can distinguish this from an unset chdir
+    # parameter.
+    if kwargs.get("chdir") == "":
```
jbedard, line 93: Won't this fail when `chdir` was not specified and we need `kwargs.get("chdir", None)`?

js/defs.bzl
```diff
@@ -83,6 +83,15 @@ def js_binary(**kwargs):
...
+        kwargs["chdir"] = "."
     _js_binary(
```
jbedard, line 95: Don't we need the same fix for `js_test`? Please add a test that fails without the fix if so...

js/private/test/image/asserts.bzl
acozzette: Yeah, I'm not exactly sure what this test is about but I think my changes somehow made a file here get bigger.

js/defs.bzl
acozzette: I believe this should be fine because if you don't specify a default then the default is just None: https://bazel.build/rules/lib/core/dict#get

jbedard: Oh, looks like you're right, I thought it failed if you didn't specify a default. If CI is green then 👍

js/private/test/image/asserts.bzl
jbedard: This PR only changes rules/macros and not the generated .sh so I don't understand why this is needed. Please double check

js/defs.bzl
acozzette: Yes, good catch, I added a test for this and made the same fix for js_test.

js/private/test/image/asserts.bzl
acozzette: Yeah, this seems to be unneeded now because the tests pass without it, so I reverted this change.

jbedard (approved):

PR #2823, opened 2026-05-04
fix: fix `js_run_binary` so tools executed with `use_execroot_entry_point` use exec-configuration runfiles

longlho:
`js_run_binary` runs its `tool` in exec config through `bazel-lib` `run_binary`, but previously hoisted the tool runfiles through target-config helper targets. When target and exec platforms differ, this can omit exec-platform files, including native optional npm packages such as `@rolldown/binding-linux-x64-gnu`.

This changes the runfiles hoist to use a helper rule with `cfg = "exec"` on the tool attribute. It also makes the launcher resolve execroot entry points with `JS_BINARY__BINDIR` when `JS_BINARY__USE_EXECROOT_ENTRY_POINT` is set, while preserving `BAZEL_BINDIR` for action context.

Repro repo: https://github.com/longlho/rules-js-exec-platform-repro

---

Fix `js_run_binary` with `use_execroot_entry_point` under differing target and execution platforms, including native optional npm dependency resolution for JS tools.

### Test plan

- New test cases added
- Manual testing:
  - `npm exec --yes @bazel/bazelisk -- test //js/private/test/js_run_binary_exec_config:run_tool_uses_exec_config_runfiles --test_output=all`
  - `npm exec --yes @bazel/bazelisk -- run //:buildifier.check`
  - `npm exec --yes @bazel/bazelisk -- test //js/private/test/js_binary_sh/... //js/private/test/js_run_binary_exec_config:run_tool_uses_exec_config_runfiles --test_output=errors`

### DIFF
--- a/js/private/js_binary.sh.tpl
+++ b/js/private/js_binary.sh.tpl
@@ -95,12 +95,36 @@ function logf_debug {
     fi
 }
 
+function _normalize_cpu {
+    case "$1" in
+    *aarch64* | *arm64*) echo "arm64" ;;
+    *amd64* | *k8* | *x86_64*) echo "x64" ;;
+    *) echo "$1" ;;
+    esac
+}
+
+function _use_exec_config_entry_point {
+    if [ "$(_normalize_cpu "${BAZEL_TARGET_CPU:-}")" != "$(_normalize_cpu "${JS_BINARY__TARGET_CPU:-}")" ]; then
+        return 0
+    fi
+    case "${BAZEL_BINDIR:-}" in
+    bazel-out/*-ST-* | bazel-out/platform-*) return 0 ;;
+    *) return 1 ;;
+    esac
+}
+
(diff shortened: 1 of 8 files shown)


acozzette:
I have been working on this same issue and came up with a fix similar to yours recently: https://github.com/aspect-build/rules_js/pull/2809 However, I came to the realization that we probably should not try to "fix" `use_execroot_entry_point` to use exec-configured runfiles. The problem is that `use_execroot_entry_point` intentionally hoists runfiles in the target configuration, and the purpose is to be compatible with frameworks like Next.js. Next.js does pre-rendering, which involves running target-platform code during the build, so this inherently requires blurring the distinction between the exec and target platforms. So I think the best solution is to turn off `use_execroot_entry_point` in cases where it is not needed, and perhaps even make it default to `False` eventually. If you set `use_execroot_entry_point = False`, does that solve the problem for you?

PR #2837, opened 2026-05-12
feat: add flag controlling the default for use_execroot_entry_point

acozzette:
This change introduces the flag `--@aspect_rules_js//js:use_execroot_entry_point`, which controls the default behavior of the `use_execroot_entry_point` option on `js_run_binary`.

We could like to recommend moving away from enabling `use_execroot_entry_point`, and this flag provides an easy way to do that at a global level, while still making it possible to override that on specific targets if necessary.

I also added CI test runs with `--@aspect_rules_js//js:use_execroot_entry_point=False` so that we have good test coverage of both modes. This required explicitly setting `use_execroot_entry_point` on some targets that only work with one more or the other.

---

There is now a `--@aspect_rules_js//js:use_execroot_entry_point` flag which can be set to `True` or `False` to determine the default `use_execroot_entry_point` behavior for `js_run_binary`.

### Test plan

- New test cases added


contrib/nextjs/defs.bzl
```diff
@@ -202,6 +202,7 @@ def nextjs_build(name, config, srcs, next_js_binary, data = [], **kwargs):
         chdir = native.package_name(),
         mnemonic = "NextJs",
         progress_message = "Compile Next.js app %{label}",
+        use_execroot_entry_point = True,
```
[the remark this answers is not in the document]
acozzette: I pushed a fix for this.

js/private/test/js_run_devserver/BUILD.bazel
```diff
@@ -28,6 +28,7 @@ js_run_devserver_test(
         "no-remote-exec",
     ],
     tool = ":jasmine",
+    use_execroot_entry_point = True,
```
jbedard: Is this an example where the `tool` should be updated to support `False` instead? Maybe just leave a `TODO` here for now and come back to it later?

acozzette: Yes, I think we eventually want to update this test so that it works with `use_execroot_entry_point = False`. I added a TODO for that.

jbedard:
I think this needs more docs in the `js_run_binary` docstring as well as maybe in a README or something like that?

Can we explain how this helps with cross-platform builds and avoids duplicating runfiles, avoids the optionDependencies bug etc, how we may switch the default in rules_js v4 etc...?

acozzette:
I updated the docstring and also added those details to docs/use_execroot_entry_point.md.

docs/use_execroot_entry_point.md
```diff
@@ -0,0 +1,73 @@
...
+
+## What `use_execroot_entry_point` does
+
+**`use_execroot_entry_point = True` (the current default):**
+The tool's runfiles are hoisted into `srcs`, which causes them to be rebuilt in
+the target configuration and land in the target-platform bin directory. The
+entry point used is the one in that output tree (the "execroot entry point"),
+rather than the copy inside the runfiles symlink tree. With everything
+consolidated in `bazel-out/<target-cfg>/bin/`, Node.js sees a single
+`node_modules` tree. This is the right choice for Next.js for two reasons:
```
jbedard: Why bring up `Next.js` at this point? We were talking about generic bazel/node/node_modules and suddenly `Next.js` is mentioned.

Can we remain generic and just use Nextjs as an example in one sentence at the end of the paragraph?

acozzette: Good point, the info about Next.js seems very abrupt. I tweaked things to cut down on the detail a little bit and make it clear that Next.js is just one example.

acozzette:
I realized while working on #2853 that there's really no problem with using `use_execroot_entry_point = True` on `js_run_devserver`, so I now think we should just leave that option alone. I reverted the parts of this PR that touched `js_run_devserver`.

jbedard:
Can you outline why it's not a problem there?

That means `js_run_devserver(use_execroot_entry_point=True)` by default still instead of `None` and the config option?

acozzette:
`js_run_devserver` always builds everything for the target platform, so as far as I can tell there's no real risk of building for the wrong platform or mixing code built for different platforms.

So for `js_run_devserver` it should be fine to keep `use_execroot_entry_point = True` as the default.

jbedard:
Would that conflict with our goal of making `use_execroot_entry_point = False` by default though? Even when exec==target does it not have some minor side-effects that we'd prefer were equal in all rules?

acozzette:
I don't really think so. Since `bazel run` doesn't use a sandbox, when you run a `js_run_devserver` you end up with sources both in the runfiles and directly in the execroot. So there's no extra hoisting happening, and all the sources are built for the same platform. I think there would be some benefit to making the option `False` by default everywhere, but to me the disruption just doesn't seem worth it in the case of `js_run_devserver`.

docs/use_execroot_entry_point.md
```diff
@@ -0,0 +1,71 @@
+# use_execroot_entry_point
```
jbedard, line 1: Maybe in a followup PR, but currently nothing actually links to this doc? Maybe https://github.com/aspect-build/rules_js/issues/2852 needs to solve that as well?

docs/use_execroot_entry_point.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Recommendation
+
+We recommend setting `use_execroot_entry_point = False` wherever possible and
+ensuring that all code executed during the build is declared as a dependency of
+the `js_binary`. The main exception is Next.js and similar frameworks that
+expect inputs and outputs in the same directory tree or that execute
+target-platform code during the build, in which case `True` is required.
+
+To disable `use_execroot_entry_point` globally, pass the build flag:
```
jbedard: It doesn't disable it globally though, it just changes the default?

acozzette: Good point, I updated this to clarify that this only changes the default.

docs/use_execroot_entry_point.md
acozzette: That's true, nothing links to this yet. Maybe in a followup I should bring back docs/README.md and point to the new file as part of fixing #2852?

jbedard: A followup with that would be great 👍

jbedard (approved):

Issue #2581, opened 2025-12-15
[Bug]: BAZEL_BINDIR requirement prevents use of compilation_mode transitions


jbedard:
#2574 is blocked while executing a `js_binary` action requires specifying `BAZEL_BINDIR` in `ctx.action.env`

acozzette:
One related issue I hit recently is that the `BAZEL_BINDIR` requirement forces us to directly invoke protoc [here](https://github.com/aspect-build/rules_js/blob/401406a1984917e4d17a1596b5c0dbced0927ff4/js/private/proto.bzl#L92) so that we can set that variable. Otherwise we could simplify that code by just calling out to `proto_common.compile()` from the protobuf repo.

acozzette:
If we just want to make `js_binary` compatible with path mapping, one possible way to do that would be to just provide a way to pass `BAZEL_BINDIR` as a command-line flag instead of an environment variable.

acozzette:
Actually as I think about it more, even passing the output bin dir as a flag will be tricky to do in a way that's compatible with path mapping. From what I understand, Make variables and things like `$(location)` do not play nicely with path mapping. The only way I can think of to get ahold of the output bin dir in a path-mapping-friendly way is to pass one of the outputs as an argument with something like `args.add(out)` as described [here](https://github.com/bazelbuild/bazel/discussions/22658) and have the wrapper script parse the bin dir from that. But that has to be done in a rule rather than a macro, so `js_run_binary` can't easily do it without new support in the underlying `_run_binary` rule from bazel-lib.

jbedard:
Fixed in https://github.com/aspect-build/rules_js/pull/2918

PR #2844, opened 2026-05-14
feat: prototype of js_run_binary changes

acozzette:
These are prototype changes just meant for discussion right now.

`npm/private/test/BUILD.bazel` contains an example Rspack build that fails with `use_execroot_entry_point = True` since it involves an OS transition. I got it working with `use_execroot_entry_point = False` and tried to set things up so that we can do so in a reasonably user-friendly way. I'm using this as a motivating example for what `js_run_binary` needs to support.

One problem is that the Rspack config file is a JavaScript source file, and the naive way of providing it as an input will always result in it landing outside the runfiles directory, preventing it from importing any other sources in the binary. We could take dependencies of the config file (such as `@rspack/cli`), and include them outside the runfiles directory as well, but then we start to run of the risk of duplicating any dependencies that are also used by the `js_binary`. To get around that problem, I updated `js_run_binary` to add an `extra_tool_data` param. This parameter takes any inputs that need to be runfiles, and we create a new executable with those runfiles added, making it possible to get all JavaScript sources into the runfiles.

The other problem is that once we have the config file in the runfiles, we need a way to pass its path as a command-line argument. As far as I can tell, the runfiles directory cannot be determined at analysis time, so the earliest we can do this is in the launcher script. I set things up so that the launcher script evaluates any appearances of `$RUNFILES_DIR` in its arguments. There's probably a lot of room for improvement here, but at least it provides a syntax that is not too horrible.

---

### Test plan

- New test cases added


js/private/js_run_binary.bzl
```diff
@@ -245,6 +295,22 @@ def js_run_binary(
...
+    if extra_tool_data:
+        augmented_tool_name = "_{}_extra_tool_data".format(name)
+        _js_run_binary_extra_tool_data(
+            name = augmented_tool_name,
+            tool = tool,
+            extra_data = extra_tool_data,
+            tags = kwargs.get("tags", []) + ["manual"],
+            testonly = kwargs.get("testonly", False),
+        )
+        effective_tool = ":{}".format(augmented_tool_name)
```
jbedard, line 312: Why would "extra tool data" replace the tool itself? 🤔

acozzette: This was the only way I could figure out to get additional runfiles in. As far as I can tell, there's no way to add runfiles to an existing executable, so instead you have to create a new executable with the added runfiles.

acozzette: If this approach seems too weird, we could instead require the `tool` to already have all necessary runfiles in it. So with the Rspack example, we could create an Rspack binary that includes `rspack_config.cjs` (plus any dependencies), and then just pass that as the `tool` to `js_run_binary`.

acozzette: Actually I gave that a try just now and it works surprisingly well: https://github.com/acozzette/rules_js/commit/93c69c9b47ef57f6df0b0603d4b32efbae063732 Amazingly, you can use `$$RUNFILES_DIR` in `fixed_args` on a `js_binary` and it already works with no new functionality required. What do you think?

jbedard: > If this approach seems too weird, we could instead require the tool to already have all necessary runfiles in it.

That seems like a reasonable restriction to me. Does that mean the `tools` attribute is useless though?

jbedard: > Actually I gave that a try just now and it works surprisingly well: https://github.com/acozzette/rules_js/commit/93c69c9b47ef57f6df0b0603d4b32efbae063732 Amazingly, you can use $$RUNFILES_DIR in fixed_args on a js_binary and it already works with no new functionality required. What do you think?

That LGTM. I would just want to make sure `$$RUNFILES_DIR` is is not deprecated and double check there's no downside to using it. Will that var still exist if/when we [drop `BAZEL_BINDIR`](https://github.com/aspect-build/rules_js/issues/2581)?

acozzette: I don't think the `tools` attribute is completely useless, but I see now that it's not a great choice for JS sources since any JS sources listed in `tools` won't be able to import from the other sources in runfiles. It seems fine for other kinds of inputs, though.

From what I understand, `$RUNFILES_DIR` is pretty well established and is set by Bazel, so it won't go away if we drop `BAZEL_BINDIR`. I think a bigger potential problem, though, is that this assumes runfiles are supported, which is not true on Windows by default. I was thinking about addressing that issue by having our launcher script take a dependency on `@bazel_tools//tools/bash/runfiles`, and then we could call the `rlocation` function which will handle proper runfiles lookups regardless of whether runfiles are fully supported. I gave this a try and it seemed to work. The syntax in the BUILD file then looks like `"$$(rlocation $(rlocationpath :rspack_config))"`. But currently Windows does not work with `use_execroot_entry_point = False` due to other runfiles issues, and so there are other things we would have to fix as well.

PR #2846, opened 2026-05-15
feat: streamline invocation of binaries from npm and update examples

acozzette:
The new example in examples/rspack started out as a test case by @jbedard and @nickschaap showing how `js_run_binary` can sometimes fail to work correctly when the execution and target platforms are different.

I was able to get it working, primarily by setting `use_execroot_entry_point = False`. The problem with the default behavior (`use_execroot_entry_point = True`) is that it runs target-platform sources during the build action, which can go haywire if the target platform is meaningfully different from the execution platform.

Getting all the sources in the right place and built for the right platform is not entirely straightforward, so I thought it would be good to have some simple examples showing this.

A small new feature the examples rely on is the ability to pass `data` and `fixed_args` to the macro that wraps `js_run_binary()` in the generated `package_json.bzl` file. The args are forwarded to the underlying `js_binary`.

---

The `bin.<binary name>()` macro generated in `package_json.bzl` now supports passing `data` and `fixed_args` through to the `js_binary` it creates.

### Test plan

- New test cases added


examples/rspack/rspack.config.cjs
```diff
@@ -0,0 +1,12 @@
+const { defineConfig } = require('@rspack/cli');
+const path = require("path");
+
+module.exports = defineConfig({
+  entry: {
+    main: './rspack_entry.js',
```
nickschaap, line 6: So is rspack resolving this from the current working directory or relative to the config file since the config file will live in a separate bazel-out tree than than this entrypoint

acozzette: I believe it's resolving it from the current working directory. `js_run_binary`'s working directory is basically `"bazel-out/<target platform>/bin/" + chdir` where in this case `chdir` is `"rspack"`.

For the output path below, I did have to replace `__dirname` with `process.cwd()` to reflect that the config file is no longer placed in the target platform bin directory.

acozzette:
I went ahead and updated a bunch of other examples in the repo to use this same pattern. Luckily it was easy to just point Claude at the original example and ask it to imitate that.

jbedard:
I think we need to find a way to still directly invoke `vite_bin.vite()` and not force `vite_bin.vite_binary()` + `js_run_binary`.

Wrapping things like tool config files+deps into `js_library` is fine though, imo.

acozzette:
I tweaked things so that now we are down to just one `js_library` for the config and one call to `vite_bin.bin()` (or whatever the particular tool might be).

js/private/coverage/bundle/BUILD.bazel
```diff
@@ -1,31 +1,38 @@
...
         ":node_modules/c8",
     ],
-    outs = [
-        "bundle.js",
-    ],
-    args = [
+    outs = ["bundle.js"],
+    chdir = package_name(),
+    data = [":config"],
+    fixed_args = [
```
jbedard, line 27: Why the switch from `args` to `fixed_args`?

jbedard: It looks like that's actually a new feature that should be documented?

examples/rspack/BUILD.bazel
```diff
@@ -0,0 +1,77 @@
...
+    name = "rspack_config",
+    srcs = ["rspack.config.cjs"],
+    deps = [":node_modules/@rspack/cli"],
+)
+
+bin.rspack(
+    name = "rspack_build",
+    srcs = ["rspack_entry.js"],
+    outs = ["rspack/main.bundle.js"],
+    chdir = package_name(),
```
jbedard, line 20: Is this required when using `$$RUNFILES_DIR`?

examples/rspack/BUILD.bazel
```diff
@@ -0,0 +1,77 @@
...
+    data = [":rspack_config"],
+    # Since :rspack_config is a dependency of the binary, we need to reference
+    # it here in fixed_args. The fixed_args param applies to the js_binary
+    # whereas args applies to js_run_binary.
+    fixed_args = [
+        "build",
+        "--config",
+        "$$RUNFILES_DIR/$(rlocationpath :rspack_config)",
+    ],
+    use_execroot_entry_point = False,
```
jbedard, line 32: Is this required? Should we leave it out and let https://github.com/aspect-build/rules_js/pull/2837 test both values on CI for `examples/*`?

js/private/coverage/bundle/BUILD.bazel
acozzette: Yes, I guess that should be documented here, is that right? https://docs.aspect.build/bazel/javascript/rules_js#use-binaries-published-to-npm

examples/rspack/BUILD.bazel
acozzette: Yes, this is still important because it affects the paths referenced in rspack.config.js.

examples/rspack/BUILD.bazel
acozzette: This particular one is required because otherwise the cross-platform transitions below will fail. But I updated all the others to leave it out so that we test both values on CI (after #2837 lands).

js/private/coverage/bundle/BUILD.bazel
acozzette: The reason for specifying the args with `fixed_args` is that we need to set them on the `js_binary`. The `js_binary` is what takes the dependency on `:config` and so it is the only place where we can evaluate `$(rlocationpath :config)`. Also `$RUNFILES_DIR` only gets evaluated if you set it with `fixed_args` on a `js_binary`--I'm not even sure if this is an intentional feature, but it works and is convenient for this use case.

I was thinking ideally I would like to eventually fix things so that you can do `$$(rlocation $(rlocationpath :config))`, or even `$(rlocation :config)`. To do that we need to have the launcher script take a dependency on the bash runfiles support library, though.

jbedard: We can't just pass `args` to the `js_binary`? What is the use case here for using both `args` and `fixed_args`?

acozzette: The thing is that `fixed_args` applies to the `js_binary` and `args` applies to the `js_run_binary`. So if you want to do `$(rlocationpath :config)` then that has to be on the `js_binary` since that is what has a dependency on `:config`. And if you wanted to do something like `$(rootpath rspack_entry.js)` then that would have to go on the `js_run_binary` which takes that file in `srcs`. We aren't currently expanding any locations in `srcs` in our examples, but it seems like the kind of thing that could often be necessary.

I was trying to think if there is some way we could get everything working with a single `args` parameter, but it's not entirely obvious if that is doable. The best way I can think of would be to add a param to `js_run_binary` for labels that can participate in variable expansion but should not be inputs to the build action. Then we could put all the args on the `js_run_binary`. But that would require either updating or forking `run_binary`, and also updating the launcher script to be able to expand `$RUNFILES_DIR` in its command-line args (or some other syntax we come up with).

jbedard: But it's basically `js_binary(args)` vs `js_run_binary(args)`, while being hidden behind a macro which makes it hard to create a clear API for it?

Is `fixed_args` used somewhere that you're getting the term from? What about `tool_args` to align with `tool` and `tools`?

acozzette: I just went with `fixed_args` since that is the name of the attribute on `js_binary`. It does technically have a built-in `args` attribute as well, but from what I can tell it is not particularly useful for us since it only applies when you execute it with either `bazel run` or `bazel test`.

`tool_args` sounds ok to me, though one thing that's slightly weird about it is that then `rollup_bin.rollup()` and `rollup_bin.rollup_binary()` would have different names for this (`tool_args` for one and `fixed_args` for the other). I'm happy to go with whichever one you prefer, though.

jbedard: If `fixed_args` is on `js_binary` then that's probably what we should go with here as well, IIUC 🤔

npm/private/npm_import.bzl
```diff
@@ -364,7 +364,8 @@ def bin_internal(name, store_target, store_target_dir, bin_path, bin_mnemonic, *
     _js_binary(
         name = "%s__js_binary" % name,
         entry_point = ":%s__entry_point" % name,
-        data = [store_target],
+        data = [store_target] + kwargs.pop("data", []),
+        fixed_args = kwargs.pop("fixed_args", []),
```
jbedard, line 368: I think this needs to be documented somewhere?

npm/private/npm_import.bzl
```diff
@@ -364,7 +364,8 @@ def bin_internal(name, store_target, store_target_dir, bin_path, bin_mnemonic, *
     _js_binary(
         name = "%s__js_binary" % name,
         entry_point = ":%s__entry_point" % name,
-        data = [store_target],
+        data = [store_target] + kwargs.pop("data", []),
```
jbedard, line 367: So `data` is considered exec-data and not target-data? I think that makes sense? 🤔

But then where does target-data go?

npm/private/npm_import.bzl
acozzette: I filed https://github.com/aspect-build/rules_js/issues/2852 for this as we were talking about offline.

npm/private/npm_import.bzl
acozzette: `js_run_binary` doesn't accept a `data` param, so luckily that works out nicely. `srcs` is the way to pass target-platform data and that will get passed through to `js_run_binary`.

jbedard (approved):

PR #2848, opened 2026-05-18
Add configurable coverage_extensions attribute to js_binary

olbapmar:
Allow overriding the file extensions used for coverage instrumentation instead of relying on the hardcoded JS/TS defaults. This is useful when js_binary invokes other language binaries (e.g., C++) whose source files use different extensions.

### DIFF
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -230,6 +230,23 @@ _ATTRS = {
         to use npm.
         """,
     ),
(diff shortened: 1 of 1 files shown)


jbedard:
Do you have any examples that we can turn into a test?

olbapmar:
I'm closing this as I've realized I need to make more changes before upstreaming some patches. Sorry for the noise!

PR #2864, opened 2026-05-28
fix(npm): correct package_store_prefix_len for links in external-repo sub-packages

oscar-meroton:
When calculating the package store prefix truncation with external modules (i.e. a repo name is included in the path), an extra slash after the repo name was unaccounted for, resulting in incorrect symlinks (and subsequently `ERR_MODULE_NOT_FOUND` error during builds).

Below is an example of the logic used to calculate the truncation length:

```python
# ../[com_github_example_repo+]/[example_package]/[node_modules/.aspect_rules_js/]foo@1.2.3/node_modules/bar
#    |        repo name       | |  package name | |         store prefix         |

path = "../com_github_example_repo+/example_package/node_modules/.aspect_rules_js/foo@1.2.3/node_modules/bar"
truncate_length = len(repo_name) + len(package_name) + len(store_prefix) + 3 # = 72
truncated_path = path[truncate_length:] # = "s/foo@1.2.3/node_modules/bar"
```

The extra `s` in the beginning of `truncated_path` should not be there and is solved by changing the `+ 3` to a `+ 4`.

### DIFF
--- a/e2e/bzlmod/BUILD.bazel
+++ b/e2e/bzlmod/BUILD.bazel
@@ -134,6 +134,9 @@ build_test(
 
         # js_run_binary() running a binary in @other_repo
         ":run_other_module_binary_test",
+
+        # js_run_binary() running a binary linked in a sub-package of @other_repo
+        ":run_other_module_subpackage_binary_test",
     ],
 )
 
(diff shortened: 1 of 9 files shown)


jbedard:
@oscar-meroton sorry I didn't see this till now. How can we reproduce this in a test?

oscar-meroton:
@jbedard Hey, sorry for the delay. I'll try to look into reproducing this in a test later this week.

jbedard (approved):

PR #2868, opened 2026-05-30
feat(npm): use pnpm publish for npm_package

luciomartinez:
## What

Changes `npm_package(publishable = True)` so the generated `<name>.publish` target runs `pnpm publish` instead of `npm publish`.

This removes the proposed `publish_tool` API and keeps the public target shape unchanged. A `pnpm_binary` override remains available for custom pnpm repository names; by default it uses `@pnpm//:pnpm`.

The publish wrapper now:

- invokes pnpm with the generated package directory from `bazel-bin`
- uses a temporary cwd when `BUILD_WORKSPACE_DIRECTORY` is available, copying in only `pnpm-workspace.yaml` and `.npmrc` when present
- adds `--no-git-checks` because Bazel publishes generated output
- keeps `include_npm = True` for pnpm <=10, where registry operations may delegate to npm

## Why

`pnpm publish` handles pnpm workspace publishing semantics such as replacing `catalog:` dependency specs before packing. `npm publish` does not understand those pnpm-specific specs.

The isolated temporary cwd exposes the workspace-level pnpm settings required for publishing without making the source workspace the child process cwd. Using pnpm directly also avoids adding a new publish-tool selection API while preserving the existing `<name>.publish` entrypoint for users.

## Notes

- The catalog regression test verifies that pnpm rewrites the published `package.json`, not only that dry-run publish exits successfully.
- User-level `.npmrc` and environment-based configuration remain available through the inherited process environment.
- pnpm 11 publishes natively; the npm fallback path is only relevant for older pnpm versions.

### DIFF
--- a/npm/private/npm_package.bzl
+++ b/npm/private/npm_package.bzl
@@ -140,6 +140,7 @@ def npm_package(
         include_runfiles = False,
         hardlink = "auto",
         publishable = False,
+        pnpm_binary = "@pnpm//:pnpm",
         verbose = False,
         **kwargs):
     """A macro that packages sources into a directory (a tree artifact) and provides an `NpmPackageInfo`.
@@ -147,8 +148,11 @@ def npm_package(
     This target can be used as the `src` attribute to `npm_link_package`.
 
     With `publishable = True` the macro also produces a target `[name].publish`, that can be run to publish to an npm registry.
-    Under the hood, this target runs `npm publish`. You can pass arguments to npm by escaping them from Bazel using a double-hyphen,
-    for example: `bazel run //path/to:my_package.publish -- --tag=next`
+    Under the hood, this target runs `pnpm publish`. You can pass arguments to pnpm by escaping them from Bazel using a double-hyphen,
+    for example: `bazel run //path/to:my_package.publish -- --tag=next`.
+
+    pnpm publishing uses the configured `pnpm_binary` and can resolve pnpm workspace settings such as `catalog` entries from
+    `pnpm-workspace.yaml`. Workspace protocol dependencies follow pnpm's own resolution requirements.
 
     Files and directories can be arranged as needed in the output directory using
     the `root_paths`, `include_srcs_patterns`, `exclude_srcs_patterns` and `replace_prefixes` attributes.
@@ -210,7 +214,7 @@ def npm_package(
 
         srcs: Files and/or directories or targets that provide `DirectoryPathInfo` to copy into the output directory.
 
-        args: Arguments that are passed down to `<name>.publish` target and `npm publish` command.
+        args: Arguments that are passed down to `<name>.publish` target and the publish command.
 
         data: Runtime / linktime npm dependencies of this npm package.
 
@@ -409,6 +413,9 @@ def npm_package(
 
         publishable: When True, enable generation of `{name}.publish` target
 
+        pnpm_binary: Label of the pnpm binary used for publishing. This is typically `@pnpm//:pnpm`
+            from the `pnpm` extension in `@aspect_rules_js//npm:extensions.bzl`.
+
         verbose: If true, prints out verbose logs to stdout
 
         **kwargs: Additional attributes such as `tags` and `visibility`
@@ -437,10 +444,11 @@ def npm_package(
             name = "{}.publish".format(name),
             entry_point = Label("@aspect_rules_js//npm/private:npm_publish_mjs"),
             fixed_args = [
+                "$(rootpath {})".format(pnpm_binary),
                 "./$(rootpath :{})".format(name),
             ],
-            data = [name],
-            # required to make npm to be available in PATH
+            data = [name, pnpm_binary],
+            # pnpm <11 may delegate registry operations to npm. This can be removed when rules_js drops pnpm <=10.
             include_npm = True,
             args = args,
             tags = kwargs.get("tags", []) + ["manual"],
--- a/npm/private/npm_package.bzl
+++ b/npm/private/npm_package.bzl
@@ -154,6 +154,16 @@ def npm_package(
     pnpm publishing uses the configured `pnpm_binary` and can resolve pnpm workspace settings such as `catalog` entries from
     `pnpm-workspace.yaml`. Workspace protocol dependencies follow pnpm's own resolution requirements.
 
+    The default `pnpm_binary` requires the consuming module to define and import the `@pnpm` repository:
+
+    ```starlark
+    pnpm = use_extension("@aspect_rules_js//npm:extensions.bzl", "pnpm")
+    pnpm.pnpm(name = "pnpm")
+    use_repo(pnpm, "pnpm")
+    ```
+
+    Pass a different `pnpm_binary` label when the pnpm repository uses another name.
+
     Files and directories can be arranged as needed in the output directory using
     the `root_paths`, `include_srcs_patterns`, `exclude_srcs_patterns` and `replace_prefixes` attributes.
 
@@ -413,8 +423,9 @@ def npm_package(
 
         publishable: When True, enable generation of `{name}.publish` target
 
-        pnpm_binary: Label of the pnpm binary used for publishing. This is typically `@pnpm//:pnpm`
-            from the `pnpm` extension in `@aspect_rules_js//npm:extensions.bzl`.
+        pnpm_binary: Label of the pnpm binary used for publishing. The default `@pnpm//:pnpm`
+            requires the consuming module to define the `pnpm` extension repository and import it
+            with `use_repo(pnpm, "pnpm")`.
 
         verbose: If true, prints out verbose logs to stdout
 
--- a/npm/private/npm_publish.mjs
+++ b/npm/private/npm_publish.mjs
@@ -1,9 +1,73 @@
 import { spawnSync } from 'node:child_process'
+import {
+    copyFileSync,
+    existsSync,
+    mkdtempSync,
+    rmSync,
+} from 'node:fs'
+import { tmpdir } from 'node:os'
+import path from 'node:path'
 
-const restArgs = process.argv.slice(2)
+const [toolPath, packageDir, ...restArgs] = process.argv.slice(2)
 
-const spawn = spawnSync('npm', ['publish', ...restArgs], {
+if (!toolPath || !packageDir) {
+    console.error(
+        'Expected publish tool path and package directory arguments.',
+    )
+    process.exit(1)
+}
+
+const packagePath = path.resolve(packageDir)
+
+let cleanupCwd
+let spawnOptions = {
     stdio: 'inherit',
-})
+}
+
+if (process.env.BUILD_WORKSPACE_DIRECTORY) {
+    cleanupCwd = mkdtempSync(path.join(tmpdir(), 'rules-js-pnpm-publish-'))
+
+    // Give pnpm only the workspace-level files it needs for publish settings.
+    for (const filename of ['pnpm-workspace.yaml', '.npmrc']) {
+        const source = path.join(process.env.BUILD_WORKSPACE_DIRECTORY, filename)
+        if (existsSync(source)) {
+            copyFileSync(source, path.join(cleanupCwd, filename))
+        }
+    }
+
+    spawnOptions = {
+        cwd: cleanupCwd,
+        env: {
+            ...process.env,
+            // The pnpm binary is itself a js_binary launcher; this keeps it runnable after
+            // changing cwd away from the Bazel output tree.
+            BAZEL_BINDIR: process.env.BAZEL_BINDIR || '.',
+        },
+        stdio: 'inherit',
+    }
+}
+
+const spawn = spawnSync(
+    path.resolve(toolPath),
+    ['publish', packagePath, '--no-git-checks', ...restArgs],
+    spawnOptions,
+)
+
+if (cleanupCwd) {
+    rmSync(cleanupCwd, {
+        force: true,
+        recursive: true,
+    })
+}
+
+if (spawn.error) {
+    console.error(spawn.error.message)
+    process.exit(1)
+}
+
+if (spawn.signal) {
+    console.error(`pnpm publish exited with signal ${spawn.signal}`)
+    process.exit(1)
+}
 
 process.exit(spawn.status)
--- a/npm/private/test/npm_package_publish/test.sh
+++ b/npm/private/test/npm_package_publish/test.sh
@@ -3,6 +3,8 @@
 readonly PUBLISH_A="$1"
 readonly PUBLISH_B="$2"
 readonly PUBLISH_PNPM="$3"
+readonly PNPM="$(cd "$(dirname "$4")" && pwd)/$(basename "$4")"
+readonly PKG_PNPM="$(cd "$(dirname "$5")" && pwd)/$(basename "$5")"
 
 # Assert that pnpm reads package.json from the package directory.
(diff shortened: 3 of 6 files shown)


jbedard:
I wonder if we should be using `pnpm publish` 100% of the time? With `js_binary(include_npm)` I assume `pnpm publish` will use the correct npm if/when it delegates to npm. Do you know any reason _not_ to use `pnpm publish`?

jbedard:
Some history, originally [it was decided](https://github.com/aspect-build/rules_js/pull/1074#discussion_r1197278219) that `pnpm publish` did not add enough value and there were concerns about how it will find `npm` when delegating to `npm publish`.

IMO your reasons for this PR show that using `pnpm publish` does add enough value today, especially with things like catalog deps, but also with pnpm 11 it seems it no longer delegates to npm at all (the note at the top of [the docs](https://pnpm.io/cli/publish)).

@luciomartinez want to try just using `pnpm publish` 100% of the time?

npm/private/npm_package.bzl
```diff
@@ -433,20 +457,37 @@ def npm_package(
...
+        if publish_tool == "pnpm":
+            js_binary(
+                name = "{}.publish".format(name),
+                entry_point = Label("@aspect_rules_js//npm/private:npm_publish_with_tool_mjs"),
+                fixed_args = [
+                    "$(rootpath {})".format(pnpm_binary),
+                    "./$(rootpath :{})".format(name),
+                ],
+                data = [name, pnpm_binary],
+                # required to make npm available because pnpm may delegate registry operations to npm
```
jbedard: Add a note that this is no longer true in pnpm 11+, so when we drop pnpm <=10 we can remove `include_npm`...

npm/private/npm_publish_with_tool.mjs
```diff
@@ -0,0 +1,28 @@
...
+        'Expected publish tool path and package directory arguments.',
+    )
+    process.exit(1)
+}
+
+const publishCwd = process.env.BUILD_WORKSPACE_DIRECTORY || process.cwd()
+
+const spawn = spawnSync(
+    path.resolve(toolPath),
+    ['publish', path.resolve(packageDir), ...restArgs],
```
jbedard: I think we need to add [--no-git-checks](https://pnpm.io/cli/publish#--no-git-checks) since we know we are publishing from `bazel-bin`?

luciomartinez:
Thanks! I pushed fixes as suggested:

- `npm_publish_with_tool.mjs` now adds `--no-git-checks` itself, so callers do not need to know about the bazel-bin/git-checks detail.
- Added the `include_npm` note for pnpm <=10; per pnpm docs, pnpm 11 publishes natively and no longer delegates to npm.
- Removed `--no-git-checks` from the test invocation so the test exercises the wrapper default.

On using `pnpm publish` 100% of the time: I do not have a technical objection from our use case. I kept this opt-in/default-npm because existing `publishable = True` users do not currently need a pnpm repo and may rely on npm publish semantics, so making pnpm the only/default path feels like a broader compatibility change. Happy to switch direction if maintainers prefer pnpm-only here.

jbedard:
@codex

luciomartinez:
CI note: two checks are currently failing:

- `test (e2e/js_image_oci)`
- `test (e2e/js_image_oci, bazel-9-no-execroot-entry-point)`

The useful error I found in the failed job logs is:

```text
ERROR: No test targets were found, yet testing was requested
```

This looks unrelated to this PR: the changes here are limited to `npm_package` publish wrapper behavior and publish tests, while the failing jobs are `e2e/js_image_oci`. Other `js_image_oci` matrix variants passed, including `bazel-8` and `bazel-9`, and the npm-related checks/tests are green. I would expect a rerun of the failed jobs to be the next useful step.

jbedard:
That might be github being flaky?
> ERROR: /home/runner/.bazel/external/jq.bzl+/jq/toolchain/BUILD:5:15: @@jq.bzl+//jq/toolchain:type depends on @@jq.bzl+//:package_metadata in repository @@jq.bzl+ which failed to fetch. no such package '@@package_metadata+//rules': java.io.IOException: Error downloading [https://github.com/bazel-contrib/supply-chain/releases/download/v0.0.6/supply-chain-v0.0.6.tar.gz] to /home/runner/.bazel/external/package_metadata+/temp8929006452875241854/supply-chain-v0.0.6.tar.gz: GET returned 502 Bad Gateway or Proxy Error

npm/private/npm_publish_with_tool.mjs
```diff
@@ -0,0 +1,28 @@
...
+        'Expected publish tool path and package directory arguments.',
+    )
+    process.exit(1)
+}
+
+const publishCwd = process.env.BUILD_WORKSPACE_DIRECTORY || process.cwd()
+
+const spawn = spawnSync(
+    path.resolve(toolPath),
+    ['publish', path.resolve(packageDir), '--no-git-checks', ...restArgs],
```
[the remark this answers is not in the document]
jbedard: Did your testing prove this bot-comment incorrect? Your testing showed `catalog:` and similar pnpm logic was done correctly?

luciomartinez:
> That might be github being flaky?

That's right. I'll re-trigger it by pushing fixes to the issue reported by Codex...

jbedard:
I manually got the CI jobs to re-run and they appear to be passing 👍

npm/private/npm_publish_with_tool.mjs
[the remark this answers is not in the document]
luciomartinez: Good catch to ask for proof here. I tightened the test so it now asserts that pnpm rewrites the catalog dependency before packing, not just that the dry-run succeeds.

The test workspace still only includes `packages/*`, while the package being published is the generated Bazel output package. The dry-run output now has to include the packed `package.json` size after pnpm resolution (`116` bytes, down from the 132-byte source manifest containing `catalog:`). That demonstrates pnpm is applying the catalog from `BUILD_WORKSPACE_DIRECTORY` even though the package directory itself is outside the workspace package globs.

jbedard:
Regarding using pnpm publish 100% of the time... a robot (2 actually) told me that the CLI args are (intentionally) almost equivalent, the differences it did point out are things that I think are not applicable in bazel or which we handled such as `--no-git-checks`. So I'm honestly tempted to just 100% switch and avoid the need for the `publish_tool` API.

If someone complains we (or I, using a robot) can maybe revert back to your `publish_tool` idea.

luciomartinez:
@jbedard I'm with you buddy! I took that direction and pushed a simpler version:

- `<name>.publish` now always invokes `pnpm publish` for `npm_package(publishable = True)`
- removed the proposed `publish_tool` API
- reused the existing internal `npm_publish.mjs` wrapper instead of adding a second wrapper
- kept `pnpm_binary` as the small escape hatch for custom pnpm repo names
- kept `include_npm = True` with the pnpm <=10 note
- updated the publish test expectations to pnpm errors and kept the catalog rewrite assertion

npm/private/npm_publish.mjs
```diff
@@ -1,9 +1,28 @@
...
+        cwd: publishCwd,
+        env: {
+            ...process.env,
+            BAZEL_BINDIR: process.env.BAZEL_BINDIR || '.',
+        },
+        stdio: 'inherit',
+    },
+)
+
+process.exit(spawn.status ?? 1)
```
jbedard: Why the new `?? 1`?

npm/private/npm_publish.mjs
```diff
@@ -1,9 +1,28 @@
...
+}
 
-process.exit(spawn.status)
+const publishCwd = process.env.BUILD_WORKSPACE_DIRECTORY || process.cwd()
+
+const spawn = spawnSync(
+    path.resolve(toolPath),
+    ['publish', path.resolve(packageDir), '--no-git-checks', ...restArgs],
+    {
+        cwd: publishCwd,
```
jbedard: Previously we didn't need the `cwd` or `env` - why are those required now?

npm/private/npm_publish.mjs
luciomartinez: It was guarding the case where `spawnSync` does not return a process exit status, for example if spawning fails or the child exits by signal. I agree `?? 1` was a bit opaque, so I made those cases explicit now: print `spawn.error`, handle `spawn.signal`, and then exit with the actual child status.

npm/private/npm_publish.mjs
luciomartinez: `cwd` is needed now because pnpm gets value specifically from workspace-level config: catalogs, workspace settings, and registry config should be discovered from the user workspace rather than from the Bazel output tree.

`env` is only there to preserve the existing environment while adding a default `BAZEL_BINDIR=.`. The pnpm binary is a `js_binary` launcher; once we run it from `BUILD_WORKSPACE_DIRECTORY`, it is no longer naturally running from the Bazel output tree, so the nested launcher needs that value to avoid its bindir check. I added short comments for both.

jbedard: So it runs from the src-dir basically, so it can do things like resolve catalogs. That means it has access to essentially the entire workspace now and I'd worry that means other side-effects could occur. Where previously it was running from within the bazel-out somewhere and only had the files declared as inputs (basically just the `npm_package` output I'd hope). But it still publishes the directory created in `bazel-bin`?

Did you debate any other ways of doing this like ensuring the `pnpm-workspace.yaml` is available in the cwd (without changing the cwd).

luciomartinez: Good point. I changed this so we no longer run pnpm from the whole source workspace for every publish.

The wrapper now keeps the normal/default publish path close to the previous behavior: pnpm inherits the launcher cwd and publishes the generated package directory from `bazel-bin`.

Only when the generated package manifest actually contains a `catalog:` dependency do we create a small temporary cwd and copy in just the workspace-level files pnpm needs (`pnpm-workspace.yaml`, plus `.npmrc` if present). The package being published is still the absolute generated package directory from `bazel-bin`; the temp cwd only gives pnpm catalog/registry config without making the whole source tree the child process cwd.

jbedard: How did `npm publish` get the `.npmrc` previously if we didn't have such a temp dir and copy step?

I wonder if should always do this step instead of "only when catalogs are used" just to be consistent? Only doing it "when required" seems fragile, and having 2 ways of doing things just complicates things imo.

Sorry I'm asking question without really investigating, but so far you're responding faster then I'd be able to answer my own questions 👍

jbedard: I think `catalog:` versions can also be in the `package.json` `pnpm` field instead of in `pnpm-workspace.yaml`? Maybe that is only the legacy way of doing it and we should simply not support it and avoid the hassle though? I'd have to check, but I _think_ that is deprecated...

luciomartinez: Great points! I simplified it again: whenever `BUILD_WORKSPACE_DIRECTORY` is available, every publish now uses the same isolated temporary cwd and copies in `pnpm-workspace.yaml` and `.npmrc` when present. There is no longer a package-content check or a separate catalog-only execution path. The published package is still the absolute generated package directory from `bazel-bin`.

On the previous npm behavior: npm already reads config from the inherited environment plus user/global config such as `$HOME/.npmrc` ([npm docs](https://docs.npmjs.com/cli/v11/configuring-npm/npmrc)). The old wrapper did not deliberately expose a source-root project `.npmrc`, since it ran from the Bazel output context. Copying a project `.npmrc` into the isolated cwd is explicit and limited; for pnpm it only carries auth and registry settings ([pnpm settings docs](https://pnpm.io/settings)).

I also checked the current pnpm 10 and 11 catalog docs. They define catalogs in `pnpm-workspace.yaml` via `catalog` / `catalogs`; package manifests contain the `catalog:` references. I do not see a current supported `package.json#pnpm` catalog-definition path that this wrapper needs to carry over ([pnpm catalogs docs](https://pnpm.io/catalogs)).

npm/private/npm_package.bzl
```diff
@@ -140,15 +140,19 @@ def npm_package(
         include_runfiles = False,
         hardlink = "auto",
         publishable = False,
+        pnpm_binary = "@pnpm//:pnpm",
```
jbedard, line 143: Does this depend on the user defining `pnpm` and doing `use_repo()` in their MODULE? Does it use the user defined pnpm (from the user MODULE) or does it use a default from rules_js?

(I forget if we have a default or how this will work if the user doesn't define it).

npm/private/npm_publish.mjs
```diff
@@ -1,9 +1,73 @@
...
+let spawnOptions = {
     stdio: 'inherit',
-})
+}
+
+if (process.env.BUILD_WORKSPACE_DIRECTORY) {
+    cleanupCwd = mkdtempSync(path.join(tmpdir(), 'rules-js-pnpm-publish-'))
+
+    // Give pnpm only the workspace-level files it needs for publish settings.
+    for (const filename of ['pnpm-workspace.yaml', '.npmrc']) {
```
jbedard, line 31: Very often large repos have the pnpm-workspace+lock in a subdir since they are large and expensive. What then? I've often seen the pnpm-workspace+lock somewhere such as `tools/{package.json,pnpm-workspace.yaml,pnpm-lock.yaml}` etc.

We might want to (maybe in the future and not this PR) look at `@npm` to determine the location of the workspace file. I think that would require `npm_translate_lock(name="npm")` to declare a target such as `@npm//:the_lockfile` which referenced the file no matter where it was.

Or we just have a `npm_package(publish_data)` attribute and users who don't put these in the root have to put data there. Lets not do it until we have to though...

npm/private/test/npm_package_publish/test.sh
```diff
@@ -2,27 +2,66 @@
...
+
+cat pub_pnpm.log | grep '"name": "@mycorp/pkg-to-publish-pnpm"'
+
+# shellcheck disable=SC2181
+if [ $? != 0 ]; then
+    echo "FAIL: expected pnpm publish dry-run output to include package name, GOT: $(cat pub_pnpm.log)"
+    exit 1
+fi
+
+# The source package.json is 132 bytes. pnpm rewrites catalog: to the concrete
```
jbedard: Can we not have a better check then this? Checking the number of bytes is very fragile and someone will just blindly update it when it changes, which makes this assertion (close to, but not completely) pointless imo

npm/private/npm_publish.mjs
jbedard: FWIW I think (and claude also claims, which increases my confidence 😅) `package.json#pnpm.catalog` was the legacy way of doing it. I think we should ignore this and keep it unsupported since it is legacy/deprecated

npm/private/npm_package.bzl
luciomartinez: Yes: `@pnpm` comes from the consuming module, not an implicit repository supplied by rules_js. The consumer needs to invoke the `pnpm` extension and import the repository with `use_repo`; the extension supplies the default repository name/version only once invoked. A differently named repository can be passed through `pnpm_binary`. I added the exact `MODULE.bazel` setup and clarified this requirement in the attribute docs.

npm/private/test/npm_package_publish/test.sh
luciomartinez: Fair. I replaced the byte-count assertion with a semantic one. The test now packs the generated package using the same Bazel-managed pnpm in a temporary catalog workspace, extracts the tarball, and checks that the packed `package.json` contains `"typescript": "5.9.3"` and no remaining `catalog:` spec. The focused test and all 94 `//npm/private/test/...` tests pass.

PR #2872, opened 2026-06-03
docs: expand and update documentation on `use_execroot_entry_point`

acozzette:
---

### DIFF
--- a/docs/use_execroot_entry_point.md
+++ b/docs/use_execroot_entry_point.md
@@ -3,8 +3,8 @@
 This page describes the `use_execroot_entry_point` option on `js_run_binary`
 and provides guidance on when to use each value. The short version is that
 `use_execroot_entry_point=True` sets up a directory layout that is more
-friendly to some JavaScript tools, but at the expense of causing problems for
-cross-platform builds.
+friendly to some JavaScript tools, but at the expense of a performance hit and
+issues with cross-platform builds.
 
 ## Background
 
@@ -82,11 +82,11 @@ behavior by default.
 
 ### Example
 
-A simple example of the recommended way to set up a `js_run_binary` target is
-[here](../examples/rspack/BUILD.bazel). This particular case *requires*
-`use_execroot_entry_point = False`, because otherwise the cross-platform build
-in that file would fail as a result of Bazel trying to use the wrong Rspack
-binary. Below is the key part, edited slightly for brevity:
+A simple demonstration of the recommended way to set up a `js_run_binary`
+target is the [Rspack example](../examples/rspack/BUILD.bazel). This particular
+case *requires* `use_execroot_entry_point = False`, because otherwise the
+cross-platform build in that file would fail as a result of Bazel trying to use
+the wrong Rspack binary. Below is the key part, edited slightly for brevity:
 
 ```
 js_library(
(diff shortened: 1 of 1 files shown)


docs/use_execroot_entry_point.md
```diff
@@ -1,37 +1,42 @@
 # use_execroot_entry_point
 
 This page describes the `use_execroot_entry_point` option on `js_run_binary`
-and provides guidance on when to use each value.
+and provides guidance on when to use each value. The short version is that
+`use_execroot_entry_point=True` sets up a directory layout that is more
+friendly to some JavaScript tools, but at the expense of causing problems for
```
jbedard: does it also have a performance hit?

"but with an impact on performance and issues on cross-platform builds" or something like that?

docs/use_execroot_entry_point.md
```diff
@@ -46,26 +51,85 @@ the build action runs from the runfiles tree, which avoids cross-platform
...
 Individual targets can still override the flag by explicitly setting
 `use_execroot_entry_point = True` or `use_execroot_entry_point = False`.
 
 In a future major version, we will likely disable the `use_execroot_entry_point`
 behavior by default.
+
+### Example
+
+A simple example of the recommended way to set up a `js_run_binary` target is
+[here](../examples/rspack/BUILD.bazel). This particular case *requires*
```
jbedard: replace `[here]` with `the [rspack example]` or something meaningful

docs/use_execroot_entry_point.md
```diff
@@ -46,26 +51,85 @@ the build action runs from the runfiles tree, which avoids cross-platform
...
+   its dependencies are built for the exec platform, which is appropriate
+   since they will run during the build action. They will land in the runfiles
+   directory adjacent to the other exec-platform sources, which will allow
+   module resolution to proceed correctly.
+- `chdir = package_name()` causes the working directory to be
+  `bazel-out/<target-cfg>/bin/examples/rspack`. This is not strictly
+   necessary, but it is convenient to have the outputs go directly in the
+   build action's current directory.
+- The config file (`rspack.config.cjs`) refers to `process.cwd()`, not
+  `__dirname`, for specifying the output path. This is key, because the config
```
jbedard, line 128: The last sentence is confusing... maybe "This is key, because the config file (and therefore `__dirname`) will be in the runfiles directory and not the output tree"

docs/use_execroot_entry_point.md
```diff
@@ -46,26 +51,85 @@ the build action runs from the runfiles tree, which avoids cross-platform
...
+  file will be in the runfiles directory and thus `__dirname` will not be
+  anywhere near the output tree.
+- We refer to `"$$RUNFILES_DIR/$(rlocationpath :rspack_config)"` in
```
jbedard, line 131: This is the part that I wish we could find a better solution for :/

jbedard, line 131: You don't explain in this paragraph what `$(rlocationpath)` does

docs/use_execroot_entry_point.md
acozzette: Ok, I added a performance hit as one of the downsides.

docs/use_execroot_entry_point.md
acozzette: Done

docs/use_execroot_entry_point.md
acozzette: Sounds good, I adopted your language here.

docs/use_execroot_entry_point.md
acozzette: I agree, I think there is room to improve this. We could potentially set up a syntax like `$(rlocation :config)` and expand that ourselves inside the implementation of `js_binary`.

acozzette: I updated this to explain that and also the reason for the double dollar signs.

jbedard: I think that might be worth it, even though I think that means more bloat in our shell wrapper script?

jbedard (approved):

docs/use_execroot_entry_point.md
acozzette: What I had in mind was that we would just replace `$(rlocation :config)` with `$$RUNFILES_DIR/$(rlocationpath :config)` entirely in Starlark, so there would be no change to the wrapper script.

I think we might eventually want to update the wrapper script to accommodate Windows builds where there is no runfiles directory, though. That will probably require having the script take a dependency on the runfiles library, and then instead of referring to `$RUNFILES_DIR` we can call `rlocation <path>` to look up the location of the config file.

acozzette: Let me resolve this conversation now so that I can submit this, but I got Claude to come up with something that looks promising here: https://github.com/aspect-build/rules_js/pull/2873 It probably needs to be cleaned up a bit but I can do that if we decide this is a good direction.

PR #2873, opened 2026-06-03
feat: add $(rlocation ...) syntax as a cleaner alternative to $$RUNFILES_DIR/$(rlocationpath ...)

acozzette:
This change also exposes the `expand_rlocation_refs` helper function in `js_lib_helpers` so that we can reuse it in `rules_webpack` and possibly other rulesets where it might be useful.

---

`js_binary` now supports the `$(rlocation ...)` syntax in `fixed_args`.

### Test plan

- New test cases added

### DIFF
--- a/examples/rspack/BUILD.bazel
+++ b/examples/rspack/BUILD.bazel
@@ -27,7 +27,7 @@ bin.rspack(
     fixed_args = [
         "build",
         "--config",
-        "$$RUNFILES_DIR/$(rlocationpath :rspack_config)",
+        "$(rlocation :rspack_config)",
     ],
     use_execroot_entry_point = False,
 )
--- a/js/private/js_helpers.bzl
+++ b/js/private/js_helpers.bzl
@@ -190,6 +190,30 @@ def gather_runfiles(
         for target in deps
     ])
 
+def expand_rlocation_refs(value):
+    """Pre-processes $(rlocation <label>) into $$RUNFILES_DIR/$(rlocationpath <label>).
+
+    After this transformation, the standard expand_locations/expand_variables chain
+    resolves the rlocationpath and converts $$ to a literal $.
+    """
+    result = []
+    remaining = value
+    prefix = "$(rlocation "
+    for _ in range(len(value)):
+        idx = remaining.find(prefix)
+        if idx == -1:
+            break
+        result.append(remaining[:idx])
+        remaining = remaining[idx + len(prefix):]
+        end = remaining.find(")")
+        if end == -1:
+            fail("Unclosed $(rlocation ...) in: " + value)
+        label_str = remaining[:end]
+        result.append("$$RUNFILES_DIR/$(rlocationpath " + label_str + ")")
+        remaining = remaining[end + 1:]
+    result.append(remaining)
+    return "".join(result)
+
 LOG_LEVELS = {
     "fatal": 1,
     "error": 2,
--- a/js/private/js_helpers.bzl
+++ b/js/private/js_helpers.bzl
@@ -205,6 +205,11 @@ def expand_rlocation_refs(value):
             break
         result.append(remaining[:idx])
         remaining = remaining[idx + len(prefix):]
+        # Parentheses are allowed in Bazel target names and could therefore
+        # theoretically appear in a label. We just look for the first closing
+        # parenthesis without handling that possibility, as this is what Bazel
+        # itself does:
+        # https://github.com/bazelbuild/bazel/blob/b4f16d988ddebc1891d7af57622ec5aff149a838/src/main/java/com/google/devtools/build/lib/analysis/LocationExpander.java#L179
         end = remaining.find(")")
         if end == -1:
             fail("Unclosed $(rlocation ...) in: " + value)
--- a/js/private/js_helpers.bzl
+++ b/js/private/js_helpers.bzl
@@ -205,6 +205,7 @@ def expand_rlocation_refs(value):
             break
         result.append(remaining[:idx])
         remaining = remaining[idx + len(prefix):]
+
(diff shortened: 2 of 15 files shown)


js/private/js_helpers.bzl
```diff
@@ -190,6 +190,30 @@ def gather_runfiles(
...
+        idx = remaining.find(prefix)
+        if idx == -1:
+            break
+        result.append(remaining[:idx])
+        remaining = remaining[idx + len(prefix):]
+        end = remaining.find(")")
+        if end == -1:
+            fail("Unclosed $(rlocation ...) in: " + value)
+        label_str = remaining[:end]
+        result.append("$$RUNFILES_DIR/$(rlocationpath " + label_str + ")")
```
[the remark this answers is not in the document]
acozzette: This is true, but not a new problem. I think we should eventually set things up to use the runfiles library so that it handles Windows builds properly.

examples/rspack/BUILD.bazel
```diff
@@ -27,7 +27,7 @@ bin.rspack(
     fixed_args = [
         "build",
         "--config",
-        "$$RUNFILES_DIR/$(rlocationpath :rspack_config)",
+        "$(rlocation :rspack_config)",
```
jbedard, line 30: So we are adding a magic `$(rlocation)` util in addition to the "standard" `$()` utils. Where/who defines those "standard" ones? Are they native bazel?

js/private/js_helpers.bzl
```diff
@@ -190,6 +190,30 @@ def gather_runfiles(
...
+    result = []
+    remaining = value
+    prefix = "$(rlocation "
+    for _ in range(len(value)):
+        idx = remaining.find(prefix)
+        if idx == -1:
+            break
+        result.append(remaining[:idx])
+        remaining = remaining[idx + len(prefix):]
+        end = remaining.find(")")
```
jbedard, line 214: Can't we have nested `$()`? That's why I don't like the idea of parsing this stuff manually :/

examples/rspack/BUILD.bazel
acozzette: Right, this is our own magic one. The standard ones are native Bazel in ctx.expand_location(): https://bazel.build/rules/lib/builtins/ctx#expand_location

js/private/js_helpers.bzl
jbedard: Why can't we just replace `$(rlocation` with `$$RUNFILES_DIR/$(rlocationpath` without ever determining the ending `)`?

jbedard:
@codex

js/private/js_helpers.bzl
acozzette: I was wondering about that too. I think this is theoretically a problem but unlikely to come up in practice. You can put parentheses in a target name, but that would be so unidiomatic that it's hard to imagine someone doing that.

Actually I just tried this with `$(location :config())` and found that it can't handle nested parentheses either. So it seems that even the builtin location functions have this problem.

acozzette: Doing the replacement without determining the ending paren is a clever idea, but I think there is a significant downside. We probably want to start using the runfiles library eventually so that we can handle Windows builds, and once we do that we will need to generate a string like `$$(rlocation $(rlocationpath :config))`. That requires adding an extra paren on the end and so we have to know where the original ending paren was.

jbedard: > Actually I just tried this with $(location :config()) and found that it can't handle nested parentheses either. So it seems that even the builtin location functions have this problem.

Can you determine why? If bazel is dumb+strict then we can be too 👍

acozzette: It seems to use roughly the same logic that I'm using, where it just looks for the first closing parenthesis without worrying about nesting. If I rename the config to `rspack_config()` in the rspack example and try to do `$(location :rspack_config())`, I get this error:
```
ERROR: /home/adam/src/rules_js/examples/rspack/BUILD.bazel:16:11: in js_binary rule //rspack:rspack_build__js_binary: label '//rspack:rspack_config(' in $(location) expression is not a declared prerequisite of this rule
```

I tried escaping the parentheses or quoting the label name a bunch of different ways, but couldn't find any way to make Bazel happy about this. So I think it just can't handle this situation of `$(location)` expansion when there's a closing parenthesis in the label.

acozzette: The relevant parsing code in Bazel appears to be [here](https://github.com/bazelbuild/bazel/blob/b4f16d988ddebc1891d7af57622ec5aff149a838/src/main/java/com/google/devtools/build/lib/analysis/LocationExpander.java#L179). That seems to confirm that it just looks for the first closing parenthesis without trying to take into account any parentheses inside the label.

jbedard: Want to reference that exact implementation in a comment to show that we are aligning with bazel conventions?

acozzette: Sure, I added a link to that in a comment here.

js/private/js_helpers.bzl
```diff
@@ -190,6 +190,36 @@ def gather_runfiles(
...
+    remaining = value
+    prefix = "$(rlocation "
+    for _ in range(len(value)):
+        idx = remaining.find(prefix)
+        if idx == -1:
+            break
+        result.append(remaining[:idx])
+        remaining = remaining[idx + len(prefix):]
+
+        # Parentheses are allowed in Bazel target names and could therefore
```
jbedard, line 209: Can they!? I thought we exclusively escaped them in rules_js because they can't....

acozzette: Yep, surprisingly they are explicitly allowed according to the documentation here: https://bazel.build/concepts/labels#target-names

PR #2879, opened 2026-06-04
fix: get `nextjs_standalone_build` working with cross-platform builds

### LINKED ISSUE #2814: nextjs_standalone_build npm deps not exec-configured, breaking cross-platform image builds with js_image_layer on macOS
---
Rules version: 3.0.3                                                                                                                                                                                             
Bazel version: 9.1.0 (Bzlmod)                                                                                                                                                                                       
Platform: macOS (darwin/amd64) building a linux/amd64 OCI image     
Note: This report was generated by an LLM; I don't fully understand the breakage but the result is a pain point in our dev flow
 
---
## Problem
When `js_image_layer` is applied to a `nextjs_standalone_server` target with a linux platform transition, the transition cascades transitively to `nextjs_standalone_build`. The `next build` action runs on the macOS exec platform but its npm inputs — including `@next/swc-*` — are now selected for the linux target platform. The linux SWC binary cannot execute on macOS, so Next.js attempts a live download via pnpm config get registry, which fails because pnpm is not in the sandboxed action PATH.
                                                                      
## Reproduction

MODULE.bazel                                                                                                                                                                                                   
```
bazel_dep(name = "rules_oci", version = "2.3.0")
  oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")                                                                                                                                                     
  oci.pull(name = "node24_slim", image = "docker.io/library/node",                                                                                                                                                 
           digest = "sha256:...", platforms = ["linux/amd64"])                                                                                                                                                     
  use_repo(oci, "node24_slim", "node24_slim_linux_amd64")
```
                                                                                                                                                                                    
 BUILD.bazel 
```                                                                                                                                                                                      
  nextjs_standalone_build(name = "standalone", ...)                                                                                                                                                                
  nextjs_standalone_server(name = "server", app = ":standalone", ...)
                                                                                                                                                                                                                   
  js_image_layer(                                                                                                                                                                                                  
      name = "server_layers",                                                                                                                                                                                      
      binary = ":server",                                                                                                                                                                                          
      root = "/app",                                        
      platform = ":linux_amd64",   # triggers the cascading transition                                                                                                                                             
  )                                                                                                                                                                                                                
                                                                                                                                                                                                                   
  bazel build //apps/web:server_layers                                                                                                                                                                             
  # Error: /bin/sh: pnpm: command not found                 
  # (next build tried to download the darwin SWC binary because linux SWC can't execute)                                                                                                                           
```

## Root cause

`nextjs_standalone_build` feeds npm packages to next build as target-configured deps. When a downstream platform transition flips the target platform to linux, those deps resolve to linux npm packages. Because `next build` is a Bazel action that executes on the exec platform (macOS), it requires exec-platform npm packages — particularly the SWC native binary. The action and its tool requirements are split across configurations.

# Expected behaviour
`js_image_layer` on macOS targeting linux/amd64 should produce a valid linux image. The next build action should always use exec-platform npm packages (specifically SWC) regardless of the downstream target platform, while the packaged runfiles in the image layers use target-platform packages.

# Suggested fix
In `nextjs_standalone_build` (or the underlying `js_run_binary/action`), mark the npm dep attributes that provide build-time tooling (SWC, webpack, etc.) with `cfg = "exec"` so they always resolve to the host platform. The build output (`.next/`, `server.js`) is platform-neutral JS and can flow through downstream target transitions unchanged.

acozzette:
I was stumped for a while about how to get Next.js working well in Bazel. Because of its pre-rendering, `next build` inherently relies on executing target-platform sources during the build. That left us with two bad options. The first is what we do now, which is to just execute target-platform sources during the build. But this blows up if the target platform is significantly different from the exec platform, since the build process involves some native binaries. The other option was to try to set things up so that we operate on target-platform sources but run only exec-platform sources. But Next.js will execute target-platform sources as part of pre-rendering, and this results in React getting resolved in more than one `node_modules` tree and crashing.

I found a third option that seems to work and is fairly simple, though. We just build for the exec platform and then copy the result verbatim to the target platform bin directory. The discussion here suggests roughly the same idea: https://github.com/vercel/next.js/discussions/93034

I thought this might result in the wrong dependencies landing in the `node_modules` directory in the output, but we actually delete that directory anyway for unrelated reasons:
https://github.com/aspect-build/rules_js/blob/48e1704c6f403c943f94d5885a2c7254ced4f5e9/contrib/nextjs/next.bazel.mjs#L34

Any NPM package dependencies are expected to be provided in a subsequent call to `nextjs_standalone_server`, and at that point they will be treated correctly as target-platform dependencies.

Fixes #2814.

---

- Breaking change (forces users to change their own code or config): yes, at least potentially

The `nextjs_standalone_build` now works correctly when the build is targeting a different OS or CPU.

### Test plan

- New test cases added

### DIFF
--- a/contrib/nextjs/defs.bzl
+++ b/contrib/nextjs/defs.bzl
@@ -42,6 +42,7 @@ For [standalone applications](https://nextjs.org/docs/app/api-reference/config/n
   [standalone directory structure guidelines](https://nextjs.org/docs/app/api-reference/config/next-config-js/output#automatically-copying-traced-files)
 """
 
+load("@bazel_lib//lib:copy_directory.bzl", "copy_directory_bin_action")
 load("@bazel_lib//lib:copy_file.bzl", "copy_file")
 load("@bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
 load("@bazel_lib//lib:directory_path.bzl", "directory_path")
@@ -434,32 +435,31 @@ def nextjs_standalone_server(name, app, pkg = None, data = [], **kwargs):
     )
 
 def _copy_exec_to_bin_impl(ctx):
-    src_files = ctx.files.src
-    if len(src_files) != 1:
-        fail("src must provide exactly one artifact; got %d" % len(src_files))
-    src = src_files[0]
-    out = ctx.actions.declare_directory(ctx.label.name)
-    ctx.actions.run_shell(
-        inputs = [src],
-        outputs = [out],
-        command = "mkdir -p {out} && cp -r {src}/. {out}/".format(
-            src = src.path,
-            out = out.path,
-        ),
-        mnemonic = "CopyExecToBin",
-        progress_message = "Copying exec-platform artifact to target bin %{label}",
+    dst = ctx.actions.declare_directory(ctx.label.name)
+    copy_directory_bin = ctx.toolchains["@bazel_lib//lib:copy_directory_toolchain_type"].copy_directory_info.bin
+    copy_directory_bin_action(
+        ctx,
+        src = ctx.file.src,
+        dst = dst,
+        copy_directory_bin = copy_directory_bin,
     )
-    return [DefaultInfo(files = depset([out]))]
+    return [
+        DefaultInfo(
+            files = depset([dst]),
+            runfiles = ctx.runfiles([dst]),
+        ),
+    ]
 
 _copy_exec_to_bin = rule(
     implementation = _copy_exec_to_bin_impl,
     attrs = {
         "src": attr.label(
             mandatory = True,
-            allow_files = True,
+            allow_single_file = True,
             cfg = "exec",
             doc = "A tree-artifact target to copy from exec-platform to target-platform bin.",
         ),
     },
+    toolchains = ["@bazel_lib//lib:copy_directory_toolchain_type"],
     doc = "Copies a tree artifact built in the exec configuration into the target-platform bin directory.",
 )
--- a/contrib/nextjs/defs.bzl
+++ b/contrib/nextjs/defs.bzl
@@ -319,9 +319,15 @@ def nextjs_standalone_build(name, config, srcs, next_js_binary, data = [], use_e
         tags = ["manual"],
     )
 
-    # `next build` of the standalone application
+    # `next build` of the standalone application.
+    # Next.js is tricky to handle in Bazel, because during its pre-rendering it
+    # executes some of the sources it is operating on. We run the risk of
+    # either trying to run code for an incompatible platform, or mixing
+    # node_modules directories from two different platforms. We work around
+    # these issues by building for the exec platform and then copying that
+    # result verbatim to the target platform bin directory.
     js_run_binary(
-        name = name,
+        name = "_%s.next_build" % name,
         tool = next_js_binary,
         env = env,
         args = ["build"],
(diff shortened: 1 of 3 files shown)


contrib/nextjs/defs.bzl
```diff
@@ -316,12 +321,19 @@ def nextjs_standalone_build(name, config, srcs, next_js_binary, data = [], use_e
...
+        testonly = testonly,
     )
 
-    # `next build` of the standalone application
+    # `next build` of the standalone application.
+    # Next.js is tricky to handle in Bazel, because during its pre-rendering it
+    # executes some of the sources it is operating on. We run the risk of
+    # either trying to run code for an incompatible platform, or mixing
+    # node_modules directories from two different platforms. We work around
+    # these issues by building for the exec platform and then copying that
```
jbedard, line 334: What does "then copying that results verbatim to the target platform" actually do? Does that carry forward things like `optionDependencies` or something?

contrib/nextjs/defs.bzl
```diff
@@ -409,3 +432,34 @@ def nextjs_standalone_server(name, app, pkg = None, data = [], **kwargs):
         visibility = ["//visibility:private"],
         tags = ["manual"],
     )
+
+def _copy_exec_to_bin_impl(ctx):
```
jbedard, line 437: Isn't there a copy rule we can use for this?

acozzette: bazel-lib has a bunch of rules for copying files around, but as far as I can tell none of them can specifically handle copying files from the exec-platform tree to the target-platform tree. I suspect I could get one of them to work if I combined it with a transition on the input, but it seemed simpler to just add my own rule.

contrib/nextjs/defs.bzl
acozzette: The output appears to exclude all NPM package dependencies due to this logic that removes the `node_modules` directory: https://github.com/aspect-build/rules_js/blob/48e1704c6f403c943f94d5885a2c7254ced4f5e9/contrib/nextjs/next.bazel.mjs#L40 So I think as a result we never end up with the wrong optional dependencies.

contrib/nextjs/defs.bzl
```diff
@@ -409,3 +432,34 @@ def nextjs_standalone_server(name, app, pkg = None, data = [], **kwargs):
...
+    src_files = ctx.files.src
+    if len(src_files) != 1:
+        fail("src must provide exactly one artifact; got %d" % len(src_files))
+    src = src_files[0]
+    out = ctx.actions.declare_directory(ctx.label.name)
+    ctx.actions.run_shell(
+        inputs = [src],
+        outputs = [out],
+        command = "mkdir -p {out} && cp -r {src}/. {out}/".format(
+            src = src.path,
```
jbedard: You should essentially never do `.path` and there should be a better way of doing this such as referring to input/output with vars in the `command`.

See the `copy_directory_bin_action` - can we use that action from within our own rule?

acozzette: Good idea, I updated it to use that and it works.

examples/nextjs/BUILD.bazel
```diff
@@ -68,6 +68,42 @@ nextjs_standalone_server(
...
+)
+
+js_image_layer(
+    name = "server_macos",
+    binary = ":server",
+    platform = ":macos",
+)
+
+build_test(
+    name = "cross_platform_build_test",
```
jbedard, line 100: This test previously failed on mac or linux?

acozzette: I believe any cross-platform build here would fail. Since I'm building on Linux, the Mac build failed for me, but I assume the reverse would fail too.

jbedard:
@codex review

jbedard (approved):

PR #2883, opened 2026-06-05
fix: overriding version of "pnpm" repo

jbedard:
The default `pnpm` repo was always taking priority and could not be overridden. See https://github.com/aspect-build/rules_js/pull/2349#discussion_r3362093839

### Test plan

- New test cases added

### DIFF
--- a/npm/private/pnpm_extension.bzl
+++ b/npm/private/pnpm_extension.bzl
@@ -1,6 +1,5 @@
 """pnpm extension logic (the extension itself is in npm/extensions.bzl)."""
 
-load("@bazel_lib//lib:lists.bzl", "unique")
 load(":pnpm_repository.bzl", "DEFAULT_PNPM_VERSION", "LATEST_PNPM_VERSION")
 load(":utils.bzl", "utils")
 load(":versions.bzl", "PNPM_VERSIONS")
@@ -27,13 +26,22 @@ def resolve_pnpm_repositories(mctx):
 
     # Collect all the module tags and associated versions/integrities/options
     integrity = {}
-    registrations = {}
+
+    # Whether npm should be bundled with pnpm, OR'd across all tags for a repo
+    # (name -> bool). Like patches, this is a per-repo setting, not per-version.
+    include_npm_by_repo = {}
 
     # Patches and patch_args are per-repo-name (NOT per-version) and only
     # accepted from the root module. Multiple tags for the same repo name
     # concatenate their patches and the last patch_args wins.
     patches_by_repo = {}
     patch_args_by_repo = {}
+
+    # Versions explicitly requested for each repo: name -> version -> is_root. Tags
+    # that carry no version (the default, e.g. the always-present registration from
+    # rules_js itself) are excluded, so they only matter as a fallback. A version
+    # requested by the root module (is_root = True) takes priority over the rest.
+    requested = {}
     for mod in mctx.modules:
         for attr in mod.tags.pnpm:
             if attr.name != DEFAULT_PNPM_REPO_NAME and not mod.is_root:
@@ -41,15 +49,23 @@ def resolve_pnpm_repositories(mctx):
                 Only the root module may override the default name for the pnpm repository.
                 This prevents conflicting registrations in the global namespace of external repos.
                 """)
-            if not registrations.get(attr.name, False):
-                registrations[attr.name] = {}
+            if attr.name not in requested:
+                requested[attr.name] = {}
+                include_npm_by_repo[attr.name] = False
+            include_npm_by_repo[attr.name] = include_npm_by_repo[attr.name] or attr.include_npm
             if mod.is_root and (getattr(attr, "patches", None) or getattr(attr, "patch_args", None)):
                 patches_by_repo.setdefault(attr.name, [])
                 patches_by_repo[attr.name].extend(getattr(attr, "patches", []) or [])
                 patch_args_by_repo[attr.name] = list(attr.patch_args)
 
-            if attr.pnpm_version_from and attr.pnpm_version and attr.pnpm_version != DEFAULT_PNPM_VERSION:
+            if attr.pnpm_version_from and attr.pnpm_version:
                 fail("Cannot specify both pnpm_version = {} and pnpm_version_from = {}".format(attr.pnpm_version, attr.pnpm_version_from))
+
+            # A tag that specifies no version (neither pnpm_version nor pnpm_version_from)
+            # carries the default version, e.g. the registration from rules_js itself.
+            # Such registrations must not influence version selection when any module
+            # explicitly requests a version.
+            is_default_registration = attr.pnpm_version == "" and attr.pnpm_version_from == None
 
             # Handle pnpm_version_from by reading package.json
             if attr.pnpm_version_from != None:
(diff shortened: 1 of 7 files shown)


jbedard:
@codex

jbedard:
@codex review

npm/private/pnpm_extension.bzl
```diff
@@ -51,6 +61,13 @@ def resolve_pnpm_repositories(mctx):
             if attr.pnpm_version_from and attr.pnpm_version and attr.pnpm_version != DEFAULT_PNPM_VERSION:
                 fail("Cannot specify both pnpm_version = {} and pnpm_version_from = {}".format(attr.pnpm_version, attr.pnpm_version_from))
 
+            # A tag that specifies no version carries the default version, e.g. the
+            # registration from rules_js itself. Such registrations must not influence
+            # version selection when any module explicitly requests a version.
+            # Note: explicitly requesting DEFAULT_PNPM_VERSION is indistinguishable
+            # from not requesting a version at all.
+            is_default_registration = attr.pnpm_version == DEFAULT_PNPM_VERSION and attr.pnpm_version_from == None
```
gzm0: TBH: I feel this is rather counter intuitive.

I'd expect the following to behave differently:

```
pnpm.pnpm(name = "pnpm") #default
pnpm.pnpm(
  name = "pnpm",
  pnpm_version = "10.33.4" # assuming this is the current default.
)
```

npm/private/pnpm_extension.bzl
```diff
@@ -97,8 +113,10 @@ def resolve_pnpm_repositories(mctx):
     notes = []
     repositories = {}
     for name, versions_map in registrations.items():
-        # Disregard repeated version numbers and convert {version:include_npm} to version[]
-        versions = unique(versions_map.keys())
+        # Versions requested by the root module take priority, then versions requested
```
gzm0, line 114: Heads-up: This might be a breaking change w.r.t. the original behavior: Based on the comment below, we'd expect a higher version registration to evict a lower version registration even if it is from the root module (whether or not this is desired is another question of course).

npm/private/pnpm_extension.bzl
```diff
@@ -97,8 +113,10 @@ def resolve_pnpm_repositories(mctx):
...
+        # by any other module, falling back to the default-version registrations when
+        # no module requested a specific version.
+        versions = root_requested[name].keys() or requested[name].keys() or versions_map.keys()
```
gzm0:
```suggestion
        versions = root_requested[name].keys() or requested[name].keys() or [DEFAULT_PNPM_VERSION]
```

IIUC should be equivalent but much easier to understand. (and if it's not equivalent I didn't understand 🤷 ).

gzm0:
FWIW: I can confirm that this PR fixes the issue we've been seeing.

npm/private/pnpm_extension.bzl
jbedard: We could make the default attribute value `""`, so we can properly distinguish between your two examples there. That sounds reasonable to me and I'll try it out...

npm/private/pnpm_extension.bzl
jbedard: I (and a robot) agree it seems equivalent, and your suggestion does seem more understandable to me

acozzette (approved):

npm/private/pnpm_extension.bzl
jbedard: That is a very good point and I'm debating what is better here. The root MODULE wins vs the highest version (of the same name) wins 🤔

gzm0: I have been thinking about this a bit and get the feeling that the overall extension / repository structure is actually more complicated then it needs to be:

In steady state, different modules might have different "invocations" of `npm_translate_lock` (I'm deliberately glossing over extension tags vs repositories vs whatever). Every invocation needs a pnpm version to work (for automated lockfile updates if they are enabled and for an enduser pnpm command to install things).

Observe that the pnpm versions used for each `npm_translate_lock` are completely independent. There is no need for "a pnpm version". From that POV, it seems it would make much more sense to "attach" the pnpm binary target(s) we currently generate in `@pnpm` to the `npm_translate_lock` extension.

This leaves us with 2 problems (I can think of):

- How to bootstrap a new package.json / pnpm-lock.yaml: Should be solvable by carefully laying out the repositories generated by the extension.
- How to ensure no two modules define the same repositories in the extension. The `npm` and the `pnpm` extension solve this problem differently right now (`npm`: no solution, it will likely just blow up, `pnpm` version merging we discuss here). The more I think about this, it is actually a bazel shortcoming: Were the repositories defined with repository rules (and not module extensions), they would be namespaced by the module defining them (but we can't do that for either extensions because we need multi repos for `npm` and facts for `pnpm`).

I hope this gives some perspective. I feel integrating `pnpm` into `npm` might be a viable option to deal with some of the (unnecessary?) complexity emerging from version selection here.

jbedard: Ideally `npm_translate_lock` never uses pnpm, especially if it is not the root MODULE.

I think `npm_translate_lock` only needs pnpm when doing:
1. `npm_translate_lock(update_pnpm_lock)`, and if it is not the root MODULE then that should probably be a noop (a module should update it's pnpm lock before publishing)
2. `npm_translate_lock(npm_package_lock|yarn_lock)` which really should only be in a root MODULE while transitioning to pure pnpm.

The other common use cases of `pnpm` would be:
1. `npm_package(publishable)`, I assume only ever run in the root MODULE
2. use of bazel-managed pnpm on the command line, I assume only ever in the root MODULE
3. by javascript code that uses pnpm for something and loads it via `rules_js` instead of `package.json` - this is the one that can cross MODULEs and I'd be most concerned about


I do think this might be over-complicated today and we really just need:
* a default pnpm for things like `npm_package(publishable)` or `npm_translate_lock(npm_package_lock|yarn_lock|update_pnpm_lock)` that you can override if you don't like the default
* custom pnpm, but it should be scoped to the MODULE that defines it and not interfere with others

The second point there (scoping to a MODULE) might be the key that would simplify everything here?

gzm0: > The second point there (scoping to a MODULE) might be the key that would simplify everything here?

Yeah, that sounds about right. It's kind of unfortunate, that the [`pnpm_repository` rule](https://github.com/aspect-build/rules_js/blob/427b524bec024eefa2b64282437dac25cf7e3788/npm/repositories.bzl#L13) essentially does this, except for facts for integrity (version extraction could probably be moved there).

jbedard: I might look into that a bit before merging this PR.

> It's kind of unfortunate, that the [pnpm_repository rule](https://github.com/aspect-build/rules_js/blob/427b524bec024eefa2b64282437dac25cf7e3788/npm/repositories.bzl#L13) essentially does this,

What do you mean? It sounds like you have a suggestion or idea for a fix...

gzm0: Ah, sorry, that was a bit brief indeed. The idea is that instead of using a module extension for pnpm, use a repository rule. Basically what rules_img does for pulling:

https://github.com/bazel-contrib/rules_img#1-pull-a-base-image

So in a user `MODULE.bazel` for a custom pnpm repo, instead of:

```bzl
pnpm = use_extension("@aspect_rules_js//npm:extensions.bzl", "pnpm")
pnpm.pnpm(name = "my_repo")
use_repo(pnpm, "pnpm")
```

Do

```bzl
pnpm = use_repo_rule("@aspect_rules_js//npm:repositories.bzl", "pnpm")
pnpm(name = "my_repo")
```

The key difference is that "my_repo" is only visible to the declaring module, so no deconflicting needs to be performed.

The default pnpm can be defined in this way in the `MODULE.bazel` of rules_js and exposed to users via an alias (say as `@aspect_rules_js//npm:pnpm`).

The big problem with this approach is that we lose the ability to use facts.

jbedard: We could potentially use that solution only for the builtin/default pnpm, but that doesn't seem to actually reduce the size of this PR so doesn't seem worth it atm :/

Issue #2267, opened 2025-07-09
js_image_layer fails on windows


peakschris:
js_image_layer is failing on windows due to a mixture of forward and backslashes in the mtree file

Example:
```
npm_link_all_packages(name = "node_modules")

js_library(
    name = "mgr-lib",
    srcs = glob([
        "lib/**/*.js",
    ]),
    data = [
        ":node_modules/colors",
    ]
)

js_binary(
    name = "mgr",
    entry_point = "x",
    data = [
        ":mgr-lib",
    ],
    patch_node_fs = False,
    preserve_symlinks_main = False,
)

js_image_layer(
    name = "layers",
    binary = ":mgr",
    root = "/app",
)
```

Results in:
```
>bazel build ... --verbose_failures
ERROR: D:/workdir/redacted/src/mgr/BUILD.bazel:126:15: JsImageLayer package_store_3p //src/mgr:layers failed: (Exit 1): tar.exe failed: error executing JsImageLayer command (from target //src/mgr:layers)
  cd /d D:/redacted/b/buusiixg/execroot/_main
  SET LC_ALL=C.UTF-8
  external\aspect_bazel_lib~~toolchains~bsd_tar_windows_amd64\tar.exe --create --file bazel-out/x64_windows-opt/bin/src/mgr/layers_package_store_3p.tar.gz --gzip --options=gzip:!timestamp @bazel-out/x64_windows-opt/bin/src/mgr/layers_package_store_3p.mtree
tar.exe: Error reading archive bazel-out/x64_windows-opt/bin/src/node-mgr/layers_package_store_3p.mtree: Can't open bazel-out/x64_windows-opt/bin/node_modules/.aspect_rules_js/@colors+colors@1.6.0/node_modules/@colors/colors/examples
ormal-usage.js
tar.exe: Error exit delayed from previous errors.
```

Mtree:
```
#mtree
./app uid=0 gid=0 time=0 mode=0755 type=dir
./app/src/mgr/node-mgr.runfiles/_main/node_modules/.aspect_rules_js/@colors+colors@1.6.0/node_modules/@colors/colors/LICENSE uid=0 gid=0 time=0 mode=0555 type=file content=bazel-out/x64_windows-opt/bin/node_modules/.aspect_rules_js/@colors+colors@1.6.0/node_modules/@colors/colors/LICENSE
[...]

PR #2884, opened 2026-06-09
Fix js_binary and js_image_layer on windows

peakschris:
Changes:

Fix js_binary to run on windows without bash; works both in js_image_layer and via bazel
Fix js_image_layer to run on windows without bash
This means windows oci images do not need to package bash
Disable failing tests due to windows or --noenable_runfiles

Fixes:
- https://github.com/aspect-build/rules_js/issues/2267


peakschris:
@alexeagle could you ask someone to help me get this one merged? Also one in rules_jest. Would really help us if we didn't need to keep our forks up to date.

PR #2886, opened 2026-06-11
test: enable runfiles tree on Windows for e2e/bzlmod

jbedard:
### Test plan


jbedard:
See BCR: https://buildkite.com/bazel/bcr-presubmit/builds/35899#019eb7dc-5836-449c-b0b1-12115a872222

acozzette:
Should we be adding support for the runfiles library to the launcher script so that the Windows tests can run normally without this flag? I have a change in progress that does that. I set it aside because it was built on top of the `$(rlocation)` idea that we are probably not going to go with. But I could fix it to work without that.

acozzette (approved):

PR #2890, opened 2026-06-15
fix(nextjs): add missing dependencies in nextjs_standalone_server

acozzette:
Adds a server_test to examples/nextjs that starts the standalone server binary and polls for an HTTP response, catching potential regressions where the build succeeds but the server fails to run.

This turned up some issues where we had some missing dependencies. These would not necessarily be easy to catch with manual testing, since "bazel run" does not use a sandbox and can therefore succeed even when important dependencies are absent. The most important thing was making sure the `js_binary` takes a dependency on the actual sources produced by `next build`.

---

Fix handling of dependencies in nextjs_standalone_server()

### Test plan

- New test cases added

### DIFF
--- a/contrib/nextjs/defs.bzl
+++ b/contrib/nextjs/defs.bzl
@@ -405,15 +405,6 @@ def nextjs_standalone_server(name, app, pkg = None, data = [], **kwargs):
     # directory name, so they must be preserved.
     app_dir = app.split(":")[-1] if ":" in app else app.split("/")[-1]
 
-    # The standalone server binary
-    js_binary(
-        name = name,
-        entry_point = ":_{}.js".format(name),
-        chdir = native.package_name(),
-        data = data,
-        **kwargs
-    )
-
     # The server entry point into the standalone directory.
     directory_path(
         name = "_{}.js".format(name),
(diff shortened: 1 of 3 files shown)


contrib/nextjs/defs.bzl
```diff
@@ -447,6 +438,15 @@ def nextjs_standalone_server(name, app, pkg = None, data = [], **kwargs):
         tags = ["manual"],
     )
 
+    # The standalone server binary
```
jbedard, line 441: This looks like more then a test. Should this PR be titled "fix"?

acozzette: That sounds fair, I updated the PR to be a fix.

examples/nextjs/server_test.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
jbedard, line 1: What actually runs this file? Normally I think `e2e/*/test.sh` gets invoked automatically but idk about `exmaples/`...

acozzette: I wrapped it in an `sh_test` so that we can run the test through Bazel.

jbedard: Oh right, the file I already reviewed 👍 😅

jbedard (approved):

PR #2896, opened 2026-06-25
fix(npm): correct package_store_prefix_len for links in external-repo sub-packages

ma-oli:
`_symlink_package_store` strips `package_store_prefix_len` characters from a dependency's `package_store_directory.short_path` to form the relative symlink target. The length omitted the "/" separators that follow the package path and the repo name, so for a link target in a non-root package of an external repository it stripped two characters too few. The leftover was the trailing "s/" of ".aspect_rules_js/", producing dangling symlinks such as `../../s/<pkg>@<ver>/node_modules/<pkg>` and `ERR_MODULE_NOT_FOUND` at runtime.

In a root module the miscount is only one character and is masked by "//" path collapsing, so it surfaces only when a module's npm packages live in a sub-package (e.g. //frontend) and that module is consumed as an external dependency.

Account for both separators: +1 after the package path and +1 after the repo name.

---

### Test plan

- New test cases added

### DIFF
--- a/npm/private/npm_package_store.bzl
+++ b/npm/private/npm_package_store.bzl
@@ -189,9 +189,9 @@ def _npm_package_store_impl(ctx):
 
     package_store_prefix_len = _PACKAGE_STORE_PREFIX_LEN
     if ctx.label.package:
-        package_store_prefix_len += len(ctx.label.package)
+        package_store_prefix_len += len(ctx.label.package) + 1  # +1 for the "/" after the package path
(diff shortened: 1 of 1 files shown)


jbedard:
Can you add a test reproducing the issue? I would think such an error would have been caught by now so there must be something unique going on here?

jbedard:
Will be done in https://github.com/aspect-build/rules_js/pull/2864

PR #2900, opened 2026-07-03
fix: resolve npm package store tar toolchain as a Label

keegancsmith:
The --incompatible_auto_exec_groups flag requires Starlark actions to declare whether their executable/tools come from an action toolchain. This is not a Bazel default today, but it is exposed through strict preset configurations and is intended to catch rules that rely on implicit action toolchain inference.

npm_package_store already declares the tar toolchain, but the extraction action passes that action toolchain as a string. That string is resolved in the consuming repository's mapping, which can fail when @tar.bzl is not visible as an apparent repository name. Use a pre-resolved Label so the action's toolchain behavior is explicit without changing its inputs or execution model.

### Test plan

- Manual testing; please provide instructions so we can reproduce: Using this patch on rules_js in our repository with --incompatible_auto_exec_groups everything starts to work.

### DIFF
--- a/e2e/repo_mapping/.bazelrc
+++ b/e2e/repo_mapping/.bazelrc
@@ -1,3 +1,5 @@
 import %workspace%/../../tools/preset.bazelrc
 import %workspace%/../e2e.bazelrc
 
+# Exercise action toolchain resolution when apparent repository names differ.
+common --incompatible_auto_exec_groups
--- a/npm/private/npm_package_store.bzl
+++ b/npm/private/npm_package_store.bzl
@@ -260,7 +260,7 @@ def _npm_package_store_impl(ctx):
                     execution_requirements = {
                         "supports-path-mapping": "1",
                     },
-                    toolchain = tar_lib.toolchain_type,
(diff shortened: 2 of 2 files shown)


jbedard:
Can we maybe add an e2e test with a .bazelrc adding the `--incompatible_auto_exec_groups`? Then enable that test only on the bazel versions that support it?

keegancsmith:
I added `--incompatible_auto_exec_groups` to the existing `e2e/repo_mapping` test workspace rather than adding a new e2e case.

That workspace already renames `aspect_rules_js` via `repo_name`, so it exercises the same apparent-repository-name failure mode that caused `@tar.bzl` to resolve incorrectly. Its `:node_modules_test` also goes through `npm_package_store`, so the regression is covered directly.

I checked the Bazel versions in our current CI matrix and the flag exists on Bazel 7.x, 8.x, and 9.x, so I didn’t add skip tags/version gating. Verified locally with:

`USE_BAZEL_VERSION=7.x bazel test //:node_modules_test`

from `e2e/repo_mapping`.

jbedard (approved):

PR #2904, opened 2026-07-08
fix: node coverage

### LINKED ISSUE #2901: [Bug]: `bazel coverage` reports 0% for `jest_test` on `rules_js` 3.x (works on 1.x)
### What happened?

`bazel coverage` on a `jest_test` gives 0% and an empty `coverage.dat`, even though the test passes and runs the code.

```
$ bazel coverage //:cov --test_output=all
 PASS  foo.test.js
All files | 0 | 0 | 0 | 0 |
```

plain js_test V8 coverage also reports LH:0/LF:0 on bazel 8 + rules_js 3.2.2; not jest-specific

jest scopes coverage to `rootDir` (the test's runfiles dir), but node resolves the runfiles symlink to the `bazel out/bin` realpath on `require`, so the instrumented path is outside `rootDir` and gets dropped:

```
rootDir:  .../tests.runfiles/_main/<pkg>
resolved: .../bazel-out/.../bin/<pkg>/foo.js
```

`node_options = ["--preserve-symlinks"]` fixes attribution but breaks the `.aspect_rules_js` store (needs realpath): `Cannot find module 'resolve-cwd'`. So first-party code wants symlinks kept and the store wants them resolved.

### Expected

`foo.js` shows as covered, with `SF:`/`LH:`/`LF:` records in `coverage.dat` - like `bazel-examples/jest` does on rules_js 1.x.

### Version

- macOS (darwin arm64)
- `bazel 8.7.0`
- `aspect_rules_js` 3.2.2, `aspect_rules_jest` 0.25.2, `aspect_rules_ts` 3.8.11
- node 20.18.0, TypeScript + Jest
- `coverage --experimental_split_coverage_postprocessing --experimental_fetch_all_coverage_outputs`

### How to reproduce

```shell
load("@aspect_rules_jest//jest:defs.bzl", "jest_test")

js_library(name = "foo", srcs = ["foo.js"])
js_library(name = "foo_test_lib", srcs = ["foo.test.js"], deps = [":foo"])

jest_test(
    name = "cov",
    config = "jest.config.js",   # module.exports = { testEnvironment: "node" }
    data = [":foo_test_lib"],
    node_modules = "//:node_modules",
)



// foo.js
module.exports = { foobar: () => "foobar" };
// foo.test.js
test("foobar", () => expect(require("./foo.js").foobar()).toBe("foobar"));


`bazel coverage //:cov` → 0%, empty `coverage.dat`.
```

### Any other information?

`bazel-examples/jest` reports 100% on Bazel 7.4.1 / rules_js 1.31 / rules_jest 0.19.5; same recipe reports 0% on 8.7 / 3.2.2 / 0.25.2. Plain `js_library` reproduces it (not ts_project-specific); babel and v8 providers both 0%; the split flags change output plumbing but not the 0%. Related: bazel-contrib/rules_nodejs#3564.

Is `bazel coverage` supposed to work with `jest_test` on 3.x? If so, what's the supported config given the store's realpath needs vs jest's `rootDir` scoping?

mitchspano:
Fixes #2901: `bazel coverage` reported 0% (empty `coverage.dat`) for JS tests whenever coverage postprocessing ran as its own action (`--experimental_split_coverage_postprocessing`, which is typical with remote execution).

The V8 -> lcov conversion ran in the `_lcov_merger` action, whose runfiles contain only the merger and node, not the instrumented sources, so c8 emitted `LF:0` for every file. The conversion now runs at the end of the test action, where the V8 data and sources coexist; the report is stashed in `$COVERAGE_DIR` and the merger simply publishes it, falling back to the old behavior when no stash exists. Adding the sources to the merger action was considered instead, but the V8 URLs still point at the test's sandbox paths and would need fragile rewriting.

---

- Searched for relevant documentation and updated as needed: no docs cover the coverage merger; none needed

> fix: `bazel coverage` no longer reports 0% for JS tests when coverage postprocessing runs as a separate action (#2901)

### Test plan

- New test cases added: `e2e/coverage/test.sh` runs `bazel coverage` in both the default and split-postprocessing modes and asserts the lcov report attributes hits and misses to `lib.js`. Fails against pre-fix sources (`coverage.dat is missing or empty`), passes with the fix.

Disclaimer: This is my first contribution on this project. I've read the code of conduct and the contributing guidelines, but if I made a mistake in the creation of this PR, please let me know! I'm happy to change the target branch or anything else that is required.

### DIFF
--- a/e2e/pnpm_repo_install/pnpm_patch_test.sh
+++ b/e2e/pnpm_repo_install/pnpm_patch_test.sh
@@ -3,14 +3,21 @@ set -o errexit -o nounset -o pipefail
 
 # --- begin runfiles.bash initialization v3 ---
 # Copy-pasted from the Bazel Bash runfiles library v3.
-set -uo pipefail; set +e; f=bazel_tools/tools/bash/runfiles/runfiles.bash
+set -uo pipefail
+set +e
+f=bazel_tools/tools/bash/runfiles/runfiles.bash
 # shellcheck disable=SC1090
-source "${RUNFILES_DIR:-/dev/null}/$f" 2>/dev/null || \
-  source "$(grep -sm1 "^$f " "${RUNFILES_MANIFEST_FILE:-/dev/null}" | cut -f2- -d' ')" 2>/dev/null || \
-  source "$0.runfiles/$f" 2>/dev/null || \
-  source "$(grep -sm1 "^$f " "$0.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
-  source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
-  { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
+source "${RUNFILES_DIR:-/dev/null}/$f" 2>/dev/null ||
+    source "$(grep -sm1 "^$f " "${RUNFILES_MANIFEST_FILE:-/dev/null}" | cut -f2- -d' ')" 2>/dev/null ||
+    source "$0.runfiles/$f" 2>/dev/null ||
+    source "$(grep -sm1 "^$f " "$0.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null ||
+    source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null ||
+    {
+        echo >&2 "ERROR: cannot find $f"
+        exit 1
+    }
+f=
+set -e
 # --- end runfiles.bash initialization v3 ---
 
 HELLO=$(rlocation "pnpm_patched/package/hello.txt")
--- a/js/private/js_binary.sh.tpl
+++ b/js/private/js_binary.sh.tpl
@@ -482,6 +482,23 @@ RESULT="$?"
 set -e
 
 # ==============================================================================
+# Generate the lcov coverage report here in the test action, where the V8
+# coverage data (NODE_V8_COVERAGE) and the instrumented sources both exist and
+# their paths match. The _lcov_merger runs in a separate action that does not
+# have the sources, so running the c8 conversion there yields 0% (see #2901).
+# The merger early-exits when COVERAGE_OUTPUT_FILE is already non-empty, keeping
+# the report produced here.
+# ==============================================================================
+if [ "${COVERAGE_DIR:-}" ] && [ "${COVERAGE_OUTPUT_FILE:-}" ]; then
+    js_coverage_report=$(find "$JS_BINARY__RUNFILES" -path '*/js/private/coverage/coverage.js' 2>/dev/null | head -1)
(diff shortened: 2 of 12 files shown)


e2e/pnpm_repo_install/pnpm_patch_test.sh
```diff
@@ -3,14 +3,21 @@ set -o errexit -o nounset -o pipefail
 
 # --- begin runfiles.bash initialization v3 ---
 # Copy-pasted from the Bazel Bash runfiles library v3.
-set -uo pipefail; set +e; f=bazel_tools/tools/bash/runfiles/runfiles.bash
```
jbedard, line 6: Did this need to change?

mitchspano: the aspect format (`shfmt`) applied this automatically. It's purely cosmetic. We can revert , but it will resurface the next time someone touches the file.

js/private/js_binary.sh.tpl
```diff
@@ -481,6 +481,23 @@ wait "$child"
...
+# ==============================================================================
+# Generate the lcov coverage report here in the test action, where the V8
+# coverage data (NODE_V8_COVERAGE) and the instrumented sources both exist and
+# their paths match. The _lcov_merger runs in a separate action that does not
+# have the sources, so running the c8 conversion there yields 0% (see #2901).
+# The merger early-exits when COVERAGE_OUTPUT_FILE is already non-empty, keeping
+# the report produced here.
+# ==============================================================================
+if [ "${COVERAGE_DIR:-}" ] && [ "${COVERAGE_OUTPUT_FILE:-}" ]; then
+    js_coverage_report=$(find "$JS_BINARY__RUNFILES" -path '*/js/private/coverage/coverage.js' 2>/dev/null | head -1)
```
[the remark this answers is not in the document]
jbedard: Another robots suggestion for this:
> The clean fix: js_binary.bzl already knows the coverage entry point label and merges _lcov_merger runfiles (js_binary.bzl:547-548). Plumb the runfiles path in as a template substitution / env var populated only when hasattr(ctx.attr, "_lcov_merger") — i.e., only for js_test. That's O(1), deterministic, and automatically excludes nested js_binary launchers since their scripts are expanded per-target at analysis time.

js/private/js_binary.sh.tpl
```diff
@@ -481,6 +481,23 @@ wait "$child"
 RESULT="$?"
 set -e
 
+# ==============================================================================
+# Generate the lcov coverage report here in the test action, where the V8
+# coverage data (NODE_V8_COVERAGE) and the instrumented sources both exist and
+# their paths match. The _lcov_merger runs in a separate action that does not
```
jbedard, line 487: robot says:
> The fix only works because js_binary.bzl:538-548 merges _lcov_merger runfiles into the test's runfiles, and that block is annotated "TODO: Remove once bazel<8 support is dropped." After this PR, removing it would silently resurrect the 0% bug (the find misses → fallback → merger without sources). The comment needs updating so nobody executes that TODO.

Which makes me think this might not be the right approach for this...?

PR #2908, opened 2026-07-09
perf: use a dict for linked package store directory membership checks

jbedard:
The 'not in' check against a list is O(direct deps) for each entry in the transitive npm_package_store_infos of the package, making the linking loop quadratic for packages with many deps.

### Test plan

### DIFF
--- a/npm/private/npm_package_store.bzl
+++ b/npm/private/npm_package_store.bzl
@@ -282,7 +282,8 @@ def _npm_package_store_impl(ctx):
                     verbose = ctx.attr.verbose,
                 )
 
-        linked_package_store_directories = []
+        # dict of linked package store directories for fast membership checks below
+        linked_package_store_directories = {}
         for dep, _dep_aliases in ctx.attr.deps.items():
             dep_info = dep[NpmPackageStoreInfo]
             dep_aliases = _dep_aliases.split(",") if _dep_aliases else [dep_info.package]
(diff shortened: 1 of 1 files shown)


dzbarsky (approved):
Did you actually measure a win? I feel like linear scan is usually faster than dict for up to 10ish elements so this might be a loss in aggregate

npm/private/npm_package_store.bzl
```diff
@@ -292,7 +293,7 @@ deps of npm_package_store must be in the same package.""" % (ctx.label.package,
                 fail(msg)
 
             if dep_package_store_directory:
-                linked_package_store_directories.append(dep_package_store_directory)
+                linked_package_store_directories[dep_package_store_directory] = None
```
acozzette: Maybe not important but I thought the usual convention was to use a value of `True` when using a dictionary as a set?

acozzette (approved):

npm/private/npm_package_store.bzl
jbedard: I'd agree with that actually 👍 

I think it makes no difference when doing `x in linked_package_store_directories` though...

jbedard:
> Did you actually measure a win? I feel like linear scan is usually faster than dict for up to 10ish elements so this might be a loss in aggregate

I can investigate this specific one more and try both.

jbedard:
>I feel like linear scan is usually faster than dict for up to 10ish elements so this might be a loss in aggregate

A robot says:
> the dict is win-or-tie everywhere — membership at D=3 (the linear scan's best case) is a tie, and inserts favor the dict ~2x because list.append() is a method call while d[k] = None isn't — so there's no size at which the list wins in aggregate.

npm/private/npm_package_store.bzl
jbedard: Updated to `True` in 292a834d (and rebased onto main).

PR #2910, opened 2026-07-09
perf: reduce per-action overhead registering NpmPackageExtract

jbedard:
Build the tar argument list with a single add_all call, hoist the constant execution_requirements dict, and skip the --exclude pass when exclude_package_contents is empty. The rendered command line is unchanged.

### Test plan


dzbarsky (approved):
curious how you measured an improvement?

acozzette (approved):

jbedard:
A robot measured it somehow. In this case I'll blindly accept that since it's literally converting 7 invocations and 1 array creation into 1-2 invocations and 1 array creation. I think that's a win no matter what? But any changes you think are questionable please flag them and I'll analyze them individually...

These were all robot suggestions after analyzing `build --nobuild --starlark_cpu_profile` on a large repo, FWIW

PR #2916, opened 2026-07-13
fix(node-patches): resolve readlink target against the real parent dir

### LINKED ISSUE #2914: [Bug]: node-patches readlink/readlinkSync computes wrong absolute path (joins target onto unresolved parent dir)
### What happened?

The Node `fs` monkey-patches in `js/private/node-patches/fs.cjs` compute the wrong absolute path for a symlink target in `readlink` and `readlinkSync`. They join the raw kernel link target onto the **lexical** parent directory of the *queried* path:

```js
// js/private/node-patches/fs.cjs — readlinkSync (and the async readlink callback)
const resolved = resolvePathLike(args[0]);
const str = path.resolve(path.dirname(resolved), origReadlinkSync(...args));
```

`path.dirname(resolved)` is the parent of the path *as the caller wrote it*, which may itself traverse other symlinks. When the caller reaches the link through symlinks whose resolution changes the directory depth, the relative (`../…`) link target is counted from the wrong base and the resulting absolute path lands one directory level too high. Because that wrong path is still *inside* the guard root, no escape is detected and the bogus path is handed back to the caller.

Any tool that resolves modules via `readlink` (here: eslint-plugin-import's `import/namespace` rule through `eslint-import-resolver-typescript` → `enhanced-resolve`) then `stat`s a path that doesn't exist and fails with `ENOENT`.

Concretely, resolving a first-party package reached through an aliased package-store path that is one path-segment shallower than the real package directory produced an absolute path missing an intermediate segment, e.g.:

```
Error: ENOENT: no such file or directory, stat
  '.../bin/<pkg>/lint_/lint.runfiles/node_modules/.aspect_rules_js/@myorg+web@0.0.0/node_modules/@myorg/base/dist/esm/subpath/index.js'
Occurred while linting .../src/foo.ts:1
Rule: "import/namespace"
```

**Expected:** `readlink`/`readlinkSync` should return a target resolved against the *real* location of the link's parent directory, so the absolute path is correct regardless of how many symlinks the caller path traversed.

### Version

Development (host) and target OS/architectures: Linux x86_64 (host + target)

Output of `bazel --version`: `bazel 7.7.1-1`

Version of the Aspect rules: `rules_js` v2.8.2 (via a vendored archive). **The buggy code is unchanged on `main` / v3.2.2** — `readlink`/`readlinkSync` in `fs.cjs` still use `path.resolve(path.dirname(resolved), …)` with the unresolved parent — so this is not fixed in the latest release.

Language(s) and/or frameworks involved: TypeScript, Node.js v22, ESLint (`eslint-plugin-import` `import/namespace` + `eslint-import-resolver-typescript`), pnpm workspaces, first-party `js_library` packages linked via `npm_link_all_packages`.

### How to reproduce

```shell
# Shape of the repro (in a rules_js pnpm-workspace monorepo):
# - first-party package @myorg/web (js_library) that re-exports from @myorg/base
# - @myorg/base has a package.json "types"/"exports" that points into ./dist
#   (so a resolver follows the dist entry, which re-exports a subpath)
# - an eslint_test on a consumer that imports @myorg/web
#
# The consumer resolves @myorg/base through the package-store alias
#   node_modules/.aspect_rules_js/@myorg+web@0.0.0/node_modules/@myorg/base
# which is reached via symlinks of a different depth than the real
# bazel-out package dir, so readlink's lexical-parent join is off by one level.

bazel test //path/to/consumer:lint   # eslint_test → ENOENT under import/namespace

# Minimal isolation (no bazel): load js/private/node-patches/register.cjs with
# JS_BINARY__FS_PATCH_ROOTS set, then fs.readlinkSync() a symlink reached through
# a parent symlink of different depth — the returned absolute path is wrong.
# Realpath-ing the parent before joining fixes it.
```

### Any other information?

Root cause is purely the lexical-vs-real parent in the two `readlink` implementations. Suggested fix (identity transform whenever the caller path is already fully resolved, so it only changes results that were previously wrong):

```js
let linkDir = path.dirname(resolved);
try { linkDir = origRealpathSyncNative(linkDir); } catch (_) {}
const str = path.resolve(linkDir, origReadlinkSync(...args));
```

Apply the same to the async `readlink` callback. Cost is one guarded `realpath` syscall per `readlink`.

mloeppky-zoox:
Fixes #2914

## Problem

`fs.readlink` / `fs.readlinkSync` in the node-patches compute the target's absolute path by joining the raw kernel link target onto the **lexical** parent of the *queried* path:

```js
const targetAbs = path.resolve(path.dirname(resolved), linkTarget)
```

`path.dirname(resolved)` is the parent of the path *as the caller wrote it*, which may itself traverse symlinks. When the caller reaches the link through symlinks whose resolution changes directory depth (e.g. an aliased `node_modules` path that is a different number of segments deep than the link's real location), a relative (`../…`) target is counted from the wrong base. The resulting absolute path lands one or more levels off — but still *inside* the guard root, so `isEscape` doesn't flag it and the bogus path is handed to the caller.

Any tool that resolves modules via `readlink` then stats a nonexistent path and fails with `ENOENT`. We hit this with ESLint's `import/namespace` rule (via `eslint-import-resolver-typescript` → `enhanced-resolve`) resolving a first-party package reached through a package-store alias one segment shallower than its real `bazel-out` dir.

## Fix

Resolve the link's parent to its **real** path before joining the target (`origRealpathSyncNative`, with a try/catch fallback to the previous behavior). This is an identity transform whenever the queried path is already fully resolved, so it only changes results that were previously wrong. Applied to both the async `readlink` callback and `readlinkSync`; regenerated the checked-in `fs.cjs` from `src/fs.cts` via the `write_source_files` target.

## Testing

- Existing `//js/private/test/node-patches:readlink_*` targets pass on node 20/22/24 (cjs + esm) — no regression.
- Verified end-to-end in a downstream monorepo: the fix resolves an `ENOENT` in an `eslint_test` that was resolving `@scope/base` through `@scope/web`'s store copy.

Isolated repro (no Bazel): load `js/private/node-patches/register.cjs` with `JS_BINARY__FS_PATCH_ROOTS` set, then `fs.readlinkSync()` a symlink reached through a parent symlink of a different depth — the returned absolute path is wrong before this change, correct after.

Happy to add a dedicated regression test to `readlink.mjs` if you'd like to guide the fixture shape (the observable difference is in the escape/`nextHop` mapping, so it needs a sandbox-style fixture with a depth-mismatched alias).

Co-Authored-By: Claude Code

### DIFF
--- a/js/private/node-patches/src/fs.cts
+++ b/js/private/node-patches/src/fs.cts
@@ -111,6 +111,10 @@ export function patcher(roots: string[]): () => void {
         .map((root) => {
             const lexical = path.resolve(root)
             let real = lexical
+            // A configured root need not exist on disk when the patch is
+            // registered (or may be inaccessible); when it can't be realpath'd
+            // we keep real == lexical, i.e. treat the root as not-symlinked so
+            // the remapping below is a no-op.
             try {
                 real = origRealpathSyncNative(lexical) as string
             } catch {}
--- a/js/private/node-patches/src/fs.cts
+++ b/js/private/node-patches/src/fs.cts
@@ -107,6 +107,34 @@ export function patcher(roots: string[]): () => void {
         .native as typeof FsType.realpathSync.native
 
     const { canEscape, isEscape } = escapeFunction(roots)
+    const rootMappings = roots
+        .map((root) => {
+            const lexical = path.resolve(root)
+            let real = lexical
+            try {
+                real = origRealpathSyncNative(lexical) as string
+            } catch {}
+            return { lexical, real }
+        })
+        .sort((a, b) => b.lexical.length - a.lexical.length)
+
+    // Keep real paths in the lexical namespace when the configured root is
+    // itself a symlink. Nested links that leave the root remain canonical.
+    function realpathInRootNamespace(p: string): string {
+        const real = origRealpathSyncNative(p) as string
+        for (const root of rootMappings) {
+            if (
+                isSubPath(root.lexical, p) &&
+                isSubPath(root.real, real)
+            ) {
+                return path.resolve(
+                    root.lexical,
+                    path.relative(root.real, real)
+                )
+            }
+        }
+        return real
+    }
 
     // =========================================================================
     // fs.lstat
@@ -310,7 +338,7 @@ export function patcher(roots: string[]): () => void {
             // cannot be realpath'd (identity transform when already resolved).
             let linkDir = path.dirname(resolved)
             try {
-                linkDir = origRealpathSyncNative(linkDir) as string
+                linkDir = realpathInRootNamespace(linkDir)
             } catch {}
             const targetAbs = path.resolve(linkDir, linkTarget)
             const escapedRoot: string | false = isEscape(resolved, targetAbs)
@@ -359,7 +387,7 @@ export function patcher(roots: string[]): () => void {
         // directory (see fs.readlink above for the rationale).
         let linkDir = path.dirname(resolved)
         try {
-            linkDir = origRealpathSyncNative(linkDir) as string
+            linkDir = realpathInRootNamespace(linkDir)
         } catch {}
         const targetAbs = path.resolve(linkDir, linkTarget)
 
--- a/js/private/node-patches/src/fs.cts
+++ b/js/private/node-patches/src/fs.cts
@@ -302,7 +302,17 @@ export function patcher(roots: string[]): () => void {
             if (err) return cb(err)
             const resolved = resolvePathLike(args[0])
             const linkTarget = p!
-            const targetAbs = path.resolve(path.dirname(resolved), linkTarget)
+            // Resolve the link target against the REAL location of the link's
+            // parent directory. `resolved` may itself traverse symlinks (e.g. an
+            // aliased node_modules path of a different depth than the real dir);
+            // joining a relative target onto the lexical parent then lands at the
+            // wrong absolute path. Falls back to the old behavior if the parent
+            // cannot be realpath'd (identity transform when already resolved).
+            let linkDir = path.dirname(resolved)
+            try {
+                linkDir = origRealpathSyncNative(linkDir) as string
+            } catch {}
+            const targetAbs = path.resolve(linkDir, linkTarget)
             const escapedRoot: string | false = isEscape(resolved, targetAbs)
             if (escapedRoot) {
                 const escapedRoots = [escapedRoot]
@@ -345,7 +355,13 @@ export function patcher(roots: string[]): () => void {
     ) {
         const resolved = resolvePathLike(args[0])
         const linkTarget = origReadlinkSync(...args)
-        const targetAbs = path.resolve(path.dirname(resolved), linkTarget)
+        // Resolve the link target against the REAL location of the link's parent
+        // directory (see fs.readlink above for the rationale).
+        let linkDir = path.dirname(resolved)
+        try {
+            linkDir = origRealpathSyncNative(linkDir) as string
+        } catch {}
+        const targetAbs = path.resolve(linkDir, linkTarget)
 
         const escapedRoot: string | false = isEscape(resolved, targetAbs)
         if (escapedRoot) {
--- a/js/private/node-patches/src/fs.cts
+++ b/js/private/node-patches/src/fs.cts
@@ -136,6 +136,20 @@ export function patcher(roots: string[]): () => void {
         return real
     }
 
+    // Resolve a relative symlink `linkTarget` against the REAL location of the
+    // link's parent directory. `resolved` is an absolute path whose final
+    // component is a symlink we just read successfully, but the path we were
+    // queried through may be an alias of a different depth than the real dir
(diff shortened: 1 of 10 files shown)


jbedard:
Is there any way to add a test that failed without the fix?

mloeppky-zoox:
> Is there any way to add a test that failed without the fix?

Done

js/private/node-patches/src/fs.cts
```diff
@@ -302,7 +302,17 @@ export function patcher(roots: string[]): () => void {
...
-            const targetAbs = path.resolve(path.dirname(resolved), linkTarget)
+            // Resolve the link target against the REAL location of the link's
+            // parent directory. `resolved` may itself traverse symlinks (e.g. an
+            // aliased node_modules path of a different depth than the real dir);
+            // joining a relative target onto the lexical parent then lands at the
+            // wrong absolute path. Falls back to the old behavior if the parent
+            // cannot be realpath'd (identity transform when already resolved).
+            let linkDir = path.dirname(resolved)
+            try {
+                linkDir = origRealpathSyncNative(linkDir) as string
```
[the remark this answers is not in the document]
mloeppky-zoox: Fixed (with additional test case)

js/private/node-patches/src/fs.cts
```diff
@@ -302,7 +330,17 @@ export function patcher(roots: string[]): () => void {
...
+            // Resolve the link target against the REAL location of the link's
+            // parent directory. `resolved` may itself traverse symlinks (e.g. an
+            // aliased node_modules path of a different depth than the real dir);
+            // joining a relative target onto the lexical parent then lands at the
+            // wrong absolute path. Falls back to the old behavior if the parent
+            // cannot be realpath'd (identity transform when already resolved).
+            let linkDir = path.dirname(resolved)
+            try {
+                linkDir = realpathInRootNamespace(linkDir)
+            } catch {}
```
jbedard: When would this throw? Can we put a comment explaining why?

This is also repeated 2x now, lets put it into the helper?

Can we have:
```
const targetAbs = realpathInRootNamespace(resolved, linkDir)
```
or something like that? Would a better name be appropriate after doing that?

mloeppky-zoox: Extracted both call sites into a helper:

```ts
function resolveTargetAgainstRealParent(resolved: string, linkTarget: string): string {
    const linkDir = realpathInRootNamespace(path.dirname(resolved))
    return path.resolve(linkDir, linkTarget)
}
```

On "when would this throw?" — digging into it, the honest answer is *it shouldn't in normal operation*. We only reach this after `origReadlink`/`origReadlinkSync` succeeded against the real fs, so the symlink (and therefore its parent dir) exists on disk and `realpath` resolves. The only ways to throw are genuine anomalies (a TOCTOU race, or `EACCES`/`ELOOP`/`ENOTDIR`).

So rather than comment the old `try/catch`, I removed it. The previous fallback resolved the target against the *lexical* parent — which is exactly the wrong-absolute-path bug this PR fixes — so silently masking a failure by falling back re-introduced the bug for that path. Now a failure just propagates, consistent with every other operation in these functions (`resolvePathLike`, `isEscape`, `nextHop`, the `path.*` calls are all un-guarded too).

On the name: I went with `resolveTargetAgainstRealParent(resolved, linkTarget)` rather than overloading `realpathInRootNamespace` — it isn't a realpath of the whole input, it's "resolve the relative target against the real location of the parent," so the intent name reads better at the call site. The rationale (depth-mismatched aliased parent, guaranteed-present parent) is in the doc comment on the helper.

js/private/node-patches/src/fs.cts
```diff
@@ -107,6 +107,48 @@ export function patcher(roots: string[]): () => void {
         .native as typeof FsType.realpathSync.native
 
     const { canEscape, isEscape } = escapeFunction(roots)
+    const rootMappings = roots
+        .map((root) => {
+            const lexical = path.resolve(root)
+            let real = lexical
+            try {
+                real = origRealpathSyncNative(lexical) as string
+            } catch {}
```
jbedard, line 120: Should we also drop this `catch`? Or put a comment in it explaining why we're ignoring it and keeping `real == lexical`...

mloeppky-zoox: Good question, but I think this one should stay (with a comment, added in 3d05c6f) — it's the opposite case from the readlink `catch` I dropped.

This runs once at patcher-init over the *configured roots*, which legitimately may not exist on disk (or be inaccessible) when the patch is registered. And the fallback here is meaningful: keeping `real == lexical` just means "treat this root as not-symlinked," so the remapping in `realpathInRootNamespace` becomes a no-op for it. That's a correct default, not a mask over a bug.

Dropping the `catch` would instead throw during patcher setup for any root that can't be realpath'd, crashing the process at startup. So unlike the per-readlink realpath (where a throw is a genuine anomaly on a path we *know* exists because the readlink just succeeded), an error here is an expected startup condition. Comment added spelling that out.

jbedard (approved):

PR #2917, opened 2026-07-13
fix(npm_package_store): link first-party (JsInfo) deps into the store node_modules

### LINKED ISSUE #1092: [Bug]: "Cannot find module" error for dependency that is a peerDep of a first-party package

### LINKED ISSUE #2915: [Bug]: npm_package_store JsInfo/first-party branch doesn't link deps into store node_modules (asymmetric with the NpmPackageInfo branch)
### What happened?

In `npm/private/npm_package_store.bzl`, `_npm_package_store_impl` handles three `ctx.attr.src` cases. The `NpmPackageInfo` (3rd-party) branch iterates `ctx.attr.deps` and creates a nested store symlink for each dependency:

```py
for dep, _dep_aliases in ctx.attr.deps.items():
    ...
    dep_symlink_path = "node_modules/{}/{}/node_modules/{}".format(
        utils.package_store_root, package_store_name, dep_alias)
    files.append(utils.make_symlink(ctx, dep_symlink_path, dep_package_store_directory.path))
```

The `elif ctx.attr.src and JsInfo in ctx.attr.src:` (first-party / `js_library`) branch does **not** do this — it only creates the single top-level symlink to the package directory and never links `ctx.attr.deps` into `node_modules/{store_root}/{store_name}/node_modules/…`.

Consequence: for a first-party package `@myorg/web` that depends on first-party `@myorg/base`, the store copy

```
node_modules/.aspect_rules_js/@myorg+web@0.0.0/node_modules/@myorg/
```

is missing the `@myorg/base` entry entirely. Any tool that resolves a first-party package **through its package-store copy** (rather than the fully-populated app-level link) then fails to find its deps — e.g. `Cannot find module`/`ENOENT` on `@myorg/base/...` reached from `@myorg/web`. This is the same symptom as closed #1092 (`Cannot find module 'third-party-dep-1' from '.../.aspect_rules_js/@first-party+leaf@0.0.0/node_modules/@first-party/leaf/...'`).

**Expected:** the first-party (JsInfo) branch should link its `deps` into the store's nested `node_modules` the same way the `NpmPackageInfo` branch does, so a first-party package's deps are resolvable through its store copy.

### Version

Development (host) and target OS/architectures: Linux x86_64 (host + target)

Output of `bazel --version`: `bazel 7.7.1-1`

Version of the Aspect rules: `rules_js` v2.8.2. **Still present on `main` / v3.2.2** — the JsInfo branch of `_npm_package_store_impl` still only makes the top-level symlink and does not iterate `ctx.attr.deps`.

Language(s) and/or frameworks involved: TypeScript, Node.js v22, pnpm workspaces, first-party `js_library` packages linked via `npm_link_all_packages`; surfaced via an ESLint `import/namespace` resolver walking a re-export chain, but applies to any store-copy-based resolution.

### How to reproduce

```shell
# In a rules_js pnpm-workspace monorepo:
# - first-party @myorg/base (js_library) with a real dependency
# - first-party @myorg/web (js_library) that deps on @myorg/base
# - build/link @myorg/web, then inspect its package-store copy:

ls bazel-bin/.../node_modules/.aspect_rules_js/@myorg+web@0.0.0/node_modules/@myorg/
# EXPECTED: contains a `base` symlink
# ACTUAL:   `base` is absent (only @myorg/web itself is present)

# Any resolver that reaches @myorg/base *through* @myorg/web's store copy
# (e.g. following an exports/types re-export chain) then ENOENTs.
```

### Any other information?

Fix is to mirror the `NpmPackageInfo` branch's dep loop into the JsInfo branch (same-root-package guard, `dep_package_store_directory` truthiness check, ref-dep fallthrough to `direct_ref_deps`):

```py
for dep, _dep_aliases in ctx.attr.deps.items():
    dep_info = dep[NpmPackageStoreInfo]
    dep_aliases = _dep_aliases.split(",") if _dep_aliases else [dep_info.package]
    dep_package_store_directory = dep_info.package_store_directory
    if dep_info.root_package != ctx.label.package:
        fail(...)  # same message as the 3p branch
    if dep_package_store_directory:
        for dep_alias in dep_aliases:
            dep_symlink_path = "node_modules/{}/{}/node_modules/{}".format(
                utils.package_store_root, package_store_name, dep_alias)
            files.append(utils.make_symlink(ctx, dep_symlink_path, dep_package_store_directory.path))
    else:
        direct_ref_deps[dep] = dep_aliases
```

Caveat we observed: this changes the output-file set of **every** first-party package store repo-wide (broad cache invalidation; also watch for an output-path collision when a dep is aliased to the package's own name — the 3p branch has the same property). In our own case this was **not** the actual crash — a separate `node-patches` `readlink` bug (filed separately) was; once that was fixed, resolution reached the fully-populated app-level link and this store gap no longer bit us. But the asymmetry is real and matches #1092/#2226, so filing for visibility.

mloeppky-zoox:
Fixes #2915

## Problem

In `_npm_package_store_impl`, the `NpmPackageInfo` (3rd-party) branch iterates `ctx.attr.deps` and links each dep into the store's nested `node_modules/{store_root}/{store_name}/node_modules/{dep}`. The `elif ctx.attr.src and JsInfo in ctx.attr.src:` (first-party / `js_library`) branch does **not** — it only creates the top-level package symlink and never links its deps.

Consequently, for a first-party package `@scope/web` that depends on first-party `@scope/base`, the store copy

```
node_modules/.aspect_rules_js/@scope+web@0.0.0/node_modules/@scope/
```

is missing `@scope/base`. The package itself is linked (both at the app level and as a package-store directory); what's absent is the package's **direct deps** inside its *own* package-store `node_modules/`. App-level `require()`s still resolve through the fully-populated app-level `node_modules`, so they work — but any tool that resolves a first-party package's deps **through its package-store copy** (rather than the app-level link) then fails with `Cannot find module` / `ENOENT`. That's the `readlink`/`realpath` resolution path in #2915 / #2914 (and the same symptom as the closed #1092).

## Fix

Mirror the `NpmPackageInfo` branch's direct-deps loop into the JsInfo branch: same same-root-package guard, `dep_package_store_directory` truthiness check, and ref-dep fallthrough to `direct_ref_deps`.

## Testing

- New assertion in `examples/linked_consumer/test_pkg_deps_linked.js`: it locates the first-party `@lib/test2` package-store link and asserts test2's own deps (`@aspect-test/e`, `alias-e`) are linked into test2's package-store `node_modules/` — which were absent before this change.
- The `*_package_store_1p.listing` image-test fixtures now include first-party packages' dep links (e.g. `uuid` under `@image-test/a` and `@image-test/d`), giving a concrete before/after diff of the files that newly become available in the sandbox.
- Existing first-party linking targets still pass (`//linked_consumer:{direct,pkg,sources_test,types_test}`, `//js_lib_pkg/b:{test_each_target,test_parent_target}`) — existing linking is unaffected.

## Notes / caveats

- This adds nested symlink outputs to **every** first-party package store, so expect broader cache invalidation. As with the 3rd-party branch, watch for an output-path collision if a dep is aliased to the package's own name.
- Related: closed #1092 (same "Cannot find module" symptom) and open #2226 (first-party peer-dep propagation, a different mechanism in `npm_link_package_store.bzl`).

Co-Authored-By: Claude Code


jbedard:
Sounds reasonable, but we need a decent test reproducing the issue...

mloeppky-zoox:
> Sounds reasonable, but we need a decent test reproducing the issue...

Done

js/private/test/image/custom_layers_nomatch_test_package_store_1p.listing
```diff
@@ -11,7 +11,9 @@ drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/@image-test+a@0.0.0/node_modules/
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/@image-test+a@0.0.0/node_modules/@image-test/
 lrwxrwxr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/@image-test+a@0.0.0/node_modules/@image-test/a -> ../../../../../js/private/test/image-fixture-a
+lrwxrwxr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/bin.runfiles/_main/node_modules/.aspect_rules_js/@image-test+a@0.0.0/node_modules/uuid -> ../../uuid@8.3.2/node_modules/uuid
```
jbedard, line 14: Why did a `uuid` link get added from this change?

examples/linked_consumer/test_pkg_deps_linked.js
```diff
@@ -1,5 +1,28 @@
...
+const storeNodeModules = path.dirname(path.dirname(packageStoreLink))
+for (const dep of ['@aspect-test/e', 'alias-e']) {
+    const depPath = path.join(storeNodeModules, dep)
+    assert.ok(
+        fs.existsSync(depPath),
+        `${dep} should be linked into the first-party package store at ${depPath}`
+    )
+}
+
 const testLib = require('@lib/test')
```
jbedard, line 26: How could these `require()`s have worked fine if the `node_modules/@lib/test2` had supposedly not existed?

Can we have a test case (most likely similar to the oci layer tests) that clearly demonstrates which new files are available in the sandbox after this change? A diff in something such as `js/private/test/image/custom_layers_nomatch_test_package_store_1p.listing` showing which new files exist after this fix would be ideal...

js/private/test/image/custom_layers_nomatch_test_package_store_1p.listing
mloeppky-zoox: `@image-test/a` and `@image-test/d` are first-party packages that depend on `uuid@8.3.2`. Before this change, a first-party package's package-store `node_modules/` received none of its own deps; after it, they do — so these two `uuid` symlinks are the fix working: the first-party package's `uuid` dep is now resolvable from its package-store copy. This `.listing` diff is effectively the before/after demonstration of which files newly appear.

examples/linked_consumer/test_pkg_deps_linked.js
mloeppky-zoox: Sorry, that framing was misleading — `@lib/test2` itself was always linked (both the app-level `node_modules/@lib/test2` and its package-store directory). What was missing is test2's **direct deps** inside test2's *own* package-store `node_modules/`.

The existing `require()`s resolve `@lib/test2` and its deps through the app-level `node_modules` tree, which is fully linked, so they pass. The gap only appears when a tool resolves a package's deps starting from its real package-store location — the `readlink`/`realpath` walk in #2914 / #2915 — which is the case that broke.

This test asserts test2's deps `@aspect-test/e` and `alias-e` are now present in test2's package-store `node_modules/` (absent before this change), and the four `*_package_store_1p.listing` fixtures enumerate the newly-linked first-party deps (e.g. the `uuid` links above).

jbedard (approved):

PR #2918, opened 2026-07-14
feat: support --bazel-bindir flag for path-mapping-compatible js_binary tools

acozzette:
BAZEL_BINDIR is normally set via the $(BINDIR) Make variable, which bakes in the real, configuration-specific bindir (e.g. bazel-out/k8-fastbuild/bin) at analysis time.  Bazel's path mapping only works for command-line arguments, so BAZEL_BINDIR as an environment variable permanently defeats action-cache sharing across configurations (e.g. -c opt vs -c fastbuild).

This adds a --bazel-bindir=<path> flag to the generated js_binary launcher (js/private/js_binary.sh.tpl), consumed as the first argument, so callers can pass a path-mapping-safe bindir value instead of relying on the environment.

e2e/path_mapping/bindir_path_mapping_check proves this works end-to-end: a test-only rule drives a real js_binary via a map_each-computed --bazel-bindir argument, and test.sh proves two builds that differ only in compilation mode share a single cached action via a shared --disk_cache.

I plan to get js_run_binary working soon, but we need to wait for a bazel-lib release before we can do that.

---

Support for running js_binary targets in a path-mapping-friendly way

### Test plan

- New test cases added


e2e/path_mapping/test.sh
```diff
@@ -0,0 +1,47 @@
...
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+scratch="$(mktemp -d)"
+trap 'rm -rf "$scratch"' EXIT
+disk_cache="$scratch/disk_cache"
+exec_log="$scratch/exec_log.json"
+
+# Start from a clean slate so this build actually executes the action, rather
+# than being served by leftover local build state from earlier CI steps.
```
jbedard: Why is that required?

e2e/output_paths/test.js
```diff
@@ -1,13 +0,0 @@
-const { Runfiles } = require('@bazel/runfiles')
```
jbedard, line 1: What was this e2e test previously testing? That seems unrelated to path mapping, why delete this?

acozzette: As far as I can tell it was just doing a smoke test that we can still build things when we pass `--experimental_output_paths=strip`, but it was not actually exercising path mapping. I found that kind of confusing and it seemed to me that my new tests already cover that anyway.

e2e/path_mapping/test.sh
acozzette: Without this the result from a previous run might be cached, and then the build action won't execute at all regardless of whether path mapping is working.

e2e/output_paths/test.js
jbedard: Does path-mapping effect runfiles in anyway such that we'd want to keep these .js assertions?

acozzette: As far as I know, runfiles are not very closely related to path mapping. But I can keep these if you prefer, what do you think?

jbedard: If we delete them lets just make sure similar tests exist somewhere? Maybe a prefactor can be moving them somewhere more appropriate if you don't think they belong here?

acozzette: OK, I think I'll just go ahead and leave the existing tests where they are then. I updated my PR to create a new e2e directory for the new tests.

e2e/path_mapping/bindir_path_mapping_check/bindir_check.mjs
```diff
@@ -0,0 +1,25 @@
...
+const bindir = process.env.BAZEL_BINDIR
+if (bindir !== 'bazel-out/cfg/bin') {
+    process.stderr.write(
+        `Expected BAZEL_BINDIR to be "bazel-out/cfg/bin", got "${bindir}"\n`
+    )
+    process.exit(1)
+}
+
+const leaked = process.argv
+    .slice(2)
```
jbedard: Do we need the `.slice`?

jbedard:
Long term I still question this approach. I want to remove the shell wrapper and instead invoke node directly or the hermetic-launcher wrapping node instead (for perf and removing the dependency on a shell)... how will this work then?

I still wonder if we can do this within node itself, either:
* `--eval "process.env.BAZEL_BINDIR=??"` added to the args
* [--import](https://nodejs.org/api/cli.html#--importmodule) or [--require](https://nodejs.org/api/cli.html#-require-module) a script that does magic

I'm not sure what "does magic" means though... how can we determine the `BAZEL_BINDIR` at runtime?

Maybe what you have here is fine for now, but please keep these ideas in mind... ideally the .sh file goes away sooner or later...

js/private/js_binary.sh.tpl
```diff
@@ -90,6 +90,28 @@ _exit() {
...
+# ==============================================================================
+# Handle --bazel-bindir flag
+# ==============================================================================
+
+# If a --bazel-bindir=<path> or --bazel-bindir <path> flag is passed it must
+# be the first argument(s). It is consumed by this launcher script and used to
+# set BAZEL_BINDIR, overriding any value already set in the environment.
+if [ $# -gt 0 ]; then
+    case "$1" in
+    --bazel-bindir=*)
```
jbedard: Do we really need both? 🤦
Can we just be strict and not accept the `=` version?

e2e/path_mapping/bindir_path_mapping_check/bindir_check.mjs
acozzette: I don't think it's strictly necessary, so I went ahead and removed it.

js/private/js_binary.sh.tpl
acozzette: Done.

jbedard:
@acozzette have you given any thought into the `--eval` or `--require` ideas [above](https://github.com/aspect-build/rules_js/pull/2918#issuecomment-5007639339)? Even if not today are those possible long term?

acozzette:
The only API Bazel provides for mapping paths from Starlark is `args.add()` and `args.add_all()`, so I think we're pretty much forced to use some kind of command-line flag. (Unless we were to find some way to avoid needing `BAZEL_BINDIR` at all, but I can't see any way to do that without making `js_binary` much harder to use.) And then since we don't want to do any real parsing in bazel-lib, we don't have much choice but to use some kind of two-argument form `--x <bin dir>`.

I asked Claude for ideas on how to keep this working when we get rid of the bash launcher and it made this recommendation which sounds promising to me:

> 1. (Recommended) Replace the bash launcher with a Node bootstrap script, invoked directly via shebang
> 
> Make the js_binary "executable" itself a #!/usr/bin/env node script (today it's #!/usr/bin/env bash). Bazel/the OS execs node directly via the shebang — one process, not bash-forks-node. The bootstrap script, in plain JS:
> 
> - Parses process.argv for the two-arg --bazel-bindir <path> form (this branch already simplified to just that form — see 910db30c), splices it out so the real entry point never sees it.
> - Sets process.env.BAZEL_BINDIR and process.chdir(path) — a direct port of js_binary.sh.tpl:100-104 / :166.  
> - Applies the existing fs-patch setup in-process (call the same function register.cjs/fs.cjs exposes today) instead of relying on --require for it.  
> - Then require()/dynamic-import()s the real entry point.
> 
> This is the true "invoke node directly" end state: it keeps --bazel-bindir working with identical argv semantics, and collapses two processes into one. The .sh.tpl template becomes a .js.tpl (or .mjs.tpl) template — same Starlark-side substitution mechanism (js_binary.bzl's launcher_subst), just generating JS instead of bash.
> 
> Caveat: the bash launcher does more than bindir handling — runfiles init ({{initialize_runfiles}}), the PWD-based execroot/bindir auto-detection heuristic (js_binary.sh.tpl:117-148), JS_BINARY__NO_CD_BINDIR opt-out, and fatal-error messaging all need JS equivalents. None of this is blocked by the flag-handling question, but it's the bulk of the actual porting effort. Windows needs an analogous node_wrapper.bat-style entry point (or a .cmd shim), and fs-patching is already disabled there today.

e2e/path_mapping/test.sh
jbedard: Can we not check the outputs somehow instead of using `--execution_log_json_file`?

acozzette: I think we could parse it from the Bazel command output because I see that it prints this: `INFO: 2 processes: 1 disk cache hit, 1 internal.` Is that what you had in mind? I don't think we could tell whether there was a disk cache hit by looking in `bazel-out/`, though.

jbedard: I was thinking of it more as "the same output files" and not "is it a cache hit", can we not just inspect to ensure there is only one `bazel-out/*` directory?

jbedard:
@codex

e2e/path_mapping/test.sh
acozzette: No, I don't think that works unfortunately. From what I understand, path mapping is a temporary thing that Bazel sets up for a build action. Once the action is over, Bazel copies or moves the outputs from `bazel-out/cfg/bin` to the real `bazel-out/<platform>/bin` directory, and `bazel-out/cfg/bin` is not persisted.

jbedard: > Once the action is over, Bazel copies or moves the outputs from bazel-out/cfg/bin to the real bazel-out/<platform>/bin directory, and bazel-out/cfg/bin is not persisted.

Oh, TIL. Which is why you're checking the execlog. I'd still prefer if we could avoid doing a `bazel clean` though, won't that clean _everything_ while we just want to force a few build actions to re-run? Can we change an env var instead or something like that?

js/private/js_binary.bzl
```diff
@@ -591,21 +591,34 @@ def _js_binary_impl(ctx):
...
+    outputs = kwargs.get("outputs")
+    if not outputs:
+        fail("js_run_binary_action requires at least one output")
+    extra_args = ctx.actions.args()
+    extra_args.add("--bazel-bindir")
+    extra_args.add_all([outputs[0]], map_each = _bazel_bindir_arg)
+
     ctx.actions.run(
-        env = dict(env) | {"BAZEL_BINDIR": ctx.bin_dir.path},
+        arguments = [extra_args] + kwargs.pop("arguments", None) or [],
```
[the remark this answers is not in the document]
acozzette: Fixed

e2e/path_mapping/test.sh
acozzette: Sure, I updated it to just change an environment variable instead of doing a full bazel clean.

e2e/path_mapping/test.sh
```diff
@@ -0,0 +1,49 @@
...
+# configurations, so each gets its own action instance the first time Bazel
+# visits it in a given build graph -- the incremental "did anything change"
+# check within one build never gets a chance to compare across them. Only an
+# explicit disk (or remote) cache lookup, keyed by the path-mapped action
+# digest, can serve the second build's action from the first build's result.
+set -o errexit -o nounset -o pipefail
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+scratch="$(mktemp -d)"
```
jbedard, line 18: We have to use a temp dir instead of just editing .bazelrc? That means when `bazel test //...` is run in here this followup `./test.sh` will not use the same cache?

js/private/js_binary.sh.tpl
```diff
@@ -132,6 +132,23 @@ _exit() {
...
 trap _exit EXIT
 
+# ==============================================================================
+# Handle --bazel-bindir flag
+# ==============================================================================
+
+# If a --bazel-bindir <path> flag is passed it must be the first two
+# arguments. It is consumed by this launcher script and used to set
+# BAZEL_BINDIR, overriding any value already set in the environment.
+if [ $# -gt 0 ] && [ "$1" = "--bazel-bindir" ]; then
```
jbedard, line 142: Do we care if a global `BAZEL_BINDIR` is already set? Such as if one `js_binary` launches another? 🤔

jbedard (approved):

e2e/path_mapping/test.sh
acozzette: We could perhaps put this in .bazelrc, but I'm not sure what a reasonable location for the disk cache would be (maybe a local directory that we put in `.gitignore`?). I actually think we are probably better off just using a temp dir, though. `--disk_cache` is another layer on top of Bazel's usual caching, so it's not very wasteful or anything to create a fresh disk cache on every run--all actions aside from the one we're trying to test will still be cached by Bazel's non-`disk_cache` caching.

js/private/js_binary.sh.tpl
acozzette: I think that should be fine. It seems to me like one would rarely want to set both `--bazel-bindir` and `BAZEL_BINDIR`, but if that does happen then the flag overrides the environment variable. In the case of one `js_binary` launching another, I would expect that only the environment variable would be set.

PR #2919, opened 2026-07-15
fix: detect execroot when output_base path contains a bazel-out segment

titusfortner:
The `js_binary` launcher infers the Bazel execroot by scanning `$PWD` for the substring `/bazel-out/`. If `--output_base` (or any ancestor directory) is named `bazel-out`, that substring matches the output-base directory instead of the real output tree, so the launcher computes the execroot too shallow. For a `js_binary` used as the tool of a `js_run_binary` this fails with `entry_point '<...>' not found`, and because the wrong branch also skips `cd "$BAZEL_BINDIR"`, tools that rely on being in `$(BINDIR)` then fail to read their inputs.

Detect a fresh build-action execroot structurally instead: `$PWD` is the execroot when the bindir hangs off it (`-d "$PWD/$BAZEL_BINDIR"`), even when its path contains a `bazel-out` segment. Otherwise scan the output tree for the execroot — the runfiles case (e.g. `js_test`), or a nested `js_binary` whose parent already `cd`'d into the bindir. A nested child inherits `BAZEL_BINDIR` but runs from the bindir, so keying off `BAZEL_BINDIR` alone is not enough; the directory check distinguishes the two. The runfiles scan also strips from the last `bazel-out` segment rather than the first, so a `bazel-out` component in the output base is not mistaken for the output tree. None of this changes behavior under the default output base.

Reproduce with any `js_run_binary` and an output base whose path contains a `bazel-out` segment:

```sh
bazel --output_base=/tmp/demo/bazel-out build //path/to:some_js_run_binary_target
```

- Searched for relevant documentation and updated as needed: yes (no docs describe this behavior)

Fix `js_binary`/`js_run_binary` failing to find the entry point when `--output_base` points at a path containing a `bazel-out` segment.

### Test plan

- Covered by existing test cases (launcher snapshot, `js_binary_sh`, `js_test` toolchain tests, and nested-execution coverage in `examples/runfiles`)
- New e2e case: `e2e/js_binary_workspace` now runs under an output base whose path contains a `bazel-out` segment (via a relative `--output_base` in its `.bazelrc`), reproducing the bug without the fix

### DIFF
--- a/e2e/js_binary_workspace/.bazelignore
+++ b/e2e/js_binary_workspace/.bazelignore
@@ -1,1 +1,2 @@
 workspace/
+.e2e_output_base/
--- a/e2e/js_binary_workspace/.bazelrc
+++ b/e2e/js_binary_workspace/.bazelrc
@@ -1,2 +1,6 @@
 import %workspace%/../../tools/preset.bazelrc
 import %workspace%/../e2e.bazelrc
+
+# Regression test for #2919: exercise execroot detection when the output base path contains a
+# "bazel-out" segment (nested so it does not clash with the bazel-out convenience symlink).
+startup --output_base=.e2e_output_base/bazel-out/base
--- a/js/private/js_binary.sh.tpl
+++ b/js/private/js_binary.sh.tpl
@@ -173,9 +173,10 @@ elif [[ "$PWD" == *"/bazel-~1/"* ]]; then
     bazel_out_segment="/bazel-~1/"
 fi
 
-# BAZEL_BINDIR is only set for build actions, which run from the execroot; only scan $PWD to find
-# the execroot in the runfiles case (BAZEL_BINDIR unset), never when it is already the execroot.
-if [[ -z "${BAZEL_BINDIR:-}" && "${bazel_out_segment:-}" ]]; then
+# When $PWD is a build action execroot the bindir hangs off it (BAZEL_BINDIR resolves from $PWD), so
+# $PWD is the execroot even if its path contains a "bazel-out" segment (e.g. a matching output base).
+# Otherwise scan the output tree for the execroot (runfiles, or a nested js_binary in the bindir).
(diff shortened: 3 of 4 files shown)


jbedard:
Is there any legit reason for `--output_base` (or any ancestor directory) to be named bazel-out?

Is there any way to reproduce this in an e2e test with a `.bazelrc`?

titusfortner:
> Is there any legit reason for --output_base (or any ancestor directory) to be named bazel-out

It's just what I happened to do. I changed it in my implementation so I'm not waiting on a fix here. I just figured I would suggest a change since it was unexpected behavior. (https://github.com/SeleniumHQ/selenium/commit/43b9ade8a04c5c57ca86fdfbccb213dd98e45184)

fixed the bug and added a test.

e2e/js_binary_workspace/.bazelignore
```diff
@@ -1 +1,2 @@
 workspace/
+.e2e_output_base/
```
jbedard, line 2: Does this also need to be git ignored?

titusfortner: The only thing inside .e2e_output_base/ is bazel-out/, and the repo's root .gitignore already has bazel-* which matches bazel-out at any depth, so I don't think it's necessary

acozzette:
@titusfortner Are you ok with signing the CLA?

titusfortner:
Oops missed that. It's signed now thanks.

e2e/js_binary_workspace/.bazelignore
jbedard: Can we name it `bazel-e2e_output_base` or something like that so the full directory gets ignored? Or maybe it doesn't make any difference? 🤷

acozzette (approved):

acozzette:
@titusfortner Thank you for this fix!

PR #2932, opened 2026-07-22
fix(coverage): generate lcov report in the test action

### LINKED ISSUE #965: [Bug]: Tests are broken with rbe and --experimental_split_coverage_postprocessing
### What happened?

With bazel 6.1.0, run a test with RBE and flags: 
* --remote_download_minimal
* --experimental_split_coverage_postprocessing
* --experimental_fetch_all_coverage_outputs
And see an error like 

```
FATAL: the entry_point '$BAZEL_OUTPUT_BASE/sandbox/linux-sandbox/17885/execroot/$REPO_NAME/bazel-out/k8-fastbuild/bin/...TARGET_PATH.../../aspect_rules_js/js/private/coverage/coverage.js' not found in runfiles
```

### Version

Output of `bazel --version`: 6.1.0 

3530bb64176cdbd54fec772bda646a76a6249655

### How to reproduce

_No response_

### Any other information?

_No response_

### LINKED ISSUE #2901: [Bug]: `bazel coverage` reports 0% for `jest_test` on `rules_js` 3.x (works on 1.x)

jbedard:
Fix #2901
Fix #965
Ref #2933
Close #2904
Close #991

Fixed `bazel coverage` returning empty (0%) reports under split-postprocessing and remote execution.

### Test plan

- New test cases added

### DIFF
--- a/js/private/js_binary.sh.tpl
+++ b/js/private/js_binary.sh.tpl
@@ -478,6 +478,29 @@ wait "$child"
 RESULT="$?"
 set -e
 
+# Generate the lcov report for a passing test; empty when not collecting coverage. See #2901.
+coverage_entry_point="{{coverage_entry_point}}"
+
+if [ -n "$coverage_entry_point" ] && [ "$RESULT" = "${JS_BINARY__EXPECTED_EXIT_CODE:-0}" ] && [ -n "${COVERAGE_DIR:-}" ] &&
+    # A test reporting its own coverage owns it, except in split mode where bazel drops it.
+    { [ ! -s "${COVERAGE_OUTPUT_FILE:-}" ] || [ "${SPLIT_COVERAGE_POST_PROCESSING:-0}" = "1" ]; }; then
+    coverage_entry_point="$JS_BINARY__RUNFILES/$coverage_entry_point"
+    if [ ! -f "$coverage_entry_point" ]; then
+        # Report empty coverage rather than fail an otherwise passing test.
+        logf_error "coverage report generator '%s' not found; code coverage requires a runfiles tree" "$coverage_entry_point"
+    else
+        # NODE_V8_COVERAGE unset so this process writes no profile of its own.
+        (
+            unset NODE_V8_COVERAGE
+            export JS_COVERAGE__RUNFILES="$JS_BINARY__RUNFILES"
+            exec "$JS_BINARY__NODE_BINARY" "$coverage_entry_point"
+        ) || {
+            logf_error "coverage report generation failed"
+            exit 1
+        }
+    fi
+fi
+
 # ==============================================================================
 # Mop up after main program
 # ==============================================================================
--- a/.github/workflows/ci-workflows.yaml
+++ b/.github/workflows/ci-workflows.yaml
@@ -243,6 +243,13 @@ jobs:
               if: matrix.workspace.path == '.' && matrix.bazel.id == 'bazel-7'
               run: aspect test --task-key=coverage-${{ matrix.workspace.slug }}-${{ matrix.bazel.id }} ${{ matrix.bazel.flags }} --bazel-flag=--collect_code_coverage --bazel-flag=--instrument_test_targets --bazel-flag=--nocache_test_results -- //js/private/test/coverage/...
 
+            # Again with the merger as its own action, as remote execution requires. The
+            # mergers assert different things in that mode -- bazel discards a report the
+            # test wrote to COVERAGE_OUTPUT_FILE itself -- so both modes need a run.
+            - name: Coverage (split postprocessing)
+              if: matrix.workspace.path == '.' && matrix.bazel.id == 'bazel-7'
+              run: aspect test --task-key=coverage-split-${{ matrix.workspace.slug }}-${{ matrix.bazel.id }} ${{ matrix.bazel.flags }} --bazel-flag=--collect_code_coverage --bazel-flag=--instrument_test_targets --bazel-flag=--nocache_test_results --bazel-flag=--experimental_split_coverage_postprocessing --bazel-flag=--experimental_fetch_all_coverage_outputs -- //js/private/test/coverage/...
+
             # Skipped on the no-execroot-entry-point variant: test.sh scripts invoke plain
             # `bazel` without matrix.bazel.flags, so that leg wouldn't exercise the flag —
             # it would only duplicate the bazel-9 run.
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -227,3 +227,75 @@ You can also reduce the size of the extracted packages themselves with
 `npm.npm_exclude_package_contents`, which removes files you consider unnecessary (docs,
 tests, etc.) at extraction time; see
 [Unnecessary npm package content](#unnecessary-npm-package-content) above.
+
+## Code coverage
+
+Under `bazel coverage`, `js_test` produces its lcov report while the test runs,
+where the V8 coverage data and the instrumented sources are both available.
+
+### Coverage runs count against the test's timeout
+
+Because the report is produced while the test runs, generating it consumes the
+target's own `timeout`/`size` budget. Converting V8 coverage to lcov takes time
+proportional to the number of instrumented files, so a test that sits near its
+timeout under `bazel test` may need a larger `size` or `timeout` to pass under
+`bazel coverage`.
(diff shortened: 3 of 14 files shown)


markarce:
### `bazel coverage` still empty for store-linked first-party imports (separate from the split-postprocessing fix)

Tested this branch via `git_override` against a real `jest_test` in our monorepo (aspect_rules_js 3.1.1 + this PR). The split-postprocessing fix works, but coverage is **still 0%** whenever first-party code is imported through the pnpm / `.aspect_rules_js` store — which is the common case. Root cause looks independent of #2932.

**Mechanism.** `js/private/coverage/coverage.js` roots c8 at:
```
src: pwd,          // the test's runfiles dir
resolve: '',
all: true,
// allowExternal defaults false -> relativePath filtering on
```
Store-resolved first-party modules execute at a realpath **outside** that `src` root, so every real script is dropped by the `relativePath`/`_shouldInstrument` filter and only the `all: true` entries survive (empty).

**Evidence (monorepo jest_test).** Raw `NODE_V8_COVERAGE` shows 128 first-party files with real hit counts, e.g.:
```
file:///.../bazel-out/darwin_arm64-fastbuild/bin/<root_dir>/packages/<pkg>/src/<filename>.js
```
That's genuine first-party source (not `node_modules`), in the `bin` tree — outside c8's runfiles `src` -> the resulting `coverage.dat` is empty.

**Minimal repro (plain `js_test`, no jest).** Same as `e2e/coverage`, but `lib` is a first-party package imported by name instead of a relative `./lib.js`:

`lib/BUILD.bazel`
```python
load("@aspect_rules_js//npm:defs.bzl", "npm_package")
js_library(name = "src", srcs = ["lib.js", "package.json"])
npm_package(name = "lib", srcs = [":src"], package = "@repro/lib", visibility = ["//visibility:public"])
```
`BUILD.bazel`
```python
load("@aspect_rules_js//js:defs.bzl", "js_test")
load("@aspect_rules_js//npm:defs.bzl", "npm_link_package")
npm_link_package(name = "node_modules/@repro/lib", src = "//lib:lib")
js_test(name = "test", data = [":node_modules/@repro/lib"], entry_point = "test.js")
```
`test.js`: `const lib = require('@repro/lib'); require('node:assert').strictEqual(lib.covered(), 'covered')`
(`lib/lib.js` = the `covered()` / `uncovered()` from `e2e/coverage`.)

`bazel coverage //:test` -> `covered()` runs (verified via `NODE_V8_COVERAGE`, URL under `node_modules/.aspect_rules_js/@repro+lib@0.0.0/...`) but `coverage.dat` is **0 bytes**. Switch `test.js` back to a relative `require('./lib.js')` and it's fully covered — so the trigger is store resolution. (In a pnpm workspace the URL is the package's `src` in `bin` as above; in this minimal `npm_package` case it lands in `.aspect_rules_js` — both outside c8's `src` root.)

**Fix direction** (in the coverage wrapper, not `--preserve-symlinks` — that breaks the `.aspect_rules_js` store, per #2901): normalize V8 script URLs into the runfiles/`src` space before c8 reads them, or root c8 at the realpath and normalize the output `SF:` paths. Complementary to the split-postprocessing fix here.

Happy to open this as a separate issue if you'd prefer.

acozzette (approved):

thesayyn:
Please see https://github.com/bazelbuild/bazel/issues/18017 there was a principled way to do this but i don't remember anymore.

jbedard:
@thesayyn I believe this workaround does the same thing you were doing in #991

jbedard:
@acozzette see [diff](https://github.com/aspect-build/rules_js/compare/93af209f7744c5d4677c0ef32818032fe4af71f4..ef95b82f53d6fcd280a163d9d2b9392270c90f33) which changes this a bit:
* the "merger" is now implemented in bash instead of node, which I think eliminates #965
* ^ simplifies `coverage.sh.tpl`, except some test assertions that might be uglier in bash then js

acozzette:
Could you add some more detail to the description? I'm finding it somewhat hard to follow all the things the PR is doing.

jbedard:
Sorry this expanded to modify multiple things, maybe wait till monday and I'll try to have something more readable or a pre-factor that does some of it.

jbedard:
Here's some prefactors to land first:
* #2950
* #2951
* #2952 
* #2953
* #2954

jbedard:
@codex

js/private/js_binary.sh.tpl
```diff
@@ -402,6 +402,27 @@ wait "$child"
...
+    coverage_entry_point="$JS_BINARY__RUNFILES/$coverage_entry_point"
+    if [ ! -f "$coverage_entry_point" ]; then
+        # Report empty coverage rather than fail an otherwise passing test.
+        logf_error "coverage report generator '%s' not found; code coverage requires a runfiles tree" "$coverage_entry_point"
+    else
+        # Subshell: NODE_V8_COVERAGE unset so this process writes no profile of its own.
+        (
+            unset NODE_V8_COVERAGE
+            export JS_COVERAGE__RUNFILES="$JS_BINARY__RUNFILES"
+            exec "$JS_BINARY__NODE_BINARY" "$coverage_entry_point"
```
jbedard, line 496: So we now execute the coverage tool in the `js_binary.sh` so it is the same bazel action as the the action `bazel coverage` is running on that emits the coverage data which bazel understands. This is what fixes #2901

This is what allows `js/private/coverage/coverage.sh.tpl` and the merger logic to be plain bash now which I believe is what fixes [#965](https://github.com/aspect-build/rules_js/issues/965)

jbedard:
@codex

acozzette (approved):

jbedard:
@codex

PR #2936, opened 2026-07-23
feat(npm): enable symbolic macros for node_modules on Bazel 9+

kirkobyte:
## Summary

Continues work from #2666. On Bazel 9+, the generated `defs.bzl` now wraps store invocations and link-package targets in symbolic macros (`_all_stores`, `_npm_link_all_packages`), preparing for future lazy macro evaluation. On Bazel 7 and 8, the generator emits the original legacy-macro code unchanged.

- Searched for relevant documentation and updated as needed: no (no user-facing docs change)

**Release notes:** On Bazel 9+, `npm_link_all_packages` now uses symbolic macros for store invocations and link-package target creation, enabling visibility encapsulation and preparing for future lazy macro evaluation.

## What changed

- **Version detection**: Uses `bazel_features.globals.macro` and `hasattr(attr, "label_list_dict")` to detect symbolic macro support. This activates on Bazel 9+ only (`macro()` exists on Bazel 8 but `attr.label_list_dict` does not).
- **Conditional code generation**: The `npm_translate_lock` code generator emits different `defs.bzl` content depending on the Bazel version:
  - **Bazel 9+**: Store calls wrapped in `_all_stores = macro(...)`, link targets wrapped in `_npm_link_all_packages = macro(...)`
  - **Bazel 7/8**: Original inline code (no behavioral change)
- **FP store template**: Added `_FP_STORE_MACRO_TMPL` with correct 4-space indentation for symbolic macro impl function bodies (fixes an indentation bug in the original PR).
- **BUILD dep**: Added `@bazel_features//:features` to `npm_translate_lock_generate` bzl_library deps.

## Fixes over #2666

- Gates symbolic macro usage on Bazel version so Bazel 7/8 are not broken
- Fixes first-party store template indentation (8→4 spaces for macro impl body)
- Adjusts `to_dict_attr` indent depth to match macro nesting level

## Test plan

- `bazel build //:node_modules` passes on Bazel 7.7.1, 8.2.1, and 9.0.0
- All `write_npm_translate_lock_*` snapshot tests pass on Bazel 7 (the repo's default `.bazelversion`)
- Snapshot tests on Bazel 9 show expected diff (symbolic macro code vs inline code)

### DIFF
--- a/npm/private/BUILD.bazel
+++ b/npm/private/BUILD.bazel
@@ -91,6 +91,7 @@ bzl_library(
     deps = [
         ":starlark_codegen_utils",
         ":utils",
+        "@bazel_features//:features",
         "@bazel_lib//lib:base64",
         "@bazel_skylib//lib:dicts",
         "@bazel_skylib//lib:partial",
--- a/npm/private/npm_translate_lock_generate.bzl
+++ b/npm/private/npm_translate_lock_generate.bzl
@@ -1,10 +1,13 @@
 """Starlark generation helpers for npm_translate_lock.
 """
 
+load("@bazel_features//:features.bzl", "bazel_features")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load(":npm_translate_lock_helpers.bzl", "helpers")
 load(":starlark_codegen_utils.bzl", "starlark_codegen_utils")
 load(":utils.bzl", "utils")
+
+_SUPPORTS_SYMBOLIC_MACROS = bazel_features.globals.macro != None and hasattr(attr, "label_list_dict")
 
 ################################################################################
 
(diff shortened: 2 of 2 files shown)


kirkobyte:
Closing — the `_all_stores` symbolic macro works, but the main invalidation benefit requires wrapping link targets too, which is blocked by the symbolic macro naming constraint (`/` is not a valid separator). Need to design a `__`-naming API to make this viable. See discussion in PR for details.

jbedard:
So that would require breaking our `:node_modules/{pkg}` convention completely? Due to what codex said:
> Bazel's symbolic macro naming rule only permits targets equal to the macro name or prefixed with node_modules_, node_modules-, or node_modules.
?

I wonder if we can convince the bazel team to add `/` to that list...

kirkobyte:
I imagine it's there because it can conflict with the other `/` symbols, e.g. how do we tell the difference between the two slashes in `:node_modules/lodash/sort` and would supporting `/` in the name open the door to conflicts?

jbedard:
Scoped packages such as `:node_modules/@angular/core` already have multiple slashes...

kirkobyte:
I’m meaning to refer to this concept and whether / is disallowed to remove label ambiguity: https://bazel.googlesource.com/bazel/+/5f083b8f5a073a3427d575aa54eac5b779226708/site/docs/guide.md
> In addition, Bazel allows a slash to be used instead of the colon required by the label syntax; this is often convenient when using Bash filename expansion. For example, foo/bar/wiz is equivalent to //foo/bar:wiz (if there is a package foo/bar) or to //foo:bar/wiz (if there is a package foo).

But I’d definitely avoid modifying all the paths if it is possible to allow / in the names

PR #2939, opened 2026-07-23
feat: add npm_package(packable) for creating the package tar.gz file

jbedard:

Add `npm_package(packable=True)` for exposing the tar.gz artifact generated by pnpm

### Test plan

- New test cases added


npm/private/npm_package.bzl
```diff
@@ -448,6 +458,34 @@ def npm_package(
...
+        # `pnpm pack` resolves --out relative to --dir (the package output
+        # directory, which is `out` or `name`). Climb back out of that directory
+        # to the Bazel package root, then down to the declared `{name}.tgz`, so
+        # the tarball lands exactly on the declared output regardless of how many
+        # path segments `out`/`name` contain.
+        pack_dir = out if out else name
+        pack_out = "{}/{}.tgz".format("/".join([".."] * len(pack_dir.split("/"))), name)
+        js_run_binary(
+            name = "{}.pack".format(name),
+            tool = Label("@pnpm//:pnpm"),
```
jbedard, line 471: Note the use of pnpm instead of npm like `npm_package(publishable)` does. Long term `publishable` should do the same thing and use this.

It is possible this is not the version of pnpm the user wants - tough luck for now, that can be a future FR and I'll keep the API minimal for now

thesayyn (approved):

PR #2943, opened 2026-07-24
fix: do not assume testonly=False in macros to support default_testonly

jbedard:
There are probably more places in other rulesets we need to be doing this...

### Test plan

- Manual testing

### DIFF
--- a/js/private/js_run_binary.bzl
+++ b/js/private/js_run_binary.bzl
@@ -280,8 +280,7 @@ def js_run_binary(
             include_npm_sources = include_npm_sources,
             # Always tag the target manual since we should only build it when the final target is built.
             tags = kwargs.get("tags", []) + ["manual"],
-            # Always propagate the testonly attribute
-            testonly = kwargs.get("testonly", False),
+            testonly = kwargs.get("testonly"),
         )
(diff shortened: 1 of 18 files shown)


js/private/test/launcher/BUILD.bazel
```diff
@@ -119,7 +119,6 @@ js_binary(
 
 js_run_binary(
     name = "exit_zero_run",
-    testonly = True,
```
jbedard, line 122: Now the `package(default_testonly = True)` at the top of this file actually works...

js/private/js_run_binary.bzl
```diff
@@ -378,8 +375,7 @@ See https://github.com/aspect-build/rules_js/tree/main/docs#using-binaries-publi
             srcs = [":{}".format(js_runfiles_lib_name)],
             # Always tag the target manual since we should only build it when the final target is built.
             tags = kwargs.get("tags", []) + ["manual"],
-            # Always propagate the testonly attribute
-            testonly = kwargs.get("testonly", False),
+            testonly = kwargs.get("testonly"),
```
thesayyn, line 378: Ahhh, this is really nice!

thesayyn (approved):

PR #2948, opened 2026-07-25
feat: support the npmrc tokenHelper auth setting

sethvargo:
Adds support for pnpm's [`tokenHelper`](https://pnpm.io/npmrc#urltokenhelper) npmrc setting. `tokenHelper` lets an executable print the registry auth token, which matters for registries that issue short-lived tokens instead of static `_authToken` values.

rules_js reads it from `npmrc`. The path can be absolute or relative to the `.npmrc`, and `${VAR}` expansion still applies. Global `tokenHelper=` and per-registry `//registry/:tokenHelper=` both work. For the same registry, `tokenHelper` wins over `_authToken`.

SECURITY: pnpm only runs `tokenHelper` from user-level config. rules_js can run one from a checked-in `.npmrc`. rules_js already runs repo-controlled executables like lifecycle hooks, so this fits the existing trust model, but the `npmrc` attribute calls it out because it has to be deliberate.

---

Support pnpm's `tokenHelper` npmrc setting. Use it when your registry issues short-lived auth tokens. rules_js runs the helper and uses stdout as the auth token. The helper can be absolute or relative to the `.npmrc`.

### Test plan

- New test cases added


npm/extensions.bzl
```diff
@@ -145,20 +162,19 @@ def _npm_translate_lock_bzlmod(module_ctx, mod, attr, exclude_package_contents_c
...
-        if home_directory:
-            home_npmrc_path = "{}/{}".format(home_directory, ".npmrc")
-            if module_ctx.path(home_npmrc_path).exists:
-                home_npmrc = parse_npmrc(module_ctx.read(home_npmrc_path))
-
-                (registries2, npm_auth2) = npm_translate_lock_helpers.get_npm_auth(home_npmrc, home_npmrc_path, module_ctx)
-                registries.update(registries2)
-                npm_auth.update(npm_auth2)
-        else:
+        auth_npmrc_path = _resolve_npmrc_auth_file(module_ctx, attr)
```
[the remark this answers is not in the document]
sethvargo: Real, but I think it's pre-existing and out of scope here. `_update_pnpm_lock` runs `pnpm install --lockfile-only` with no `environment` set (line 499) - it doesn't forward the `npmrc` attribute or `use_home_npmrc` to that pnpm subprocess either, so the update process has never seen the auth rules_js parses. This PR only wires up the auth rules_js itself uses to fetch packages.

The env var form (`PNPM_CONFIG_NPMRC_AUTH_FILE`) does reach pnpm, since `rctx.execute` inherits the ambient env - it's only the attribute that isn't forwarded, same as every other auth input today. Forwarding auth into the update subprocess is a separate change; I'd rather do it as a follow-up than grow this PR.

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -227,6 +267,21 @@ def _get_npm_auth(npmrc, npmrc_path, rctx):
...
+            # //registry.corp.com/:tokenHelper=/abs/path
+            # registry: registry.corp.com
+            registry = k.removeprefix("//").removesuffix(_NPM_TOKEN_HELPER).removesuffix(":").removesuffix("/")
+            token_helpers[registry] = v
+
+    # tokenHelper runs an executable, so only honor it from a trusted user-level auth
+    # file. Run helpers last so they win over a static token for the same registry.
+    if allow_token_helper:
+        for (registry, helper) in token_helpers.items():
+            token = _run_token_helper(helper, npmrc_path, rctx)
```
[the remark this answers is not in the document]
sethvargo: This is deliberate. Auth is parsed eagerly up front today - `_authToken` and `_auth` all get resolved into the auth dict before any package URL is selected - and I kept `tokenHelper` consistent with that so precedence stays deterministic.

You're right that pnpm runs the helper lazily per request, so a flaky helper for a registry this lockfile doesn't touch would break the build. But making it lazy means threading helpers down into per-URL selection during import generation, which is a bigger cross-cutting change than this PR should carry. And since this only reads the user's own trusted auth file, a broken helper there is a config error I'd rather surface loudly. Leaving it eager for now; lazy execution is a fair follow-up.

e2e/npm_translate_lock_auth/test.sh
```diff
@@ -19,6 +19,12 @@ cp -f .npmrc ~/.npmrc
 rm .npmrc
 _sedi 's#npmrc = "//:.npmrc",#use_home_npmrc = True,#' MODULE.bazel
 
+# Replace the @aspect-build registry token with a tokenHelper that prints it instead,
```
jbedard: Doesn't that remove the testing we were doing? I think we need to essentially clone this e2e test but use the new attribute?

sethvargo: Restored that test to its `_authToken` form in 8e732699 and added a parallel `e2e/npm_translate_lock_auth_token_helper` that drives `tokenHelper` through the new `npmrc_auth_file` attribute. You're right - the in-place swap had quietly dropped the static-token coverage.

sethvargo: Mind taking another look when you get a chance?

npm/private/npm_translate_lock.bzl
```diff
@@ -280,6 +280,23 @@ For more about how to use npm_translate_lock, read [pnpm and rules_js](/docs/pnp
             Because of the repository rule invalidation issue, using the home `.npmrc` is not recommended.
             `.npmrc` settings should generally go in the `npmrc` in your repository so they are shared by all
             developers. The home `.npmrc` should be reserved for authentication settings for private npm repositories.
+
+            The `tokenHelper` auth setting is only run when found in this file (or the file set by
+            `npmrc_auth_file`), never in the `npmrc` file.
+        """),
+        "npmrc_auth_file": attr.string(doc = """
+            Path to the trusted user-level auth file to read when `use_home_npmrc` is True.
```
jbedard: Why can't it be used without `use_home_npmrc`? Why can't it be a local path (which might then run/import something global or read a CI env var etc)?

e2e/npm_translate_lock_auth_token_helper/token-helper.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -o errexit -o nounset -o pipefail
+
+# Print the auth token on stdout. A real helper would mint or refresh a token here.
+printf '%s\n' "${ASPECT_GH_PACKAGES_AUTH_TOKEN:-}"
```
jbedard, line 5: So this depends on CI setting the variable... why do we need the `sed` logic in `test.sh` at all?

Can `npm_translate_lock` not reference this `token-helper.sh` directly and then the regular `bazel test //...` (which depends on CI setting the `ASPECT_GH_PACKAGES_AUTH_TOKEN` var) will work and we can drop `test.sh`?

npm/private/npm_translate_lock.bzl
sethvargo: I was mirroring pnpm's trust model. `tokenHelper` runs an executable, so pnpm only honors it from user-level config and requires an absolute path. That keeps a checked-in `.npmrc` from sneaking in `tokenHelper=/bin/evil`.

But yea, `npmrc_auth_file` already comes from checked-in `MODULE.bazel`, so the boundary is looser here. I lean toward relaxing this and accepting a label or relative path. Before i change it, do you want pnpm-strict or Bazel-native?

e2e/npm_translate_lock_auth_token_helper/token-helper.sh
sethvargo: You're right. This caught a real bug: CI runs `bazel test //...` before optional `test.sh`, so the main leg would hit `__NPMRC_AUTH_FILE__` unsubstituted.

The `sed` is only there because `tokenHelper` and `npmrc_auth_file` need absolute paths we don't know until checkout. If `npmrc_auth_file` is a label and `tokenHelper` resolves relative to it, `npm_translate_lock` can use the checked-in file and plain `bazel test //...` works.

I think this is the same design call as the other thread. Can we settle that first, then I'll drop `test.sh` with it?

jbedard: > I think this is the same design call as the other thread. Can we settle that first, then I'll drop test.sh with it?

Yes, however...

How does the other test workaround this then? Does it just have `tags = ["manual"]` so it doesn't actually test anything on the plain `bazel test //...`? What we do here depends on (a) what the other test does and (b) our decision in the other thread...

npm/extensions.bzl
```diff
@@ -145,20 +162,19 @@ def _npm_translate_lock_bzlmod(module_ctx, mod, attr, exclude_package_contents_c
...
+        if not auth_npmrc_path:
             # buildifier: disable=print
             print("""
 WARNING: Cannot determine home directory in order to load home `.npmrc` file in module extension `npm_translate_lock(name = "{attr_name}")`.
 """.format(attr_name = attr.name))
+        elif module_ctx.path(auth_npmrc_path).exists:
+            home_npmrc = parse_npmrc(module_ctx.read(auth_npmrc_path))
```
jbedard: Don't we need to _execute_ the auth file? Not just `read()` it? Just based on the https://pnpm.io/npmrc#urltokenhelper docs...

jbedard: Actually `_get_npm_auth` already does that... so why do we even need the new `npmrc_auth_file` attribute?

npm/private/npm_translate_lock.bzl
sethvargo: Done in 2840087a. tokenHelper comes from the regular `npmrc` now, and the path can be relative to it, so the gate and absolute-path requirement are gone. The security tradeoff is called out on the `npmrc` attr - a checked-in helper runs during setup, which pnpm doesn't allow.

e2e/npm_translate_lock_auth_token_helper/token-helper.sh
sethvargo: The sibling doesn't use `tags = ["manual"]` - its `bazel test //...` authenticates for real off the checked-in `.npmrc` with `${ENV}` tokens. Same here now in 2840087a: `.npmrc` names a relative `tokenHelper`, `test.sh` is gone, plain `bazel test //...` runs it.

npm/extensions.bzl
sethvargo: Right, it was redundant - `_get_npm_auth` already runs the helper. Dropped `npmrc_auth_file` in 2840087a; tokenHelper reads straight from the `npmrc` file.

npm/private/npm_translate_lock.bzl
```diff
@@ -195,6 +195,14 @@ For more about how to use npm_translate_lock, read [pnpm and rules_js](/docs/pnp
             When set, the `.npmrc` file specified is parsed and npm auth tokens and basic authentication configuration
             specified in the file are passed to the Bazel downloader for authentication with private npm registries.
 
+            A [`tokenHelper`](https://pnpm.io/npmrc#urltokenhelper) is also honored: rules_js runs the executable it
+            names and uses the token it prints on stdout. The helper path may be absolute or relative to this
+            `.npmrc` file, and may reference environment variables.
+
+            SECURITY: unlike pnpm, which only runs a `tokenHelper` from user-level config, rules_js runs one declared
```
jbedard, line 202: Should we revert this by only doing `allow_token_helper = True` for the home npmrc? Does that make the test setup harder?

jbedard:
@codex

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -102,8 +102,44 @@ def _gather_values_from_matching_names(additive, keyed_lists, *names):
...
+        npmrc_dir = str(rctx.path(str(npmrc_path)).dirname)
+        helper = str(rctx.path(npmrc_dir + "/" + helper))
+
+    result = rctx.execute([helper])
+    if result.return_code != 0:
+        fail("tokenHelper \"{helper}\" in \"{npmrc}\" exited with {code}:\nSTDOUT:\n{stdout}\nSTDERR:\n{stderr}".format(
+            helper = helper,
+            npmrc = npmrc_path,
+            code = result.return_code,
+            stdout = result.stdout,
```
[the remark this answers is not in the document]
jbedard: We should probably just drop stdout in this case? Just to be safe...

npm/private/npm_translate_lock_helpers.bzl
```diff
@@ -227,6 +268,21 @@ def _get_npm_auth(npmrc, npmrc_path, rctx):
...
+            registry = k.removeprefix("//").removesuffix(_NPM_TOKEN_HELPER).removesuffix(":").removesuffix("/")
+            token_helpers[registry] = v
+
+    # Run token helpers last so a helper wins over a static token for the same registry.
+    # allow_token_helper guards against the (unused) state.bzl auth parse executing helpers.
+    if allow_token_helper:
+        for (registry, helper) in token_helpers.items():
+            token = _run_token_helper(helper, npmrc_path, rctx)
+            if registry not in auth:
+                auth[registry] = {}
```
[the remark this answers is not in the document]
jbedard: Is this a problem before your change as well and an issue should be logged?

npm/private/npm_translate_lock.bzl
sethvargo: I don't think we should make `allow_token_helper` home-only. A real use case here is a hermetic build where the repo owns the npmrc. If `tokenHelper` only works from $HOME, CI depends on an undeclared file again.

That's why i split these into separate functions before.

The tradeoff is a checked-in `tokenHelper` runs during setup, which pnpm doesn't allow. I think that's acceptable because rules_js already runs repo-controlled executables like lifecycle hooks. But if you want to go back to the home npmrc iteration, i can. Want me to?

npm/private/npm_translate_lock_helpers.bzl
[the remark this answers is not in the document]
sethvargo: Fixed in 08fa6c9a - tokenHelper failures now report only the exit code and stderr, not stdout.

npm/private/npm_translate_lock_helpers.bzl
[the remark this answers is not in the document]
sethvargo: Yes, this was pre-existing, and I fixed it in 08fa6c9a. _select_npm_auth stopped at the first match instead of keeping the longest prefix, so I dropped the break and added a test.

npm/private/npm_translate_lock.bzl
sethvargo: I also think we're possibly going in circles here, so I'd like to align on the overall goals:

1. We have a project that uses npm.
2. We want a project-level npmrc config that is controlled by the project.
3. We want the project to get its auth credentials from a tokenHelper (it happens to be a binary, but it could be any script).

A `git clone` should just work, and we don't want to fuss with the Developer's system.

jbedard: This is fine with me then.

From what I can see/understand I think what we have now is greatly simpler then the original PR. Do you agree?

sethvargo: Yea, i agree. Much simpler.

`tokenHelper` just rides the existing `npmrc` attribute now, without the extra `npmrc_auth_file`/`use_home_npmrc` path or the `test.sh` sed workaround. Thanks for sticking with this one - it's in much better shape.

Is there anything else you'd like to see changed? Or is this ready to merge?

jbedard (approved):

jbedard:
Thanks for the new feature! 👍

PR #2949, opened 2026-07-26
refactor: start moving launcher functionality from bash into JS

acozzette:
We would eventually like to get rid of the bash launcher used by `js_binary`, so this change starts that process by moving some of the easy stuff out of the bash script and into a JS file passed to Node via `--require`.

We already have such a JS file that we currently use primarily for applying patches to `fs`, so this change just adds onto that.

---

### Test plan

### DIFF
--- a/js/private/devserver/src/js_run_devserver.mjs
+++ b/js/private/devserver/src/js_run_devserver.mjs
@@ -13,7 +13,6 @@ import {
 
 // Environment constants
 const {
-    BAZEL_BINDIR,
     JS_BINARY__EXECROOT,
     JS_BINARY__BINDIR,
     JS_BINARY__WORKSPACE,
@@ -22,34 +21,6 @@ const {
 } = process.env
 
 const RUNFILES_ROOT = path.join(JS_BINARY__RUNFILES, JS_BINARY__WORKSPACE)
-
-// node-bootstrap.cjs applies JS_BINARY__CHDIR (if this target has a chdir configured) before this
-// entry_point runs, then clears it, since it's a one-time launcher concern and would otherwise be
-// wrongly re-applied by anything that forks a child process inheriting it (see node-bootstrap.cjs
-// for details). We still need that value here to apply the same chdir inside our own sandbox
-// below, so infer it by comparing our own (already chdir'd) cwd against the roots node-bootstrap
-// would have resolved it relative to: RUNFILES_ROOT (when running from within a test, sandboxed
-// or not) or BAZEL_BINDIR under the execroot (a plain `bazel run`).
-function inferJsBinaryChdir() {
-    const cwd = process.cwd()
-    const bindir = BAZEL_BINDIR || JS_BINARY__BINDIR
-    // RUNFILES_ROOT must be checked first: a target's runfiles directory lives under its own
-    // BAZEL_BINDIR location (".../bin/PACKAGE/target.runfiles"), so in runfiles mode BAZEL_BINDIR
-    // is *also* a valid but overly shallow ancestor of cwd, giving a longer, wrong result.
-    const roots = [
-        RUNFILES_ROOT,
-        path.join(JS_BINARY__EXECROOT, bindir),
-    ]
-    for (const root of roots) {
-        const rel = path.relative(root, cwd)
-        if (rel && !rel.startsWith('..') && !path.isAbsolute(rel)) {
-            return rel
-        }
-    }
-    return undefined
-}
-
-const JS_BINARY__CHDIR = inferJsBinaryChdir()
 
 // Globals
 const syncedTime = new Map()
@@ -476,8 +447,8 @@ async function main(args, sandbox) {
 
     const config = JSON.parse(await fs.promises.readFile(configPath))
 
-    const cwd = JS_BINARY__CHDIR
-        ? path.join(sandbox, JS_BINARY__CHDIR)
+    const cwd = process.env.JS_BINARY__CHDIR
+        ? path.join(sandbox, process.env.JS_BINARY__CHDIR)
         : sandbox
 
     const tool = config.tool
@@ -785,7 +756,7 @@ async function cycleSyncRecurse(cycle, file, isDirectory, sandbox, writePerm) {
         const sandboxMain = path.join(sandbox, JS_BINARY__WORKSPACE)
 
         // Intentionally synchronous; see comment on mkdirpSync
-        mkdirpSync(path.join(sandboxMain, JS_BINARY__CHDIR || ''))
+        mkdirpSync(path.join(sandboxMain, process.env.JS_BINARY__CHDIR || ''))
         await main(process.argv.slice(2), sandboxMain)
     } catch (e) {
         console.error(e)
--- a/js/private/devserver/src/js_run_devserver.mjs
+++ b/js/private/devserver/src/js_run_devserver.mjs
@@ -13,6 +13,7 @@ import {
 
 // Environment constants
 const {
+    BAZEL_BINDIR,
     JS_BINARY__EXECROOT,
     JS_BINARY__BINDIR,
     JS_BINARY__WORKSPACE,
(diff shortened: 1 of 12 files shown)


js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -0,0 +1,103 @@
...
+// on the main thread: cwd is process-wide (already inherited by any worker thread once the main
+// thread has changed it), and process.chdir() throws if called from within a worker thread.
+//
+// A process that inherited NODE_OPTIONS from an already-running js_binary process (for example a
+// worker pool that forks using process.execPath, which we override below to the node wrapper)
+// re-runs this script with the same JS_BINARY__CHDIR still set. JS_BINARY__CHDIR_APPLIED records
+// the value we last chdir'd to for this env lineage, so such a re-invocation is a no-op instead
+// of retrying a chdir that's already been done (or, worse, one relative to a cwd the caller set
+// deliberately, e.g. an explicit `cwd` option on a spawned child unrelated to this package).
+if (JS_BINARY__CHDIR && isMainThread && process.env.JS_BINARY__CHDIR_APPLIED !== JS_BINARY__CHDIR) {
```
acozzette: I don't love the fact that this requires adding yet another `JS_BINARY__` environment variable, but I haven't been able to figure out a better solution yet.

e2e/js_image_oci/README.md
```diff
@@ -34,9 +34,9 @@ app
 |   |-- aspect_rules_js
 |   |   `-- js
 |   |       `-- private
-|   |           `-- node-patches
+|   |           `-- node-bootstrap
```
jbedard, line 37: While I agree with the rename... want to do that first in its own PR so the diff is actually visible to show what you're changing?

js/private/node-bootstrap/bootstrap.cjs
jbedard: Can we `delete process.env["JS_BINARY__CHDIR"]` instead?

e2e/js_image_oci/README.md
acozzette: Sure, I just sent out #2958 with just the rename.

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -0,0 +1,102 @@
+const path = require('path')
```
jbedard, line 1: Any value writing this in typescript alongside `fs.cts` now that it is growing? 🤷

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -0,0 +1,102 @@
...
+const { isMainThread } = require('worker_threads')
```
jbedard: Will worker threads inherit the `--require`? 🤔

js/private/node-bootstrap/bootstrap.cjs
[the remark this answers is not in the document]
jbedard: Is this fixed?

jbedard:
I think we should add `windows-macos` to your branch name so those 2 platforms get triggered on CI for this. Want to change this PR to only do the rename for now and a new branch can make the actual changes with `windows-macos` in the branch name?

EDIT: or now I see #2958 - you can also edit the ci yaml file to always run on windows+macos instead of creating a new branch+PR for now (and revert it before merge) - because I think there's no way to keep the PR (and all the valuable comments 😅) while changing the branch name

js/private/devserver/src/js_run_devserver.mjs
```diff
@@ -13,6 +13,7 @@ import {
 
 // Environment constants
 const {
+    BAZEL_BINDIR,
```
jbedard: Is this to move toward dropping `JS_BINARY__BINDIR`? Should we use `BINDIR` instead to also move toward dropping `BAZEL_BINDIR`?

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -17,6 +24,38 @@ if (!process.env.JS_BINARY__NODE_PATCHES_DEPTH) {
...
+              JS_BINARY__CHDIR
+          )
+        : JS_BINARY__CHDIR
+    if (JS_BINARY__LOG_DEBUG) {
+        console.error(
+            `DEBUG: ${JS_BINARY__LOG_PREFIX}: changing directory to user specified package ${JS_BINARY__CHDIR}`
+        )
+    }
+    process.chdir(target)
+    // Clear this so a process that inherits NODE_OPTIONS from this one (for example a worker
```
jbedard: Tell claude to stfu. I assume this can be summarized in 1-2 lines? Same with a lot of comments...

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -0,0 +1,103 @@
...
+// fs patches
+if (JS_BINARY__PATCH_NODE_FS && JS_BINARY__PATCH_NODE_FS != '0') {
+    // Configure fs patch roots for node fs patches which are run via --require in the node
+    // wrapper. Don't override an already-set JS_BINARY__FS_PATCH_ROOTS in case a js_binary
+    // such as js_run_deverser runs another js_binary tool.
+    const fsPatchRoots =
+        JS_BINARY__FS_PATCH_ROOTS ||
+        (JS_BINARY__EXECROOT && JS_BINARY__RUNFILES
+            ? `${JS_BINARY__EXECROOT}:${JS_BINARY__RUNFILES}`
+            : undefined)
```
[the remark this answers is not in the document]
jbedard: ☝️is that ok? What if one `js_binary` invokes another via some ugly `js_binary(args=$(location :other_binary)` etc...

js/private/node-bootstrap/bootstrap.cjs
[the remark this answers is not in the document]
acozzette: I think this is still a problem. I got rid of `JS_BINARY__CHDIR_APPLIED` in favor of unsetting `JS_BINARY__CHDIR`, but I think Codex is still correct that this will not work correctly if the js_binary spawns another js_binary launcher. I need to think about the best way to fix this.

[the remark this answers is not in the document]
acozzette: The `chdir` functionality seems pretty complicated, so I decided to leave that out of this PR for now.

js/private/node-bootstrap/bootstrap.cjs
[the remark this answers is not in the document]
acozzette: I think this is a real problem, but I have now fixed it by exporting `JS_BINARY__FS_PATCH_ROOTS` so that this propagates to any other binaries that might be called.

js/private/node-bootstrap/bootstrap.cjs
acozzette: That appeared to be the case, though luckily we don't have to worry about it for now since I've removed the chdir change from this PR.

js/private/devserver/src/js_run_devserver.mjs
acozzette: The rationale for this was just to have js_run_devserver.mjs infer `JS_BINARY__CHDIR` so that we could avoid needing to propagate that variable. But since I gave up on changing chdir in this PR, I have reverted the changes to this file.

js/private/node-bootstrap/bootstrap.cjs
acozzette: Yeah, some of Claude's comments were a bit verbose, but I think I got them under control now.

js/private/node-bootstrap/bootstrap.cjs
acozzette: That SGTM but could I send that as a follow-up just to make the diff easier to read?

acozzette:
> I think we should add `windows-macos` to your branch name so those 2 platforms get triggered on CI for this. Want to change this PR to only do the rename for now and a new branch can make the actual changes with `windows-macos` in the branch name?
> 
> EDIT: or now I see #2958 - you can also edit the ci yaml file to always run on windows+macos instead of creating a new branch+PR for now (and revert it before merge) - because I think there's no way to keep the PR (and all the valuable comments 😅) while changing the branch name

I edited the yaml file to temporarily run on Mac and Windows, and confirmed that all the tests passed.

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -29,13 +32,36 @@ if (process.platform == 'win32') {
...
+// Put the node wrapper directory first on the PATH so that child processes resolve `node` to it.
+if (JS_BINARY__NODE_WRAPPER) {
+    const wrapperDir = path.dirname(JS_BINARY__NODE_WRAPPER)
+    const entries = (process.env.PATH || '')
+        .split(path.delimiter)
+        .filter(Boolean)
+
+    // Child processes re-run this bootstrap, so don't keep re-prepending the same directory.
+    if (entries[0] !== wrapperDir) {
+        process.env.PATH = [wrapperDir, ...entries].join(path.delimiter)
```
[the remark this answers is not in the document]
acozzette: This seems to me like a theoretical issue that we can probably live with. We're moving toward replacing the bash launcher with hermetic-launcher, and that is not going to give us any way to update PATH before invoking Node.

Issue #2933, opened 2026-07-22
bazel coverage empty for first-party code imported by name through an npm_package


jbedard:
Split off from #2901 / #2932 ([comment](https://github.com/aspect-build/rules_js/pull/2932#issuecomment-5041637627)). The split-postprocessing fix in #2932 does **not** cover this; it is a distinct root cause.

### Symptom

A `js_test` that imports first-party code **by package name**, where the package is provided via `npm_package` + `npm_link_package`, produces an empty (0%) `coverage.dat` — even though the code runs. The same code imported via a `js_library` linked by name, or via a relative `require("./…")`, reports correctly.

### Root cause

The coverage manifest lists the instrumented source path (e.g. `pkg/lib.js`), but `npm_package` **repackages** the sources into the `.aspect_rules_js` store, so node executes — and V8 records — the store realpath:

```
…/runfiles/_main/node_modules/.aspect_rules_js/@repro+lib@0.0.0/node_modules/@repro/lib/lib.js
```

c8 drops it twice over: it is under `node_modules` (excluded by default) and its path shares no token with the manifest entry (`pkg/lib.js`) beyond the basename. A `js_library` avoids this because its store entry is a symlink to the real directory and node (no `--preserve-symlinks`) resolves back to the manifest path. Contrast confirmed from live V8 data:

| Provisioning | V8 URL (relative to c8 `src`) | Result |
| --- | --- | --- |
| relative `require("./lib.js")` | `lib.js` | counted |
| first-party `js_library` | `pkg/lib.js` | counted |
| first-party `npm_package` | `node_modules/.aspect_rules_js/@repro+lib@0.0.0/node_modules/@repro/lib/lib.js` | dropped |

### Why it cannot be fixed in `coverage.js` alone

Attributing the store file to `pkg/lib.js` needs a store→source path mapping, which is not derivable from the paths (package name `@repro/lib` ≠ bazel package `pkg`) and is not present in the test's runfiles (only the repackaged copy is). The fix is rule-level.

### Fix directions (needs a decision on the `SF:` convention)

1. **Report against the (normalized) store/package path** — the rule passes the first-party package-store directories to `coverage.js`, which includes them and overrides the `node_modules` exclusion for exactly those. `SF:` becomes the package path rather than `pkg/lib.js`.
2. **Remap to the source path** — thread a store→source map and add the first-party sources to the coverage inputs so c8 can read them for line/branch structure.

### Repro

[...]

PR #2953, opened 2026-07-26
test: e2e coverage workspace covering the supported scenarios

jbedard:
Records what `bazel coverage` produces today for a js_test: a relative import, a first-party js_library linked by name, and a passing test with a non-zero expected_exit_code. Reports are checked by looking for the covered/uncovered function names rather than an SF: path, so the assertions survive changes to how sources resolve.

Two scenarios are asserted as not working, so that fixing either one has to touch this file:

- Split post-processing (--experimental_split_coverage_postprocessing, required for remote execution) reports empty coverage for every target, #2901.
- First-party code repackaged with npm_package and imported by name resolves to a copy in the store with no link back to source, #2933. That one is a documented limitation rather than a bug; use js_library linking instead.

### Test plan

- New test cases added

### DIFF
--- a/e2e/coverage/test.sh
+++ b/e2e/coverage/test.sh
@@ -1,0 +1,111 @@
+#!/usr/bin/env bash
+set -o errexit -o nounset -o pipefail
+
+# Coverage behavior for `bazel test` / `bazel coverage`.
+#
+# Every target executes the same covered()/uncovered() code and differs only in how
+# the source is provided or the test exits — see BUILD.bazel for what each exercises.
+#
+# //:first_party_npmpkg_test is an UNSUPPORTED scenario, kept as executable
+# documentation: first-party code repackaged with npm_package runs from a copy in
+# the .aspect_rules_js store with no link back to source, so coverage comes out
+# empty. Use js_library linking (//:first_party_jslib_test) for first-party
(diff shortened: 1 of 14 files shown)


e2e/coverage/test.sh
```diff
@@ -0,0 +1,111 @@
...
+bazel coverage --nocache_test_results "${ALL_TARGETS[@]}"
+for t in "${WORKING_TARGETS[@]}"; do assert_coverage "$t" inline; done
+for t in "${UNSUPPORTED_TARGETS[@]}"; do check_unsupported "$t" inline; done
+
+bazel coverage --nocache_test_results \
+    --experimental_split_coverage_postprocessing \
+    --experimental_fetch_all_coverage_outputs \
+    "${ALL_TARGETS[@]}"
+readonly ISSUE_2901=https://github.com/aspect-build/rules_js/issues/2901
+for t in "${WORKING_TARGETS[@]}"; do check_broken "$t" split "$ISSUE_2901"; done
```
jbedard, line 110: Note how this says `check_broken` and is what the followup actually fixes, while the `check_unsupported` below might never be fixed: https://github.com/aspect-build/rules_js/issues/2933

acozzette (approved):

PR #2954, opened 2026-07-28
test: assert coverage merger behavior in bash, not appended JS

jbedard:
The mergers here were built by genrule-appending assertions onto coverage.js. That never worked for the case the `fail` test was meant to cover: coverage.js calls process.exit(0) as soon as it sees a non-empty COVERAGE_OUTPUT_FILE, so execution never reached the appended process.on('exit') line and no handler was ever registered. The test only proved the merger does not crash.

Assert in bash instead, appended to the merger's launcher after it runs, where COVERAGE_DIR and COVERAGE_OUTPUT_FILE are both set and no early exit can skip it. The case is now covered by a real test program that writes its own report, so `fail` becomes `preexisting` after what it exercises: a report the test wrote itself must survive. See https://github.com/aspect-build/rules_js/pull/430.

### Test plan

- New test cases added


js/private/coverage/coverage.sh.tpl
```diff
@@ -55,3 +55,5 @@ fi
 # ==============================================================================
 
 "$node" "$entry_point"
```
jbedard, line 57: The [next PR](https://github.com/aspect-build/rules_js/pull/2932) removes this node invocation and replaces it with raw bash, that's the reason for moving the test assertions to bash as a pre-factor...

acozzette (approved):

PR #2956, opened 2026-07-29
feat: add js_run_binary_action helper for invoking js_binary from custom rules

acozzette:
js_binary's launcher requires BAZEL_BINDIR to be set when invoked as a build tool, but rule authors writing their own ctx.actions.run call currently have to deal with this implementation detail manually. Add js_run_binary_action() (js/libs.bzl, implemented in js/private/js_binary.bzl) to hide it, and update examples/js_binary/custom_rule.bzl and examples/worker/defs.bzl's pi_rule to use it.

The main motivation for this is that adding proper support for path mapping is going to require conveying the output bin directory via a `--bazel-bindir` command-line flag rather than an environment variable. Ideally we do not want users to specify that flag directly, but as long as they call our helper function we can set the flag there ourselves.

Another issue is that there are custom rules setting environment variables that we intended to be private, such as `JS_BINARY__STDOUT_OUTPUT_FILE`. If we can get all such rules to use our helper instead, then we can potentially have more flexibility to change how we handle these private variables.

---

Add the `js_run_binary_action` helper for invoking a `js_binary` from a rule implementation

### Test plan

### DIFF
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -591,6 +591,29 @@ def _js_binary_impl(ctx):
         ),
     ]
 
+def js_run_binary_action(ctx, **kwargs):
+    """Runs a `js_binary` as a tool in a custom rule's action.
+
+    `js_binary`'s launcher script requires the `BAZEL_BINDIR` environment variable to be
+    set to the root of the Bazel output tree whenever it is run as a build tool (as
+    opposed to via `bazel run`). This wraps `ctx.actions.run`, setting `BAZEL_BINDIR`
+    correctly so rule authors invoking a `js_binary` (or similar) executable don't have
+    to know this detail.
+
+    We recommend that rule authors use this helper function whenever possible, instead of
+    directly invoking a `js_binary`.
+
+    Args:
+        ctx: the rule context
+        **kwargs: additional arguments forwarded to `ctx.actions.run`, e.g. `executable`,
+            `arguments`, `inputs`, `outputs`, `mnemonic`, `execution_requirements`, `env`
+    """
+    env = kwargs.pop("env", None) or {}
+    ctx.actions.run(
+        env = dict(env) | {"BAZEL_BINDIR": ctx.bin_dir.path},
+        **kwargs
+    )
+
 js_binary_lib = struct(
     attrs = _ATTRS,
     create_launcher = _create_launcher,
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -594,14 +594,9 @@ def _js_binary_impl(ctx):
 def js_run_binary_action(ctx, **kwargs):
     """Runs a `js_binary` as a tool in a custom rule's action.
 
-    `js_binary`'s launcher script requires the `BAZEL_BINDIR` environment variable to be
-    set to the root of the Bazel output tree whenever it is run as a build tool (as
-    opposed to via `bazel run`). This wraps `ctx.actions.run`, setting `BAZEL_BINDIR`
-    correctly so rule authors invoking a `js_binary` (or similar) executable don't have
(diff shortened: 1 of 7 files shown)


js/private/js_binary.bzl
```diff
@@ -591,6 +591,26 @@ def _js_binary_impl(ctx):
...
+
+    `js_binary`'s launcher script requires the `BAZEL_BINDIR` environment variable to be
+    set to the root of the Bazel output tree whenever it is run as a build tool (as
+    opposed to via `bazel run`). This wraps `ctx.actions.run`, setting `BAZEL_BINDIR`
+    correctly so rule authors invoking a `js_binary` (or similar) executable don't have
+    to know this detail.
+
+    Args:
+        ctx: the rule context
+        env: a dict of extra environment variables to set; merged with `BAZEL_BINDIR`
```
jbedard: The idea is that we will hopefully replace this with `args` and appending `--bazel-bin` and no one will notice?

If so... lets drop it from the method signature and arguments here and do `kwargs.pop("env", {})` instead so the method signature + docs don't change when we switch from `env` to `args`?

js/private/js_binary.bzl
```diff
@@ -591,6 +591,26 @@ def _js_binary_impl(ctx):
         ),
     ]
 
+def run_js_binary_action(ctx, env = {}, **kwargs):
```
jbedard: Should we name it `js_run_binary_action`? 🤔

jbedard: Robot ideas:


> js_run_binary_action(ctx, ...)
> 
> Matches convention already in repo and upstream:
> - copy_js_file_to_bin_action(ctx, file) — js/private/js_helpers.bzl:77
> - bazel-lib: copy_file_action, copy_directory_action, copy_to_directory_action — action twins of same-named rules
> 
> So js_run_binary (rule) : js_run_binary_action (action) reads as sibling pair, zero guessing for rules_ts/rules_jest authors.
> 
> Alternatives, weaker:
> - run_js_binary — verb-first reads nicer, but breaks the <rule>_action pairing and invites confusion with js_run_binary
> - js_binary_action — hides that it's the action form of js_run_binary, not of js_binary
> - js_helpers.run_binary namespaced — fine if you want a struct, but repo exports flat helpers today
> 
> Export surface: js/libs.bzl (helper API) not js/defs.bzl (rules). If you want it grouped, add to js_lib_helpers struct as .run_binary_action.

js/private/js_binary.bzl
acozzette: Yes, that's what I was thinking. Sounds good, I went and made the change to remove `env` from the function signature.

js/private/js_binary.bzl
acozzette: `js_run_binary_action` sounds good to me, so I renamed it to that.

js/private/js_binary.bzl
```diff
@@ -591,6 +591,29 @@ def _js_binary_impl(ctx):
         ),
     ]
 
+def js_run_binary_action(ctx, **kwargs):
+    """Runs a `js_binary` as a tool in a custom rule's action.
+
+    `js_binary`'s launcher script requires the `BAZEL_BINDIR` environment variable to be
```
jbedard: Do we want to say what the script explicitly requires? I think the main reason for this helper is so that people don't know it what it requires and we can change from `env` to `args` without updating the docs?

jbedard (approved):

js/private/js_binary.bzl
acozzette: Good point, I updated the docstring to just say that it handles implementation details, but avoid saying exactly what it does.

jbedard (approved):

PR #2959, opened 2026-07-29
refactor: strip test coverage logic from js_binary.sh launcher for non-test targets

jbedard:
This should make all `js_binary` shell scripts smaller while keeping `js_test` the same.

### Test plan

- New test cases added


jbedard:
Is this worth it? Will this (conditionally adding stuff to shell scripts) be a good thing long term or become a PITA?

acozzette:
> Is this worth it? Will this (conditionally adding stuff to shell scripts) be a good thing long term or become a PITA?

I kind of lean toward thinking it's not worth it. It adds some extra complexity to save a little bit of size in every `js_binary` launcher, but it's not clear whether there's a significant performance benefit. If we had some benchmark results then that might be convincing, though. What do you think?

jbedard:
This will have no performance impact at all really, it just removes a snippet of coverage logic from non-test targets.

jbedard:
not worth it for now

PR #2960, opened 2026-07-30
fix: js_run_binary_action drops --bazel-bindir value when outputs[0] …

acozzette:
…is a directory

Args.add_all's default expand_directories=True expands a declared directory into its (not-yet-known, empty at analysis time) contents before map_each runs, instead of calling map_each on the directory itself. This silently dropped the --bazel-bindir argument's value whenever outputs[0] was a directory output.

This change adds test coverage for directory outputs by updating the test rule in e2e/path_mapping to produce a directory instead of a file. We have test coverage elsewhere for rules that produce individual files, so we're not losing any coverage by tweaking the test this way.

---

### Test plan

- New test cases added

### DIFF
--- a/e2e/path_mapping/bindir_path_mapping_check/bindir_path_mapping_check.bzl
+++ b/e2e/path_mapping/bindir_path_mapping_check/bindir_path_mapping_check.bzl
@@ -4,7 +4,7 @@
 load("@aspect_rules_js//js:libs.bzl", "js_run_binary_action")
 
 def _bindir_path_mapping_check_impl(ctx):
-    output = ctx.actions.declare_file(ctx.label.name + ".ok")
+    output = ctx.actions.declare_directory(ctx.label.name)
 
     # Baking the value of this env var (set via --action_env) into the action
     # allows test.sh to invalidate this one action without the need for a full
--- a/js/private/js_binary.bzl
+++ b/js/private/js_binary.bzl
@@ -615,7 +615,10 @@ def js_run_binary_action(ctx, **kwargs):
         fail("js_run_binary_action requires at least one output")
     extra_args = ctx.actions.args()
     extra_args.add("--bazel-bindir")
-    extra_args.add_all([outputs[0]], map_each = _bazel_bindir_arg)
+
+    # Set expand_directories = False to ensure this works correctly if
+    # outputs[0] is a directory.
+    extra_args.add_all([outputs[0]], map_each = _bazel_bindir_arg, expand_directories = False)
 
     ctx.actions.run(
         arguments = [extra_args] + (kwargs.pop("arguments", None) or []),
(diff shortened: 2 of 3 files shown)


e2e/path_mapping/bindir_path_mapping_check/bindir_path_mapping_check.bzl
```diff
@@ -4,7 +4,7 @@
 load("@aspect_rules_js//js:libs.bzl", "js_run_binary_action")
 
 def _bindir_path_mapping_check_impl(ctx):
-    output = ctx.actions.declare_file(ctx.label.name + ".ok")
+    output = ctx.actions.declare_directory(ctx.label.name)
```
jbedard, line 7: Should we not have tests for both?

acozzette: We do, we have the file case covered in examples/js_binary/custom_rule.bzl and examples/worker/defs.bzl.

js/private/js_binary.bzl
```diff
@@ -615,7 +615,10 @@ def js_run_binary_action(ctx, **kwargs):
         fail("js_run_binary_action requires at least one output")
     extra_args = ctx.actions.args()
     extra_args.add("--bazel-bindir")
-    extra_args.add_all([outputs[0]], map_each = _bazel_bindir_arg)
+
+    # Set expand_directories = False to ensure this works correctly if
+    # outputs[0] is a directory.
+    extra_args.add_all([outputs[0]], map_each = _bazel_bindir_arg, expand_directories = False)
```
jbedard, line 621: Today do any tests fail if `expand_directories = True`? Lets ensure we have one.

I think that would just be a case where an action does `declare_directory` and then puts >1 file in it?

acozzette: We don't have any such tests today, but I updated e2e/path_mapping/bindir_path_mapping_check/bindir_path_mapping_check.bzl to do that and ensured that it fails if we change `expand_directories` back to `True`.

jbedard: Wouldn't we need to update `e2e/path_mapping/bindir_path_mapping_check/bindir_check.mjs` to output more then one file to the directory output to test that? Or have you not pushed something?

acozzette: Oh sorry, I guess I didn't read your comment carefully. It only writes one file. However, empirically that is enough to trigger the bug. I think that makes sense because the bug was an analysis-time issue where we were not creating the command-line args correctly, so it shouldn't matter how many files we create at execution time. I can have it create more than one if you think that is more realistic, though. What do you think?

jbedard: I was wondering if `e2e/path_mapping/bindir_path_mapping_check/bindir_check.mjs` writes two files and have `expand_directories = True` will it list both files? Then the bash script reading `--bazel-bin first-out second-out` will only read 1/2 of those and the other will be passed through to the real binary...?

I think just adding a one extra output file within the directory output would make it a bit more realistic of an "output directory" and satisfy my request...

acozzette: Sure, I went ahead and updated that to output two files to the directory.

jbedard (approved):

PR #2962, opened 2026-08-01
feat(js_run_binary): add support for path mapping

acozzette:
This change adds support for path mapping in `js_run_binary` by updating it to use the new `--bazel-bindir` flag instead of setting the `BAZEL_BINDIR` environment variable. The only mechanism Bazel provides for determining path-mapped paths is by adding inputs directly to an `Args` object from `ctx.actions.args()`. We can now compute the output Bazel bin directory that way using a special case added to `run_binary()` for this purpose in https://github.com/bazel-contrib/bazel-lib/pull/1269.

The `js_run_binary` macro currently sets several other problematic environment variables: BAZEL_BUILD_FILE_PATH, BAZEL_COMPILATION_MODE, BAZEL_TARGET_CPU, BAZEL_TARGET, BAZEL_WORKSPACE, BAZEL_PACKAGE, and BAZEL_TARGET_NAME. Most of these are determined by evaluating Make variables, and `run_binary` conservatively assumes that path mapping is not safe if any location or Make variable expansion occurs. This change guards those variables behind a new `set_legacy_environment_variables` parameter (default `True`) so they can be disabled without a breaking change, ahead of flipping the default in a future release. To enable path mapping on a particular `js_run_binary` target, make sure to set this parameter to `False`.

---

Support for path mapping in `js_run_binary`

### Test plan

- New test cases added

### DIFF
--- a/js/private/js_run_binary.bzl
+++ b/js/private/js_run_binary.bzl
@@ -309,6 +309,7 @@ def js_run_binary(
     # These environment variables are deprecated and will default to not being set in a future
     # release; see the `set_legacy_environment_variables` docstring.
     if set_legacy_environment_variables:
+        fixed_env["BAZEL_BINDIR"] = "$(BINDIR)"
         fixed_env["BAZEL_BUILD_FILE_PATH"] = "$(BUILD_FILE_PATH)"
         fixed_env["BAZEL_COMPILATION_MODE"] = "$(COMPILATION_MODE)"
         fixed_env["BAZEL_TARGET_CPU"] = "$(TARGET_CPU)"
--- a/js/private/js_run_binary.bzl
+++ b/js/private/js_run_binary.bzl
@@ -54,8 +54,8 @@ def js_run_binary(
     The following environment variables are made available to the Node.js runtime based on available Bazel [Make variables](https://bazel.build/reference/be/make-variables#predefined_variables):
 
     * BAZEL_BINDIR: the bazel bin directory; equivalent to the `$(BINDIR)` Make variable of the `js_run_binary` target
-    * BAZEL_TARGET_CPU (legacy; see `set_legacy_environment_variables`): the target cpu architecture; equivalent to `$(TARGET_CPU)` Make variable of the `js_run_binary` target
     * BAZEL_COMPILATION_MODE (legacy; see `set_legacy_environment_variables`): One of `fastbuild`, `dbg`, or `opt` as set by [`--compilation_mode`](https://bazel.build/docs/user-manual#compilation-mode); equivalent to `$(COMPILATION_MODE)` Make variable of the `js_run_binary` target
+    * BAZEL_TARGET_CPU (legacy; see `set_legacy_environment_variables`): the target cpu architecture; equivalent to `$(TARGET_CPU)` Make variable of the `js_run_binary` target
 
     The following environment variables are made available to the Node.js runtime based on the rule context:
 
--- a/js/private/js_run_binary.bzl
+++ b/js/private/js_run_binary.bzl
@@ -17,6 +17,24 @@ load(":js_helpers.bzl", _envs_for_log_level = "envs_for_log_level")
 load(":js_info_files.bzl", _js_info_files = "js_info_files")
 load(":js_library.bzl", _js_library = "js_library")
 
+def _build_file_path():
+    """Returns the package-relative path to the current package's BUILD file.
+
+    Equivalent to `ctx.build_file_path`, but computed at macro-evaluation time since macros have
+    no `ctx`. The build file's name isn't known in advance (`BUILD` or `BUILD.bazel`), so this
+    uses `glob()` to detect it; if both are present, `BUILD.bazel` wins, matching Bazel's own
+    precedence.
+    """
+    matches = native.glob(["BUILD.bazel", "BUILD"], allow_empty = True)
+    if "BUILD.bazel" in matches:
+        build_file_name = "BUILD.bazel"
+    elif "BUILD" in matches:
+        build_file_name = "BUILD"
+    else:
+        fail("Could not find a BUILD or BUILD.bazel file in package '{}'".format(native.package_name()))
+    package = native.package_name()
+    return "{}/{}".format(package, build_file_name) if package else build_file_name
+
 def js_run_binary(
         name,
         tool,
@@ -61,9 +79,9 @@ def js_run_binary(
 
     * BAZEL_PACKAGE: the package of the bazel target being run; equivalent to `ctx.label.package` of the `js_run_binary` target's rule context
     * BAZEL_TARGET_NAME: the full label of the bazel target being run; a stringified version of `ctx.label` of the `js_run_binary` target's rule context
-    * BAZEL_BUILD_FILE_PATH (legacy; see `set_legacy_environment_variables`): the path to the BUILD file of the bazel target being run; equivalent to `ctx.build_file_path` of the `js_run_binary` target's rule context
-    * BAZEL_TARGET (legacy; see `set_legacy_environment_variables`): the name of the bazel target being run; equivalent to `ctx.label.name` of the  `js_run_binary` target's rule context
-    * BAZEL_WORKSPACE (legacy; see `set_legacy_environment_variables`): the bazel repository name; equivalent to `ctx.workspace_name` of the `js_run_binary` target's rule context
+    * BAZEL_BUILD_FILE_PATH: the path to the BUILD file of the bazel target being run; equivalent to `ctx.build_file_path` of the `js_run_binary` target's rule context
+    * BAZEL_TARGET: the name of the bazel target being run; equivalent to `ctx.label.name` of the  `js_run_binary` target's rule context
+    * BAZEL_WORKSPACE: the bazel repository name; equivalent to `ctx.workspace_name` of the `js_run_binary` target's rule context
 
     Args:
         name: Target name
@@ -254,9 +272,8 @@ def js_run_binary(
 
             Refer to https://bazel.build/rules/lib/builtins/actions#run for more details.
 
-        set_legacy_environment_variables: Whether to set the legacy `BAZEL_BUILD_FILE_PATH`,
-            `BAZEL_COMPILATION_MODE`, `BAZEL_TARGET_CPU`, `BAZEL_TARGET` and `BAZEL_WORKSPACE`
-            environment variables.
+        set_legacy_environment_variables: Whether to set the legacy `BAZEL_COMPILATION_MODE`
+            and `BAZEL_TARGET_CPU` environment variables.
 
             These variables are deprecated and setting them will default to False in a future
             release. Set this to False to opt out now.
@@ -305,19 +322,27 @@ def js_run_binary(
         extra_srcs.append(":{}".format(copy_to_bin_name))
 
     # Automatically add common and useful make variables to the environment for js_run_binary build targets
+    label = native.package_relative_label(name)
     fixed_env = {
+        "BAZEL_BUILD_FILE_PATH": _build_file_path(),
         "BAZEL_PACKAGE": native.package_name(),
         "BAZEL_TARGET_NAME": name,
+        "BAZEL_TARGET": "{}//{}:{}".format(
+            "@" + label.repo_name if label.repo_name else "",
+            label.package,
+            label.name,
+        ),
+        # This variable used to be derived from ctx.workspace_name and was the
+        # name from the WORKSPACE file, but with bzlmod it is always just
+        # "_main".
+        "BAZEL_WORKSPACE": "_main",
     }
 
     # These environment variables are deprecated and will default to not being set in a future
     # release; see the `set_legacy_environment_variables` docstring.
     if set_legacy_environment_variables:
-        fixed_env["BAZEL_BUILD_FILE_PATH"] = "$(BUILD_FILE_PATH)"
         fixed_env["BAZEL_COMPILATION_MODE"] = "$(COMPILATION_MODE)"
         fixed_env["BAZEL_TARGET_CPU"] = "$(TARGET_CPU)"
-        fixed_env["BAZEL_TARGET"] = "$(TARGET)"
-        fixed_env["BAZEL_WORKSPACE"] = "$(WORKSPACE)"
 
     # Configure working directory to `chdir` is set
     if chdir != None:
--- a/js/private/js_run_binary.bzl
+++ b/js/private/js_run_binary.bzl
@@ -45,6 +45,7 @@ def js_run_binary(
         patch_node_fs = True,
         allow_execroot_entry_point_with_no_copy_data_to_bin = False,
         use_default_shell_env = False,
+        set_legacy_environment_variables = True,
         **kwargs):
     """Wrapper around @bazel_lib `run_binary` that adds convenience attributes for using a `js_binary` tool.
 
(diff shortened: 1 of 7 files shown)


js/private/js_run_binary.bzl
```diff
@@ -298,16 +306,19 @@ def js_run_binary(
 
     # Automatically add common and useful make variables to the environment for js_run_binary build targets
     fixed_env = {
-        "BAZEL_BINDIR": "$(BINDIR)",
-        "BAZEL_BUILD_FILE_PATH": "$(BUILD_FILE_PATH)",
-        "BAZEL_COMPILATION_MODE": "$(COMPILATION_MODE)",
         "BAZEL_PACKAGE": native.package_name(),
-        "BAZEL_TARGET_CPU": "$(TARGET_CPU)",
         "BAZEL_TARGET_NAME": name,
-        "BAZEL_TARGET": "$(TARGET)",
```
jbedard, line 307: Why would `TARGET`, `WORKSPACE` or `BUILD_FILE_PATH` effect path-mapping?

js/private/js_run_binary.bzl
```diff
@@ -253,6 +254,13 @@ def js_run_binary(
 
             Refer to https://bazel.build/rules/lib/builtins/actions#run for more details.
 
+        set_legacy_environment_variables: Whether to set the legacy `BAZEL_BUILD_FILE_PATH`,
```
jbedard: Should we aim to drop these in 4.x? Aim to make `js_run_binary` as close as possible to the `run_binary` API?

Or instead reword this to be something such as `supports_path_mapping = False` or `cross_compilation = True`?

js/private/js_run_binary.bzl
acozzette: The information those variables contain is not inherently hostile to path mapping, but the problem is that `run_binary` conservatively assumes that any Make variable expansion is unsafe for path mapping since it might produce an invalid path. We got around this for `$(BINDIR)` by adding a special case in `run_binary`, but we don't have that ability for `$(TARGET)` or the others.

js/private/js_run_binary.bzl
acozzette: Yes, I think it would be good to flip the default in 4.x. If we want to drop even more variables then we could look at dropping `BAZEL_PACKAGE` and `BAZEL_TARGET_NAME` as well, but I left those alone since they're not implemented with Make variables which are problematic for path mapping.

I hesitate to reword this as `supports_path_mapping` because unsetting these variables is not strictly necessary or sufficient for path mapping. You could opt into path mapping with `execution_requirements = {"supports-path-mapping": "1"}` even with these variables set, if you know it's safe. It would not perform well in terms of caching, but technically path mapping would be enabled. Or even if you do unset these variables, it's not necessarily safe to enable path mapping since the particular `js_run_binary` target might have other problematic expansions such as `$(execpath ...)`.

js/private/js_run_binary.bzl
jbedard: We don't we have that ability for `$(TARGET)` or the others? Does `run_binary` just have to add those?

js/private/js_run_binary.bzl
jbedard: Don't we still need a permanent way to opt-in/out of path-mapping? `set_legacy_environment_variables` sounds different then that?

js/private/js_run_binary.bzl
acozzette: There's actually another problem, which is that Bazel doesn't provide any way to do path mapping for environment variables, so if we wanted to support these other variables then we would have to move them onto the command line similar to what we're doing with `--bazel-bindir`.

We could in principle support these other variables in `run_binary`, but I suspect it wouldn't be worth the trouble. With `$(BINDIR)` we have a very limited special case which is that we handle it in a path-mapping-friendly way only for an arg exactly equal to the string `$(BINDIR)` (so something like `--x=$(BINDIR)` will not work). It might be tough to justify doing the same thing for variables that aren't strictly necessary.

Some of these other variables we could potentially support without help from `run_binary`. E.g. we could maybe compute `BAZEL_WORKSPACE` with something like `native.repo_name()` if it seems important enough. But I think `TARGET_CPU` and especially `COMPILATION_MODE` we should explicitly try to avoid, since those will reduce the caching benefits of path mapping.

js/private/js_run_binary.bzl
acozzette: `run_binary` will already enable path mapping if it can determine that it's safe to do so, and it doesn't provide a way to opt out. So my thought is that we can just rely on that existing logic. And then in some cases we may need to explicitly opt in because we know it's safe even though `run_binary` can't determine that, and in that situation we just need to set `execution_requirements = {"supports-path-mapping": "1"}` on the `js_run_binary`.

js/private/js_run_binary.bzl
acozzette: Actually scratch what I said about needing to move any variables to the command line. I remembered that this is not relevant for any of the variables here since none of them contain paths that could need to be path-mapped. The issue instead is just that `run_binary` conservatively assumes that any variable expansion *might* result in a path that's invalid with respect to path mapping, so the use of these variables such as `$(TARGET)` will cause it to disable path mapping. That's still a bit awkward to fix in `run_binary` without trying to parse these variable expressions, though.

jbedard: I think anything actually using `CPU` or `COMPILATION_MODE` basically imply that you can't share and don't want path-mapping.

But knowing the `TARGET` or `WORKSPACE` seems perfectly reasonable to me. Can `run_binary` not be updated to support those similar to `BINDIR`?

js/private/js_run_binary.bzl
```diff
@@ -298,16 +306,19 @@ def js_run_binary(
...
-        "BAZEL_WORKSPACE": "$(WORKSPACE)",
     }
 
+    # These environment variables are deprecated and will default to not being set in a future
+    # release; see the `set_legacy_environment_variables` docstring.
+    if set_legacy_environment_variables:
+        fixed_env["BAZEL_BUILD_FILE_PATH"] = "$(BUILD_FILE_PATH)"
+        fixed_env["BAZEL_COMPILATION_MODE"] = "$(COMPILATION_MODE)"
+        fixed_env["BAZEL_TARGET_CPU"] = "$(TARGET_CPU)"
+        fixed_env["BAZEL_TARGET"] = "$(TARGET)"
```
jbedard, line 314: what's the difference between `$TARGET` and the `TARGET_NAME`? One has the `//full/path:`?

js/private/js_run_binary.bzl
jbedard: But we need some API to opt-in to the compilation-mode specific flags (default true for now, false in 4.x - so we have path-mapping enabled by default in 4.x).

We'll want to keep that API even in 4.x though so things like webpack can opt-in to have `COMPILATION_MODE` published and it can decide to minify based on that. So I don't think this flag should be labelled as "legacy"?

js/private/js_run_binary.bzl
acozzette: Luckily I was able to get these working without special support in `run_binary`, so I think that should solve the problem.

js/private/js_run_binary.bzl
acozzette: I would actually vote for getting rid of these completely in a future major version. Anyone can always do `js_run_binary(..., env = {"BAZEL_COMPILATION_MODE": "$(COMPILATION_MODE)"})`, so I don't think there's a strong reason for us to have these particular variables built in, especially since most targets don't need them. `webpack_bundle()` in particular doesn't rely on these variables since it uses a custom rule rather than `js_run_binary`.

js/private/js_run_binary.bzl
acozzette: Right, the only difference is that `TARGET_NAME` is qualified with the full path and the `@workspace` at the beginning.

js/private/js_run_binary.bzl
jbedard: > I would actually vote for getting rid of these completely in a future major version.

If we go that direction then I would think _all_ the variables should be behind the `set_legacy_environment_variables` flag?

So if you want to be 4.x compatible:
* set `set_legacy_environment_variables` and manage vars yourself

If you want path-mapping:
* set `set_legacy_environment_variables` and manage path-mapping compatible vars yourself

?

acozzette: Sounds good to me, I updated this to move the other related variables behind the flag as well.

js/private/js_run_binary.bzl
```diff
@@ -53,16 +54,16 @@ def js_run_binary(
     The following environment variables are made available to the Node.js runtime based on available Bazel [Make variables](https://bazel.build/reference/be/make-variables#predefined_variables):
 
     * BAZEL_BINDIR: the bazel bin directory; equivalent to the `$(BINDIR)` Make variable of the `js_run_binary` target
```
jbedard, line 56: Should we drop the docs for this? Or should it also be deprecated and "legacy"?

Should `--bazel-bindir` only set `BINDIR` and drop the `BAZEL_` prefix which seems to be the direction we're going...?

js/private/js_run_binary.bzl
```diff
@@ -53,16 +54,16 @@ def js_run_binary(
...
 
     The following environment variables are made available to the Node.js runtime based on the rule context:
 
-    * BAZEL_BUILD_FILE_PATH: the path to the BUILD file of the bazel target being run; equivalent to `ctx.build_file_path` of the `js_run_binary` target's rule context
-    * BAZEL_PACKAGE: the package of the bazel target being run; equivalent to `ctx.label.package` of the `js_run_binary` target's rule context
-    * BAZEL_TARGET_NAME: the full label of the bazel target being run; a stringified version of `ctx.label` of the `js_run_binary` target's rule context
-    * BAZEL_TARGET: the name of the bazel target being run; equivalent to `ctx.label.name` of the  `js_run_binary` target's rule context
-    * BAZEL_WORKSPACE: the bazel repository name; equivalent to `ctx.workspace_name` of the `js_run_binary` target's rule context
+    * BAZEL_BUILD_FILE_PATH (legacy; see `set_legacy_environment_variables`): the path to the BUILD file of the bazel target being run; equivalent to `ctx.build_file_path` of the `js_run_binary` target's rule context
```
jbedard: Since every one of these is now "(legacy; ...)" should we move that repeated statement into the parent sentence/section-intro?

js/private/js_run_binary.bzl
acozzette: I don't think we can easily deprecate this one unfortunately. The tricky thing about it is that it is important for it to persist in the environment so that one `js_binary` can invoke other `js_binary`s. (So that would prevent us from doing something simple like relying on it in our launcher script but unsetting it before calling node.) And also I suspect that it's so widely used that it would just be a big can of worms to try to rename it or deprecate it.

js/private/js_run_binary.bzl
acozzette: Good point, I made that simplification.

js/private/js_run_binary.bzl
jbedard: That doesn't mean we can't deprecate the public version of it though?

> I suspect that it's so widely used that it would just be a big can of worms to try to rename it or deprecate it.

So widely used where? In random scripts in random repos? In rulesets?

js/private/js_run_binary.bzl
```diff
@@ -57,13 +57,13 @@ def js_run_binary(
     * BAZEL_COMPILATION_MODE (legacy; see `set_legacy_environment_variables`): One of `fastbuild`, `dbg`, or `opt` as set by [`--compilation_mode`](https://bazel.build/docs/user-manual#compilation-mode); equivalent to `$(COMPILATION_MODE)` Make variable of the `js_run_binary` target
     * BAZEL_TARGET_CPU (legacy; see `set_legacy_environment_variables`): the target cpu architecture; equivalent to `$(TARGET_CPU)` Make variable of the `js_run_binary` target
```
jbedard: Isn't this also deprecated and users should just manually use `$(TARGET_CPU)` instead? Same with `BAZEL_COMPILATION_MODE`.

`BAZEL_BINDIR` is still questionable, we still need it, but can we deprecate the public API and make it internal essentially 🤷

js/private/js_run_binary.bzl
acozzette: We document in the README.md that one has to set `BAZEL_BINDIR` for every node action, so in some sense I think it has to be public just for that reason. I imagine that not many projects are reading the value of `BAZEL_BINDIR` but a significant number must be setting it in order to run `js_binary` targets.

js/private/js_run_binary.bzl
acozzette: Yes, there ones are now deprecated too.

What do you have in mind as far as deprecating `BAZEL_BINDIR`? I think it's important for it to stay around, and in some situations users may need to set it themselves. Should we say that you can set it but just shouldn't read it?

jbedard: I question if we should just make it `BINDIR` instead if we're dropping all the other `BAZEL_` ones? I need to think about that more, and open to ideas...

acozzette: Honestly I don't think it would be worth it given how much work would be involved. We would probably need to support both `BINDIR` and `BAZEL_BINDIR` for a transitional period, get everyone to switch over to `BINDIR`, and then drop `BAZEL_BINDIR`. I think it would be doable with some effort but the benefit seems very small to me.

js/private/js_run_binary.bzl
```diff
@@ -296,17 +304,18 @@ def js_run_binary(
...
-        "BAZEL_TARGET_NAME": name,
-        "BAZEL_TARGET": "$(TARGET)",
-        "BAZEL_WORKSPACE": "$(WORKSPACE)",
-    }
+    fixed_env = {}
+
+    # These environment variables are deprecated and will default to not being set in a future
+    # release; see the `set_legacy_environment_variables` docstring.
+    if set_legacy_environment_variables:
+        fixed_env["BAZEL_BUILD_FILE_PATH"] = "$(BUILD_FILE_PATH)"
```
jbedard, line 311: Should `BAZEL_BINDIR` still be here? So we aren't depending on what `--bazel-bindir` does?

acozzette: SGTM, I added that one back in.

js/private/js_run_binary.bzl
```diff
@@ -296,17 +304,19 @@ def js_run_binary(
...
-        "BAZEL_TARGET_CPU": "$(TARGET_CPU)",
-        "BAZEL_TARGET_NAME": name,
-        "BAZEL_TARGET": "$(TARGET)",
-        "BAZEL_WORKSPACE": "$(WORKSPACE)",
-    }
+    fixed_env = {}
+
+    # These environment variables are deprecated and will default to not being set in a future
+    # release; see the `set_legacy_environment_variables` docstring.
+    if set_legacy_environment_variables:
```
thesayyn, line 309: i actually do support this. It hurts the cacheability of the whole rules_js. and we even have a patch in js_image_layer to stub these values out.

thesayyn (approved):

js/private/js_run_binary.bzl
jbedard: Great 👍

js/private/js_run_binary.bzl
jbedard: After your last PR doesn't one have to either a) set `BAZEL_BINDIR` or b) pass `--bazel-bindir`?

Note that this code no longer sets `BAZEL_BINDIR` unless `set_legacy_environment_variables` is true so I think this one should be marked as legacy just like all the others...

acozzette: > After your last PR doesn't one have to either a) set `BAZEL_BINDIR` or b) pass `--bazel-bindir`?

Yes, that's right.

> Note that this code no longer sets `BAZEL_BINDIR` unless `set_legacy_environment_variables` is true so I think this one should be marked as legacy just like all the others...

SGTM, I marked this one as legacy like the others.

jbedard (approved):

PR #2965, opened 2026-08-04
refactor(js_run_binary): support path mapping with use_execroot_entry_point

acozzette:
use_execroot_entry_point hoisted a js_binary tool's entire runfiles tree (including its generated launcher, node/npm wrappers, and node-patches bootstrap) into a target-configuration rebuild, so that the entry point could be resolved from the target-platform bin directory. The launcher script bakes `$(COMPILATION_MODE)/$(TARGET_CPU)/$(BINDIR)` into its content, so this hoisted copy differed between -c fastbuild and -c opt, defeating cross-config path mapping even though none of that toolchain scaffolding is ever read from the target-platform bin directory at runtime (only the entry point and its data are).

js_binary now exposes a leaner "execroot_data_files" output group (entry point + data, without the toolchain scaffolding), and js_run_binary hoists that instead of the tool's full runfiles.

---

Allow path mapping to work with js_run_binary even with use_execroot_entry_point enabled

### Test plan

- New test cases added

### DIFF
--- a/e2e/path_mapping/js_run_binary_path_mapping_check/BUILD.bazel
+++ b/e2e/path_mapping/js_run_binary_path_mapping_check/BUILD.bazel
@@ -2,8 +2,8 @@ load("@aspect_rules_js//js:defs.bzl", "js_binary", "js_run_binary")
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
 
 # Verifies that a js_run_binary target supports path mapping. This requires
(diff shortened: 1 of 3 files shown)


jbedard (approved):

jbedard:
IMO this is a `refactor` or potentially `perf` and should not be advertised as a `feat`, especially since path-mapping hasn't even been released yet? 🤷 

LGTM though 👍

PR #2967, opened 2026-08-05
feat(js_run_binary): add flag controlling default of set_legacy_envir…

acozzette:
…onment_variables

Adds the //js:set_legacy_environment_variables build flag, mirroring //js:use_execroot_entry_point, so users can flip the default for the legacy BAZEL_* env vars repo-wide via .bazelrc instead of per-target.

---

Adds the //js:set_legacy_environment_variables build flag

### Test plan

- New test cases added

### DIFF
--- a/docs/path_mapping.md
+++ b/docs/path_mapping.md
@@ -30,9 +30,10 @@ determines it is safe to do so, but you may need to explicitly opt in (see
 below). To set up a `js_run_binary` target for path mapping:
  - Pass `set_legacy_environment_variables = False`. This will disable the
    setting of environment variables such as `BAZEL_COMPILATION_MODE` that would
-   otherwise be problematic. Alternatively, you can turn off this behavior by
-   default globally by passing
-   `--@aspect_rules_js//js:set_legacy_environment_variables=False`.
+   otherwise be problematic. Alternatively, you can change the default for all
+   targets by passing `--@aspect_rules_js//js:set_legacy_environment_variables=False`
+   on the command line; individual targets can still override this default by
+   setting the attribute explicitly.
  - Avoid using [Make
    variables](https://bazel.build/reference/be/make-variables) or expressions
    such as `$(execpath :target)`. Such variables and expressions are not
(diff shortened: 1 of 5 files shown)


docs/path_mapping.md
```diff
@@ -30,7 +30,9 @@ determines it is safe to do so, but you may need to explicitly opt in (see
 below). To set up a `js_run_binary` target for path mapping:
  - Pass `set_legacy_environment_variables = False`. This will disable the
    setting of environment variables such as `BAZEL_COMPILATION_MODE` that would
-   otherwise be problematic.
+   otherwise be problematic. Alternatively, you can turn off this behavior by
```
jbedard: Shouldn't we say that you can change the default via flag, not "turn off globally"? Because individual targets can still override the flag via the attribute.

js/private/js_run_binary.bzl
```diff
@@ -305,15 +308,27 @@ def js_run_binary(
...
-        fixed_env["BAZEL_BUILD_FILE_PATH"] = "$(BUILD_FILE_PATH)"
-        fixed_env["BAZEL_COMPILATION_MODE"] = "$(COMPILATION_MODE)"
-        fixed_env["BAZEL_TARGET_CPU"] = "$(TARGET_CPU)"
-        fixed_env["BAZEL_TARGET"] = "$(TARGET)"
-        fixed_env["BAZEL_WORKSPACE"] = "$(WORKSPACE)"
-        fixed_env["BAZEL_PACKAGE"] = native.package_name()
-        fixed_env["BAZEL_TARGET_NAME"] = name
+    # When set_legacy_environment_variables is None, behavior is controlled by the
+    # //js:set_legacy_environment_variables flag.
+    legacy_env_values = {
```
jbedard: Is there any way to flip this to not construct it by default? Any less/less-complicated code that way?
```
legacy_env = {}
if set_legacy_environment_variables != Flase:
   legacy_env = { ... the vars ... }
   if set_legacy_environment_variables == None:
       legacy_env = select({ ... })
```

I guess performance doesn't matter, so whatever is less code or whatever you think looks nicer 🤷

jbedard (approved):

docs/path_mapping.md
acozzette: That's true, I reworded it to clarify that the flag just changes the default.

js/private/js_run_binary.bzl
acozzette: Sure, I updated it to use your approach, so at least this way we don't construct it if the param is explicitly set to `False`.

PR #2972, opened 2026-08-14
perf: invoke node directly from the js_binary launcher

acozzette:
The launcher ran the program through our Node wrapper script, which does nothing but pass its arguments along to Node and add a `--require` flag for our fs patches.

Let's just inline that into our main launcher script so that we avoid spawning another bash process unnecessarily.

We still need to keep the Node wrapper around, though, so that if we ever shell out to `node` then we get our wrapper with its patches.

---

### Test plan

### DIFF
--- a/js/private/js_binary.sh.tpl
+++ b/js/private/js_binary.sh.tpl
@@ -385,7 +385,7 @@ for ARG in ${ALL_ARGS[@]+"${ALL_ARGS[@]}"}; do
     esac
 done
 
-# Configure JS_BINARY__FS_PATCH_ROOTS for node fs patches which are run via --require in the node wrapper.
+# Configure JS_BINARY__FS_PATCH_ROOTS for node fs patches which are run via --require below.
 # Don't override JS_BINARY__FS_PATCH_ROOTS if already set by an outer js_binary incase a js_binary such
 # as js_run_deverser runs another js_binary tool.
 if [ -z "${JS_BINARY__FS_PATCH_ROOTS:-}" ]; then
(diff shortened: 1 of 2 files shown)


js/private/js_binary.sh.tpl
```diff
@@ -463,8 +463,14 @@ fi
 # Run the main program
 # ==============================================================================
 
+# Invoke node directly rather than through $JS_BINARY__NODE_WRAPPER. This way
+# we avoid spawning an extra bash process on every launch. The wrapper is
+# still put on the PATH as `node` so that child processes get the patched
+# runtime.
+node_cmd=("$JS_BINARY__NODE_BINARY" --require "$JS_BINARY__NODE_PATCHES")
```
jbedard: What if we put the `--require` onto `JS_BINARY__NODE_OPTIONS`? Then can a child process also avoid the `JS_BINARY__NODE_WRAPPER`?

jbedard: Or even `NODE_OPTIONS`. Does that also mean it runs sooner then `--require`? I think I saw codex mentioning that somewhere recently...?

jbedard:
Make sure we have tests covering this scenario 👍

jbedard (approved):

js/private/js_binary.sh.tpl
acozzette: I like the `NODE_OPTIONS` idea, because that could potentially let us drop the node wrapper altogether. According to the docs, options from that variable are treated as if they were specified at the beginning of the command line, so if I understand right they would run before any other `--require`.

I gave it a try though, and unfortunately it seemed non-trivial to get it working. For example we have to be careful to avoid nested launchers adding a second `--require` onto `NODE_OPTIONS`. So I think we should leave it for a potential follow-up but it does seem promising.

PR #2973, opened 2026-08-14
perf: write captured stdout/stderr straight to the output file

acozzette:
It is advantageous to write the captured stderr or stdout straight to its intended destination. This eliminates a cleanup step after Node exits, making it possible to invoke Node via `exec` and thus avoid forking an extra process.

There is a drawback, which is that if the action fails then the captured output will be lost, making it somewhat harder to troubleshoot. This seems like an acceptable cost given the performance benefit, though.

One other caveat is that we can only do this when `silent_on_success` is not enabled, but that parameter defaults to `True` for `js_run_binary`. Under the default behavior, we have to store stdout and stderr in temporary files so that we are prepared to print them out if the action fails. So to get any improvement from this change, the action has to opt in by setting `silent_on_success = False`, or else invoke the `js_binary` without using `js_run_binary`.

This commit makes one other significant change, which is to always print out the launcher script's fatal- and error-level log messages instead of capturing them. This behavior is logical because if such a message is getting logged then the action is almost certainly going to fail, in which case there is no point in attempting to capture the log message anyway. We might as well print out the message to help the user understand what went wrong.

---

### Test plan

### DIFF
--- a/js/private/js_binary.sh.tpl
+++ b/js/private/js_binary.sh.tpl
@@ -15,8 +15,46 @@ set -o pipefail -o errexit -o nounset
 {{envs}}
 
 # ==============================================================================
+# Handle --bazel-bindir flag
+# ==============================================================================
+
+# If a --bazel-bindir <path> flag is passed it must be the first two
+# arguments. It is consumed by this launcher script and used to set
+# BAZEL_BINDIR, overriding any value already set in the environment.
+if [ $# -gt 0 ] && [ "$1" = "--bazel-bindir" ]; then
+    if [ $# -lt 2 ]; then
+        echo "ERROR: --bazel-bindir flag requires a value" >&2
+        exit 1
+    fi
+    BAZEL_BINDIR="$2"
+    export BAZEL_BINDIR
+    shift 2
+fi
+
+# ==============================================================================
 # Prepare stdout capture, stderr capture && logging
 # ==============================================================================
+
+# Convert stdout, stderr and exit_code capture outputs paths to absolute paths.
+# A path not starting with "bazel-out/" is relative to the bin directory; joining
+# it with BAZEL_BINDIR here (rather than baking in bazel-out/<config>/bin at
+# analysis time) keeps it correct when path mapping is active.
+function resolve_capture_path {
+    case "$1" in
+    bazel-out/*) echo "$PWD/$1" ;;
+    *) echo "$PWD/${BAZEL_BINDIR:-$JS_BINARY__BINDIR}/$1" ;;
+    esac
+}
+
+if [ "${JS_BINARY__STDOUT_OUTPUT_FILE:-}" ]; then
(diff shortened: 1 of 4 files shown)


js/private/js_binary.sh.tpl
```diff
@@ -14,73 +14,116 @@ set -o pipefail -o errexit -o nounset
...
+
+# A stream with a declared output file is written straight to that file so that
+# nothing has to run after node exits and this shell can exec node.
+#
+# A stream that is only subject to silent_on_success has no final destination, so
+# it is buffered in a temp file and replayed on failure.
+if [ "${JS_BINARY__STDOUT_OUTPUT_FILE:-}" ]; then
+    STDOUT_CAPTURE="$JS_BINARY__STDOUT_OUTPUT_FILE"
+    : >"$STDOUT_CAPTURE"
+elif [ "${JS_BINARY__SILENT_ON_SUCCESS:-}" ]; then
```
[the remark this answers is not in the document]
acozzette: This is an accepted cost. If a target captures stdout or stderr then we are OK with losing that output in the event the action fails.

js/private/js_binary.sh.tpl
```diff
@@ -14,73 +14,116 @@ set -o pipefail -o errexit -o nounset
...
+fi
+
+# A stream with a declared output file is written straight to that file so that
+# nothing has to run after node exits and this shell can exec node.
+#
+# A stream that is only subject to silent_on_success has no final destination, so
+# it is buffered in a temp file and replayed on failure.
+if [ "${JS_BINARY__STDOUT_OUTPUT_FILE:-}" ]; then
+    STDOUT_CAPTURE="$JS_BINARY__STDOUT_OUTPUT_FILE"
+    : >"$STDOUT_CAPTURE"
```
jbedard: What on earth does this line do? lol

acozzette: Good question, apparently `:` is a builtin that basically does nothing, but we're using it to truncate the output file. I don't think this is necessary though, so I removed these lines.

js/private/js_binary.sh.tpl
```diff
@@ -14,73 +14,114 @@ set -o pipefail -o errexit -o nounset
 
 {{envs}}
 
+# ==============================================================================
```
jbedard, line 17: These only moved and didn't change? Why?

Would the PR be any easier to review if you moved them in a separate commit (but same PR and we can squash-on-merge)?

acozzette: Sure, I restructured things so that the first commit is just moving code blocks around and the second commit has the interesting stuff.

The main reason for moving stuff around is that since we're now writing straight the the stdout and stderr capture files, we need to know their absolute paths earlier on, and that requires BAZEL_BINDIR, so we need to parse the `--bazel-bindir` flag earlier too.

jbedard (approved):

PR #2974, opened 2026-08-14
refactor: use hermetic_launcher to invoke node

acozzette:
---

### Test plan

### DIFF
--- a/js/private/node-bootstrap/launcher.cjs
+++ b/js/private/node-bootstrap/launcher.cjs
@@ -1,0 +1,430 @@
+// Preload for the hermetic launcher, which invokes node directly rather than
+// through js_binary.sh.tpl. It reconstructs the parts of that script's runtime
+// contract that a process which only execve's cannot set up, then hands off to
+// bootstrap.cjs.
+//
+// Everything here is derived from process.execPath, process.execArgv, __dirname,
+// process.cwd() and the environment, so this file needs no per-target generation.
+// __dirname is what makes that work: this file always sits at
+// <runfiles>/<repo>/js/private/node-bootstrap/, whatever the repo is called.
+//
+// The per-target constants the bash launcher bakes in -- JS_BINARY__WORKSPACE,
+// JS_BINARY__TARGET, JS_BINARY__PACKAGE, JS_BINARY__BUILD_FILE_PATH,
+// JS_BINARY__COMPILATION_MODE, JS_BINARY__TARGET_CPU, JS_BINARY__BINDIR -- are
+// deliberately not reconstructed. They are unreachable from here.
+//
+// Ordering matters: bootstrap.cjs destructures process.env at module load, so
+// every variable it reads has to be set before the require at the bottom.
+
+const fs = require('fs')
+const path = require('path')
+const { pathToFileURL } = require('url')
+
+// A module loaded under two path spellings runs twice, and fs.cjs throws rather
+// than patch twice. Make the second run a no-op.
+const GUARD = Symbol.for('aspect_rules_js.hermetic_launcher')
+
+// Node worker threads inherit execArgv, so this preload runs in every one of them, and
+// two of the things it does belong to the main thread alone:
+//
+//   - Changing directory. process.chdir throws ERR_WORKER_UNSUPPORTED_OPERATION in a
+//     worker, which does not need it anyway: it inherits the cwd the main thread
+//     already moved to.
+//   - Redirecting the main module. A worker's main is the script it was started with
+//     rather than the js_binary entry point, and process.argv[1] is not even set there.
+//
+// Everything else here still has to happen in a worker, because the fs patches apply per
+// realm.
+//
+// The bash launcher never met any of this: the shell did its cd before node started, and
+// nothing inside node changed directory or chose a main afterwards. rollup running terser
+// is a real target that does both.
+const IS_MAIN_THREAD = require('worker_threads').isMainThread
+
+const LOG_PREFIX = 'aspect_rules_js[js_binary]'
+
+// The runfiles directory name of the main repository. rules_js is bzlmod-only, and
+// bzlmod always names it `_main` -- which is what both bazel-lib's to_rlocation_path and
+// hermetic_launcher itself already assume. If it ever were something else the derived
+// entry point would not exist, and redirectMainToExecroot fatals naming the path it
+// looked for rather than running the wrong file.
+const MAIN_REPO_PREFIX = '_main/'
+
+// Features of the bash launcher that this one cannot implement: each needs work after
+// the program exits, and this launcher is execve'd over. Nothing in rules_js sets them
+// -- js_run_binary's stdout, stderr, exit_code_out and silent_on_success go through
+// run_binary's wrapper, and expected_exit_code keeps a js_binary off this launcher
+// altogether -- so reaching one means an `env` that asked for the script's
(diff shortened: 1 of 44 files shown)


purkhusid:
This is a pretty cool change. Would allow us to put the js_binary targets directly onto distroless images that don't have `bash` or `coreutils` installed.

js/private/test/js_binary_sh/BUILD.bazel
```diff
@@ -14,8 +14,16 @@ js_binary(
...
+# The assertions below are on the launcher script's own diagnostics, including the
+# per-target constants it bakes in, so this target has to run the script. Wrapping the
+# empty `args` in a select() is what holds it there: the js_run_binary macro cannot see
+# through one, so it declines the hermetic launcher, while the command line the select
+# resolves to is still empty. Without this the target would take the hermetic launcher
+# whenever --//js:use_execroot_entry_point is False, and that launcher prints none of
+# this -- JS_BINARY__WORKSPACE and the rest are documented as unreachable from it.
 js_run_binary(
     name = "capture_stderr",
+    args = select({"//conditions:default": []}),
```
jbedard, line 26: Should this be its own target specifically describing the edge case it's testing? Is that a test that can merge on its own now?

js/private/worker/src/BUILD.bazel
```diff
@@ -27,16 +27,21 @@ rollup_bin.rollup(
         ":node_modules/google-protobuf",
     ],
     outs = ["bundle.js"],
+    # Only the config path needs `$RUNFILES_DIR` expanded; the rest are ordinary `args`,
```
jbedard, line 30: Prefactor (without mentioning hermetic launcher which is an implementation detail)? Does this comment belong on the `fixed_args` instead of `args`?

js/private/node-bootstrap/launcher.cjs
```diff
@@ -0,0 +1,483 @@
...
+            )
+        }
+        debug(`changing directory to BAZEL_BINDIR (root of Bazel output tree) ${bindir}`)
+        process.chdir(bindir)
+    }
+    return execroot
+}
+
+// The path this file was preloaded from. Not __dirname, which node has already resolved
+// through the runfiles symlink back to the source tree; the files beside this one have to
```
jbedard, line 199: To the source tree? Is that only on macos?

js/private/node-bootstrap/launcher.cjs
```diff
@@ -0,0 +1,483 @@
...
+    // Everything below is computed against the directory we started in, which
+    // resolveExecroot may change.
+    const startCwd = process.cwd()
+
+    const isActionLaunch = takeBazelBindir()
+    rejectNodeOptions()
+
+    const runfiles = resolveRunfiles(startCwd)
+    process.env.RUNFILES_DIR = runfiles
+    process.env.JS_BINARY__RUNFILES = runfiles
```
jbedard, line 420: Should this be deprecated and tell people to use `RUNFILES_DIR`? Maybe that's a bigger discussion, and would be breaking so we'd have to keep it for now anyway...

js/private/node-bootstrap/launcher.cjs
```diff
@@ -0,0 +1,483 @@
...
+//
+// Everything else here still has to happen in a worker, because the fs patches apply per
+// realm.
+//
+// The bash launcher never met any of this: the shell did its cd before node started, and
+// nothing inside node changed directory or chose a main afterwards. rollup running terser
+// is a real target that does both.
+const IS_MAIN_THREAD = require('worker_threads').isMainThread
+
+const LOG_PREFIX = 'aspect_rules_js[js_binary]'
```
jbedard, line 47: What if this is `js_test`? 😅

js/private/node-bootstrap/launcher.cjs
```diff
@@ -0,0 +1,483 @@
...
+//   - Redirecting the main module. A worker's main is the script it was started with
+//     rather than the js_binary entry point, and process.argv[1] is not even set there.
+//   - Turning COVERAGE_DIR into NODE_V8_COVERAGE, which takes a fresh node. A worker
+//     inherits an environment that already has the variable, and replacing the process
+//     out from under the main thread is not something it could survive anyway.
+//
+// Everything else here still has to happen in a worker, because the fs patches apply per
+// realm.
+//
+// The bash launcher never met any of this: the shell did its cd before node started, and
```
jbedard, line 42: Why are we mentioning "the bash launcher"? This comment will last longer then todays "bash launcher" and shouldn't reference it

js/private/worker/src/BUILD.bazel
acozzette: I'm actually inclined to think that if we go with the direction in this PR we should probably find a way to increase the fixed-arg limit so that changes like this one aren't necessary. Hermetic_launcher has a limit of 10 built-in args, and we use five of them ourselves, so that only leaves five for users.

js/private/node-bootstrap/launcher.cjs
```diff
@@ -0,0 +1,483 @@
...
+        }
+        dir = parent
+    }
+}
+
+// Starts node over again on `mainScript`, with the same options and the same arguments,
+// and does not return. The two things a preload cannot do to the node it is already
+// running in both end up here.
+//
+// process.execve replaces this process rather than adding one, which is what the launcher
```
jbedard, line 261: don't reference the history of the repo in a comment

js/private/node-bootstrap/launcher.cjs
```diff
@@ -0,0 +1,483 @@
...
+// script's `exec` did and costs the same nothing. A node too old to have it forks and
+// waits instead -- and that is the same vintage of node that makes the ESM caller below
+// necessary at all.
+function reExec(mainScript, env, reason) {
+    debug(`re-executing node on ${mainScript}: ${reason}`)
+    const argv = [...process.execArgv, mainScript, ...process.argv.slice(2)]
+    if (typeof process.execve === 'function') {
```
jbedard, line 268: When would this not be true? That might be worth a (1 line) comment)

js/private/worker/src/BUILD.bazel
jbedard: Is that the reason for this? Does something (hermetic-launcher action or js_binary or something) throw an error outlining issue if/when that happens?

jbedard: Maybe open an issue to discuss it on the hermetic-launcher repo? If you can't find any explanation for the limit in the code or existing (closed) issues...

acozzette: > Is that the reason for this? Does something (hermetic-launcher action or js_binary or something) throw an error outlining issue if/when that happens?

Yes, hermetic-launcher will fail to stamp out a launcher if you try to give it more than 10 embedded args. I believe it's due to the fact that hermetic-launcher is implemented like a binary template so that launchers can be stamped out without requiring a compiler invocation. As a result there's a fixed number of slots and there's a size penalty to every launcher if we increase that number.

I filed an issue here last week but haven't gotten a response yet: https://github.com/hermeticbuild/hermetic-launcher/issues/76

Since filing that issue I have mostly changed my mind, though. I think we are going to end up needing some form of generated launcher, because for example we need a way to set the `NODE_V8_COVERAGE` environment variable before invoking node. As long as we need a launcher anyway, we might as well put the embedded args there, and then we can have an unlimited number of them without needing any changes in hermetic_launcher.

PR #2976, opened 2026-08-18
test: explicitly set `silent_on_success` on most `js_run_binary` targets

acozzette:
This change sets `silent_on_success = True` on a few usages that generate a lot of output, but otherwise sets `silent_on_success = False` almost everywhere. The macros in the generated `package_json.bzl` are left alone, since we cannot change those without affecting users in an unpredictable way.

The main goal of this is to improve test coverage. When `silent_on_success` is `False`, we often exercise a different code path because (depending on other options) we can sometimes `exec` node and avoid needing to do any cleanup after it returns.

I am also working on setting up `hermetic_launcher` as an alternative to the bash launcher, but we cannot easily use `hermetic_launcher` if `silent_on_success` is enabled. Turning that off in more places should unlock many more opportunities to test `hermetic_launcher`.

---

### Test plan

### DIFF
--- a/npm/private/npm_package.bzl
+++ b/npm/private/npm_package.bzl
@@ -483,6 +483,7 @@ def npm_package(
             tags = kwargs.get("tags", []) + ["manual"],
             testonly = kwargs.get("testonly"),
             visibility = kwargs.get("visibility", None),
+            silent_on_success = False,
         )
 
     _npm_package(
--- a/npm/private/npm_package.bzl
+++ b/npm/private/npm_package.bzl
@@ -483,7 +483,6 @@ def npm_package(
             tags = kwargs.get("tags", []) + ["manual"],
             testonly = kwargs.get("testonly"),
             visibility = kwargs.get("visibility", None),
-            silent_on_success = False,
         )
 
     _npm_package(
--- a/contrib/nextjs/defs.bzl
+++ b/contrib/nextjs/defs.bzl
@@ -212,6 +212,7 @@ def nextjs_build(name, config, srcs, next_js_binary, data = [], use_execroot_ent
         mnemonic = "NextJs",
         progress_message = "Compile Next.js app %{label}",
         use_execroot_entry_point = use_execroot_entry_point,
+        silent_on_success = kwargs.pop("silent_on_success", True),
         **kwargs
     )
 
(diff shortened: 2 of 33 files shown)


npm/private/npm_package.bzl
```diff
@@ -483,6 +483,7 @@ def npm_package(
             tags = kwargs.get("tags", []) + ["manual"],
             testonly = kwargs.get("testonly"),
             visibility = kwargs.get("visibility", None),
+            silent_on_success = False,
```
jbedard: This means this is more then just for tests and the PR title should change?

acozzette: Good point, this part's not critical so I just reverted the change to this file.

jbedard (approved):

jbedard:
We could potentially add an option to change the default via `.bazelrc` config? Then users can opt-in to the fast-path globally by default by editing `.bazelrc`?

This is a good start either way, being an explicit `True|False` avoids a `select()` and should be done where possible 👍

acozzette:
> We could potentially add an option to change the default via `.bazelrc` config? Then users can opt-in to the fast-path globally by default by editing `.bazelrc`?
> 
> This is a good start either way, being an explicit `True|False` avoids a `select()` and should be done where possible 👍

Yeah, that might be a good thing to do soon. I wonder if we should flip the default on `silent_on_success` in the next major release so that we can avoid the need for the bash launcher (or spawn_binary) on every js_run_binary target. I guess arguably we could even do that without a major version increment, since this wouldn't exactly break anything--in the worst-case it would just spam the terminal with more output than usual.

jbedard:
Since `silent_on_success` requires an extra process (bash or spawn_binary or anything) I think it should be disabled by default, but that is big enough I'd rather not do it now even though it's not "breaking".

I think we should provide an option for people to opt-in to the fast-path logic via config flags for now, then swap the default in the next major release while still allowing the config flag to override that default if they need to reverse the default-swap etc.

I would think the ideal case is that tools output to stderr on failure, and don't output anything to stdout. Maybe instead of `silent_on_success` we can have `stdout = False|"output_file"` and it is simply `False` to suppress stdout by default? While stderr goes through and gets outputted by default...

PR #2980, opened 2026-08-20
refactor: delegate js_run_binary output capture to run_binary

acozzette:
`run_binary` now has its own implementations of `stdout`, `stderr`, `exit_code_out` and `silent_on_success`, so we can have `js_run_binary` use them without reimplementing that functionality.

I previously attempted this and had to roll back the change, but that was because I deleted the corresponding logic from the bash launcher, and it turns out that some targets rely on that to work even when `js_run_binary` is not involved. This commit adds test coverage of that use case to ensure it keeps working.

This change will cause most `js_run_binary` invocations to be wrapped with `spawn_binary`, which is the tool `run_binary` uses to capture stdout, stderr, and the exit code. This is due to the fact that `silent_on_success` defaults to `True` on `js_run_binary`. Performance-wise, this is mostly a wash: we get the extra `spawn_binary` process but also allow the bash launcher to `exec` node instead of spawning a new process for it.

The main motivation here is to move toward having `js_run_binary` use `hermetic_launcher` instead of our bash launcher. Unlike the bash launcher, `hermetic_launcher` has no way to capture the output or exit code of the binary, but that problem goes away if we can let `spawn_binary` handle it.

---

### Test plan

- New test cases added

### DIFF
--- a/js/private/js_run_binary.bzl
+++ b/js/private/js_run_binary.bzl
@@ -16,28 +16,6 @@ load("@bazel_lib//lib:utils.bzl", bazel_lib_utils = "utils")
 load(":js_helpers.bzl", _envs_for_log_level = "envs_for_log_level")
 load(":js_info_files.bzl", _js_info_files = "js_info_files")
 
-def _bindir_relative_out(out):
-    """Path of a declared output relative to the bin directory.
-
-    `run_binary` declares `outs` with `attr.output_list`, so the label is always in the
-    current package and the path is fully computable here. Returning a literal keeps the
-    value free of `$(execpath)`, which would otherwise both disable path mapping in
-    `run_binary` and bake in a `bazel-out/<config>/bin` prefix that is wrong when path
-    mapping is active. The launcher recognizes the path as bin-relative because it does
-    not start with `bazel-out/`, and joins it with the path-mapped `--bazel-bindir`.
-    """
-
-    # `out` may be a string or a Label; str() normalizes both to a form ending in
-    # ":<name>" or, for a plain relative string, to the name itself.
-    path = str(out).split(":")[-1]
-    package = native.package_name()
-    if package:
-        path = "{}/{}".format(package, path)
-    repo = native.repo_name()
-    if repo:
-        path = "external/{}/{}".format(repo, path)
-    return path
-
 def js_run_binary(
         name,
         tool,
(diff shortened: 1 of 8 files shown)


js/private/js_run_binary.bzl
```diff
@@ -373,22 +351,6 @@ def js_run_binary(
                 normalized_chdir = "external/{}/{}".format(repo, chdir)
         fixed_env["JS_BINARY__CHDIR"] = normalized_chdir
 
-    # Configure capturing stdout, stderr and/or the exit code
-    extra_outs = []
-    if stdout:
-        fixed_env["JS_BINARY__STDOUT_OUTPUT_FILE"] = _bindir_relative_out(stdout)
```
jbedard, line 379: So we never use these env vars anymore, but the bash launcher still supports them for the time being?

Should we add `TODO(4.0): drop support` to the bash launcher script for those?

js/private/test/launcher/direct_capture.bzl
```diff
@@ -0,0 +1,71 @@
...
+
+load("@aspect_rules_js//js:libs.bzl", "js_binary_lib")
+
+def _direct_capture_impl(ctx):
+    stdout = ctx.outputs.stdout
+    stderr = ctx.outputs.stderr
+    exit_code = ctx.outputs.exit_code_out
+    outputs = [stdout, stderr, exit_code]
+
+    js_binary_lib.run_binary_action(
```
jbedard, line 18: Is there any way to add `js_binary_lib.run_binary_action(stderr, stdout, ...)` and delegate that to `spawn_binary` when they are set? Or is `spawn_binary` something that must be done at macro/rule declaration and not within a rule implementation like this?

acozzette: I believe that could work, but it would be a little bit tricky because it would require the rule to specify the toolchain for `spawn_binary`. So it would be a bit of a leaky abstraction because the rule would have to know that we're using `spawn_binary` and make sure to declare that toolchain. Or alternatively we could maybe provide a list of toolchains we need and ask rule authors to rely on that.

js/private/js_run_binary.bzl
acozzette: Right, we don't use these anymore but we need to keep supporting them because that are projects out there that rely on setting these variables directly. I went ahead and added a TODO to drop them in 4.0.

js/private/test/launcher/direct_capture.bzl
jbedard: > the rule would have to know that we're using spawn_binary and make sure to declare that toolchain

Ok, maybe that won't be a great idea then lol

jbedard (approved):

PR #2984, opened 2026-08-21
refactor: generate the coverage report from node, not the launcher

acozzette:
This change removes the coverage report generation from the bash launcher and puts it in a Node exit listener instead. The exit listener spawns another Node process to run coverage.js, much like the bash launcher did.

The main benefit of this is that it means we generally no longer have to wait for Node to exit so that we can do any cleanup afterward. In almost all cases the bash launcher can just `exec` Node and that's it.

The only cases where we cannot do that are when someone has directly set an environment variable such as `JS_BINARY__STDOUT_OUTPUT_FILE` when invoking a `js_binary`. These variables were meant as internal implementation details, but we will support this use case until the next major version.

---

### Test plan

- New test cases added


js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -43,3 +43,95 @@ if (
...
+//
+// JS_BINARY__COVERAGE_REPORT is set by js_binary.bzl for a test target under
+// `bazel coverage` only, and holds the runfiles-relative path of coverage.js.
+const { JS_BINARY__COVERAGE_REPORT } = process.env
+if (JS_BINARY__COVERAGE_REPORT) {
+    // Child node processes re-enter this bootstrap with an inherited environment. Only
+    // the root process reports, once every child has exited and written its profile.
+    delete process.env.JS_BINARY__COVERAGE_REPORT
+}
+if (JS_BINARY__COVERAGE_REPORT && process.env.COVERAGE_DIR) {
```
jbedard: We never actually use `process.env.COVERAGE_DIR`?

js/private/test/coverage/BUILD.bazel
```diff
@@ -81,6 +82,61 @@ coverage_pass_test(
...
+    ],
+    enable_runfiles = select({
+        "@bazel_lib//lib:enable_runfiles": True,
+        "//conditions:default": False,
+    }),
+    entry_point = "child.js",
+    # child.js spawns process.execPath, which bootstrap.cjs only patches to the node
+    # wrapper off Windows.
+    target_compatible_with = select({
+        "@platforms//os:windows": ["@platforms//:incompatible"],
```
jbedard: Why?

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -43,3 +43,95 @@ if (
...
+    // launcher used to run after the process was gone. A program that writes its own
+    // report to COVERAGE_OUTPUT_FILE from an 'exit' listener therefore also gets a
+    // report generated here. What ends up published is unchanged: the merger prefers
+    // the program's own report, except under split post-processing where bazel discards
+    // it and ours is published instead.
+    process.on('exit', (code) => {
+        // A test reporting its own coverage owns it, except in split mode where bazel
+        // drops it.
+        if (
+            env.SPLIT_COVERAGE_POST_PROCESSING !== '1' &&
```
jbedard: Should we read+record these vars at startup instead of reading them within the `exit` hook so they can't be changed on us? Or do we want to support them changing?

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -43,3 +43,95 @@ if (
...
+        JS_BINARY__COVERAGE_REPORT
+    )
+
+    // NB: this runs before any 'exit' listener the program registers, whereas the
+    // launcher used to run after the process was gone. A program that writes its own
+    // report to COVERAGE_OUTPUT_FILE from an 'exit' listener therefore also gets a
+    // report generated here. What ends up published is unchanged: the merger prefers
+    // the program's own report, except under split post-processing where bazel discards
+    // it and ours is published instead.
+    process.on('exit', (code) => {
```
jbedard: Want to give this method a name so it shows up in stack traces? `function onExitCoverageReport(code) {` or something like that...

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -43,3 +43,95 @@ if (
...
+        // A test reporting its own coverage owns it, except in split mode where bazel
+        // drops it.
```
jbedard: This robot has a 1980s screen size, can we avoid wrapping comments just for 2 words like this?

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -43,3 +43,95 @@ if (
...
+    const env = {
+        ...process.env,
+        JS_COVERAGE__RUNFILES: process.env.JS_BINARY__RUNFILES,
+        NODE_V8_COVERAGE: '',
+    }
+    const cwd = process.cwd()
+    // Not process.execPath, which was patched above to the node wrapper; that would
+    // re-apply this bootstrap in the reporter process.
+    const nodeBinary = process.env.JS_BINARY__NODE_BINARY
+    const report = require('path').resolve(
```
jbedard: I would vote to use the `node:` prefix on all these native node imports just to avoid/reduce-likelyhood of `require` hacks interfering with them...

js/private/test/image/non_ascii/custom_layer_groups_test_node.listing
```diff
@@ -9,7 +9,7 @@ drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/non_
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/non_ascii/bin2.runfiles/_main/js/
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/non_ascii/bin2.runfiles/_main/js/private/
 drwxr-xr-x  0 0      0           0 Jan  1  1970 ./app/js/private/test/image/non_ascii/bin2.runfiles/_main/js/private/node-bootstrap/
--r-xr-xr-x  0 0      0        1460 Jan  1  1970 ./app/js/private/test/image/non_ascii/bin2.runfiles/_main/js/private/node-bootstrap/bootstrap.cjs
+-r-xr-xr-x  0 0      0        5738 Jan  1  1970 ./app/js/private/test/image/non_ascii/bin2.runfiles/_main/js/private/node-bootstrap/bootstrap.cjs
```
jbedard: That's a lot bigger :/

js/private/node-bootstrap/bootstrap.cjs
acozzette: We are actually doing that already on line 71, so that should prevent these variables from unexpectedly changing.

docs/troubleshooting.md
```diff
@@ -252,6 +253,14 @@ For the same reason the report is produced by the target's own `node_toolchain`,
 not by the one used for build actions. A test pinned to an old Node version runs
 the coverage reporter under that version too.
 
+### Code run from the program's own `exit` listeners is not covered
+
+The reporter is registered by a `--require` preload, so it necessarily runs
```
jbedard, line 258: What if we did `setTimeout` within the `exit` hook? Or is that risky or asking for trouble? :/

js/private/node-bootstrap/coverage.cjs
```diff
@@ -0,0 +1,86 @@
...
+    try {
+        // A test reporting its own coverage owns it, except under split
+        // post-processing, where bazel drops what the test wrote.
+        if (env.SPLIT_COVERAGE_POST_PROCESSING !== '1' && env.COVERAGE_OUTPUT_FILE) {
+            const stat = require('node:fs').statSync(env.COVERAGE_OUTPUT_FILE, {
+                throwIfNoEntry: false,
+            })
+            if (stat && stat.size > 0) return
+        }
+        if (!require('node:fs').existsSync(report)) {
```
jbedard: Maybe put the `require()`s at the root of the `onExitCoverageReport` method to (a) group them and (b) remove duplicates?

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -43,3 +43,97 @@ if (
...
+        JS_BINARY__COVERAGE_REPORT
+    )
+
+    // NB: this runs before any 'exit' listener the program registers, whereas the
+    // launcher used to run after the process was gone. A program that writes its own
+    // report to COVERAGE_OUTPUT_FILE from an 'exit' listener therefore also gets a
+    // report generated here. What ends up published is unchanged: the merger prefers
+    // the program's own report, except under split post-processing where bazel discards
+    // it and ours is published instead.
+    process.on('exit', (code) => {
```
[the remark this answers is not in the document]
acozzette: This is true but there's not much we can do about it, so I think we should accept it as a known drawback of this approach.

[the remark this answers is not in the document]
acozzette: I think this is OK, we just need to rely on the test to be somewhat well-behaved.

js/private/node-bootstrap/bootstrap.cjs
acozzette: No, and actually I think it was unnecessary to refer to it here, so I removed the reference to it.

js/private/test/coverage/BUILD.bazel
acozzette: According to Claude, this would not work on Windows because on Windows we don't patch `process.execPath` to point to the node wrapper: https://github.com/aspect-build/rules_js/blob/ac921051504861aebaa0726814d7c7dca803f41a/js/private/node-bootstrap/bootstrap.cjs#L21 I went ahead and deleted this, though. I believe we have a lot of stuff that doesn't work on Windows yet, so it's probably not worth cluttering things with this logic, especially since Windows is mostly not covered on CI anyway.

js/private/node-bootstrap/bootstrap.cjs
acozzette: Good idea, this is done now (and moved to coverage.cjs).

js/private/node-bootstrap/bootstrap.cjs
acozzette: Done

js/private/node-bootstrap/bootstrap.cjs
acozzette: Done

js/private/test/image/non_ascii/custom_layer_groups_test_node.listing
acozzette: I was able to undo most of the size difference by moving the new coverage logic to coverage.cjs.

docs/troubleshooting.md
acozzette: Hmm, interesting idea but I kinda suspect it's probably not worth the trouble. It seems to me that we probably won't miss anything important by losing coverage of the exit listeners.

js/private/node-bootstrap/coverage.cjs
acozzette: Done

docs/troubleshooting.md
jbedard: I'm fine with leaving this as a known limitation for now 👍

js/private/node-bootstrap/coverage.cjs
```diff
@@ -0,0 +1,92 @@
...
+// NB: this runs before any 'exit' listener the program registers, whereas the launcher used
+// to run after the process was gone. A program that writes its own report to
+// COVERAGE_OUTPUT_FILE from an 'exit' listener therefore also gets a report generated here.
+// What ends up published is unchanged: the merger prefers the program's own report, except
+// under split post-processing where bazel discards it and ours is published instead. For
+// the same reason the profile is snapshotted before those listeners run, so code a program
+// executes from its own 'exit' listener is reported as uncovered.
+process.on('exit', function onExitCoverageReport() {
+    try {
+        // Required here rather than at module scope so that loading this module, which
```
jbedard: slop comments (imo)

acozzette: I agree, I was just fixing up some comments in this file and pushed a commit to delete the one here.

jbedard (approved):

Issue #2888, opened 2026-06-13
[Bug]: vite 8 (rolldown) fails under the sandbox because native-fs path canonicalization bypasses the rules_js `fs` patches


macourteau:
## What happened?

I have a Bazel project that uses rules_js to build a vite app. It builds fine with vite 7, but fails with vite 8. To isolate it I made a repro with two packages that are identical except for the vite major version (and the matching `@vitejs/plugin-react` major, since plugin-react 6.x requires vite 8).

`bazel build //vite8:build` fails in `vite:build-html`:

```
[plugin vite:build-html]
Error: The "fileName" or "name" properties of emitted chunks and assets must
be strings that are neither absolute nor relative paths, received
"../../../../../../../../../execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/vite8/index.html".
    at PluginContextImpl.emitFile (.../rolldown/dist/shared/bindingify-input-options-....mjs:978:23)
    at PluginContextImpl.generateBundle (.../vite/dist/node/chunks/node.js:22686:10)
```

The `../../../...` part is the giveaway: the emitted `index.html` fileName has been computed against a root that lives outside the sandbox working directory.

## What I think is going on

As far as I can tell, rules_js launches node with patched `fs` APIs (`lstat`, `readlink`, `realpath`, `readdir`, and so on) so that JS code can't resolve the sandbox input symlinks back out to the real output tree. In effect this emulates `--preserve-symlinks` for everything in the execroot, which is what keeps Node tooling working inside the sandbox.

vite 7 and earlier bundle through rollup plus esbuild's JS pipeline, whose path handling goes through Node's (patched) `fs`. Paths stay inside the sandbox, the `index.html` fileName is computed relative to the project root, and the build works.

vite 8 bundles through rolldown, whose resolver and fs access are native Rust. It seems the native code calls the OS `realpath` directly, which bypasses the rules_js patches. The `index.html` entry, which is a sandbox symlink, gets canonicalized to its real path under the output tree, while `config.root` stays the unresolved sandbox working directory. `vite:build-html` then emits with `path.relative(root, id)`, which produces the `../../../...` escape above, and rolldown rejects it.

So from the rules_js side, the surface seems to be this: the `fs`-patching strategy that emulates `--preserve-symlinks` doesn't reach a bundler that does its fs access in native code. rolldown is the first mainstream one I've hit that does this, but it probably won't be the last.

[...]

PR #2988, opened 2026-08-25
fix: sandbox npm package store for devservers

imanmalekian31:
Add an opt-in `package_store_mode = "sandbox"` mode to `js_run_devserver`.

By default, npm package symlinks in the custom devserver sandbox point back to the Bazel execroot. Bundlers that canonicalize paths using native `realpath()` and enforce a project-root boundary, such as Next.js with Turbopack, reject packages resolved outside the sandbox.

The new mode materializes the npm package store inside the sandbox and rewrites or dereferences `node_modules` symlinks so their real paths remain within the sandbox root.

Files are copied with `COPYFILE_FICLONE`, which uses a copy-on-write filesystem clone when supported and falls back to a regular copy. Both strategies keep devserver writes isolated from Bazel outputs. The sandbox parent directory can be configured with `JS_RUN_DEVSERVER_SANDBOX_DIR`; placing it on the same filesystem as the execroot increases the chance that copy-on-write cloning is available.

The existing behavior remains the default, so current users are unaffected unless they opt in.

Related: #2888

---

### Suggested release notes

`js_run_devserver` now supports `package_store_mode = "sandbox"` for bundlers such as Turbopack that require resolved npm package paths to remain within the project root.

### Test plan

- Added unit coverage for package-store path and sandbox symlink resolution.
- Added integration tests verifying that:
  - scoped and unscoped npm packages resolve inside the sandbox;
  - package-store directories are materialized rather than linked to the execroot;
  - no `node_modules` symlink, including `.bin` and package-internal links, escapes the sandbox;
  - granting sandbox write permissions does not modify execroot files;
  - a devserver that chmods and writes a package file without pre-granted write permissions cannot mutate the execroot file.
- Extended the `js_run_devserver` end-to-end test to exercise incremental syncing with `package_store_mode = "sandbox"`.
- Verified JavaScript syntax, focused unit tests, and diff checks locally.

### DIFF
--- a/js/private/devserver/src/js_run_devserver.mjs
+++ b/js/private/devserver/src/js_run_devserver.mjs
@@ -27,6 +27,19 @@ const syncedTime = new Map()
 const syncedChecksum = new Map()
 const mkdirs = new Set()
 
+// Set of all data file paths (workspace-relative, posix separators) that the sandbox is expected to
+// contain after the current sync. Used to resolve node_modules symlinks to their in-sandbox targets
+// without depending on the order in which entries happen to be synced.
+const entryPaths = new Set()
+
+// When true the package store is materialized inside the sandbox and node_modules symlinks are
+// pointed at it instead of at the execroot. Set from the rule config in main().
+let sandboxPackageStore = false
+
+// How many levels of symlinked directories will be dereferenced into the sandbox before giving up.
+// Guards against symlinked directories that point at one another, which would recurse forever.
+const MAX_DEREF_DEPTH = 32
+
 // Ensure that a directory exists. If it has not been previously created or does not exist then it
 // creates the directory, first recursively ensuring that its parent directory exists. Intentionally
 // synchronous to avoid race conditions between async promises. If we use `await fs.promises.mkdir(p)`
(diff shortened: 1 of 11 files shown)


js/private/devserver/src/js_run_devserver.mjs
```diff
@@ -66,16 +79,76 @@ export function isNodeModulePath(p) {
...
+            break
+        }
+        rel = parent
+    }
+    return undefined
+}
+
+// Stats a path following symlinks, or returns null for a broken link so that the caller can
+// recreate it as a symlink rather than failing the sync.
+async function statFollowingLinks(src, file) {
```
jbedard, line 143: This method only uses `file` in a debug message - is this worth a method instead of inlining it in the 2 use cases I see? Since the method is just  a `.stat(src)` call...

js/private/devserver/src/js_run_devserver.mjs
```diff
@@ -142,9 +215,16 @@ export function friendlyFileSize(bytes) {
...
     let symlinkMeta = ''
-    if (isNodeModulePath(file)) {
+    if (sandboxSrc) {
+        // The target is synced into the sandbox, so point at it rather than at the execroot. This
+        // keeps realpath() of the link inside the sandbox root, which bundlers that enforce a
+        // project-root boundary (e.g. Turbopack) require.
+        src = sandboxSrc
+        symlinkMeta = 'sandbox'
+    }
+    if (!symlinkMeta && isNodeModulePath(file)) {
```
jbedard: change to `else if isNodeModulePath(file)`?

js/private/devserver/src/js_run_devserver.mjs
```diff
@@ -196,14 +276,33 @@ async function syncDirectory(file, src, sandbox, writePerm) {
...
+// inode, so a devserver can chmod or modify the sandbox file without mutating the Bazel output.
+async function materializeFile(src, dst, copyOnWrite) {
+    await fs.promises.copyFile(
+        src,
+        dst,
+        copyOnWrite ? fs.constants.COPYFILE_FICLONE : 0
+    )
+}
+
+async function syncFile(
```
jbedard: This line just got auto-formatted but didn't change? That's annoying :/

js/private/devserver/src/js_run_devserver.mjs
```diff
@@ -218,9 +317,11 @@ async function syncFile(file, src, dst, exists, lstat, writePerm) {
         mkdirpSync(path.dirname(dst))
     }
 
+    const copyOnWrite = sandboxPackageStore && isPackageStorePath(file)
```
jbedard: Why not use `fs.constants.COPYFILE_FICLONE` 100% of the time? Why only packages? I'm just curious if we should always be doing this...

js/private/devserver/src/js_run_devserver.mjs
```diff
@@ -357,20 +523,17 @@ async function syncFiles(files, sandbox, writePerm, doSync) {
...
-            // Node module file
-            if (is1pPackageStoreDep(file)) {
-                // 1p package store dep
-                packageStore1pDeps.push(fileInfo)
-            } else {
-                // Other node_modules file
-                otherNodeModulesFiles.push(fileInfo)
-            }
+        if (isPackageStorePath(file)) {
+            // Package store dep; 1p always, 3p as well when the package store is synced into the
```
jbedard: I see no logic about 1p vs 3p here, why is it mentioned in the comment?

js/private/devserver/src/js_run_devserver.mjs
```diff
@@ -750,8 +947,15 @@ async function cycleSyncRecurse(cycle, file, isDirectory, sandbox, writePerm) {
     onProcessEnd(() => sandbox && removeSandbox(sandbox) && (sandbox = null))
 
     try {
+        // The sandbox lives under the OS temp dir by default. JS_RUN_DEVSERVER_SANDBOX_DIR moves it
+        // elsewhere; placing it on the same filesystem as the execroot may allow package-store
+        // files to use copy-on-write clones rather than full copies.
+        const sandboxParent =
```
jbedard, line 924: I wonder if `js_run_devserver` can switch from the node `os.tmpdir()` to the bazel `actions.declare_directory()`? 🤔

That might be a separate discussions/PR, but you bring up some good reasons for trying it...

jbedard: Oh, I think just because this is a `bazel run` and not a `bazel build` that can declare output directories 😢

js/private/devserver/src/js_run_devserver.mjs
imanmalekian31: `statFollowingLinks` also centralizes the dangling-symlink behavior: both sync protocols should treat a broken link as non-dereferenceable and optionally log it instead of aborting the sync. I kept the helper so the regular and watch protocol call sites retain identical behavior.

js/private/devserver/src/js_run_devserver.mjs
imanmalekian31: Done in `d568a60f`; this is now an `else if`.

js/private/devserver/src/js_run_devserver.mjs
imanmalekian31: Agreed — I collapsed the unchanged `syncFile` signature back to one line in `d568a60f` to remove the formatting-only noise.

js/private/devserver/src/js_run_devserver.mjs
imanmalekian31: Good point. `COPYFILE_FICLONE` already falls back to a regular copy when cloning is unsupported, so `d568a60f` now uses it for every file copy and removes the package-only conditional.

js/private/devserver/src/js_run_devserver.mjs
imanmalekian31: Right — since this is a `bazel run`, there is no build action that can own a declared output directory. I kept the runtime temp directory plus the optional `JS_RUN_DEVSERVER_SANDBOX_DIR` override; changing the broader sandbox location strategy can be considered separately.

js/private/devserver/src/js_run_devserver.mjs
imanmalekian31: Updated in `d568a60f`. The comment now explains that this loop only orders package-store entries; the rule filters the entries list according to `package_store_mode`.

acozzette (approved):

imanmalekian31:
Could you review my PR? @jbedard

PR #2989, opened 2026-08-25
refactor: replace the bash launcher with a JavaScript launcher

acozzette:
This change takes our bash launcher, translates it more-or-less literally into JavaScript, and puts hermetic_launcher in front of that, removing our dependence on bash for launching Node executables. (Technically bash may still be involved in some cases, because we have a tiny bash wrapper for node at `js/private/node_bin/node` and a similar one for npm. These wrappers are made available on the PATH, but we do not call them ourselves.)

Currently hermetic_launcher invokes node to run the generated JS launcher, which in turn execs node again to run the real entry point. In certain situations we will need to keep exec'ing node twice, but I think in the common case we can put in a fast path that skips the second exec and goes straight to the main entry point without that.

---

Replace the generated bash launcher with a JavaScript launcher

### Test plan

- New test cases added


e2e/js_image_oci/src/BUILD.bazel
```diff
@@ -62,16 +62,14 @@ js_image_layer(
 
 oci_image(
     name = "image",
-    # Since js_binary depends on bash we have to bring in a base image that has bash
     base = "@debian",
     # This is `/[js_image_layer 'root']/[package name of js_image_layer 'binary' target]/[name of js_image_layer 'binary' target]`
-    cmd = ["/app/src/bin"],
```
jbedard, line 68: This looks like a breaking change already...

Issue #2937, opened 2026-07-23
[Bug]: node compile cache pollutes disk


jbedard:
### What happened?

At some point node enabling [compile caching](https://nodejs.org/api/module.html#module-compile-cache) by default, this saves intermediate files to a cache directory which keys based off file path. When file paths are in a temporary bazel sandbox this may lead to polluting the disk on every bazel invocation.

### Version

rules_js: *
nodejs: >=22.1 contains the compiler cache, maybe it was enabled by default in a later version though

### How to reproduce

```shell

```

PR #2991, opened 2026-08-27
refactor: move node compile cache logic into bootstrap.cjs

acozzette:
The bash launcher currently disables node's compile cache only when `NODE_COMPILE_CACHE` is unset (see #2937). When we replace the bash launcher with hermetic_launcher, we are no longer going to have a place to put this logic so that it executes before the Node process starts. This change therefore disables the cache unconditionally with an environment variable but re-enables it at runtime if `NODE_COMPILE_CACHE` is set. This is compatible with hermetic_launcher because we can embed a command line of something like `env NODE_DISABLE_COMPILE_CACHE=1 node ...`

One drawback of this approach is that it inverts the priority of the `NODE_COMPILE_CACHE` and `NODE_DISABLE_COMPILE_CACHE` variables. Node treats the latter as overriding the former, but this change reverses that. I think this is acceptable because it will not be an issue unless the user sets both variables. We cannot easily detect this situation because we do not have a good way to tell whether `NODE_DISABLE_COMPILE_CACHE` was set by the user or just set by us.

---

### Test plan

- New test cases added

### DIFF
--- a/js/private/node-bootstrap/bootstrap.cjs
+++ b/js/private/node-bootstrap/bootstrap.cjs
@@ -1,3 +1,10 @@
+// The launcher exports NODE_DISABLE_COMPILE_CACHE unconditionally, and then we re-enable
+// the cache here if necessary.
+if (process.env.NODE_COMPILE_CACHE) {
+    delete process.env.NODE_DISABLE_COMPILE_CACHE
+    require('node:module').enableCompileCache?.(process.env.NODE_COMPILE_CACHE)
+}
+
 const patchfs = require('./fs.cjs').patcher
 const {
     JS_BINARY__FS_PATCH_ROOTS,
--- a/js/private/js_binary.sh.tpl
+++ b/js/private/js_binary.sh.tpl
@@ -402,9 +402,8 @@ if [ "${COVERAGE_DIR:-}" ]; then
 fi
 
 # Disable Node's module compile cache by default (aspect-build/rules_js#2937).
-if [ -z "${NODE_COMPILE_CACHE:-}" ] && [ -z "${NODE_DISABLE_COMPILE_CACHE:-}" ]; then
-    export NODE_DISABLE_COMPILE_CACHE=1
-fi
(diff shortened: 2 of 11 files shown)


jbedard:
Did you manually verify this? If you do `NODE_COMPILE_CACHE=~/my/cache bazel run //js:binary` will it fill that dir with cache artifacts?

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -1,3 +1,10 @@
+// The launcher exports NODE_DISABLE_COMPILE_CACHE unconditionally, and then we re-enable
+// the cache here if necessary.
+if (process.env.NODE_COMPILE_CACHE) {
+    delete process.env.NODE_DISABLE_COMPILE_CACHE
+    require('node:module').enableCompileCache?.(process.env.NODE_COMPILE_CACHE)
```
jbedard, line 5: When did node start supporting the `?.`? Maybe the `NODE_COMPILE_CACHE` var is new enough we can assume it will be ok using this syntax?

jbedard (approved):

acozzette:
> Did you manually verify this? If you do `NODE_COMPILE_CACHE=~/my/cache bazel run //js:binary` will it fill that dir with cache artifacts?

I tried just now with `NODE_COMPILE_CACHE=$HOME/tmp/node_cache bazel run //js/private/test/js_binary_sh:one` and it seemed to work, as I see that cache directory now has a subdirectory `v22.22.0-x64-9ac5647c-1000` with a few files in it.

js/private/node-bootstrap/bootstrap.cjs
acozzette: I believe this syntax is called optional chaining and was added in Node 14, so we should be able to rely on it: https://nodejs.org/en/blog/release/v14.0.0

jbedard: Wow, I didn't realize it was actually that old lol

PR #2992, opened 2026-08-27
feat(js_run_binary): add flag controlling default of silent_on_success

acozzette:
`silent_on_success = True` has a cost: the action must run the program under the `spawn_binary` wrapper so that stdout and stderr can be buffered and replayed on failure.

This change adds the `//js:silent_on_success` build flag, which sets the default behavior for `silent_on_success`. Individual targets can still override it by setting the attribute explicitly.

---

The flag `--@aspect_rules_js//js:silent_on_success` now controls the default behavior of `silent_on_success` for `js_run_binary`.

### Test plan

- New test cases added

### DIFF
--- a/js/private/js_run_binary.bzl
+++ b/js/private/js_run_binary.bzl
@@ -154,10 +154,6 @@ def js_run_binary(
             If `None` (the default), the behavior is controlled by the `//js:silent_on_success`
             build flag, which defaults to `True`.
 
-            Enabling this has a cost: the action must run the program under a wrapper process that
-            buffers both streams so they can be replayed if the program fails, rather than simply
-            `exec`ing Node.
-
         use_execroot_entry_point: Use the `entry_point` script of the `js_binary` `tool` that is in the execroot output tree
             instead of the copy that is in runfiles.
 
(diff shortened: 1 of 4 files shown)


js/private/js_run_binary.bzl
```diff
@@ -414,6 +417,13 @@ See https://github.com/aspect-build/rules_js/tree/main/docs#using-binaries-publi
         "//conditions:default": {},
     }) if use_execroot_entry_point == None else {}
 
+    # When silent_on_success is None, behavior is controlled by the //js:silent_on_success flag.
+    if silent_on_success == None:
+        silent_on_success = select({
+            Label("//js:_silent_on_success_true"): True,
+            "//conditions:default": False,
```
jbedard, line 424: This is basically `True if x == True else False`? But I guess we must use a `select()` for this to work and can't just directly use a label for a bool flag? 🤦

jbedard (approved):

js/private/js_run_binary.bzl
acozzette: Yeah, unfortunately I think this is the only way to access the flag from a macro, since the flag doesn't have a real value at load time.

PR #2994, opened 2026-08-28
refactor: apply chdir from the node bootstrap, not the launcher

acozzette:
This is one of the more complex behaviors remaining in the bash launcher, so let's move it into the node bootstrap in preparation for adopting hermetic_launcher.

bootstrap.cjs now cds into `JS_BINARY__CHDIR` before deleting that variable. Deleting it prevents child processes or worker threads from attempting to cd to that same directory a second time.

This necessitated updating `js_run_devserver` to pass the normalized chdir value via the devserver's JSON config instead of through `JS_BINARY__CHDIR`.

---

### Test plan

- New test cases added

### DIFF
--- a/js/private/js_run_devserver.bzl
+++ b/js/private/js_run_devserver.bzl
@@ -91,9 +91,6 @@ def _js_run_devserver_impl(ctx):
     if ctx.attr.grant_sandbox_write_permissions:
         config["grant_sandbox_write_permissions"] = "1"
     if launcher.chdir:
-        # The devserver runs the tool in a sandbox that mirrors the runfiles tree, so it has to
-        # reproduce the working directory there itself. It cannot read JS_BINARY__CHDIR at
-        # runtime: bootstrap.cjs consumes that variable when it applies the chdir.
         config["chdir"] = launcher.chdir
 
     ctx.actions.write(config_file, json.encode(config))
--- a/js/private/js_run_devserver.bzl
+++ b/js/private/js_run_devserver.bzl
@@ -90,6 +90,11 @@ def _js_run_devserver_impl(ctx):
         config["command"] = ctx.attr.command
     if ctx.attr.grant_sandbox_write_permissions:
         config["grant_sandbox_write_permissions"] = "1"
+    if launcher.chdir:
+        # The devserver runs the tool in a sandbox that mirrors the runfiles tree, so it has to
+        # reproduce the working directory there itself. It cannot read JS_BINARY__CHDIR at
+        # runtime: bootstrap.cjs consumes that variable when it applies the chdir.
+        config["chdir"] = launcher.chdir
 
     ctx.actions.write(config_file, json.encode(config))
 
--- a/js/private/node-bootstrap/bootstrap.cjs
+++ b/js/private/node-bootstrap/bootstrap.cjs
@@ -29,11 +29,6 @@ if (JS_BINARY__CHDIR) {
               JS_BINARY__CHDIR
           )
         : JS_BINARY__CHDIR
-    if (JS_BINARY__LOG_DEBUG) {
-        console.error(
-            `DEBUG: ${JS_BINARY__LOG_PREFIX}: changing directory to user specified package ${dir}`
-        )
-    }
     try {
         process.chdir(dir)
     } catch (e) {
--- a/js/private/node-bootstrap/bootstrap.cjs
+++ b/js/private/node-bootstrap/bootstrap.cjs
@@ -18,18 +18,10 @@ const {
     JS_BINARY__PATCH_NODE_FS,
 } = process.env
 
-// working directory
-//
-// js_binary(chdir) / js_run_binary(chdir). The launcher leaves us in the root of the output
-// tree (or in the runfiles tree for a js_test) and hands the requested directory over in
-// JS_BINARY__CHDIR; the move itself happens here so that it does not depend on the launcher
-// being a shell script.
-//
-// Nothing in this process may observe the old cwd, so this runs before the fs patches and
-// before coverage.cjs, which snapshots process.cwd() for the reporter it later spawns.
+// Change directory as indicated by the chdir option on js_binary or js_run_binary.
 if (JS_BINARY__CHDIR) {
     // An "external/<repo>" chdir names a path in the bin directory rather than one relative to
-    // the cwd we were given, which for a js_test is the runfiles tree. See #2827.
+    // the cwd we were given, which for a js_test is the runfiles tree.
     const dir = JS_BINARY__CHDIR.startsWith('external/')
         ? require('node:path').join(
               JS_BINARY__EXECROOT,
@@ -50,13 +42,9 @@ if (JS_BINARY__CHDIR) {
         )
         process.exit(1)
     }
-    // process.chdir() does not maintain PWD, which bash cd did and which some tools read.
+    // process.chdir() does not maintain PWD, so let's update it here.
     process.env.PWD = process.cwd()
-    // The variable is an instruction to this process, and it has now been carried out, so
-    // consume it. A child node process re-enters this bootstrap through the node wrapper's
-    // --require, and a worker thread gets a copy of this environment; both already start in
-    // the directory we just moved to, and applying a relative chdir on top of it would
-    // compound (in a worker, process.chdir() throws outright).
+    // Prevent child processes and worker threads from attempting to cd a second time.
     delete process.env.JS_BINARY__CHDIR
 }
 
--- a/js/private/node-bootstrap/bootstrap.cjs
+++ b/js/private/node-bootstrap/bootstrap.cjs
@@ -7,12 +7,58 @@ if (process.env.NODE_COMPILE_CACHE) {
 
 const patchfs = require('./fs.cjs').patcher
 const {
+    BAZEL_BINDIR,
+    JS_BINARY__BINDIR,
+    JS_BINARY__CHDIR,
+    JS_BINARY__EXECROOT,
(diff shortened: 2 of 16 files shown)


js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -7,13 +7,47 @@ if (process.env.NODE_COMPILE_CACHE) {
...
 } = process.env
 
+// Change directory as indicated by the chdir option on js_binary or js_run_binary.
+if (JS_BINARY__CHDIR) {
+    // An "external/<repo>" chdir names a path in the bin directory rather than one relative to
+    // the cwd we were given, which for a js_test is the runfiles tree.
+    const dir = JS_BINARY__CHDIR.startsWith('external/')
+        ? require('node:path').join(
+              JS_BINARY__EXECROOT,
+              BAZEL_BINDIR || JS_BINARY__BINDIR,
```
jbedard, line 28: Why do we need to support both here?

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -7,13 +7,47 @@ if (process.env.NODE_COMPILE_CACHE) {
...
+              JS_BINARY__CHDIR
+          )
+        : JS_BINARY__CHDIR
+    if (JS_BINARY__LOG_DEBUG) {
```
jbedard: WDYT about just dropping this instead of moving it?

js/private/node-bootstrap/bootstrap.cjs
acozzette: `BAZEL_BINDIR` gets used for regular build actions, but for tests and `bazel run` we rely on `JS_BINARY__BINDIR`.

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -7,13 +7,47 @@ if (process.env.NODE_COMPILE_CACHE) {
...
+    JS_BINARY__EXECROOT,
     JS_BINARY__FS_PATCH_ROOTS,
     JS_BINARY__LOG_DEBUG,
     JS_BINARY__LOG_PREFIX,
     JS_BINARY__NODE_WRAPPER,
     JS_BINARY__PATCH_NODE_FS,
 } = process.env
 
+// Change directory as indicated by the chdir option on js_binary or js_run_binary.
+if (JS_BINARY__CHDIR) {
```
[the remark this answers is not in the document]
acozzette: I think this is a real issue but probably unlikely to be a problem in practice. Let's not worry about it because it would be difficult to do the necessary surgery on NODE_OPTIONS to put our preload first, especially with hermetic_launcher.

js/private/js_run_devserver.bzl
```diff
@@ -90,6 +90,8 @@ def _js_run_devserver_impl(ctx):
         config["command"] = ctx.attr.command
     if ctx.attr.grant_sandbox_write_permissions:
         config["grant_sandbox_write_permissions"] = "1"
+    if launcher.chdir:
+        config["chdir"] = launcher.chdir
```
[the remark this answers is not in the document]
acozzette: I think this OK since it seems unlikely that anyone would want to use a `js_run_devserver` as a `tool` of a `js_run_binary`.

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -7,13 +7,47 @@ if (process.env.NODE_COMPILE_CACHE) {
...
+            `DEBUG: ${JS_BINARY__LOG_PREFIX}: changing directory to user specified package ${dir}`
+        )
+    }
+    try {
+        process.chdir(dir)
+    } catch (e) {
+        console.error(
+            `FATAL: ${JS_BINARY__LOG_PREFIX}: could not change directory to '${dir}': ${e.message}`
+        )
+        process.exit(1)
```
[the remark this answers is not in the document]
acozzette: This is a real behavior change but I think it's acceptable because it's such a narrow situation--a failed `chdir` can now influence `expected_exit_code`.

js/private/node-bootstrap/bootstrap.cjs
acozzette: That sounds good, I went ahead and dropped it.

js/private/node-bootstrap/bootstrap.cjs
acozzette: This follows the pattern we use in a few places in the launcher script: https://github.com/aspect-build/rules_js/blob/fb007625f7401fead4bb1880cb73c1b1fcb5c2ee/js/private/js_binary.sh.tpl#L134

jbedard: WDYT about dropping that pattern in a future PR? We can still set it and deprecate it, but stop using it ourselves so we can delete it in 4.0?

js/private/node-bootstrap/bootstrap.cjs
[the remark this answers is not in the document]
jbedard: Didn't this basically exist before as well?

js/private/node-bootstrap/bootstrap.cjs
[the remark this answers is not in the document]
jbedard: Should we use a different exit code? 2? https://www.baeldung.com/linux/status-codes#misuse-shell-builtin

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -7,13 +7,42 @@ if (process.env.NODE_COMPILE_CACHE) {
...
+    }
+    // process.chdir() does not maintain PWD, so let's update it here.
+    process.env.PWD = process.cwd()
```
jbedard, line 41: Why `process.cwd()` instead of `dir`?

jbedard: I assume `process.cwd()` is more reliable/accurate, just curious...

jbedard (approved):

js/private/test/chdir/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
...
+
+js_test(
+    name = "cwd_test",
+    chdir = package_name(),
+    entry_point = "cwd.mjs",
+)
+
+js_binary(
+    name = "cwd_bin",
+    chdir = package_name(),
```
jbedard, line 35: what about testing `"../chdir"` or `"../chdir/" + package_Name()` or `"."`.

js/private/node-bootstrap/bootstrap.cjs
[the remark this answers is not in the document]
acozzette: I don't think so, because I think Codex is saying that without this PR the bash launcher cd's to `chdir` before invoking Node at all, so all preloads will see the correct working directory. With my PR a preload set via `NODE_OPTIONS` could theoretically observe the working directory before we cd to `chdir`.

Or maybe what you're saying is that it's already true that users can run preloads before our bootstrap if they use `NODE_OPTIONS`? In that case I agree, and it seems like that could in principle cause problems.

js/private/node-bootstrap/bootstrap.cjs
acozzette: I don't think `dir` would work because it's usually a relative path whereas `PWD` should be absolute.

js/private/node-bootstrap/bootstrap.cjs
acozzette: That sounds good to me, let me send out a follow-up PR for that.

The behavior here is actually kind of weird, because if chdir starts with `external/` then we cd to an absolute path, which is not necessarily inside the current working directory. I believe this would happen in particular for tests and `bazel run`. In those cases the working directory would be the runfiles directory, but if chdir starts with `external/` then we will cd out of the runfiles to the bin directory + `/external/...`.

I think if we're in the runfiles directory then we should make sure to stay there. If we make that behavior change then that will hopefully remove the need to reference `BAZEL_BINDIR` and `JS_BINARY__BINDIR` here.

js/private/test/chdir/BUILD.bazel
acozzette: Sure, I added a couple tests with variations on those. I didn't add any that start with `..` but only because that would take us out of the bin directory.

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -35,7 +35,9 @@ if (JS_BINARY__CHDIR) {
         console.error(
             `FATAL: ${JS_BINARY__LOG_PREFIX}: could not change directory to '${dir}': ${e.message}`
         )
-        process.exit(1)
+        // Exit 2 rather than 1 so that a broken chdir is less likely to collide with an
```
jbedard: This seems like slop. Put at most one line saying why `2`, it is unrelated to `expected_exit_code`

acozzette: Oh, I thought this comment was correct. Were you thinking of choosing exit code 2 not because of `expected_exit_code` but just because it's a more accurate code for this situation?

jbedard: I was just thinking 2 is more appropriate given my 30s of research, please google it yourself and/or ask claude what the best error code is. Nothing to do with `expected_exit_code` really

acozzette: Claude recommended 1 and I think it had a pretty convincing rationale. 1 is a general-purpose error code and is what `cd nonexist_dir` will return. 2 is more for invalid options passed to a builtin and is what you get if you do `cd --nonexistent_option`. So I went back to returning 1.

PR #2995, opened 2026-08-28
fix: resolve chdir inside the tree the launcher leaves us in

acozzette:
Currently we have an inconsistency in the way `js_binary(chdir)` is handled. If the `js_binary` target that sets `chdir` is in the main module, then we just cd directly from there into a subdirectory. However, if it is in a foreign module, then we cd to an absolute path within the bin directory. In the case of `bazel run` or a `js_test` that means we cd out of the runfiles tree in that situation.

This change fixes the problem by making sure we stay within the current directory regardless of whether the `chdir` came from a foreign module or the main module.

I also cleaned up the launcher script slightly by removing the behavior of falling back on `JS_BINARY__BINDIR` when `BAZEL_BINDIR` is unset.

---

Always resolve chdir inside the tree the launcher leaves us in.

### Test plan

- New test cases added

### DIFF
--- a/js/private/js_binary.sh.tpl
+++ b/js/private/js_binary.sh.tpl
@@ -43,7 +43,7 @@ fi
 function resolve_capture_path {
     case "$1" in
     bazel-out/*) echo "$PWD/$1" ;;
-    *) echo "$PWD/${BAZEL_BINDIR:-$JS_BINARY__BINDIR}/$1" ;;
+    *) echo "$PWD/$BAZEL_BINDIR/$1" ;;
     esac
 }
 
@@ -131,9 +131,9 @@ function logf_debug {
 function resolve_execroot_bin_path {
     local short_path="$1"
     if [[ "$short_path" == ../* ]]; then
-        echo "$JS_BINARY__EXECROOT/${BAZEL_BINDIR:-$JS_BINARY__BINDIR}/external/${short_path:3}"
+        echo "$JS_BINARY__EXECROOT/$BAZEL_BINDIR/external/${short_path:3}"
     else
-        echo "$JS_BINARY__EXECROOT/${BAZEL_BINDIR:-$JS_BINARY__BINDIR}/$short_path"
+        echo "$JS_BINARY__EXECROOT/$BAZEL_BINDIR/$short_path"
     fi
 }
 
--- a/js/private/node-bootstrap/bootstrap.cjs
+++ b/js/private/node-bootstrap/bootstrap.cjs
@@ -7,28 +7,27 @@ if (process.env.NODE_COMPILE_CACHE) {
 
 const patchfs = require('./fs.cjs').patcher
 const {
-    BAZEL_BINDIR,
-    JS_BINARY__BINDIR,
+    BUILD_WORKSPACE_DIRECTORY,
     JS_BINARY__CHDIR,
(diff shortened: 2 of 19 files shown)


js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -7,28 +7,27 @@ if (process.env.NODE_COMPILE_CACHE) {
...
-          )
-        : JS_BINARY__CHDIR
+    let dir = JS_BINARY__CHDIR
+    // chdir is relative to the root of the output tree, where an external repository's package sits
+    // under "external/<repo>". The runfiles tree instead gives every repository a top-level
+    // directory beside our own, so re-point the path there rather than leaving the tree. Bazel sets
+    // TEST_SRCDIR for a test and BUILD_WORKSPACE_DIRECTORY for `bazel run`, and these are the two
+    // situations where we will be in the runfiles tree.
+    if ((TEST_SRCDIR || BUILD_WORKSPACE_DIRECTORY) && dir.startsWith('external/')) {
+        dir = '../' + dir.slice('external/'.length)
```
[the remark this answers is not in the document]
acozzette: This is true, but it's not a regression because this situation already didn't work. Parent and child processes both doing a `chdir` also fails even when there's no external repository involved. So I don't think it's worth trying to fix this. I think we should eventually deprecate and remove `chdir` on `js_binary` because it's just confusing and hard to support. `chdir` makes a lot of sense on an action invoking a `js_binary` (such as `js_run_binary`), but not on the `js_binary` itself.

js/private/js_binary.sh.tpl
```diff
@@ -43,7 +43,7 @@ fi
 function resolve_capture_path {
     case "$1" in
     bazel-out/*) echo "$PWD/$1" ;;
-    *) echo "$PWD/${BAZEL_BINDIR:-$JS_BINARY__BINDIR}/$1" ;;
+    *) echo "$PWD/$BAZEL_BINDIR/$1" ;;
```
jbedard, line 46: So now `BAZEL_BINDIR` is the only one used, and it was already essentially required, so that implies nothing changed here basically?

Is `JS_BINARY__BINDIR` still _assigned_ though? I know it is often used within scripts...

acozzette: These helper functions are only used when `use_execroot_entry_point` is enabled or when runfiles are not enabled (which would be just Windows). `use_execroot_entry_point` mostly just applies to `js_run_binary`, which always sets `BAZEL_BINDIR`, so we can safely rely on that. On Windows I'm not as confident but I suspect a lot of stuff is already broken on Windows anyway since we're not testing it comprehensively on CI.

`JS_BINARY__BINDIR` is still getting assigned and I see that the one place we still use it is in js_run_devserver.mjs: https://github.com/aspect-build/rules_js/blob/c3c809d0300e02d0f98f547480d79f99fac267ea/js/private/devserver/js_run_devserver.mjs#L410 I don't know how hard it would be to fix that.

jbedard: I mean it's used by customers, so I just want to make sure we're still setting it even if not using it

acozzette: Oh I see, yes I've left it alone and haven't done anything to unset it.

jbedard (approved):

PR #2998, opened 2026-09-01
fix: make js_image_layer tar output independent of the spawn strategy

acozzette:
This change enables `--experimental_use_hermetic_linux_sandbox` on CI and makes a small fix to `js_image_layer` in order to do so.

I think this CI coverage would be valuable to have, because without this flag it is still possible to escape the `linux-sandbox` by following symlinks. Ideally we want to make sure we are limiting ourselves to declared inputs and not inadvertently relying on files that are supposed to be outside the sandbox.

---

### Test plan

- New test cases added

### DIFF
--- a/js/private/js_image_layer.mjs
+++ b/js/private/js_image_layer.mjs
@@ -143,7 +143,14 @@ function _mtree_file_line(key, content) {
     const dest = vis(key)
     // Due to filesystems setting different bits depending on the os we have to opt-in
     // to use a stable mode for files.
-    return `${dest} uid={{UID}} gid={{GID}} time=0 mode={{FILE_MODE}} type=file content=${vis(
+    //
+    // nlink=1 keeps the archive independent of how the sandbox materializes our inputs.
+    // Two entries may share a `content` file (the js_binary launcher appears both as the
+    // image entrypoint and inside the runfiles tree); bsdtar collapses such a pair into a
+    // hard link, but only when the source file's st_nlink > 1. The standard linux sandbox
+    // symlinks inputs (nlink 1) while the hermetic one hard links them (nlink 2), so
+    // without this the tar bytes would differ by spawn strategy.
+    return `${dest} uid={{UID}} gid={{GID}} time=0 mode={{FILE_MODE}} nlink=1 type=file content=${vis(
         content
     )}`
 }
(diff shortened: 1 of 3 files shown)


thesayyn (approved):

js/private/js_image_layer.mjs
```diff
@@ -143,7 +143,14 @@ function _mtree_file_line(key, content) {
...
     // to use a stable mode for files.
-    return `${dest} uid={{UID}} gid={{GID}} time=0 mode={{FILE_MODE}} type=file content=${vis(
+    //
+    // nlink=1 keeps the archive independent of how the sandbox materializes our inputs.
+    // Two entries may share a `content` file (the js_binary launcher appears both as the
+    // image entrypoint and inside the runfiles tree); bsdtar collapses such a pair into a
+    // hard link, but only when the source file's st_nlink > 1. The standard linux sandbox
+    // symlinks inputs (nlink 1) while the hermetic one hard links them (nlink 2), so
+    // without this the tar bytes would differ by spawn strategy.
+    return `${dest} uid={{UID}} gid={{GID}} time=0 mode={{FILE_MODE}} nlink=1 type=file content=${vis(
```
jbedard, line 153: Where did you/robot learn about this fix? Has this been done anywhere else (in the bazel ecosystem)?

acozzette: Claude somehow figured out this fix. The issue is that we call bsdtar and give it an mtree spec to tell it the files we want to put it in the output .tar file representing an image layer. Apparently if the mtree spec defines two or more entries that both point to the same file via `content=somefile` and `somefile` has more than one link, then bsdtar will encode them as hard links, unless we explicitly say we only want one link via `nlink=1`.

There does seem to be a precedent for this as I see that tar.bzl does the same thing: https://github.com/bazel-contrib/tar.bzl/pull/99

I suspect we should do the same thing in rules_py if we want to make it compatible with `--experimental_use_hermetic_linux_sandbox`, because I found that a bunch of tests under e2e/cases/oci there fail with what appears to be the same issue.

jbedard: Would you be able to make the same change there? Also enabling the hermetic sandbox if it can be done easily but otherwise just create a github issue for it?

jbedard (approved):

js/private/js_image_layer.mjs
acozzette: Sure, I'm working on a PR right now to do this for rules_py.

PR #3002, opened 2026-09-02
feat: add hermetic launcher as opt-in alternative to bash launcher

acozzette:
This change introduces a new launcher based on hermetic_launcher. For now the bash launcher is still the default, but you can opt in to the new launcher with `--@aspect_rules_js//js:hermetic_launcher`.

This launcher consists of a small native binary produced with hermetic_launcher that execs a generated JS launcher. If possible, the JS launcher jumps straight to the main entry point. However, this is not possible when `expected_exit_code` is used, in which case we have to exec node a second time. We also have to re-exec node when certain environment variables or flags are embedded in the js_binary that have an effect at node startup.

---

An experimental new launcher based on hermetic_launcher is now available: opt in by passing `--@aspect_rules_js//js:hermetic_launcher`.

### Test plan

- New test cases added


e2e/js_image_oci/src/BUILD.bazel
```diff
@@ -65,13 +65,12 @@ oci_image(
     # Since js_binary depends on bash we have to bring in a base image that has bash
     base = "@debian",
     # This is `/[js_image_layer 'root']/[package name of js_image_layer 'binary' target]/[name of js_image_layer 'binary' target]`
-    cmd = ["/app/src/bin"],
```
jbedard, line 68: If it's off by default why was this required? Or was this just bad practice and could be done in its own "refactor" PR beforehand?

docs/hermetic_launcher.md
```diff
@@ -0,0 +1,109 @@
...
+This is the first step towards replacing the bash launcher outright. The
+JavaScript launcher is deliberately an almost literal translation of the bash
+one -- same order, same messages, same decisions -- so that the two can be read
+side by side. `js/private/test/snapshots/launcher.sh` and
+`js/private/test/snapshots/launcher.cjs` are checked-in expansions of both, kept
+up to date by `//js/private/test:write_launcher` and
+`//js/private/test:write_launcher_js`, and diffing them is how a change to
+either is reviewed.
+
+## What is not implemented
```
jbedard, line 32: Lets label this as "Unsupported Deprecated Features", and maybe declare these features on the bash launcher as deprecated? Changing behaviour at runtime via env vars should be deprecated and the use of `run_binary` should be used instead basically?

docs/hermetic_launcher.md
```diff
@@ -0,0 +1,109 @@
...
+    result is not re-split on whitespace. A single-quoted segment of a `fixed_arg`
+    is left alone, as bash would have left it.
+-   **A custom rule built on `js_binary_lib.create_launcher` must republish
+    `launcher_js`.** That output group is how `js_image_layer` tells a
+    hermetic-launcher binary from a bash-launcher one: the two keep the values
+    that have to be rewritten for hermeticity in different files.
+
+## Keeping the two launchers in sync
+
+The JavaScript launcher is a transliteration of the bash one and has to stay that
```
jbedard, line 90: "The JavaScript launcher"? Shouldn't that be "The hermetic [JavaScript?] launcher" or something like that?

docs/hermetic_launcher.md
```diff
@@ -0,0 +1,109 @@
...
+either is reviewed.
+
+## What is not implemented
+
+The JavaScript launcher does not implement stdout capture, stderr capture, exit
+code capture or `silent_on_success`. It ignores `JS_BINARY__STDOUT_OUTPUT_FILE`,
+`JS_BINARY__STDERR_OUTPUT_FILE`, `JS_BINARY__EXIT_CODE_OUTPUT_FILE` and
+`JS_BINARY__SILENT_ON_SUCCESS` rather than honoring them.
+
+Nothing in rules_js asks the launcher for those anymore: `js_run_binary`
```
jbedard, line 39: This paragraph doesn't seem belong in public docs? Would anyone care? I guess we do need to mention what users should do moving forward (use `run_binary`) if they want those deprecated features but this paragraph says far more then that.

docs/hermetic_launcher.md
```diff
@@ -0,0 +1,109 @@
...
+forwards `stdout`, `stderr`, `exit_code_out` and `silent_on_success` to
+bazel-lib's `run_binary`, which captures through its own spawn wrapper -- a
+process that outlives the program and can do the work they need once it has
+exited. The launcher's implementation of them is legacy compatibility for code
+outside rules_js that sets the variables by hand (#2955), and it stays in the
+bash launcher.
+
+`expected_exit_code` _is_ implemented, since it is a `js_binary` attribute with
```
jbedard, line 47: spam

docs/hermetic_launcher.md
```diff
@@ -0,0 +1,109 @@
...
+no other home.
+
+Everything else the bash launcher does is reproduced: the `--bazel-bindir` flag,
```
jbedard, line 50: Is this paragraph necessary? I think just mentioning the difference implies "everything else is the same"

docs/hermetic_launcher.md
```diff
@@ -0,0 +1,109 @@
...
+## Keeping the two launchers in sync
+
+The JavaScript launcher is a transliteration of the bash one and has to stay that
+way until it replaces it. The rest of the suite cannot check that: a target gets
+one launcher per configuration, so every other both-launcher test skips one side
+and CI covers the other by running the whole suite again with the flag on. That
+catches breakage but not drift -- a bash-launcher change that was never ported
+can potentially leave both launchers passing every test.
+
+`//js/private/test/launcher_sync` closes that gap. A configuration transition
```
jbedard, line 97: We probably shouldn't be mentioning internal tests in a public doc?

js/private/test/entry_point_quoting/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
+load("@bazel_lib//lib:testing.bzl", "assert_contains")
```
jbedard, line 1: Lets add this test in a separate PR ahead of time?

jbedard: Along with a lot of others like js/private/test/fixed_args/BUILD.bazel I think?

js/BUILD.bazel
```diff
@@ -51,6 +51,21 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
+# This flag selects the experimental hermetic launcher: a native launcher binary
+# stamped by hermetic_launcher which execs node on a generated JavaScript launcher,
+# in place of the generated bash launcher script. See docs/hermetic_launcher.md.
+bool_flag(
+    name = "hermetic_launcher",
```
jbedard, line 58: `use_hermetic_launcher`? I think we've used that terminology elsewhere, although idk what's best?

docs/hermetic_launcher.md
```diff
@@ -0,0 +1,109 @@
+# The hermetic launcher
```
jbedard, line 1: This entire doc seems to be slop written for a rules_js developer outlining the history the launcher. This should be written for a rules_js API user, not developer, someone who only knows the public APIs with little or not implementation details...

js/private/js_binary.cjs.tpl
```diff
@@ -0,0 +1,681 @@
+// This JavaScript file is the launcher for the NodeJS JavaScript file
```
jbedard, line 1: WDYT about trying to reduce the size of the template to something minimal (just the templated info) that then does a `require()` on the larger file with the real code? Then it doesn't get cloned for every single `js_binary()` etc?

jbedard: Or maybe more here is templated then I thought such as all the "Values baked in at analysis time"

e2e/js_image_oci/src/BUILD.bazel
acozzette: It's off by default, but we do exercise it in CI. I split this off into a separate refactor PR.

js/private/test/entry_point_quoting/BUILD.bazel
acozzette: Sounds good, I sent out #3013 for this. I think the other tests can't be landed in advance without also including code changes to make the tests pass.

js/BUILD.bazel
acozzette: Good point, I went with `use_hermetic_launcher`. That seems consistent with our other flag names here that mostly start with a verb.

docs/hermetic_launcher.md
acozzette: Done

docs/hermetic_launcher.md
acozzette: Good point, I fixed the references in this file.

docs/hermetic_launcher.md
acozzette: Yeah, I deleted this paragraph.

docs/hermetic_launcher.md
acozzette: Good point, deleted.

docs/hermetic_launcher.md
acozzette: Fixed

docs/hermetic_launcher.md
acozzette: I agree, I cut out most of the contents and just left the important stuff that a user might want to be aware of.

js/private/js_binary.cjs.tpl
acozzette: I think this is a good idea, because most of this template is not specific to a single js_binary. I would lean toward leaving it as a follow-up, though, just to keep this PR from getting too big and complicated.

js/private/js_binary.bzl
```diff
@@ -432,7 +561,142 @@ def _bash_launcher(ctx, nodeinfo, entry_point_path, log_prefix_rule_set, log_pre
...
+                for fixed_arg in fixed_args
+                for token in _shell_tokenize(fixed_arg)
+            ]),
+            "{{log_prefix_rule_set}}": _quote(log_prefix_rule_set),
+            "{{log_prefix_rule}}": _quote(log_prefix_rule),
+            "{{node_options}}": "\n".join([
+                _NODE_OPTION_JS.format(quoted_value = _quote(value))
+                for value in node_options
+            ]),
+            "{{node_patches}}": _quote(ctx.file._node_patches.short_path),
```
jbedard, line 689: Does this break path-mapping?

js/private/js_binary.cjs.tpl
jbedard: Another idea is to dump the target specific data into a JSON file. But then every invocation adds a `fs.read*` + `json.decode` so maybe doing that as a bazel template like you have is better?

jbedard:
@codex

js/private/js_binary.bzl
acozzette: I don't think this line is a problem for path mapping since it uses `short_path`, which is the same regardless of the compilation mode or platform. But path mapping already doesn't work for producing a `js_binary`, because we embed variables like `JS_BINARY__BINDIR` and `JS_BINARY__COMPILATION_MODE` that differ depending on the compilation mode.

We could potentially improve this, but it's not obvious to me that it would be a significant performance benefit. I would guess that producing a js_binary launcher script is usually pretty cheap compared to invoking it, and when it comes to invoking it we do have good support for path mapping.

js/private/js_binary.cjs.tpl
acozzette: Yeah, I'll have to think about it...the thing I want to do next after this PR is set things up so that if you're not using `expected_exit_code`, we run everything in a single node process instead of spawning or exec'ing node a second time. That will probably involve moving some logic around, maybe moving some things out of the generated JS launcher and into `js/private/node-bootstrap`. Hopefully after that is done it will be easier to look at what's left and see what would be the best opportunities to shorten the generated JS launcher further.

jbedard:
Do you have some data measuring how much performance gain this has? What's the bash overhead vs hermtic-launcher+.js overhead?

jbedard (approved):
LGTM as long as you've verified the js_launcher is actually faster then the bash launcher...

acozzette:
> LGTM as long as you've verified the js_launcher is actually faster then the bash launcher...

It seems to be a modest performance regression on its own and won't be an improvement until I put in the fast path. Should I go ahead and add in the fast path to this PR?

Claude tells me it did some benchmarking and found that the bash launcher took about 73 ms, the hermetic_launcher in this PR took about 83 ms, and the hermetic_launcher with fast path took about 46 ms.

jbedard:
> hermetic_launcher with fast path

When would we not use that fast path?

acozzette:
> > hermetic_launcher with fast path
> 
> When would we not use that fast path?

We can't do it when `expected_exit_code` is used, since that inherently requires spawning another process and waiting for it to return. Actually maybe we could if we're willing to use an exit listener for this, but I would have to see if there's a way to do it reliably.

The other blockers are the `node_options` attribute on `js_binary` and setting various `NODE_*` environment variables. Those are tricky to handle with the fast path since they generally need to be set before node starts in order to take effect.

jbedard:
`expected_exit_code` is an odd edge case and we can be slow or handle that differently (wrap it bash script even)

For the `node_options` attribute on `js_binary` don't we have to convert that into direct arguments on the node/hermetic-launcher binary and have no runtime logic?

acozzette:
> For the `node_options` attribute on `js_binary` don't we have to convert that into direct arguments on the node/hermetic-launcher binary and have no runtime logic?

I was thinking about doing that, but this won't necessarily work in cases where the node options involve evaluating an environment variable such as the case here: https://github.com/aspect-build/rules_js/blob/40006e068d6ee5f274db1adb18d4c7b397b882e8/examples/stack_traces/BUILD.bazel#L17 In that particular case it seems to be just doing runfiles resolution, which hermetic-launcher supports, so we might be able to handle that situation if we can detect it.

acozzette:
@jbedard I think this is now ready for another look. I added in the fast path so that we now usually need only one node invocation.

docs/hermetic_launcher.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+With the flag on, a `js_binary`'s executable is a small native binary stamped by
+[hermetic_launcher](https://github.com/hermeticbuild/hermetic-launcher) which
+does nothing but resolve its runfiles and `execve` node on a generated
+JavaScript file, `<name>_/<name>.cjs`.
+
+## Unsupported deprecated features
+
+The hermetic launcher does not implement stdout capture, stderr capture, exit
```
jbedard, line 23: "does not implement deprecated ..."? (just insert the word deprecated in this sentence..?)

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,735 @@
...
+        (_match, braced, bare) => process.env[braced || bare] || ''
+    )
+}
+
+function setEnv(name, value) {
+    process.env[name] = expandEnvRefs(value)
+}
+
+function setEnvIfUnset(name, value) {
+    if (!process.env[name]) {
```
jbedard, line 83: What if it is set but empty? Does that count as "unset"?

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,735 @@
...
+    // rather than preloaded. Report the arguments the exec path would have used, so that a
+    // tool forwarding execArgv to a worker still reproduces the patched runtime.
+    process.execArgv = ['--require', process.env.JS_BINARY__NODE_PATCHES, ...nodeOptions]
+
+    // Required here rather than baked into the stub as a --require: the patches read
+    // JS_BINARY__ variables that only exist once the launcher above has run.
+    require(process.env.JS_BINARY__NODE_PATCHES)
+
+    // Runs the entry point as the main module, so that `require.main === module` holds for it.
+    require('node:module').runMain()
```
jbedard, line 625: Does this return anything? Or return at all? Should we `process.exit` on the next line?

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,735 @@
...
+    }
+    return `${process.env.JS_BINARY__EXECROOT}/${shortPath}`
+}
+
+function exitWith(exitCode) {
+    logfDebug(`exit code: ${exitCode}`)
+    process.exit(exitCode)
+}
+
+process.on('uncaughtException', (err) => {
```
jbedard, line 212: Are uncaught exceptions not going to exit the program anyway?

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,735 @@
...
+    logfFatal(String((err && err.message) || err))
+    logfDebug(String((err && err.stack) || err))
+    exitWith(1)
+})
+
+// Ends this process the way node ended, so that callers see a signal-terminated
+// process rather than an interposed 128+N exit code. That is what they would
+// have seen had this launcher been able to exec node instead of spawning it.
+function reraiseSignal(signal, exitCode) {
```
jbedard, line 221: This is only used in the non-fastpath, can we move it there just to isolate that code to the end of the file?

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,735 @@
...
+runfiles = normalizePath(runfiles)
+if (!path.isAbsolute(runfiles)) {
+    // Must be absolute: the runfiles path may be relative to the cwd, and we may
+    // be about to change directory.
+    runfiles = normalizePath(path.join(cwd(), runfiles))
+}
+process.env.JS_BINARY__RUNFILES = runfiles
+// Set RUNFILES_DIR if not already set so that tools such as @bazel/runfiles can
+// locate runfiles.
+process.env.RUNFILES_DIR = process.env.RUNFILES_DIR || runfiles
```
jbedard, line 269: If `RUNFILES_DIR` was already set why did we calculate `runfiles`? When/why would they be different and is this the same behaviour the bash launcher has?

docs/hermetic_launcher.md
acozzette: Good catch, done — the section is now "Unsupported deprecated features" and the sentence reads "does not implement the deprecated stdout capture, ...".

js/private/test/snapshots/launcher.cjs
acozzette: Yes, set-but-empty counts as unset. That is deliberate: the bash launcher writes `if [[ -z "${VAR:-}" ]]` for the same `env` entry, and `-z` is true for empty as well as unset, so `!process.env[name]` matches it exactly (every `process.env` value is a string, so `""` is the only falsy one — `"0"` is truthy). I added a comment saying so.

js/private/test/snapshots/launcher.cjs
acozzette: It returns, but we must not exit after it. `runMain()` returns as soon as the entry point's *top level* finishes; anything asynchronous is still queued at that point. Verified:

```
$ cat main.cjs
setTimeout(() => console.log("ASYNC_WORK_RAN"), 50)
console.log("TOP_LEVEL_DONE")
$ node driver.cjs        # sets process.argv then calls runMain()
TOP_LEVEL_DONE
runMain RETURNED
ASYNC_WORK_RAN
```

So a `process.exit` on the next line would truncate every async program. Letting it fall off the end is also what the exec path does — node exits on its own once the event loop drains. I extended the comment to say this.

js/private/test/snapshots/launcher.cjs
acozzette: They are — node exits 1 on an uncaught exception either way. The handler is not there to exit, it is there so that an unexpected failure in the launcher itself reports as `FATAL: ...` like the dozen explicit `logfFatal` call sites do, instead of a raw stack trace through launcher internals (the stack is still printed under `JS_BINARY__LOG_DEBUG`). Its counterpart is the `process.removeAllListeners('uncaughtException')` on the in-process path, without which the program's own exceptions would be reported as launcher failures. That rationale had been trimmed out of the comment, which is what invited the question — I put it back.

js/private/test/snapshots/launcher.cjs
acozzette: Done — moved into the `else` branch next to `forwardSignal`, which is the only place that calls it.

acozzette:
Sorry, Claude replied to a bunch of stuff without asking me. I haven't looked at its comments so I don't know if they make sense.

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,735 @@
...
+}
+
+// When the cwd is a build action execroot the bindir hangs off it (BAZEL_BINDIR resolves from the
+// cwd), so the cwd is the execroot even if its path contains a "bazel-out" segment (e.g. a matching
+// output base). Otherwise scan the output tree for the execroot (runfiles, or a nested js_binary in
+// the bindir).
+if (
+    bazelOutSegment &&
+    (!process.env.BAZEL_BINDIR ||
+        !isDirectory(path.join(cwd(), process.env.BAZEL_BINDIR)))
```
jbedard, line 291: Wouldn't `BAZEL_BINDIR` not being a directory be a bigger problem? Can we drop that check or put it in a more generic check somewhere higher up?

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,735 @@
...
+        logfFatal(`Unexpected RUNFILES_MANIFEST_FILE value ${manifest}`)
+        exitWith(1)
+    }
+}
+if (!runfiles) {
+    logfFatal('RUNFILES_DIR environment variable is not set')
+    exitWith(1)
+}
+runfiles = normalizePath(runfiles)
+if (!path.isAbsolute(runfiles)) {
```
jbedard, line 261: I assume we don't need to normalize the path before using `path.*`.

We're mixing a lot of `path.*` methods and path normalizing that seems very over complicated here 🤔

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,735 @@
...
+    // Must be absolute: the runfiles path may be relative to the cwd, and we may
+    // be about to change directory.
+    runfiles = normalizePath(path.join(cwd(), runfiles))
+}
+process.env.JS_BINARY__RUNFILES = runfiles
+// Set RUNFILES_DIR if not already set so that tools such as @bazel/runfiles can
+// locate runfiles.
```
jbedard, line 268: Tell claude to stfu, and allow comments to be wider then a 1980s terminal so we don't wrap just for 2 words

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,735 @@
...
+        logfDebug(
+            `inheriting JS_BINARY__EXECROOT ${process.env.JS_BINARY__EXECROOT} from parent js_binary process as JS_BINARY__USE_EXECROOT_ENTRY_POINT is set`
+        )
+    } else {
+        // We are in execroot or in some other context all together such as a nodejs_image or a manually run js_binary
+        process.env.JS_BINARY__EXECROOT = cwd()
+    }
+
+    if (!process.env.JS_BINARY__NO_CD_BINDIR) {
+        if (!process.env.BAZEL_BINDIR) {
```
jbedard, line 326: We seem to do this check in multiple places, yet only this one has the giant explanation error message. Isn't this var required (via `--bazel-bindir`) and we can validate it once up top?

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,735 @@
...
+        '--',
+        entryPoint,
+        ...args,
+    ]
+
+    if (process.env.JS_BINARY__LOG_INFO) {
+        logfInfo(['running', process.env.JS_BINARY__NODE_BINARY, ...nodeArgs].join(' '))
+    }
+
+    if (!expectedExitCode) {
```
jbedard, line 645: Can we just force the bash launcher if we have `expectedExitCode`? Even in the future when we drop the bash launcher, can we just add a bash wrapper if `js_binary(expected_exit_code)` is used?

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,735 @@
...
+    }
+}
+
+// ==============================================================================
+// Environment
+// ==============================================================================
+
+setEnv("JS_BINARY__BINDIR", "bazel-out/k8-fastbuild/bin")
+setEnv("JS_BINARY__COMPILATION_MODE", "fastbuild")
+setEnv("JS_BINARY__TARGET_CPU", "k8")
```
jbedard, line 119: Doesn't this (and the 2 above) break path-mapping?

jbedard:
@codex review

js/private/test/snapshots/launcher.cjs
jbedard: Sounds over-complicated...

js/private/test/snapshots/launcher.cjs
acozzette: This logic and the huge error message are copied nearly verbatim from the bash launcher. Maybe we should cut down on the size of this message in both places?

It's kind of complicated because this variable is not strictly required. `bazel run` and `bazel test` don't set it and this is allowed because they run in the runfiles directory. I would be tempted to make `BAZEL_BINDIR` optional because I don't think there is a good reason to require it except maybe when `use_execroot_entry_point` is enabled. But that's not specific to this launcher as both launchers do the same thing in this regard.

js/private/test/snapshots/launcher.cjs
acozzette: We could do that, though honestly I don't think it would buy us very much. We still have to keep the spawn path in the generated JS launcher to use as a fallback for old node versions and Windows, so it wouldn't simplify things a whole lot.

jbedard: "old node version and Windows"? I don't see any docs about that...?

js/private/test/snapshots/launcher.cjs
acozzette: It breaks path mapping for the build action that stamps out the launcher template, but that is already broken for the bash launcher for the same reason. I haven't really tried to get path mapping working for this, because I suspect that creating the launcher is probably pretty cheap compared to running it.

js/private/test/snapshots/launcher.cjs
acozzette: It's in the comment a few lines below:
```
        // process.execve is POSIX-only and was added in Node 22.15; when it is
        // unavailable we fall through to spawning node below.
```

js/private/test/snapshots/launcher.cjs
jbedard: You're certain it doesn't cause things like `js_run_binary` consuming this to also break path-mapping? Can `js_run_binary` still support path-mapping even if the `tool` being run differs between these?

Your right the `expand_template` (or whatever cheap action it is) for creating this file is fine, it's more running it I'm concerned about...

acozzette: Yes, I'm pretty sure because we have good test coverage for mapping. I believe the reason it's not an issue for js_run_binary is that tools that run as part of the build are always built in opt mode for whatever the host CPU is, so they don't vary based on the target platform.

js/private/js_image_layer.bzl
```diff
@@ -180,23 +180,60 @@ The default layer groups are as follows and always created.
...
+        fail("expected {} to have exactly one launcher_js file, got {}".format(binary.label, launchers))
+    return launchers[0]
+
+def _write_js_launcher(ctx, launcher_js):
+    "Sanitizes the JavaScript launcher the way _write_launcher does the bash one."
+    launcher = ctx.actions.declare_file("%s_launcher.cjs" % ctx.label.name)
+
+    substitutions = {
+        "// " + _PATCH_POINT: _JS_LAUNCHER_PREAMBLE,
+        'setEnv("JS_BINARY__BINDIR", "%s")' % launcher_js.root.path: 'setEnv("JS_BINARY__BINDIR", process.cwd())',
```
jbedard: Can we not replace something more obvious/predictable? If the top of the js has `const FOO = undefined` it would be less awkward to search+replace then this...

jbedard: Or maybe the `setEnv("JS_BINARY__BINDIR", ...)` statement can do `etEnv("JS_BINARY__BINDIR", global.JS_IMAGE_LAYER_BINDIR ? process.pwd() : ....)` and then we only have to inject `global.JS_IMAGE_LAYER_BINDIR=true` into the script instead of this awkward search+replace?

js/private/test/launcher/worker_fork_probe.mjs
```diff
@@ -12,24 +12,42 @@ const child = execFileSync(
     [...process.execArgv, '-e', 'console.log(process.env.PATH)'],
     { encoding: 'utf8' }
 )
-parentPort.postMessage({ execArgv: process.execArgv, childPath: child.trim() })
+parentPort.postMessage({
```
jbedard, line 15: Is this a test that can or should be merged into main ahead of time?

js/private/test/snapshots/launcher.cjs
```diff
@@ -0,0 +1,520 @@
...
+    )
+}
+
+function setEnv(name, value) {
+    process.env[name] = expandEnvRefs(value)
+}
+
+// An empty value counts as unset, matching the `[[ -z ]]` test the bash launcher
+// writes for the same env entry.
+function setEnvIfUnset(name, value) {
```
jbedard: Some of these could potentially go into `util.cjs`? Is there any reason not to?

js/private/test/snapshots/launcher.cjs
acozzette: Yeah, I agree. It doesn't seem worth going this far out of our way just to try to improve the error message if the launcher crashes. I went ahead and deleted this logic.

js/private/test/snapshots/launcher.cjs
acozzette: Good question, I'm not really sure. This logic was copied from the bash launcher. I went ahead and simplified this quite a bit because I realized that we can rely on the hermetic_launcher stub to set `RUNFILES_DIR` for us. All we really need to do then is set `JS_BINARY__RUNFILES` to the absolute path.

js/private/test/snapshots/launcher.cjs
acozzette: Fixed

js/private/test/snapshots/launcher.cjs
acozzette: Good point, this does seem more complicated than necessary. I simplified it down to just one line here.

js/private/test/snapshots/launcher.cjs
acozzette: Actually I don't think it would necessarily be unusual for `BAZEL_BINDIR` to not exist, because it's a relative path, and if this js_binary was spawned by another one then we could be in an arbitrary working directory. But this change isn't part of my PR anymore anyway, because this logic moved into js/private/node-bootstrap/launcher.cjs in my previous PR.

js/private/js_binary.bzl
```diff
@@ -432,7 +605,153 @@ def _bash_launcher(ctx, nodeinfo, entry_point_path, log_prefix_rule_set, log_pre
...
+                for token in _shell_tokenize(fixed_arg)
+            ]),
+            "{{log_prefix_rule_set}}": _quote(log_prefix_rule_set),
+            "{{log_prefix_rule}}": _quote(log_prefix_rule),
+            "{{node_options}}": "\n".join([
+                _NODE_OPTION_JS.format(quoted_value = _quote(value))
+                for value in node_options.all
+            ]),
+            "{{stub_node_options}}": json.encode(node_options.stub),
+            "{{env_configures_node_startup}}": json.encode(_env_configures_node_startup(envs)),
```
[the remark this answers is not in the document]
acozzette: This is an interesting one. It seems to me that in a case where `NODE_OPTIONS=--require=$JS_BINARY__RUNFILES/...` is set via the `env` attr, the value set in `RunEnvironmentInfo` is wrong because the variable is not evaluated. This isn't a problem for the bash launcher because it will just overwrite `NODE_OPTIONS` with the correct value at runtime. But presumably if we start up node with this variable set incorrectly then it will crash before we have a chance to fix the variable.

I'll have to think about this but I'm currently inclined to say we should just disallow variable evaluation in the `env` and `node_options` attributes, since that is difficult to support with hermetic_launcher.

js/private/js_binary.bzl
```diff
@@ -289,31 +300,174 @@ _ENV_SET = """export {var}={quoted_value}"""
...
+# attribute on js_binary, the hermetic launcher needs to re-launch node for them to take
+# effect.
+#
+# TZ, FORCE_COLOR, and NO_COLOR are intentionally absent from this list. Node relies on these
+# variables but picks up changes made to them at runtime.
+_NODE_STARTUP_ENV_PREFIXES = [
+    "NODE_",
+    "UV_",
+    "V8_",
+    "OPENSSL_",
```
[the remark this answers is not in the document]
acozzette: I don't think this is an issue since Bazel unsets `HOME` in build actions to keep them hermetic. `HOME` would be passed through to `bazel run`, but even there we should be OK since `RunEnvironmentInfo` will set everything the user provided to the `env` attribute.

js/private/node-bootstrap/util.cjs
```diff
@@ -110,10 +146,86 @@ function checkExecutableFile(what, p) {
...
+// ==============================================================================
+// Signals
+// ==============================================================================
+
+// Node does not forward termination signals to any child process, so the signals are
+// trapped and forwarded manually. The handlers are removed on the first signal so that a
+// second one terminates the launcher.
+function forwardSignals(child) {
+    const forward = (signal) => () => {
+        process.removeAllListeners('SIGTERM')
```
jbedard: Should we only remove our own listener and not "all"?

js/private/node-bootstrap/util.cjs
```diff
@@ -110,10 +146,86 @@ function checkExecutableFile(what, p) {
...
+        process.removeAllListeners('SIGINT')
```
jbedard: Should we only remove `signal` and not both `SIGTERM` and `SIGINT`?

js/private/test/launcher/worker_fork_probe.mjs
acozzette: Yes, I split this off into a new PR: #3020.

js/private/test/snapshots/launcher.cjs
acozzette: Good idea, I had Claude move all the helper functions it could into util.cjs so that we're not duplicating them for every launcher.

js/private/node-bootstrap/util.cjs
acozzette: Yes, that seems reasonable. I made that change and also split it off into a separate PR since that is a behavior change to the bash launcher that we might as well land first: #3023

js/private/node-bootstrap/util.cjs
acozzette: That makes sense, I updated it to just remove our own listeners.

js/private/js_image_layer.bzl
acozzette: Yeah, this does seem messier than necessary. I like your `const` idea and went with that.

docs/hermetic_launcher.md
```diff
@@ -0,0 +1,30 @@
+# The hermetic launcher
+
+A `js_binary` is normally invoked through a generated bash script
+(`js/private/js_binary.sh.tpl`), which works out where node, the fs patches and
+the entry point are, exports a set of `JS_BINARY__*` variables, changes into the
+root of the output tree, and finally execs node. That is a shell process and a
+few hundred lines of path resolution on every invocation, and it cannot run at
+all where there is no bash.
+
+The hermetic launcher is an experimental alternative, off by default:
```
jbedard, line 10: This says "off by default:" and then shows how to enable it. It should be "off by default, to enable:"?

js/private/node-bootstrap/util.cjs
```diff
@@ -0,0 +1,245 @@
...
+
+// Emit a log line to stderr.
+//
+// We use fs.writeSync rather than console.error, so that the line is flushed before
+// the process exits.
+//
+// The message is collapsed onto one line, since callers wrap long messages across source
+// lines and may log an exception.
+function log(level, message) {
+    const collapsed = message.trim().replace(/\s+/g, ' ')
```
jbedard, line 27: Why the `.replace()`? If we keep it it should be `\s\s+` I think?

js/private/test/fixed_args/BUILD.bazel
```diff
@@ -510,34 +535,39 @@ shell_expansion_case(
...
+    bash_only = True,
     fixed_args = ["a|b"],
     want_exit_code = "2",
     want_stderr = "syntax error",
 )
 
 shell_expansion_case(
     name = "redirect_breaks_launcher",
+    bash_only = True,
     fixed_args = ["a>b"],
```
jbedard, line 563: Shouldn't this properly be quoted? I don't understand why a lot of these are "bash only"?

js/private/js_binary.bzl
```diff
@@ -431,7 +630,159 @@ def _bash_launcher(ctx, nodeinfo, entry_point_path, log_prefix_rule_set, log_pre
...
 
-    return launcher, toolchain_files, normalized_chdir
+    return launcher
+
+def _compile_stub(ctx, embedded_args, transformed_args, output_file):
+    """Stamps a launcher binary from the prebuilt template stub.
+
+    This is `hermetic_launcher.compile_stub` reimplemented so the finalizer toolchain can
+    be named by Label; see the comment on _FINALIZER_TOOLCHAIN_TYPE. Drop this in favour
+    of the upstream helper once it takes Labels.
```
jbedard, line 640: Is there an issue or PR to reference for this?

js/private/test/fixed_args/BUILD.bazel
acozzette: Most of these with `want_exit_code` set are bash-only because they will crash the bash launcher but not the hermetic launcher. I think the hermetic launcher behavior behavior is more correct in this respect because arguments like `a>b` shouldn't cause a crash. The arguments are written as raw unquoted strings in the bash launcher, so it's easy to crash it with shell syntax like this.

Ideally we should probably fix this by wrapping the arguments in double quotes in the bash launcher, but this may be tricky to get right without breaking anything. Currently you can do things like `fixed_args = ["$(rootpaths :foo)"]` and this will expand to multiple arguments but would break if we quoted it in a naive way. So we would need to tokenize it at analysis time, which is what this PR does for the hermetic launcher.

js/private/js_binary.bzl
acozzette: Yes, the upstream fix actually just got merged recently: https://github.com/hermeticbuild/hermetic-launcher/pull/75 I added a note here saying we can drop this workaround once that fix makes it into a release.

js/private/node-bootstrap/util.cjs
acozzette: Good point, this is no longer necessary and I removed it. In an earlier version of this PR it had an effect but it no longer does anything useful.

docs/hermetic_launcher.md
acozzette: This is now fixed.

acozzette:
I asked Claude to benchmark the overhead of the two launchers and it found the bash launcher's mean launch time to be 82.4 ms. The hermetic launcher with the in-process fast path was 58.2 ms and with the slow path (re-execing Node) it was 95.7 ms.

jbedard:
> I asked Claude to benchmark the overhead of the two launchers and it found the bash launcher's mean launch time to be 82.4 ms. The hermetic launcher with the in-process fast path was 58.2 ms and with the slow path (re-execing Node) it was 95.7 ms.

So with modern node and not using the odd-features this is a net-win for startup overhead? 👍

js/private/test/fixed_args/BUILD.bazel
jbedard: If the hermetic one is an improvement then we can ignore/document the improvement and bash bugs - but this diff says `bash_only = True` implying that the test doesn't work with the hermetic one?

acozzette: Yeah, the test doesn't pass with the hermetic one because the test is asserting that this results in an error. Maybe instead I should assert that this succeeds and just mark it `hermetic_only`?

jbedard: > Maybe instead I should assert that this succeeds and just mark it hermetic_only?

Or pass a flag saying "only works on hermetic" or "is broken with bash". But right now `shell_expansion_case(bash_only=True)` implies it only _works_ on with bash...

acozzette: OK, I fixed these so that they test the hermetic launcher only and verify that it doesn't error out on this shell syntax.

jbedard (approved):

PR #3005, opened 2026-09-04
perf: drive the coverage report from the manifest, not a runfiles glob

acozzette:
As of today, we effectively glob the whole runfiles directory and then match each entry against the coverage manifest. This is slow because it ends up being O(number of runfiles * length of coverage manifest).

This change speeds up the coverage reporting by using the coverage manifest instead of globbing.

---

### Test plan

- New test cases added

### DIFF
--- a/js/private/coverage/bundle/c8.js
+++ b/js/private/coverage/bundle/c8.js
@@ -41,9 +41,15 @@ const pwd = path.join(
 )
 process.chdir(pwd)
 
+// Same list as COVERAGE_EXTENSIONS in js/private/coverage/extensions.bzl, which is what
+// decides the contents of COVERAGE_MANIFEST. c8 drops a manifest entry whose extension is
+// not listed here, so the two must agree; c8's own default list has neither .mts nor .cts.
+const extensions = new Set(['.mjs', '.mts', '.cjs', '.cts', '.ts', '.js', '.jsx', '.tsx'])
+
 const report = new Report({
     include: include,
     exclude: include.length === 0 ? ['**'] : [],
+    extension: [...extensions],
     reportsDirectory: process.env.COVERAGE_DIR,
     tempDirectory: process.env.COVERAGE_DIR,
     resolve: '',
@@ -51,15 +57,6 @@ const report = new Report({
     all: true,
     reporter: ['lcovonly'],
 })
-
-// js_binary's instrumentation lists .mts and .cts, so they reach COVERAGE_MANIFEST, but
-// c8's default extension list has neither and would drop them from the report.
-for (const ext of ['.mts', '.cts']) {
-    if (!report.exclude.extension.includes(ext)) {
-        report.exclude.extension.push(ext)
-    }
-}
-const extensions = new Set(report.exclude.extension)
 
 // Bazel already computed the exact set of instrumented files and handed it to us in
 // COVERAGE_MANIFEST, so membership is a lookup. c8 otherwise answers the same question by
--- a/js/private/coverage/bundle/c8.js
+++ b/js/private/coverage/bundle/c8.js
@@ -2,9 +2,30 @@ import { Report } from 'c8'
 import fs from 'fs'
 import path from 'path'
 
-// Runs in the test action, the only place the V8 data and instrumented sources are
+// Runs in the test action, the only place the V8 data and the instrumented sources are
 // both present. coverage.sh.tpl reads back this exact filename; keep them in sync.
 const stash = path.join(process.env.COVERAGE_DIR, '_rules_js_report.lcov')
+
+const debug = !!process.env.JS_BINARY__LOG_DEBUG
+const timings = []
+
+// Report generation is charged against the test's own timeout, so when something is slow
+// this is the only place that says which part. See docs/troubleshooting.md.
+function logDebug(message) {
+    if (debug) {
+        console.error(`DEBUG: ${process.env.JS_BINARY__LOG_PREFIX}: ${message}`)
+    }
+}
+
+function timed(label, fn) {
+    if (!debug) return fn()
+    const start = process.hrtime.bigint()
+    const value = fn()
+    timings.push(`${label}=${(Number(process.hrtime.bigint() - start) / 1e6).toFixed(0)}ms`)
+    return value
+}
+
+const started = process.hrtime.bigint()
 
 const include = fs
     .readFileSync(process.env.COVERAGE_MANIFEST)
@@ -20,7 +41,7 @@ const pwd = path.join(
 )
 process.chdir(pwd)
 
-new Report({
+const report = new Report({
     include: include,
     exclude: include.length === 0 ? ['**'] : [],
     reportsDirectory: process.env.COVERAGE_DIR,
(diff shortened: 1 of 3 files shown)


js/private/coverage/bundle/c8.js
```diff
@@ -2,10 +2,31 @@ import { Report } from 'c8'
...
+
+// Report generation is charged against the test's own timeout, so when something is slow
+// this is the only place that says which part.
+function logDebug(message) {
+    if (debug) {
+        console.error(`DEBUG: ${process.env.JS_BINARY__LOG_PREFIX}: ${message}`)
+    }
+}
+
+function timed(label, fn) {
```
jbedard: Is this worth keeping?

js/private/coverage/bundle/c8.js
```diff
@@ -2,10 +2,31 @@ import { Report } from 'c8'
 import fs from 'fs'
 import path from 'path'
 
-// Runs in the test action, the only place the V8 data and instrumented sources are
+// Runs in the test action, the only place the V8 data and the instrumented sources are
 // both present. coverage.sh.tpl reads back this exact filename; keep them in sync.
 const stash = path.join(process.env.COVERAGE_DIR, '_rules_js_report.lcov')
 
+const debug = !!process.env.JS_BINARY__LOG_DEBUG
```
jbedard: rename to `LOG_DEBUG` just to make it really loud and obviously constant?

js/private/coverage/bundle/c8.js
```diff
@@ -23,22 +44,51 @@ process.chdir(pwd)
...
+// Bazel already computed the exact set of instrumented files and handed it to us in
+// COVERAGE_MANIFEST, so membership is a lookup. c8 would otherwise answer the same
+// question by globbing the whole runfiles tree and matching every hit against every
+// manifest entry, which is O(files in runfiles x manifest entries) in each test action.
+// Both hooks below replace that with the set Bazel already knows.
+const instrumented = new Set(include.map((f) => path.resolve(pwd, f)))
+
+report.exclude.shouldInstrument = function shouldInstrument(filename) {
+    const resolved = path.resolve(pwd, filename)
+    return extensions.has(path.extname(resolved)) && instrumented.has(resolved)
```
jbedard: The `extensions.has()` check probably doesn't need the resolved filename, should we do that check first to avoid the `path.resolve` if the file extension isn't even supported?

js/private/coverage/bundle/c8.js
```diff
@@ -23,22 +44,51 @@ process.chdir(pwd)
...
+}
+
+report.exclude.globSync = function globSync() {
```
jbedard: Is there an equivalent `globAsync`? What does this method even do?

acozzette: The upstream function we're replacing is here: https://github.com/istanbuljs/test-exclude/blob/3a37faa17cc4f0f602a7c1ec23ef0b0fcf44ab37/index.js#L105 There's no `globAsync` but there is an async `glob` function right below that. I think what they are both intended to do is just to return all files that match the provided criteria.

jbedard: I think they are intended to use `fs.foo` and `fs.fooSync` methods, so we should override both? Since we are doing no `fs` operations you can just make the async version invoke the sync one?

Should we support the `cwd` param? Or just throw if it's not what we expect?

Lets also leave a link in a comment to the method we're overriding? To the docs and/or code, the code might be more relevant to anyone reading our code?

js/private/coverage/bundle/c8.js
```diff
@@ -23,22 +35,80 @@ process.chdir(pwd)
...
+//
+// They are methods of TestExclude, which c8 constructs from the Report options above:
+// https://github.com/istanbuljs/test-exclude/blob/3a37faa17cc4f0f602a7c1ec23ef0b0fcf44ab37/index.js
+const instrumented = new Set(include.map((f) => path.resolve(pwd, f)))
+
+// Replaces shouldInstrument (index.js#L76), which minimatches the filename against
+// every include pattern -- and prepGlobPatterns expands each manifest entry into
+// several. c8 asks this of every executed script, so the cost is manifest size times
+// scripts loaded, in every test action.
+report.exclude.shouldInstrument = function shouldInstrument(filename) {
```
jbedard: Now that I look at [the overridden implementation](https://github.com/istanbuljs/test-exclude/blob/3a37faa17cc4f0f602a7c1ec23ef0b0fcf44ab37/index.js#L76-L103) - why do we need to override it? I think all it does is basic globbing based on `this.include/exclude`...

js/private/coverage/bundle/c8.js
```diff
@@ -23,22 +35,80 @@ process.chdir(pwd)
...
+    // and silently reporting the wrong paths would be worse than failing.
+    if (cwd !== pwd) {
+        throw new Error(
+            `coverage report requested for ${cwd}, but the manifest describes ${pwd}`
+        )
+    }
+    return include.filter((f) => fs.existsSync(path.resolve(pwd, f)))
+}
+
+// TestExclude pairs glob (index.js#L119) with globSync the way fs does. Report only
```
jbedard: I think this paragraph can be summed up in 1 line?

js/private/coverage/bundle/c8.js
```diff
@@ -23,22 +35,80 @@ process.chdir(pwd)
...
+report.exclude.globSync = function globSync(cwd = pwd) {
+    // c8 passes an entry of `src`, and we give it exactly one. Manifest entries are
+    // relative to that directory, so another cwd cannot be answered from the manifest
+    // and silently reporting the wrong paths would be worse than failing.
+    if (cwd !== pwd) {
+        throw new Error(
+            `coverage report requested for ${cwd}, but the manifest describes ${pwd}`
+        )
+    }
+    return include.filter((f) => fs.existsSync(path.resolve(pwd, f)))
```
jbedard: Should we do the `extensions.has(path.extname(...))` check before `existsSync`? Or notice how the [default implementation](https://github.com/istanbuljs/test-exclude/blob/3a37faa17cc4f0f602a7c1ec23ef0b0fcf44ab37/index.js#L116C34-L116C50) always calls `shouldInstrument()` at the end after globbing the fs - should we call `shouldInstrument()` before `existsSync`?

js/private/coverage/bundle/c8.js
acozzette: The thing is that the original implementation linearly checks against everything in `this.include`, so a single `shouldInstrument` call is O(coverage manifest length). Our version should be much faster since we do a proper set lookup.

jbedard: Ok, if it makes a performance difference then sure 👍

I wonder if any of this can be upstreamed? Or if we'd be better of doing `report.exclude = { ... }` and just implementing our own version of the class instead of patching it?

js/private/coverage/bundle/c8.js
acozzette: `extensions.has(path.extname(f)` shouldn't be necessary because it should always return true. We give Bazel the same list of extensions to instrument (e.g. [here](https://github.com/aspect-build/rules_js/blob/b0a07e8c31997648762abaea4dc30635705e0ad7/js/private/js_binary.bzl#L560)) so everything in the coverage manifest should already have one of those extensions.

js/private/coverage/bundle/c8.js
acozzette: Upstreaming a fix sounds good to me, but just for the short term I think it would be good to get this workaround in.

I like the idea of implementing our own version of the class and I went ahead and did that. I also removed the async glob method because we know it's not getting called. I only implemented a subset of the original class, just enough for our use case. This could break in a future version of c8 but at least that will result in a loud error that we can fix.

js/private/coverage/bundle/c8.js
acozzette: Yeah, this does seem overly verbose. I ended up removing the whole thing since I removed the async glob method anyway.

js/private/coverage/bundle/c8.js
```diff
@@ -23,22 +33,70 @@ process.chdir(pwd)
...
+        this.extension = [...extensions]
+    }
+
+    shouldInstrument(filename) {
+        if (!this.extensions.has(path.extname(filename))) return false
+        return this.instrumented.has(path.resolve(this.root, filename))
+    }
 
-new Report({
-    include: include,
```
jbedard, line 29: You're certain we want to remove the `include/exclude/extension` from the `new Report`?

jbedard (approved):

js/private/coverage/bundle/c8.js
acozzette: Yes, because the only thing they're used for is for constructing the `Exclude` object that we are going to replace with our own implementation anyway: https://github.com/bcoe/c8/blob/ae5a0cf4349b92cac910cf7275e724ab7a725b9d/lib/report.js#L21 This does leave us tightly coupled to the internals of the `Report` class, but that seems inevitable unless/until we can upstream a proper fix.

PR #3011, opened 2026-09-08
fix: resolve npmrc credentials before running pnpm install for update_pnpm_lock

### LINKED ISSUE #3004: [Bug]: npm_translate_lock doesn't use credential helper / tokenHelper for update_pnpm_lock
### Problem

When `update_pnpm_lock = True`, running `pnpm install --lockfile-only` fails to authenticate with private registries that use `tokenHelper` in .npmrc, even though package downloads work fine.

### Current Behavior

`npm_translate_lock` handles authentication differently for package downloads vs lock file updates:

1. Package downloads (`npm_import`): aspect_rules_js parses .npmrc, executes tokenHelper, and passes credentials to Bazel's downloader. This works correctly.
2. Lock file updates (`_update_pnpm_lock`): aspect_rules_js runs pnpm install --lockfile-only directly without any .npmrc processing or tokenHelper execution. pnpm then fails with 401 Unauthorized.

Additionally, pnpm itself disallows tokenHelper in project-level .npmrc for security reasons, so even if the file is available, pnpm won't use it:

```
[ERROR] tokenHelper must not be configured in project-level .npmrc
```

### Expected Behavior

`npm_translate_lock` should handle authentication for `pnpm install --lockfile-only` the same way it does for package downloads - by parsing .npmrc, executing tokenHelper if configured, and making credentials available to pnpm (e.g., by writing a temporary .npmrc with _authToken populated).

### Workaround

We're using a preupdate script that:
1. Invokes our Bazel credential helper to get the auth token
2. Replaces the symlinked .npmrc with a regular file containing the original content plus _authToken

```javascript
// Invoke credential-helper, parse response, inject into .npmrc
const output = execSync(`${credentialHelper} get`, { input: JSON.stringify({ uri: REGISTRY_URI }), ... });
const { headers } = JSON.parse(output);
const token = headers.Authorization[0].replace("Bearer ", "");

// Replace symlink with regular file to avoid modifying source tree
if (lstatSync(".npmrc").isSymbolicLink()) {
  unlinkSync(".npmrc");
}
writeFileSync(".npmrc", `${originalNpmrc}\n${REGISTRY_URL}:_authToken=${token}\n`);
```

### Suggested Solution

In _update_pnpm_lock, before executing pnpm:
1. Parse the .npmrc (already done elsewhere in aspect_rules_js)
2. If tokenHelper is configured for any registry, execute it and capture the token
3. Write a modified .npmrc in the external repository with _authToken populated
4. Then run pnpm install --lockfile-only

This would make update_pnpm_lock work seamlessly with private registries that use tokenHelper, matching the behavior of package downloads.

### Version

Output of `bazel --version`: 9.2.0

```
bazel_dep(name = "aspect_rules_js", version = "3.4.1")
```

JavaScript

### How to reproduce

```shell

```

### Any other information?

_No response_

jbedard:
Rewrite the external repository copy of the .npmrc handed to pnpm: tokenHelper entries become _authToken (or _auth for a Basic header) with the helper output, and `$VAR` credentials become their resolved values. The source file is untouched.

Fix #3004

### Test plan

- New test cases added

### DIFF
--- a/e2e/npm_translate_lock_auth_token_helper/MODULE.bazel
+++ b/e2e/npm_translate_lock_auth_token_helper/MODULE.bazel
@@ -19,8 +19,10 @@ npm = use_extension(
 )
 npm.npm_translate_lock(
     name = "npm",
+    data = ["//:package.json"],
     npmrc = "//:.npmrc",
     pnpm_lock = "//:pnpm-lock.yaml",
+    update_pnpm_lock = True,
     verify_node_modules_ignored = "//:.bazelignore",
 )
 use_repo(npm, "npm")
--- a/e2e/npm_translate_lock_auth_token_helper/README.md
+++ b/e2e/npm_translate_lock_auth_token_helper/README.md
@@ -6,3 +6,7 @@ covers a [`tokenHelper`](https://pnpm.io/npmrc#urltokenhelper) declared in the `
 `token-helper.sh` prints the token from `ASPECT_GH_PACKAGES_AUTH_TOKEN`, and
 `ASPECT_NPM_AUTH_TOKEN` must have permission to pull the `@aspect-priv-npm` scope, same as
 the sibling test.
+
+`update_pnpm_lock` is enabled and `test.sh` changes `package.json` so that `pnpm install`
+runs with the private registry, which pnpm can only authenticate with because rules_js
+resolves the `tokenHelper` into the `.npmrc` copy it hands to pnpm.
--- a/e2e/npm_translate_lock_auth_token_helper/test.sh
+++ b/e2e/npm_translate_lock_auth_token_helper/test.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+set -o errexit -o nounset -o pipefail
+
+# Integration test for update_pnpm_lock with a tokenHelper in the project .npmrc
+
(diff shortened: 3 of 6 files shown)


jbedard:
@codex

acozzette (approved):

rafikk:
This resolves our issue reported in #3004.

We were previously working around this with a custom preupdate script that writes to `.npmrc`. Contents in this gist: https://gist.github.com/rafikk/437729019fe61d8f2133d2006f1c1189

Even with this change, I think we'll stick with our current workaround. The reason being: newer versions of pnpm allow you to set registry configuration and scopes in `pnpm-workspace.yaml`. With the preupdate route, we can delete `.npmrc` entirely: no registry or tokenHelper configuration, and just have the preupdate script write a disposable .npmrc file in the output directory with the auth token.

This also allows us to gitignore .npmrc in case someone's tooling introduces an auth token into it, making the risk of the preupdate script even lower.

Thanks for looking into this @jbedard!

jbedard:
>  newer versions of pnpm allow you to set registry configuration and scopes in `pnpm-workspace.yaml`

This makes me think this hack to work with `.npmrc` is not worth merging...

jbedard:
Replaced with #3018

PR #3015, opened 2026-09-14
refactor: move most js_binary launcher logic from bash into launcher.cjs

acozzette:
This change moves most logic out of the bash launcher, keeping only the bare minimum functionality there. It sets things up so that launcher.cjs is responsible for everything that needs to run just once per launch, and bootstrap.cjs is responsible for things that also need to happen for worker threads and child processes created with `child_process.fork`. I moved the `chdir` logic into `launcher.cjs` accordingly, since this is something that should happen just once per launch.

The motivation for this is to make it easier to move away from the bash launcher altogether. We are likely going to have an experimental bash-free launcher based on hermetic_launcher, and once we have two launchers we will want to keep the duplication between them to a minimum until we can drop the bash launcher.

---

### Test plan

- New test cases added


js/private/node-bootstrap/launcher.cjs
```diff
@@ -0,0 +1,404 @@
...
+    process.env.JS_BINARY__FS_PATCH_ROOTS = `${process.env.JS_BINARY__EXECROOT}${path.delimiter}${process.env.JS_BINARY__RUNFILES}`
+}
+
+// Put the node wrapper directory and optionally the npm wrapper directory on the path so that
+// child processes can find them.
+let PATH = process.env.PATH || ''
+if (npmBinDir) {
+    PATH = `${npmBinDir}${path.delimiter}${PATH}`
+}
+process.env.PATH = `${path.dirname(nodeWrapper)}${path.delimiter}${PATH}`
```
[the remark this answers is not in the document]
acozzette: I think this is OK. It's inevitable that users can front-run our `--require` by using `NODE_OPTIONS`, and there's not much we can do it about it since we're moving toward removing the bash script entirely.

js/private/node-bootstrap/launcher.cjs
```diff
@@ -0,0 +1,404 @@
...
+
+// Emit a log line to stderr.
+//
+// We use fs.writeSync rather than console.error, so that the line is flushed before a
+// process.exit that may follow it.
+function logTo(level, message) {
+    fs.writeSync(
+        2,
+        `${level}: ${process.env.JS_BINARY__LOG_PREFIX}: ${message}\n`
+    )
```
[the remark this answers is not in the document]
acozzette: This should be OK for now. `JS_BINARY__STDERR_OUTPUT_FILE` is deprecated and will be going away anyway.

js/private/js_binary.bzl
```diff
@@ -278,6 +278,11 @@ _ATTRS = {
         allow_single_file = True,
         default = Label("@aspect_rules_js//js/private/node-bootstrap:bootstrap.cjs"),
     ),
+    # The per-launch setup the bash launcher runs as node's first --require
```
jbedard: Delete?

js/private/node-bootstrap/bootstrap.cjs
```diff
@@ -2,7 +2,10 @@
 // (aspect-build/rules_js#2937). Note that setting NODE_DISABLE_COMPILE_CACHE
 // at runtime has no effect unless module.enableCompileCache() is subsequently
 // called, in which case it will prevent the cache from being enabled.
-if (!process.env.NODE_COMPILE_CACHE && !process.env.NODE_DISABLE_COMPILE_CACHE) {
```
jbedard, line 5: Auto-formatted? Revert?

js/private/node-bootstrap/launcher.cjs
```diff
@@ -0,0 +1,404 @@
...
+
+// ==============================================================================
+// Logging
+// ==============================================================================
+
+// Emit a log line to stderr.
+//
+// We use fs.writeSync rather than console.error, so that the line is flushed before a
+// process.exit that may follow it.
+function logTo(level, message) {
```
jbedard: Doesn't this always log to `2` and should be renamed `logError`?

js/private/node-bootstrap/bootstrap.cjs
acozzette: Done

js/private/js_binary.bzl
acozzette: Done

jbedard:
How does this effect `js_[run_]binary` launch performance?

js/private/node-bootstrap/launcher.cjs
acozzette: It does always log to stderr, but I actually think the current name makes sense because all log levels (info, debug, fatal, etc.) go to stderr. If we named it `logError` then I would find that confusing because it would seem like "error" was the log level.

jbedard: I still think it needs to be renamed to something, you can't say `logTo()` and then not have a param indicating what you're logging to...?

jbedard (approved):
LGTM if you've verified this doesn't have any performance hit when launching a `js_binary` (directory or via `js_run_binary`).

acozzette:
> How does this effect `js_[run_]binary` launch performance?

Claude benchmarked it and told me that for runs that use the runfiles directory as the working directory (so `js_test` and `bazel run`), the wall time improved by about 6 ms or 7%. However, for `js_run_binary` there was about a 4.5ms increase in CPU time that did not affect wall time. So it seems to me that it's pretty much a wash and neither change is particularly large.

js/private/node-bootstrap/launcher.cjs
acozzette: That makes sense, I renamed it to just `log()` since it always logs to the same place.

PR #3021, opened 2026-09-16
ci: run the 32-core workspaces on Aspect Workflows runners

jeffpignataro:
The root, `examples` and `e2e/js_image_oci` workspaces move from GitHub's `ubuntu-22.04-32core` to the Aspect Workflows `aspect-large` runner pool. Fork PRs keep the GitHub runner: they get no repo secrets and must not run on self-hosted capacity, so the matrix builder emits the runner as JSON and picks by `github.event.pull_request.head.repo.fork`.

**A warming archive nobody writes leaves every job cold.** The pool restores a Bazel output base from an archive that only `aspect ci warming` produces, so `warming.yaml` runs it daily on the same label, mirroring bazel-examples.

**setup-aspect v2026.23.2 hard-fails on Workflows runners.** Its only step there writes `/etc/bazel.bazelrc`, which the runner user cannot; v2026.26.9 writes `~/.bazelrc` instead and only warns when it cannot. All four uses move to v2026.26.9.

`test` gains `timeout-minutes: 90`; self-hosted jobs otherwise run to GitHub's 6h ceiling.

---

### Test plan

- Ran the matrix-builder script locally: unset `IS_FORK` yields `["self-hosted","aspect-workflows","aspect-large"]` for the three large workspaces and `"ubuntu-latest"` for the rest; `IS_FORK=true` yields `"ubuntu-22.04-32core"`. Output parses with `jq`.
- CI on this PR runs the 11 large jobs on the new pool.


.github/workflows/ci-workflows.yaml
```diff
@@ -136,25 +136,35 @@ jobs:
                       slug=${slug//+/_}
                       slug=${slug//./_}
                     fi
-                    # Slow workspaces get a larger runner.
-                    runner=ubuntu-latest
+                    # Slow workspaces get a larger runner: the Aspect Workflows `aspect-large`
+                    # pool (32 vCPU spot, silo-aws). Fork PRs get no repo secrets and must
```
jbedard: Lets not mention "silo" or "aws" when neither of those references are required here.

jeffpignataro: Yeah good call.

jbedard (approved):

PR #3024, opened 2026-09-17
fix(ci): use Bazel 9 for cache warming

jeffpignataro:
The warming job inherits Bazel 7.7.1 from `.bazelversion`, which cannot negotiate the runner cache's RE API 2.3 requirement. In [the failed run](https://github.com/aspect-build/rules_js/actions/runs/35246486449), analysis finishes in 14 seconds, but the CLI stalls before archiving until cancellation.

Pin `USE_BAZEL_VERSION` to `9.0.0` for the warming job. The repository default and Bazel 7/8/9 test matrix remain unchanged.

### Test plan

- Parsed the workflow YAML and checked the warming job's version override; `git diff --check` passes.
- In an isolated Linux reproduction with CLI 2026.33.2 and a mock cache advertising RE API 2.3, Bazel 7.7.1 finishes analysis but hangs until the timeout kills it. Bazel 9.0.0 completes warming in 1.7 seconds. The mock reports requested blobs as already cached; this does not validate real cache or archive uploads.
- Full-repository local validation attempted at `54dda4d02a43a995922b9254f8fe6aff5fbed16f` with CLI 2026.33.2, Bazel 9.0.0, `CI=true`, and the exact warming command. Native Linux ARM64 cannot select the `main_toolchain_24` fixture; emulated Linux x86-64 reaches all 926 targets but fails fetching npm packages because the downloaded `coreutils chmod` exits 139 (twice, on different packages). Both x86-64 attempts report failure and exit normally. Full local warming is therefore not verified.
- Pending: dispatch the warming workflow on a native x86-64 Workflows runner and verify cache population and archive upload.


acozzette (approved):

jbedard:
Can bazel 7+8 use the bazel 9 cache?

jeffpignataro:
> Can bazel 7+8 use the bazel 9 cache?

They can use the cache, but the warming job use the new RE protocol which is incompatible with Bazel 7.

PR #3029, opened 2026-09-21
test: update fixed_args test cases to account for old bash version on macOS

acozzette:
These tests are failing on macOS because it ships a very old version of bash (3.2), which does not correctly set `$?` in the exit handler. This change fixes it by adding a new `macos_want_exit_code` parameter and setting that to 0 where necessary.

---

### Test plan


jbedard (approved):
Any idea what is incompatible?

If it's just different exit codes can we have `macos_want_exit_code` instead of skipping them? 🤷

acozzette:
I haven't looked into it very closely yet, but apparently macOS ships an extremely old bash version in order to stay on GPLv2, so I'm guessing that's the reason for the incompatibility.

acozzette:
> If it's just different exit codes can we have `macos_want_exit_code` instead of skipping them? 🤷

Sure, I updated the change to do that instead.

jbedard (approved):

